PR #5, opened 2018-10-11
Use logger instead of print

Glutexo:
Using print is fine for a development beginning, but not very appropriate for actual logging and debugging in the future. Switched to a logger before the prints will sprinkle around the whole codebase.

### DIFF
--- a/inventory/views.py
+++ b/inventory/views.py
@@ -1,8 +1,12 @@
 from django.db.models import Q
+from logging import getLogger
 from rest_framework import viewsets, status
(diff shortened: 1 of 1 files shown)


Glutexo:
Maybe @kylape would like to review?

Glutexo:
This PR is no longer relevant as the app [switched to Flask](
https://github.com/RedHatInsights/insights-host-inventory/commit/c67fb0ea0021203aaf3a2a06b05c6c8a54d0d070) recently.

PR #6, opened 2018-10-17
Use dynamic-rest library to display models

PaulWay:
The `dynamic-rest` library allows more comprehensive query handling and consistent data serialisation across paging systems.

The insights-advisor-api would like to use the `django-rest-models` library to access the inventory API as if it were a native model in Django.  This requires the use of `dynamic-rest` views and serializers in the target API.


jhjaggars:
Since the project is using Connexion and Flask I'm going to close this PR.

PR #10, opened 2018-10-29
Allow to send request without canonical facts

dLobatog:
At the moment, the try block doesn't catch the possibility that the user
is not sending the canonical facts on the first request to create the
host. In that case, 'validated_data' will not contain the key, and an
KeyError will be thrown.

```

File "/code/inventory/views.py" in create
  16.         cf = serializer.validated_data["canonical_facts"]

Exception Type: KeyError at /api/hosts/
Exception Value: 'canonical_facts'

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,26 @@ want to destroy that data do the following:
 docker-compose down
 ```
 
+## Basic testing
+
+To test creating a sample host, you may simply send a simple JSON request like
+this:
+
+```
+curl -H 'X_RH_IDENTITY: foobar' -H 'Content-Type: application/json' -X POST localhost:8000/api/hosts/ -d '{ "display_name": "foo", "account": "bar" }'
+```
+
+After that, you can run the following command to ensure the test host 'foo' was
+registered and appears in the list of hosts.
+
+
+```
+curl -H 'X_RH_IDENTITY: foobar' -v http://localhost:8000/api/hosts/?format=json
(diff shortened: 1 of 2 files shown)


Glutexo:
As long as there is a default value (empty dict) for the canonical fact, so it saves to the database, this makes sense to me. Your host doesn’t have to have any other deduplication facts if it comes from a well-behaved source. Maybe @jhjaggars has thoughts on this?

Glutexo:
At least one canonical fact is still required, but this is properly catched. If I try to post a host without any canonical facts, I get eror _400 Bad Request_ with an explanatory message.

```
POST http://localhost:8080/api/inventory/v1/hosts

HTTP/1.0 207 MULTI STATUS
Content-Type: application/json
Content-Length: 275
Server: Werkzeug/0.15.2 Python/3.6.7
Date: Mon, 15 Apr 2019 15:23:44 GMT

{
  "data": [
    {
      "detail": "At least one of the canonical fact fields must be present.",
      "host": {
        "account": "abc123"
      },
      "status": 400,
      "title": "Invalid request",
      "type": "about:blank"
    }
  ],
  "errors": 1,
  "total": 1
}


Response code: 207 (MULTI STATUS); Time: 22ms; Content length: 275 bytes
```

Closing as resolved. If you disagree, please reopen this pull requests and drop a comment. Thanks!

PR #12, opened 2018-10-30
Fix the database name in the app init script

Glutexo:
The Docker Compose file [uses](https://github.com/RedHatInsights/insights-host-inventory/blob/d8f46c7908a6e6d90b8e7a9cd455b60845263bee/docker-compose.yml#L9) “insights” as a database name. Changed the _INVENTORY_DB_NAME_ configuration [value](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:db_name_fix?expand=1#diff-1f9f743421417fb2794c88447083a031R16) default to match this name.


Glutexo:
Rebased on current master.

Glutexo:
Rebased again. Ready for a review.

karelhala:
I like this! I had problems running this locally with `docker-compose -f dev.yml up` becaue it automatically created `insights` db but tries to connect to `test_db`.

Glutexo:
Rebased on current [_master_](https://github.com/RedHatInsights/insights-host-inventory/tree/5540dec3c697a67403c3122455472ffee7aa3008). Replaced the name in README as well. Not only that the _insights_ is the name of the database in the [Docker Compose file](https://github.com/RedHatInsights/insights-host-inventory/blob/5540dec3c697a67403c3122455472ffee7aa3008/dev.yml), but its also far better name than _test_db_. It’s not only non-descriptive, even more if it’s not the only database in the Postgres instance, but it’s a bit misleading too: it’s not a db dedicated to the tests.

Glutexo:
Rebased again. I think this can be merged together with #169.

Glutexo:
Yaai! 🦊

PR #13, opened 2018-10-30
Initialize a database with the first migration

Glutexo:
Added Alembic default files generated by `manage.py db init`. Created [the first migration](
https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:db_schema?expand=1#diff-38a1839a77a8d0a25574d61d97372f3)f that creates the first attempt on a ["hosts" table](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:db_schema?expand=1#diff-38a1839a77a8d0a25574d61d97372f3fR21).

### DIFF
--- a/migrations/README
+++ b/migrations/README
@@ -1,0 +1,1 @@
+Generic single-database configuration.
--- a/migrations/alembic.ini
+++ b/migrations/alembic.ini
@@ -1,0 +1,45 @@
+# A generic, single database configuration.
+
+[alembic]
+# template used to generate migration files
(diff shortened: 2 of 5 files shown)


quarckster:
@Glutexo Do I understand right, that `manage.py db migrate` doesn't work currently due to missing code on this PR?

Glutexo:
@quarckster Yes. There were no migrations in the repository, so there were no way to create the database schema.

PR #15, opened 2018-10-30
Fix PEP 8 errors

Glutexo:
Fixed all PEP 8 errors that Flake8 found. Sorry for such an invasive change, @dehort. I was a bit afraid that if we won’t enforce the style now, we’d never get rid of the mess.


jhjaggars:
How about using https://github.com/ambv/black to enforce style and formatting?

Glutexo:
@jhjaggars I‘d go for it. Or at least for a flake8 PR linting. I just wanted submit a first cleanup to begin with.

Glutexo:
This won’t be necessary if #22 gets merged. There are still some unanswered concerns. Once that is settled, let’s close this.

Glutexo:
#22 is merged. Closing.

PR #22, opened 2018-11-01
using black for formatting

jhjaggars:
This PR shows what it'd be like to use black for formatting


Glutexo:
Hmmm… I get different results from Black in its default configuration:

* Puts operators on the beginning of the continuation line of multi-line expressions.
* Replaces single-quotes by double-quotes.
* Removes some blank lines, typically the first blank line in a class definition.

It looks like that if we want to use Black for linting, we should make sure it’s configured the same way. Black itself [discourages](https://github.com/ambv/black#pyprojecttoml) configuration though.

But 👍 – I’d be willing to use it.

PR #23, opened 2018-11-01
moving from a single canonical facts field to explict fields

jhjaggars:


PaulWay:
So how are these going to be stored in the database?  A set of related tables?  Are you going to manage the `canonical_facts` property internally?

PR #26, opened 2018-11-02
Update python_version according to docker-compose

dLobatog:
Currently, 'FROM python:3' in docker-compose.yml is installing the app in 
a container which only has python 3.7.1 available, so the mismatch between
that and the version in Pipfile throw a warning and make it unable to start.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -23,4 +23,4 @@ request = "*"
 pytest = "*"
 
 [requires]
-python_version = "3.6"
+python_version = "3.7.1"


Glutexo:
What do you think, @dehort @jhjaggars, can we merge this?

Glutexo:
There is no _docker-compose.yml_ now. But in the [_Pipfile_](https://github.com/RedHatInsights/insights-host-inventory/blob/24fb0c9847e17e8e64c6b1327cd2974520828fbf/Pipfile) the Python version is already [specified](https://github.com/RedHatInsights/insights-host-inventory/blob/24fb0c9847e17e8e64c6b1327cd2974520828fbf/Pipfile) as _3.6_. Closing as resolved.

PR #30, opened 2018-11-05
Hook a dummy auth validation to requests

Glutexo:
Created [an Authentication Manager](https://github.com/RedHatInsights/insights-host-inventory/pull/30/files#diff-3b16d374ef08532ca6b01246c5f0fa41R17) that hooks to every request. It parses and validates an identity HTTP header and makes the payload available to views. Currently the parsing and validation is only [a dummy stub](https://github.com/RedHatInsights/insights-host-inventory/pull/30/files#diff-3b16d374ef08532ca6b01246c5f0fa41R9): every request is valid if the header is present.

Replaces #7. Asking @dehort for a review.

I created [a new file](
https://github.com/RedHatInsights/insights-host-inventory/pull/30/files#diff-13729a38e16bcaf94b928e33e5dd468a) with unit tests, keeping [_test_api.py_](
https://github.com/RedHatInsights/insights-host-inventory/pull/30/files#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2) only for API calls. When writing new tests, please ensure that the new [_bypass_auth_](https://github.com/RedHatInsights/insights-host-inventory/pull/30/files#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2R27) patch is applied to every call that makes requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Running the tests is quite simple:
 
 ```
 ./test_api.py
+./test_unit.py
 ```
 
 ## Running the server
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,6 +1,8 @@
 import os
 import logging
 import connexion
+
+from app.auth import AuthManager
 from flask_api import FlaskAPI
 from flask_sqlalchemy import SQLAlchemy
 from connexion.resolver import RestyResolver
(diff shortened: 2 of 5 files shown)


Glutexo:
Pushed a new commit. I utilized the [flask-login’s](
https://github.com/maxcountryman/flask-login) approach using the Werkzeug‘s request context. See [read](https://github.com/maxcountryman/flask-login/blob/3e521a326696cafbfbebfbb80a2fbffed68e6cf3/flask_login/utils.py#L348) and [write](https://github.com/maxcountryman/flask-login/blob/master/flask_login/login_manager.py#L295).

The identity from the HTTP header is accessible in the views:

```python
>>> from app.auth import current_identity
>>> print(current_identity)
"abc"
```

PR #9, opened 2018-10-26
[dropped] Validate HTTP auth header value

Glutexo:
Used a new [identity module](
https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:validate_http_auth_format?expand=1#diff-d1c1270f7ffbc4d11254157ca1f93e88) to validate the HTTP header. The validation is only a dummy that is going to be replaced by a real functionality. (Or at least with the more real, but still dummy one from #8.) Currently it only checks whether the header is not empty.

This PR is based on #7. It builds on the dummy middleware and delegates the validation to the new [module](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:validate_http_auth_format?expand=1#diff-d1c1270f7ffbc4d11254157ca1f93e88). After #7 gets merged, I’d like to ask @dehort for a review.

Dependencies:

* [x] #7 Use a dummy authentication middleware

PR #31, opened 2018-11-06
Bind validation to an identity module interface

Glutexo:
This is a follow-up pull-request to #30. Replaces #9.

Created a dummy [Identity module](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:auth_with_dummy_identity?expand=1#diff-aaeb53fba2b4225a7fde4430b4b26033). This module does the same primitive validation as the original __validate_identity_ function. [Bound](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:auth_with_dummy_identity?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R29) to the Authentication Manager so it can be replaced by a real working module.

After #30 gets merged, would appreciate a review from @dehort.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Running the tests is quite simple:
 
 ```
 ./test_api.py
+./test_unit.py
 ```
 
 ## Running the server
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,6 +1,8 @@
 import os
 import logging
 import connexion
+
+from app.auth import AuthManager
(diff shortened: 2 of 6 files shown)


dehort:
Does this need to be merged instead of #30 or on top of #30?  It looks like it contains #30 but I'd like to confirm that.

Glutexo:
It goes from #30, but I made it a separate PR, so the changes are more atomic. If this is merged first, #30 is included though.

PR #33, opened 2018-11-06
Fix issues related to docker and docker-compose

SteveHNH:
Fixed a few issues that were keeping the inventory service
from properly launching when called via docker-compose.

There's also a bug fix in here related to connexion and swagger_ui.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
-FROM python:3
+FROM python:3.6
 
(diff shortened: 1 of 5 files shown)


SteveHNH:
Learned a bunch about how some of this works between the time I made this PR and now. I'll take another swing at it and produce a better version of this. Closing for now

PR #37, opened 2018-11-07
Add gunicorn dependency and fix a bug

SteveHNH:
A couple bugs pop up when running behind gunicorn, so I fixed those
as well.

 - Validate responses currently produces a bunch of errors. Disabled
   that for now
 - Set swagger_ui to False so we don't get that error


Glutexo:
Just noting that I’d love to see a follow-up issue reminding to remove that workaround.

PR #38, opened 2018-11-08
Allow having views without authentication

Glutexo:
Created a [_bypass_auth_](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:bypass_auth?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R47) decorator. Marked views won’t require having the identity HTTP header. This is useful for diagnostic requests that don’t touch the database, e.g. **a health probe**.

~~I’ve not written tests for this yet. Pushing here in case I wouldn’t make it before my departure.~~ There are currently only [unit tests](https://github.com/RedHatInsights/insights-host-inventory/pull/38/files#diff-13729a38e16bcaf94b928e33e5dd468a), because it’s not possible to test it functionally without an actual unauthenticated route. I’ll prepare another PR that allows to test this even before making the actual health check endpoint.

### DIFF
--- a/app/auth/__init__.py
+++ b/app/auth/__init__.py
@@ -8,6 +8,10 @@ __all__ = ["bypass_auth", "init_app", "current_identity"]
 _IDENTITY_HEADER = "x-rh-identity"
 
 
+class NoIdentityError(RuntimeError):
+    pass
+
+
 def _validate_identity(payload):
     """
     Identity payload validation dummy. Fails if the data is empty.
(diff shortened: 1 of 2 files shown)


Glutexo:
Rebase on [current _master_](https://github.com/RedHatInsights/insights-host-inventory/tree/53768932260e9e9ab8924718ec7667fd9dceb3be). Added unit tests. Functional test will come in a separate PR, because it’s not currently possible to make them.

Glutexo:
Ready for a review. Asking @dehort.

#45 will allow to make functional tests for this.

Glutexo:
Replaced by #50. This approach by requiring authentication by default would make it difficult to access the Swagger UI console. It also contradicts the OpenAPI specification, where there are explicitly specified those methods that require authentication.

PR #39, opened 2018-11-08
Create a primitive health probe

Glutexo:
Added a new view [_/health_](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:health_probe?expand=1#diff-b5704d15302288389d3aef6d5ab2a4f3R210) that simply responds with 200. This can serve as a responder for a health probe. It still requires the request to be authenticated though.

~~I didn’t write tests for this yet. Pushing here so it’s ready for you to use.~~

After #38 gets merged, the health probe can be modified so it won’t require authentication:

```python
from app.auth import bypass_auth

@bypass_auth
def health():
    …
```

### DIFF
--- a/api/__init__.py
+++ b/api/__init__.py
@@ -1,0 +1,5 @@
+__all__ = ["health"]
+
+
+def health():
+    return "", 200
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -233,6 +233,16 @@ paths:
       responses:
         "200":
           description: Successfully added or removed a tag
+  /health:
+    get:
+      operationId: api.health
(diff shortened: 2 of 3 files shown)


Glutexo:
Rebased on [current _master_](https://github.com/RedHatInsights/insights-host-inventory/tree/53768932260e9e9ab8924718ec7667fd9dceb3be) and added [tests](https://github.com/RedHatInsights/insights-host-inventory/pull/39/files#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2R696). The request doesn’t touch the database at all, so I split the test case classes to make this a bit faster.

PR #45, opened 2018-11-21
Make create_app return the connexion app

Glutexo:
Change the return value of the [_app.create_app_](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:return_connexion_app?expand=1#diff-1f9f743421417fb2794c88447083a031R29) method to return the whole Connexion app and not only the Flask app inside. That makes possible to have control of the API specification in tests.

This will help #38, allowing to specify an unauthenticated operation in the tests and thus to make functional tests for the authentication bypassing.

Asking @dehort for a review.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -58,4 +58,4 @@ def create_app(config_name):
 
     db.init_app(flask_app)
 
-    return flask_app
+    return connexion_app
--- a/manage.py
+++ b/manage.py
@@ -7,9 +7,11 @@ from app import models
 # import models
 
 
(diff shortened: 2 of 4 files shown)


Glutexo:
This is no longer necessary as #52 is now closed. @dehort pointed out that returning the whole Connexion app goes against Flask paradigms. I [concluded](
https://github.com/RedHatInsights/insights-host-inventory/pull/52#issuecomment-442767584) that it’s actually not necessary to have a separate API specification for the test.

PR #47, opened 2018-11-21
[waiting] Test an operation with bypassed authentication

Glutexo:
Added a new [API specification](
https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:bypass_auth_functional_tests?expand=1#diff-2b12b514b8cc8a8dbe7883281c5ea68d) that is [used](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:bypass_auth_functional_tests?expand=1#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2R74) only in [tests](
https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:bypass_auth_functional_tests?expand=1#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2). This [specification](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:bypass_auth_functional_tests?expand=1#diff-2b12b514b8cc8a8dbe7883281c5ea68d) contains an [operation](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:bypass_auth_functional_tests?expand=1#diff-2b12b514b8cc8a8dbe7883281c5ea68dR15) that [does not require authentication](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:bypass_auth_functional_tests?expand=1#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2R28). It is used in [the new tests](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:bypass_auth_functional_tests?expand=1#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2R685) that verify that authentication is not required for marked views.

This custom operation won’t be needed once there will be the unauthenticated health check endpoint. That means once #39 gets merged.

This pull request is based on #38 and #45. A review and merge will be possible after those two get merged.

PR #52, opened 2018-11-28
[wip] Test operations with and without authentication

Glutexo:
Added a new [API specification](
https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:requires_identity_functional_tests?expand=1#diff-2b12b514b8cc8a8dbe7883281c5ea68d) that is used only in [tests](
https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:requires_identity_functional_tests?expand=1#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2). This [specification](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:requires_identity_functional_tests?expand=1#diff-2b12b514b8cc8a8dbe7883281c5ea68d) contains [an operation](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:requires_identity_functional_tests?expand=1#diff-2b12b514b8cc8a8dbe7883281c5ea68dR13) that does require the identity header and [an operation](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:requires_identity_functional_tests?expand=1#diff-2b12b514b8cc8a8dbe7883281c5ea68dR19) that doesn’t. It is used in [the new tests](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:requires_identity_functional_tests?expand=1#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2R66) that
verify that authentication is required only for marked views.

This provides test for #50 and is based on [that branch](https://github.com/Glutexo/insights-host-inventory/tree/requires_identity). Contains merged #45 that allows to write the functional tests. Replaces #47 that tested the original authentication approach. Asking @dehort for a review after the dependent PRs get reviewed and merged.

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -2,7 +2,7 @@ import os
 import logging
 from enum import Enum
 from app.models import Host
-from app.auth import current_identity
+from app.auth import current_identity, requires_identity
 from app import db
 from flask import current_app
 
@@ -12,6 +12,7 @@ FactOperations = Enum("FactOperations", ["merge", "replace"])
 logger = logging.getLogger(__name__)
 
 
+@requires_identity
 def addHost(host):
     """
     Add or update a host
@@ -56,6 +57,7 @@ def addHost(host):
         return found_host.to_json(), 200
 
 
+@requires_identity
 def getHostList(tag=None, display_name=None, page=1, per_page=100):
     """
     Get the list of hosts.  Filtering can be done by the tag or display_name.
@@ -119,6 +121,7 @@ def findHostsByDisplayName(account, display_name, page, per_page):
     return (total, found_host_list)
 
 
+@requires_identity
 def getHostById(hostId, page=1, per_page=100):
     current_app.logger.debug("getHostById(%s, %d, %d)" % (hostId, page, per_page))
     query_results = Host.query.filter(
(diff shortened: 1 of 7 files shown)


dehort:
I need to review this more closely.  At a quick glance, I'm not sure I see a big benefit to adding a test only api to test un-protected methods.  The host api will have the /health endpoint which will be unprotected.  I believe there is at least one test that calls this method.  

In order to add another (test only) api, we have to modify the create_app method to return the connexion object which goes against what appears to be a defacto standard in flask apps.  (Not a huge deal, but more complexity/possible confusion is added).  This change causes a ripple effect through the users of the create_app.  Again, not a huge deal, but something to think about.

Glutexo:
You’re right that once we have at least one unauthenticated endpoint (the health check) and at least one authenticated (any hosts-related one) then we can use those. Since #39 is now merged (thanks!) I will update this pull request to use the health check endpoint instead.

And it’s true that I didn’t realize that not returning the Flask app from _create_app_ breaks the common paradigm.

Once I’ll finish the automatic flagging based on the parameters in the OpenAPI specification, I‘d like to have tested both the route and method parameters. This will be however covered by the unit tests. It’s not necessary to cover this by the functional tests if there are no routes that would be affected.

Glutexo:
Closing this as it can be effectively replaced by the new simpler pull request #53.

PR #54, opened 2018-11-29
simplifying auth handling

jhjaggars:
Auth is now implemented as a single decorator.  This has
several advantages in that it is easier to test and understand.

Removed tests for the old `_before*` hook.
Added tests to cover new auth methods.
Removed old populate script.
Introduced py.test


jhjaggars:
I also added coverage integration:

```
(insights-inventory-UWHZwXDL) ➜  insights-inventory git:(master) ✗ pytest --cov=.
=================================================================================================================================== test session starts ===================================================================================================================================
platform darwin -- Python 3.6.4, pytest-4.0.1, py-1.7.0, pluggy-0.8.0
rootdir: /Users/jjaggars/code/insights-inventory, inifile:
plugins: cov-2.6.0
collected 50 items

test_api.py ....................................                                                                                                                                                                                                                                    [ 72%]
test_unit.py ..............                                                                                                                                                                                                                                                         [100%]

---------- coverage: platform darwin, python 3.6.4-final-0 -----------
Name                   Stmts   Miss  Cover
------------------------------------------
api/__init__.py            3      0   100%
api/host.py              129     11    91%
app/__init__.py           34      2    94%
app/auth/__init__.py      41      2    95%
app/auth/identity.py      19      0   100%
app/models.py             96      2    98%
app/utils.py              61     11    82%
manage.py                 11     11     0%
run.py                    11     11     0%
test_api.py              375      6    98%
test_unit.py              92      7    92%
------------------------------------------
TOTAL                    872     63    93%
```

app/auth/__init__.py
```diff
@@ -12,44 +14,38 @@ class NoIdentityError(RuntimeError):
...
-    try:
-        identity = from_encoded(payload)
-    except (KeyError, TypeError, ValueError):
-        abort(Forbidden.code)
 
+def _validate(identity):
     try:
         validate(identity)
-    except ValueError:
+    except Exception:
```
Glutexo: What’s wrong about _ValueError_? I‘d say that having a specific exception in case of validation fail would be a good thing. We can then tell whether the header is invalid or whether some error occurred during the validation.

Glutexo:
I‘m a bit sad about this simplification, even though it’s definitely concise and elegant. As I stated in my PRs, the decorator was a temporary measure as I was working on reading the information automatically from the OpenAPI specification. The attribute assignment accompanied with the __before_request_ was a pretty good fit for that. I’m afraid that I don’t see any other way how to achieve this than reintroducing the _before_request_ hook. If you have any ideas, please, tell me. Thanks!

Glutexo:
Oh, maybe there is a way. If I apply the decorator from outside like this, it might work. I’ll give it a try:

```python
from importlib import import_module
module = import_module(view_func.__module__)
setattr(module, view_func.__name__, requires_identity(view_func))
```

PR #57, opened 2018-12-05
Require identity header by API specs

Glutexo:
Some operations require the identity header authentication and some don’t. This is described by the [OpenAPI specification](
https://github.com/RedHatInsights/insights-host-inventory/blob/6662e113ddf55b175fcd01375374d2b1489512db/swagger/api.spec.yaml). [Parse](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:auth_by_spec_simple?expand=1#diff-928a71ff2d667396888de19dbef2c2c5R36) [this spec](
https://github.com/RedHatInsights/insights-host-inventory/blob/6662e113ddf55b175fcd01375374d2b1489512db/swagger/api.spec.yaml) to populate the authenticated paths. The [_requires_identity_](https://github.com/RedHatInsights/insights-host-inventory/blob/6662e113ddf55b175fcd01375374d2b1489512db/app/auth/__init__.py#L39) decorator is [applied](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:auth_by_spec_simple?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R66) to the all operations in paths that require the identity
header. These decorated view functions are then picked by Connexion.

Caveats/TODO – will do in subsequent pull requests.

* Doesn’t support overriding the parameter by the method specification.
* Doesn’t use the RestyResolver, but the default Connexions way of describing the exact module.function.

There is some black magic going on with the [_decorate_](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:auth_by_spec_simple?expand=1#diff-5609c27c7d83fbaa57c74f5491ef464bR122) method. It patches the method in the module as if the decorator was applied at the time of its definition. I don’t like this _that much_ and it can be probably better done by using a custom resolver. We can decide later how much a refactor is worth it.

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -14,7 +14,6 @@ FactOperations = Enum("FactOperations", ["merge", "replace"])
 
 @api_operation
 @metrics.api_request_time.time()
-@requires_identity
 def add_host(host):
     """
     Add or update a host
@@ -112,7 +111,6 @@ def update_existing_host(existing_host, input_host):
 
 @api_operation
 @metrics.api_request_time.time()
-@requires_identity
 def get_host_list(tag=None, display_name=None, fqdn=None, page=1, per_page=100):
     """
     Get the list of hosts.  Filtering can be done by the tag, display_name, or fqdn.
@@ -198,7 +196,6 @@ def find_hosts_by_canonical_facts(account_number, canonical_facts, page, per_pag
 
 @api_operation
 @metrics.api_request_time.time()
-@requires_identity
 def get_host_by_id(host_id_list, page=1, per_page=100):
     current_app.logger.debug(
             "get_host_by_id(%s, %d, %d)" % (host_id_list, page, per_page)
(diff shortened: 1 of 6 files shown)


Glutexo:
Rebased on the current [_master_](https://github.com/RedHatInsights/insights-host-inventory/tree/6662e113ddf55b175fcd01375374d2b1489512db). Asking @dehort for a review.

Glutexo:
Rebased once again, so it’s mergeable.

Glutexo:
Created some follow-up issues so they are not forgotten: #78, #79

Glutexo:
Created a copy of this (#87) that does not use [_unittest.mock_](https://docs.python.org/3.7/library/unittest.mock.html) in tests. Please pick whichever you like better.

Glutexo:
Rebased on [current _master_](https://github.com/RedHatInsights/insights-host-inventory/tree/142588853c96554443b5782a4b9db42d90d8f84a). Confirmed working, so mergeable.

dehort:
I am not planning on merging this in for a while.  The approach that we are currently using is working.  We need to be focusing on stabilizing things at this point, so I do not want to change things unless we must.

Glutexo:
Closing this as we have #57. Will keep the branch alive.

Glutexo:
Ok. I’ll keep it without touching until we decide it’s time to reconsider.

PR #62, opened 2018-12-10
Split base path from API path

Glutexo:
The base path of the whole application doesn’t include the _api/v1_ part. It is useful for the endpoints that are not parts of the API itself, e.g. health check or metrics. Extract [a new method](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:split_path?expand=1#diff-1f9f743421417fb2794c88447083a031R29) from [__get_api_path_](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:split_path?expand=1#diff-1f9f743421417fb2794c88447083a031R38) to use when defining the non-API endpoints.

Asking @dehort for a review.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -26,14 +26,22 @@ def _get_db_pool_size():
     return int(os.getenv("INVENTORY_DB_POOL_SIZE", "5"))
 
 
-def _get_api_path():
-    app_name = os.getenv("APP_NAME", "inventory")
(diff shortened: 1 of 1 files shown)


Glutexo:
This is no longer relevant as #58 exposes the management endpoints at _/_ instead of _/r/insights/platform/inventory/. Closing.

PR #63, opened 2018-12-10
[waiting] Use Prometheus to gather metrics

Glutexo:
[Installed]((https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:prometheus?expand=1#diff-1e61a31bf9b94805f869dc4137ec1885R22)) the Prometheus client and [exposed](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:prometheus?expand=1#diff-1f9f743421417fb2794c88447083a031R91) its default metrics (Python information and statistics) to the new _/metrics_ endpoint.

This endpoint is not a part of the API and because of that it’s not under the _api/v1_ prefix. I’ll DRY this once #62 gets merged. The pull request is based on #60 that uses application dispatching to combine the Flask app with the Prometheus one.

Will ask @dehort for review once #60 gets merged.

PR #64, opened 2018-12-10
[waiting] DRY the Prometheus code to use the base path

Glutexo:
Make [use](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:prometheus_dry?expand=1#diff-1f9f743421417fb2794c88447083a031R94) of the new base path helper [function](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:prometheus_dry?expand=1#diff-1f9f743421417fb2794c88447083a031R31) to build the metrics path for Prometheus. It is not a part of the API, so its URL is shorter.

This is based on #63 and #62. Will ask @dehort for a review once these two get merged.


Glutexo:
Prometheus is introduced by #58. This is no longer needed. Closing.

PR #76, opened 2018-12-21
DRY the identity pick and validation

Glutexo:
Introduced a new custom [Exception](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:dry_identity?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R20) to DRY the identity pick and validation. If the header is missing, undecodeable or invalid, this new [Exception](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:dry_identity?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R20) is raised. Aborting the request with a Forbidden status code is now unified in [one place](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:dry_identity?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R56).

This will help with counting invalid logins, which now must accompany every _abort_ call.

Asking @dehort for a review.

### DIFF
--- a/test_unit.py
+++ b/test_unit.py
@@ -20,10 +20,6 @@ import pytest
 from werkzeug.exceptions import Forbidden
 
 
-def _abort():
-    raise Forbidden
-
-
 class ApiOperationTestCase(TestCase):
     """
     Test the API operation decorator that increments the request counter with every
@@ -271,7 +267,7 @@ class AuthRequiresIdentityTestCase(TestCase):
     def _dummy_view_func(self):
         pass
 
-    @patch("app.auth.abort", wraps=_abort)
+    @patch("app.auth.abort", side_effect=Forbidden)
     @patch("app.auth._validate")
     @patch("app.auth._pick_identity", side_effect=InvalidIdentityError)
     def test_request_is_aborted_with_forbidden_if_identity_pick_fails(
@@ -281,7 +277,7 @@ class AuthRequiresIdentityTestCase(TestCase):
             self._dummy_view_func()
         abort.assert_called_once_with(Forbidden.code)
 
-    @patch("app.auth.abort", wraps=_abort)
+    @patch("app.auth.abort", side_effect=Forbidden)
     @patch("app.auth._validate")
     @patch("app.auth._pick_identity", side_effect=InvalidIdentityError)
     def test_request_context_is_untouched_if_identity_pick_fails(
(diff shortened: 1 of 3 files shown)


test_unit.py
```diff
@@ -210,14 +213,129 @@ def test_invalid(self):
...
+    @patch("app.auth.from_encoded")
+    @patch("app.auth.request", headers={})
+    def test_identity_is_not_valid_if_header_is_missing(self, _, from_encoded):
+        with self.assertRaises(InvalidIdentityError):
+            _pick_identity()
+        from_encoded.assert_not_called()
+
+    @patch("app.auth.from_encoded")
+    @patch("app.auth.request", headers={_IDENTITY_HEADER: Mock()})
+    def test_identity_is_decoded(self, request, from_encoded):
```
dehort: I'm not sure I see much value in this test case.  Isn't this same area of the code going to be hit in the valid/invalid decode testing?

test_unit.py
```diff
@@ -210,14 +213,129 @@ def test_invalid(self):
...
+        _pick_identity()
+        from_encoded.assert_called_once_with(request.headers[_IDENTITY_HEADER])
+
+    @patch("app.auth.request", headers={_IDENTITY_HEADER: Mock()})
+    def test_identity_is_not_valid_if_decode_fails(self, _):
```
dehort: Maybe we should make the exception handling more generic (catch the top level Exception class) in the _pick_identity method and remove or simplify this test case.

test_unit.py
```diff
@@ -210,14 +213,129 @@ def test_invalid(self):
...
+
+    @patch("app.auth.from_encoded")
+    @patch("app.auth.request", headers={_IDENTITY_HEADER: Mock()})
+    def test_decoded_identity_is_returned(self, _, from_encoded):
+        result = _pick_identity()
+        self.assertEqual(result, from_encoded.return_value)
+
+
+@patch("app.auth._request_ctx_stack")
+class AuthRequiresIdentityTestCase(TestCase):
```
dehort: This test case seems very complicated.  This seems like a lot of (somewhat brittle) code.  I say brittle because if a method name changes in the impl, we have to chase down all of those name changes in here and the compiler cannot help us out due to the fact that they are just strings.  Maybe I am misunderstanding things.

At the least, it looks like a few of the tests could be combined.

In the end, I think we're really just testing for the case where:

- the auth header is not set
- the auth header cannot be decoded
- the auth header is invalid

test_unit.py
```diff
@@ -210,14 +216,102 @@ def test_invalid(self):
...
+    def test_request_is_not_aborted_on_valid_identity(self, _ca, abort, _rcs):
+        with self._patch_request_valid():
+            self._dummy_view_func()
+        abort.assert_not_called()
+
+    def test_valid_identity_is_stored(self, _ca, _a, _rcs):
+        with self._patch_request_valid():
+            self._dummy_view_func()
+        self.assertEqual(Identity(account_number="some number"), _get_identity())
+
```
dehort: It looks like test_request_is_not_aborted_on_valid_identity and test_valid_identity_is_stored are essentially testing the same thing. Is there a reason these tests are not combined?

Glutexo:
This is now outdated. Replaced by #203.

PR #82, opened 2019-01-02
minor improvement to swagger spec regarding host identifiers

jharting:


Glutexo:
I like this. Merge? @dehort

swagger/api.spec.yaml
```diff
@@ -323,7 +323,7 @@ definitions:
       - type: object
         properties:
           id:
-            description: An internal Inventory ID of the host.
+            description: A durable and reliable platform-wide host identifier as defined in IPP-6. Applications should use this identifier to reference hosts.
```
dehort: I don't think we should mention the IPP doc here.  I believe this swagger api will eventually be available to external customers whereas I do not think the IPP doc will be.  Mentioning the IPP doc here might lead to confusion.

jharting:
@dehort updated

PR #77, opened 2018-12-21
Count login failures

Glutexo:
Add a new Prometheus [Counter](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count?expand=1#diff-09a4c1f30c5458aa77651a6d5b4d2d2fR6) for login failures. [Increment](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R44) the counter whenever authentication fails: either if the header is [undecodeable](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R29) or [invalid](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R36).

This doesn’t cover the case when the header is completely [missing](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R24). Such case is [caught](https://github.com/RedHatInsights/insights-host-inventory/blob/dc3d424bbe873f9598958f9e53ac6181518e884d/swagger/api.spec.yaml#L16) by Connexion and doesn’t get to the authentication [module](
https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41) at all.

Asking @dehort for a review.

PR #83, opened 2019-01-03
Count login failures, now without mocking

Glutexo:
Add a new Prometheus [Counter](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count_without_mocking?expand=1#diff-09a4c1f30c5458aa77651a6d5b4d2d2fR6) for login failures. [Increment](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count_without_mocking?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R45) the counter whenever an authentication fails: either if the header is [undecodable](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count_without_mocking?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R31) or [invalid](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count_without_mocking?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R38).

This doesn’t cover the case when the header is completely [missing](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count_without_mocking?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41R25). Such case is [caught](https://github.com/RedHatInsights/insights-host-inventory/blob/e56670337bb84a9bc1368e8f0a8018223c051faf/swagger/api.spec.yaml#L16) by Connexion and doesn’t get to the authentication [module](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:login_failure_count_without_mocking?expand=1#diff-3b16d374ef08532ca6b01246c5f0fa41) at all.

This is a version of #77, but without any mocking in the tests. Otherwise, the code is identical. I made this in response to @dehort’s [review](
https://github.com/RedHatInsights/insights-host-inventory/pull/70#issuecomment-44808645) of #70. Please review and pick either this or the original PR #77 depending on which approach you like better.

### DIFF
--- a/api/metrics.py
+++ b/api/metrics.py
@@ -4,3 +4,4 @@ api_request_time = Summary("inventory_request_processing_seconds", "Time spent p
 api_request_count = Counter("inventory_request_count", "The total amount of API requests")
 create_host_count = Counter("inventory_create_host_count", "The total amount of hosts created")
 update_host_count = Counter("inventory_update_host_count", "The total amount of hosts updated")
+login_failure_count = Counter("inventory_login_failure_count", "The total amount of failed login attempts")
--- a/app/auth/__init__.py
+++ b/app/auth/__init__.py
@@ -1,5 +1,6 @@
 import os
 from functools import wraps
+from api.metrics import login_failure_count
 from app.auth.identity import from_encoded, validate, Identity
 from flask import abort, request, _request_ctx_stack
 from werkzeug.local import LocalProxy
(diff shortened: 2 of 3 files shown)


Glutexo:
Oh. It’s possible that Prometheus already has a decorator that I could have used. I’ll examine.

Glutexo:
Moved the test descriptions from the noisy comments to the test names. Also rebased on [current _master_](https://github.com/RedHatInsights/insights-host-inventory/tree/142588853c96554443b5782a4b9db42d90d8f84a).

Now when we agreed that mocking is not inherently bad, I think that we can ditch this PR. Still, it’d be nice to pull some tests from here to #77 to test the whole module and not only the tiniest units. I‘ll close this once I do the move.

Glutexo:
Closing this as  #77 should be ok now. I’ll keep the branch there for reference until that PR gets merged.

PR #87, opened 2019-01-08
Require identity header by API specs – now without unittest.mock

Glutexo:
Some operations require the identity header authentication and some don’t. This is described by the [OpenAPI specification](https://github.com/RedHatInsights/insights-host-inventory/blob/e56670337bb84a9bc1368e8f0a8018223c051faf/swagger/api.spec.yaml). [Parse](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:auth_by_spec_simple_without_mocking?expand=1#diff-928a71ff2d667396888de19dbef2c2c5R36) this spec to populate the authenticated paths. The [_requires_identity_](https://github.com/RedHatInsights/insights-host-inventory/blob/e56670337bb84a9bc1368e8f0a8018223c051faf/app/auth/__init__.py#L39) decorator is applied to all operations in paths that require the identity header. These decorated view functions are then picked by Connexion.

Caveats/TODO:

* Doesn’t support overriding the parameter by the method specification.
* Doesn’t use the _RestyResolver_, but the default Connexions way of describing the exact module.function.

This is a version of #57, but without using [_unittest.mock_](https://docs.python.org/3.7/library/unittest.mock.html) in tests. Asking @dehort for a review – pick whichever version you like better.

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -4,7 +4,7 @@ from enum import Enum
 from flask import current_app
 
 from app.models import Host
-from app.auth import current_identity, requires_identity
+from app.auth import current_identity
 from app import db
 from api import metrics
 
@@ -16,7 +16,6 @@ logger = logging.getLogger(__name__)
 
 
 @metrics.api_request_time.time()
-@requires_identity
 def addHost(host):
     """
     Add or update a host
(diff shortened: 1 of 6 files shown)


dehort:
I am not planning on merging this in for a while.  The approach that we are currently using is working.  We need to be focusing on stabilizing things at this point, so I do not want to change things unless we must.

Glutexo:
Closing as we have #57. Will keep the branch alive for now.

PR #100, opened 2019-01-10
Add ability to filter by fqdn

bsquizz:
Can now query with GET /hosts?fqdn="something.com" -- `get_host_by_canonical_facts` will be used. If the host exists, a list of 1 host is returned. Otherwise, empty list is returned.


bsquizz:
Changing this to return multiple hosts

bsquizz:
this is ready @dehort

PR #107, opened 2019-01-18
Use logstash for logging

Glutexo:
Created a new logging configuration file that specifies Logstash to be used for logging.

The gist here is the [_logconfig.ini_](
https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:logstash?expand=1#diff-b289960fa46262613513c74bcd6b9471) file. If you pass it as `--log-config=logconfig.ini` to the _gunicorn_ run command, it makes it log all the messages in the Logstash format. Including those that Gunicorn outputs not only before the app itself gets initialized, but even those before the [_run.py_](
https://github.com/RedHatInsights/insights-host-inventory/blob/654ecd661d186a6da9b9645865140ebcef8b8c3a/run.py) script is actually executed.

Like this, it’s possible to add arbitrary fields to the message objects and still log plain strings.


Glutexo:
I was fiddling with that a bit and the problem was that there were still some messages that weren’t formatted by Logstash. That was because they were emitted between the configuration came into effect. Like this the Gunicorn uses it by itself. And sice we are [using](https://github.com/RedHatInsights/insights-host-inventory/blob/654ecd661d186a6da9b9645865140ebcef8b8c3a/run.py#L15) the Gunicorn handlers for our app’s logging, these are formatted by Logstash too for free.

Glutexo:
cc @dehort

PR #108, opened 2019-01-21
Logging configuration changes

dehort:
Modified the logging so that we can enable and disable logging in different parts of the application.  In other words, we can enable debug logging for the "api" and "app" parts of the application without enabling debug logging in flask, connexion, etc.

Added an environment variable (INVENTORY_LOGGING_CONFIG_FILE) that points to the logging configuration file.  This file can be shared with guicorn (gunicorn --log-config=$INVENTORY_LOGGING_CONFIG_FILE run).  This logging config file is used to "bootstrap" the logging configuration before the flask app is created.

This approach allows us to be able to configure the logging subsystem outside of the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ Configuration system properties:
  INVENTORY_DB_NAME="test_db"
  INVENTORY_DB_POOL_TIMEOUT="5"
  INVENTORY_DB_POOL_SIZE="5"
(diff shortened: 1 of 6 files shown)


Glutexo:
A review right after merging. Yay. I wanted to request some little changes (see my comments), but those can be made in a subsequent pull-requests. I can make those myself.

PR #118, opened 2019-02-18
Resolve display name

dehort:


app/models.py
```diff
@@ -62,12 +62,20 @@ def convert_dict_to_json_facts(fact_dict):
     return fact_list
 
 
+def _set_display_name_on_save(context):
```
Glutexo: I find this function name misleading. It doesn’t set anything, it determines the default value for the field. What about __display_name_default_?

app/models.py
```diff
@@ -62,12 +62,20 @@ def convert_dict_to_json_facts(fact_dict):
...
+    if not context.get_current_parameters()['display_name']:
```
Glutexo: This condition is untested.

PR #119, opened 2019-02-18
Searching changes

dehort:


api/host.py
```diff
@@ -193,6 +204,30 @@ def find_hosts_by_canonical_facts(account_number, canonical_facts, page, per_pag
...
+    logger.debug("find_hosts_by_hostname_or_id(%s)", hostname)
+    filter_list = [Host.display_name.comparator.contains(hostname),
+                   Host.canonical_facts['fqdn'].astext.contains(hostname), ]
+
+    try:
+        uuid.UUID(hostname)
+        host_id = hostname
+        filter_list.append(Host.id == host_id)
+        logger.debug("Adding id (uuid) to the filter list")
+    except Exception as e:
```
Glutexo: 😥

test_api.py
```diff
@@ -578,7 +589,60 @@ def test_query_using_display_name_substring(self):
...
+    def test_query_using_id(self):
+        host_list = self.added_hosts
+
+        self._base_query_test(host_list[0].id, 1)
+
+    def test_query_using_non_existent_hostname(self):
+        self._base_query_test("NotGonnaFindMe", 0)
+
+    def test_query_using_non_existent_id(self):
+        self._base_query_test(str(uuid.uuid4()), 0)
```
Glutexo: Since UUID4 is truly random, it’s not guaranteed that it will be different from the existing ones. Only extremely improbable.

PR #120, opened 2019-02-26
Add external_id canonical fact

Ladas:
source_ref is a host’s reference in the external source e.g.
AWS EC2, Azure, OpenStack, etc.

For AWS EC2, source_ref maps to 'instance-id' attribute
(e.g. i-1234567890abcdef0), which can be discovered by insights
client by e.g. using:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html

Topological Inventory '/vms' API endpoint has a source_ref
attribute that maps to this value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ The canonical facts are:
 * ip_addresses
 * fqdn
 * mac_addresses
-* source_ref
+* external_id
 
 Hosts are added and updated by sending a POST to the /hosts endpoint.
 (See the API Documentation for more details on the POST method).
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ The canonical facts are:
 * ip_addresses
 * fqdn
 * mac_addresses
+* source_ref
 
 Hosts are added and updated by sending a POST to the /hosts endpoint.
 (See the API Documentation for more details on the POST method).
--- a/app/models.py
+++ b/app/models.py
@@ -20,7 +20,7 @@ CANONICAL_FACTS = (
     "ip_addresses",
     "fqdn",
     "mac_addresses",
-    "source_ref",
+    "external_id",
 )
 
 
--- a/app/models.py
+++ b/app/models.py
@@ -20,6 +20,7 @@ CANONICAL_FACTS = (
     "ip_addresses",
     "fqdn",
     "mac_addresses",
+    "source_ref",
 )
 
 
--- a/app/utils.py
+++ b/app/utils.py
@@ -79,6 +79,14 @@ class HostWrapper:
         self.__data["mac_addresses"] = cf
 
     @property
+    def source_ref(self):
+        return self.__data.get("source_ref")
+
+    @source_ref.setter
+    def source_ref(self, cf):
+        self.__data["source_ref"] = cf
+
+    @property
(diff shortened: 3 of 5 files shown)


Ladas:
@dehort @Glutexo can you review please? cc @agrare @Fryguy

The first implementation is sending just the `source_ref` and `mac_addresses` canonical facts (and also display_name), since we consider these the most stable. And we will just do the initial create that will do the link with Topological Inventory and we won't update the hosts in Host Inventory further.

Once you guys allow us to update using the primary key, we can also add other canonical facts (namely ip_addresses and fqdn), but we would set these with the update query, not using them for matching the hosts. And we'd add updates based on changes (like adding new interface), so the inventories are always in sync. Then we should talk about what other facts we should set if any.

Glutexo:
Although the code works fine, I have concerns about storing this source_ref just as an external ID. Semantically we need a **source type** too. We can expect that different services have different formats of the ID, but that’s not guaranteed. What if there were a host in a different cloud with the same ID?

Thoughts? @Ladas @Fryguy

Ladas:
@Glutexo it should work as the other canonical facts, right? So we will send it with `mac_addresses`, so the combination should be unique enough. On our side we can still have duplicates, e.g. when you add 2 creds as 2 sources with overlapping VMs, then the Host Inventory will dedup it for us.

The same issue would be if you'd send just `ip_addresses` canonical fact, then you can have many VMs having e.g. `10.142.0.2`. So you need to combine with other canonical facts.

That being said, we might encode more info into the source_ref, e.g. I think Azure's default format of source ref looks like  `<subscription_id>/<region_name>/<vm_id>` which should be really unique. Some providers have UUID as the source_ref.

Not sure if `source_type` can be a canonical fact? It has low cardinality (basically few names of a few providers). Most of the canonical facts now have high cardinality and are somewhat unique (except private ip addresses)

Glutexo:
If you have a host with a _source_ref_ and a _mac_address_, then when doing an update, both values must match. Otherwise a new host is created.

I like the idea of having more information in the field. Like that, we can go even a bit further and prepend it with a cloud provider name. Like `"azure/<subscription_id>/<region_name>/<vm_id>"` or `"aws/…/…/…`. It is denormalized, but the Host Inventory doesn’t care about the meaning, only about the matching. So that’s no problem and adding as many data as possible to identify the source seems like a good way to me.

Side note: The tags were intended to be stored the same way. `namespace/key:value`

dehort:
I think the code looks fine.  I would like to rename the field to "external_id" though.  I think that is a bit more generic and provides a bit more information as to meaning of the field.

dehort:
@Ladas please change the "source_ref" field to "external_id" and I'll get this merged in.  Thanks!

Ladas:
@dehort done

Ladas:
@Glutexo I think for now we want to `external_id` to match `source_ref` in the Topology Inventory. We might modify the source_ref format there and since we want to use primary_key for updating the Host Inventory, the change would be reflected here.

We'll need to discuss with @agrare and @Fryguy if it'll make sense to prepend also the `source_type`, I am not sure right now. :-)

PR #121, opened 2019-03-01
Set the display_name to either the fqdn or id if the display_name is an empty string

dehort:
Adding more tests around resolving the display_name

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -63,10 +63,16 @@ def convert_dict_to_json_facts(fact_dict):
 
 
 def _set_display_name_on_save(context):
+    """
+    This method sets the display_name if it has not been set previously.
+    This logic happens during the saving of the host record so that
+    the id exists and can be used as the display_name if necessary.
+    """
     if not context.get_current_parameters()['display_name']:
-        try:
-            return context.get_current_parameters()["canonical_facts"]["fqdn"]
-        except Exception as e:
+        fqdn = context.get_current_parameters()["canonical_facts"].get("fqdn", None)
+        if fqdn:
+            return fqdn
+        else:
             return context.get_current_parameters()['id']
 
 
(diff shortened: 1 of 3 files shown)


app/models.py
```diff
@@ -63,10 +63,16 @@ def convert_dict_to_json_facts(fact_dict):
...
+    """
+    This method sets the display_name if it has not been set previously.
+    This logic happens during the saving of the host record so that
+    the id exists and can be used as the display_name if necessary.
+    """
     if not context.get_current_parameters()['display_name']:
-        try:
-            return context.get_current_parameters()["canonical_facts"]["fqdn"]
-        except Exception as e:
+        fqdn = context.get_current_parameters()["canonical_facts"].get("fqdn", None)
```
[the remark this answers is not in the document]
dehort: Sounds like a good idea to me.  If you create a PR, I'll get it merged in.

test_db_model.py
```diff
@@ -35,14 +35,27 @@ def app():
...
+    created_host = _create_host(fqdn=fqdn, display_name="")
+    assert created_host.display_name == fqdn
+
+
+def test_create_host_with_display_name_and_fqdn_as_empty_str(app):
+    # Verify that the display_name is populated from the id
+    created_host = _create_host(fqdn="", display_name="")
+    assert created_host.display_name == str(created_host.id)
+
+
```
Glutexo: There is no test that would check that a present _display_name_ is not overwritten. Namely testing this line: https://github.com/RedHatInsights/insights-host-inventory/pull/118/files#diff-413e9aa8961126d20e2d90befa13e0f1R66

Glutexo: There are no tests for the case that the _display_name_ is already there. It should not be overwritten.

PR #122, opened 2019-03-05
Simplify display_name default

Glutexo:
Simplified the function determining the default value for the _display_name_ field. It’s now more concise.

Arose from an earlier [discussion](https://github.com/RedHatInsights/insights-host-inventory/pull/121#discussion_r262115538). The `context.get_current_parameters()` call is now no longer repeated. Also used _or_ instead of _if/else_ statement to improve expressiveness.


dehort:
Looks good.

PR #125, opened 2019-03-08
Openapi 3 conversion

dehort:
Converted the swagger 2 interface doc to openapi 3.

Removed the more manual approach to x-rh-identity header authentication.

https://projects.engineering.redhat.com/browse/RHCLOUD-146

PR #126, opened 2019-03-09
Openapi 3 conversion remove tags from api

dehort:
This PR depends on PR#125.  This removes the tag cruft from the api.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,21 +82,6 @@ _prometheus_multiproc_dir_ environment variable. This is done automatically.
 python run_gunicorn.py 
 ```
 
-#### Disable authentication checks
-
-It is also possible to disable authentication (mainly useful for developement).
-This is accomplished by setting the FLASK_DEBUG=1 and NOAUTH=1 environment 
-variables.
-
-```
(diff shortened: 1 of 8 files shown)


Glutexo:
I don’t necessarily think that this needs to depend on #125. At least the [_api.host_](https://github.com/RedHatInsights/insights-host-inventory/pull/126/files#diff-a5c55b46cc9c536070cb56798c97adfaL119) changes can be standalone. I believe though that even the removal of the tag related parameters from the OpenAPI [specification](
https://github.com/RedHatInsights/insights-host-inventory/pull/126/files#diff-b5704d15302288389d3aef6d5ab2a4f3) could be done without any bigger overhead. The V2 vs. the V3 formats are not that different and a removal is quite simple operation. Yes, there will be conflicts, but they will be easily resolvable – just remove the same stuff from a similar place.

In this state, this Pull Requests is much harder reviewable. Both the changes from #125 and from here are intermingled in the [_Files changed_](https://github.com/RedHatInsights/insights-host-inventory/pull/126/files) tab. Even picking the related commits is not quite a piece of cake because of the merge commit in the tree – see the [log](https://github.com/dehort/insights-host-inventory/commits/openapi_3_conversion_remove_tags_from_api).

If I can humbly ask, don’t be afraid of merge conflicts and of rebasing. They add little work on one side and save a lot of trouble on the other side. Thanks for considering. 🙇🏻‍♂️

PR #127, opened 2019-03-09
Openapi 3 conversion bulk insert

dehort:
This PR changes the interface so that hosts can be created/updated in bulk.  This is a change to the POST /hosts method.

This PR depends on PR#126.

https://projects.engineering.redhat.com/browse/RHIPLAT1-826

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,21 +82,6 @@ _prometheus_multiproc_dir_ environment variable. This is done automatically.
 python run_gunicorn.py 
 ```
 
-#### Disable authentication checks
-
-It is also possible to disable authentication (mainly useful for developement).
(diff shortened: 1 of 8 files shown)


Glutexo:
My [concerns](
https://github.com/RedHatInsights/insights-host-inventory/pull/126#issuecomment-471506180) about dependency on another Pull Request applies to this one too. I can to some extent see the motivation on depending on the OpenAPI upgrade (#125), but there is no need to include #126 too. Even with the V3 PR (#125) I think that the changes to the [specification file](
https://github.com/RedHatInsights/insights-host-inventory/pull/127/files#diff-b5704d15302288389d3aef6d5ab2a4f3) would be easily replicable for the upgraded version, even though it would require some conflict resolving.

PR #128, opened 2019-03-09
Openapi 3 conversion bulk insert preshared key authn

dehort:
This PR contains only the authentication related changes related to the bulk insert.

This PR adds onto PR#127.

https://projects.engineering.redhat.com/browse/RHIPLAT1-825

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,21 +82,6 @@ _prometheus_multiproc_dir_ environment variable. This is done automatically.
 python run_gunicorn.py 
 ```
 
-#### Disable authentication checks
-
-It is also possible to disable authentication (mainly useful for developement).
-This is accomplished by setting the FLASK_DEBUG=1 and NOAUTH=1 environment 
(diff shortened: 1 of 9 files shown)


Glutexo:
I believe that the token authentication needs not to depend on the tag removal (#126), or the bulk insert (#127), even though it’s the latter that will be using it. I would even say that it doesn’t have to depend on the OpenAPI specification upgrade (#125).

If I am not mistaken, the [OpenAPI V2](https://swagger.io/specification/v2/) supports the [Security definitions](https://swagger.io/specification/v2/#securityDefinitionsObject) in the same (or at least very similar) way as [V3](https://swagger.io/specification/). So it should be possible to make this change independent on #125 too, even though it would create some merge conflicts.

There is some overlap with the removal of the option to disable authentication (included in #125). I wouldn’t mind seeing those in a single PR, but they can be separate as well.

PR #139, opened 2019-03-14
DRY getting the query results

Glutexo:
It’s not necessary to extract the total number of records and the records itself in each find function. Modified those to return the result set itself. The code is now more concise and expressive.


dehort:
Much better!  Thanks!

PR #144, opened 2019-03-15
Feedback on system profile facts API

beav:


beav:
@dehort I think this covers everything so far. I left the packages as strings since its easy to parse the NEVRA back out from the string and the string is more compact to store.

A couple of the arrays are really sets, I suppose we could sort them before saving to avoid updates that simply re-order the arrays but I suspect that won't happen often.

dehort:
Sounds good.  We can tweak it as we go.

PR #145, opened 2019-03-18
DRY pagination and logging from find methods

Glutexo:
Extracted the _.paginate_ calls as well as the _logger.debug_ calls from the specific host find methods. They were all the same, while the methods are called from only a single place. [Moved](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:dry_get?expand=1#diff-a5c55b46cc9c536070cb56798c97adfaR167) these calls to the [_get_host_list_](https://github.com/RedHatInsights/insights-host-inventory/blob/581d9fea1cd40de0c88072fb9b131e74c659b932/api/host.py#L145) operation method.

Not only the code is more concise, but also saves two parameters from four methods.

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -146,23 +146,26 @@ def get_host_list(display_name=None, fqdn=None,
         hostname_or_id=None, insights_id=None,
         page=1, per_page=100):
     if fqdn:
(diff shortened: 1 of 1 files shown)


dehort:
Looks good.  Merged.

PR #153, opened 2019-03-21
Add SQLAlchemy logging configuration

Glutexo:
Added SQLAlchemy’s default logging [configuration](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:sqlalchemy_logging?expand=1#diff-b289960fa46262613513c74bcd6b9471R26) to the [_logconfig.ini_](
https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:sqlalchemy_logging?expand=1#diff-b289960fa46262613513c74bcd6b9471). With that it’s easy to change its loglevel to INFO to see the queries or to DEBUG to see even the results.

See [the docs](https://docs.sqlalchemy.org/en/latest/core/engines.html?highlight=log%20level#configuring-logging) for reference.

### DIFF
--- a/logconfig.ini
+++ b/logconfig.ini
@@ -1,5 +1,5 @@
 [loggers]
-keys=root, gunicorn.error, gunicorn.access, app, api
+keys=root, gunicorn.error, gunicorn.access, sqlalchemy.engine, app, api
 
 [handlers]
 keys=logstash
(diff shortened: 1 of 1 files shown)


dehort:
Good idea!

PR #172, opened 2019-04-03
Change pagination to LIMIT/OFFSET

Glutexo:
Changed the pagination parameters from _[page](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:limit_offset?expand=1#diff-b5704d15302288389d3aef6d5ab2a4f3L254)/[per_page](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:limit_offset?expand=1#diff-b5704d15302288389d3aef6d5ab2a4f3L263)_ to standard _[limit](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:limit_offset?expand=1#diff-b5704d15302288389d3aef6d5ab2a4f3R254)/[offset](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:limit_offset?expand=1#diff-b5704d15302288389d3aef6d5ab2a4f3R264)_. The _offset_ starts with 0 and that’s also the default. The _limit_ is capped at 100 and its default is 50, just as with the former _per_page_ parameter. If the offset is out of bound, error _404 Not Found_ is [raised](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:limit_offset?expand=1#diff-a5c55b46cc9c536070cb56798c97adfaR195).

This pull request doesn’t aim to completely implement the complete standard pagination as described in the Cloud Services Platform [documentation](https://platform-docs.cloud.paas.upshift.redhat.com/rest/pagination.html). It’s rather the first step towards it.

I generally only transformed the existing pagination to the new one, without any major changes to the architecture. As a result,

* [x] There are some untested calls. [\[1\]](https://github.com/RedHatInsights/insights-host-inventory/blob/5f1a55e9c103de462140b5b8aa9ef9a7c74c409c/test_api.py#L1014) [\[2\]](https://github.com/RedHatInsights/insights-host-inventory/blob/5f1a55e9c103de462140b5b8aa9ef9a7c74c409c/test_api.py#L933) Those were already untested before my changes. Resolved by #175.
* [ ] The [_api.host_](https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/host.py) module would utilize a bit refactoring to DRY the code building the response.

I‘d do both right away as new, independent PRs. The tests are definitely more important. These are not needed for the ticket to be closed.

Solves [RHCLOUD-344](https://projects.engineering.redhat.com/browse/RHCLOUD-344).

Steps to reproduce:

[...]

PR #175, opened 2019-04-05
Test query results

Glutexo:
Resolved the FIXMEs in the tests: added checks for the query results.

* When querying hosts themselves.
* When querying hosts system profiles.

The multiple hosts with system profiles test checks the pagination too.

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -720,7 +720,35 @@ class BulkCreateHostsTestCase(DBAPITestCase):
                 i += 1
 
 
-class CreateHostsWithSystemProfileTestCase(DBAPITestCase):
+class PaginationTestCase(BaseAPITestCase):
+    def _base_paging_test(self, url, expected_number_of_hosts):
+        def _test_get_page(page, expected_count=1):
+            test_url = inject_qs(url, page=page, per_page="1")
+            response = self.get(test_url, 200)
+
+            self.assertEqual(len(response["results"]), expected_count)
+            self.assertEqual(response["count"], expected_count)
+            self.assertEqual(response["total"], expected_number_of_hosts)
+
+        if expected_number_of_hosts == 0:
+            _test_get_page(1, expected_count=0)
+            return
+
+        i = 0
+
+        # Iterate through the pages
+        for i in range(1, expected_number_of_hosts + 1):
+            with self.subTest(pagination_test=i):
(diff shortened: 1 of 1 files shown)


Glutexo:
Made a minor fixup:

* Added one more assert to check the system profile query results.
* Removed unrelated style changes.

Glutexo:
This is baaad! The [_HostWrapper.data_](https://github.com/RedHatInsights/insights-host-inventory/blob/2b57f9a0e7b3f7b0b8b59a7f8c1575985945c3f9/app/utils.py#L9) method returns a reference to the original dictionary of the object. That means that [replacing](https://github.com/RedHatInsights/insights-host-inventory/blob/master/test_api.py#L1026) the _"id"_ key in the original object ([_added_hosts_list\[n\]_](https://github.com/RedHatInsights/insights-host-inventory/blob/2b57f9a0e7b3f7b0b8b59a7f8c1575985945c3f9/test_api.py#L1020)) manifests in the extracted dictionary ([_expected_host_list_](https://github.com/RedHatInsights/insights-host-inventory/blob/master/test_api.py#L1021)). I never liked the [_HostWrapper_](https://github.com/RedHatInsights/insights-host-inventory/blob/2b57f9a0e7b3f7b0b8b59a7f8c1575985945c3f9/app/utils.py#L4) class, but this is super-cryptic.

It‘s clearly my fault that I missed the failing tests, but seeing the cause, I am not that much surprised. Will fix. Thanks for pointing out!

Glutexo:
Mutable objects are pure evil. I leveraged [your use](https://github.com/RedHatInsights/insights-host-inventory/pull/175/files#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2R1050) of _copy.deepcopy_ and used it once again to [populate](https://github.com/RedHatInsights/insights-host-inventory/pull/175/files#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2R1051) the _expected_host_list_ with new dictionaries instead of just references to the original one. This saves the juggling with the values, putting them back and forth, silently propagating among reference objects.

Tests are now 🍏!

Glutexo:
This relates to #172. I promised query results checking tests, so we know that the pagination didn’t break. Some rebasing may be needed after merging one of these PRs.

PR #177, opened 2019-04-05
do not validate URIs on baseurl

beav:
remove URI validation for baseurl

note: this PR is untested! I do not have a full inventory svc dev environment, but I think this is all that is needed.


beav:
closed in favor of https://github.com/RedHatInsights/insights-host-inventory/pull/198

PR #180, opened 2019-04-09
Add indexes on db columns

dehort:
Add index on the account column
Add index on the insights_id field within the canonical_facts jsonb column

### DIFF
--- a/migrations/versions/ed001bfc9c22_add_index_on_canonical_facts_insights_id.py
+++ b/migrations/versions/ed001bfc9c22_add_index_on_canonical_facts_insights_id.py
@@ -1,0 +1,31 @@
+"""empty message
+
+Revision ID: ed001bfc9c22
+Revises: 328011606777
+Create Date: 2019-04-09 16:06:14.667298
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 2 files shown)


Glutexo:
Related to #168.

Glutexo:
Because of https://github.com/RedHatInsights/insights-host-inventory/commit/3949538e00f219a051ccc86e7954b629f3bfb219, merging this will cause Alembic to have more heads. It’s necessary either to manually rebase the new migrations so they form a straight history, or to create a merge migrations by run-in `python manage.py db merge heads`.

migrations/versions/ed001bfc9c22_add_index_on_canonical_facts_insights_id.py
```diff
@@ -0,0 +1,31 @@
...
+table_name = "hosts"
+index_name = "idx_cf_insights_id"
+
+
+def upgrade():
+    op.create_index(index_name,
+                    table_name,
+                    [sa.text(("""(canonical_facts ->> 'insights_id')"""))],
+                    )
+                    #postgresql_using='gin')
```
Glutexo: Briefly – use GIN when adding index on a whole composite field (btree, hstore, json(b) etc.) and not when indexing a plain scalar value. Here, you’re creating index on a function result, which is a TEXT, so no GIN. 🥃

PR #182, opened 2019-04-10
Use limit/offset pagination in /hosts

Glutexo:
Changed the pagination parameters from _page/per_page_ to standard _limit/offset_. The _offset_ starts with 0 and that’s also the default. The _limit_ is capped at 100 and its default is 50, just as with the former _per_page_ parameter. If the offset is out of bound, error _404 Not Found_ is raised.

This pull request doesn’t aim to completely implement the complete standard pagination as described in the Cloud Services Platform [documentation](https://platform-docs.cloud.paas.upshift.redhat.com/rest/pagination.html). It’s rather the first step towards it.

I generally only transformed the existing pagination to the new one, without any major changes to the architecture. As a result,

* There are some untested calls. [\[1\]](https://github.com/RedHatInsights/insights-host-inventory/blob/5f1a55e9c103de462140b5b8aa9ef9a7c74c409c/test_api.py#L1014) [\[2\]](https://github.com/RedHatInsights/insights-host-inventory/blob/5f1a55e9c103de462140b5b8aa9ef9a7c74c409c/test_api.py#L933) Those were already untested before my changes.
* The [_api.host_](https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/host.py) module would utilize a bit refactoring to DRY the code building the response.

I‘d do both right away as new, independent PRs. The tests are definitely more important. These are not needed for the ticket to be closed.

Solves [RHCLOUD-344](https://projects.engineering.redhat.com/browse/RHCLOUD-344).

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -3,6 +3,8 @@ import sqlalchemy
 import uuid
 
 from enum import Enum
+from flask import abort
+from flask_api.status import HTTP_404_NOT_FOUND
 from marshmallow import ValidationError
 
 from app import db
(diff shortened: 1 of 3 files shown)


Glutexo:
Not necessary to open this one.

PR #201, opened 2019-04-15
Adding account_number to logstash log message

Jason-RH:
Adding the account_number to the logstash message (similar to request_id). Account number is defaulted when using token auth.


Jason-RH:
@dehort Can't add reviewers because of permissions, but let me know if there are any problems here.

dehort:
I had to rework this a bit as we changed the way the inventory logging is working (a couple of times now).  I merged this as part of pr #226.

PR #202, opened 2019-04-15
Adding support for legacy api.

Jason-RH:
When the INVENTORY_LEGACY_API_URL environment variable is set to something other than an empty string then that api path will be added.


Jason-RH:
@dehort Refactored into a for loop and added logging. Let me know if that logging isn't sufficient.

fijshion:
Tested this locally and the code looks good. IMO it's ready to merge. However, I don't have permissions to review or merge. @dehort could you click the button?

PR #205, opened 2019-04-16
Add `cloud_provider` to system profile

beav:
This commit adds a new cloud provider string field to the system
profile.


dehort:
Looks good.

PR #216, opened 2019-04-19
Allow negative offset

Glutexo:
When using _LIMIT/OFFSET_ pagination together with **consistent** first/next/previous/last links, it’s necessary to allow negative offsets. Otherwise, it wouldn’t be possible to keep the _LIMIT_ value across the pages.

Added support for negative _OFFSET_, converting the API _LIMIT/OFFSET_ values to the PostgreSQL ones, keeping the pagination cursor consistent. Any offset that would return no records yields _404 Not Found_. In case of having no records at all, 0 is the only allowed _OFFSET_.

Please see a [Gist](https://gist.github.com/Glutexo/54efb5601b837f533953a5b2c390bfd3#file-pagination-example-md) with examples that illustrate this graphically. I put quite some thought into this, consulted with co-workers and concluded that this is the only way how to do it without arbitrary decisions that would only lead to inconsistencies.

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -191,11 +191,18 @@ def get_host_list(display_name=None, fqdn=None,
 
 def _paginate_host_list_query(query, limit, offset):
     total = query.count()
+
     max_offset = max(0, total - 1)  # Zero offset is always allowed
-    if offset > max_offset:
+    min_offset = -max_offset
+    if offset > max_offset or offset < min_offset:
         raise IndexError
 
-    query = query.order_by(Host.created_on, Host.id).limit(limit).offset(offset)
+    db_offset = max(offset, 0)
+    db_limit = limit if offset >= 0 else limit + offset
+    if not db_limit:
+        raise IndexError
+
+    query = query.order_by(Host.created_on, Host.id).limit(db_limit).offset(db_offset)
     query_results = query.all()
     logger.debug(f"Found hosts: {query_results}")
     return total, query_results
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -267,7 +267,6 @@ components:
       required: false
       schema:
         type: integer
-        minimum: 0
         default: 0
   schemas:
     BulkHostOut:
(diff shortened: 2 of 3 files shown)


fijshion:
Interesting approach to solve this. It took me a little bit to understand why this is necessary which makes me a little concerned it is not intuitive and will lead to confusion (might just be me though). Another approach would be to deny requests with offsets that are not a multiple of the limit. By deny I mean return 400 with a descriptive message. Curious if you considered this approach and if so why you didn't go that way?

Glutexo:
Thanks for your comment, @fijshion! I considered the suggested approach only briefly. The main reason why I didn’t take that into account was that in that case, there would be no advantage of having the _limit/offset_ pagination at all. It would essentially be equivalent to the _page/per_page_. And I suppose that we didn't switch to this more flexible system if we were to support only in the old rigid way.

I agree that it is difficult to wrap your head around the problem. I’m actually not sure what are the use cases of those links and why do we need them in the first place. Some less consistent solutions may be appropriate for various uses: UI pagination, aggregated export, … But as these links are generic property of the API, I’d expect them to be predictable and thus consistent.

Glutexo:
Adding that the Cost Management/Koku example in the guidelines implies that non-multiplier offsets are allowed:

```json
{
  "meta": {
    "count": 11
  },

  "links": {
    "first": "/r/insights/platform/myapp/v1/object/?limit=10&offset=0",
    "next": "/r/insights/platform/myapp/v1/object/?limit=10&offset=10",
    "previous": null,
    "last": "/r/insights/platform/myapp/v1/object/?limit=10&offset=1"
  },

  "data": [
    { }
  ]
}
```

See the _last_ key.

fijshion:
Thanks, that makes sense. Last night I realized that restricting the limit/offsets like I said would essentially be page/page_size like you said. I guess I'm just not sure what the use case is for arbitrary limit/offsets. That's a bigger question though, I'm cool with going this route.

Could you add a more descriptive message to the 404 response? Something like "No resources found with provided limit and offset". Also maybe add a sentence or two to the API doc to describe negative offsets?

Glutexo:
Created a disturbingly descriptive Spike targeting the pagination problem: RHCLOUD-1064. I’ll take a look what I can do with the error message.

Glutexo:
Created #240 incorporating your suggestion. It should cleanly merge with this. If not, I’ll rebase.

Glutexo:
Merge, @dehort?

PR #219, opened 2019-04-22
Adding logic to handle cert auth on production

Jason-RH:
1. Added environment variables to configure ssl usage - configured automatically when on prod
2. Added logging for cert auth

@dehort Check out https://github.com/RedHatInsights/e2e-deploy/pull/252/ to see where these new variables are being set.


Jason-RH:
@fijshion Here's the inventory PR related to https://github.com/RedHatInsights/e2e-deploy/pull/252/

app/config.py
```diff
@@ -13,8 +13,13 @@ def __init__(self, config_name):
         self._db_password = os.getenv("INVENTORY_DB_PASS", "insights")
         self._db_host = os.getenv("INVENTORY_DB_HOST", "localhost")
         self._db_name = os.getenv("INVENTORY_DB_NAME", "insights")
+        self._db_sslmode = os.getenv("INVENTORY_SSL_MODE", "")
+        self._db_cert = os.getenv("INVENTORY_CERT", "")
 
-        self.db_uri = f"postgresql://{self._db_user}:{self._db_password}@{self._db_host}/{self._db_name}"
+        if self._db_sslmode:
```
fijshion: I think you want `if self.__db_sslmode == verify-full` here. Otherwise any value in INVENTORY_SSL_MODE is considered valid.

app/config.py
```diff
@@ -44,3 +49,8 @@ def _log_configuration(self):
             self.logger.info("Management URL Path Prefix: %s" % self.mgmt_url_path_prefix)
             self.logger.info("DB Host: %s" % self._db_host)
             self.logger.info("DB Name: %s" % self._db_name)
+            self.logger.info("DB Connection URI: %s" % self.db_uri)
+        if self._db_sslmode:
```
fijshion: Same as above, `if self.__db_sslmode == verify-full`

app/config.py
Jason-RH: Yeah I definitely should be more restrictive here. Do you think there’s a use case for other Postgres ssl methods or is having two cases (verify-full or none) enough?

fijshion: I can't think of a use case other than verify-full/none.

fijshion:
Looks good. @dehort will need to click the button after the conflict is resolved.

app/config.py
```diff
@@ -13,8 +13,13 @@ def __init__(self, config_name):
         self._db_password = os.getenv("INVENTORY_DB_PASS", "insights")
         self._db_host = os.getenv("INVENTORY_DB_HOST", "localhost")
         self._db_name = os.getenv("INVENTORY_DB_NAME", "insights")
+        self._db_sslmode = os.getenv("INVENTORY_SSL_MODE", "")
+        self._db_cert = os.getenv("INVENTORY_CERT", "")
 
-        self.db_uri = f"postgresql://{self._db_user}:{self._db_password}@{self._db_host}/{self._db_name}"
+        if self._db_sslmode == "verify-full":
```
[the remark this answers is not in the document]
Jason-RH: Agreed. Would it be useful to have some other ssl_mode values as constants too? Or is verify-full the only ssl_mode we need?

Jason-RH:
@dehort I believe all requested changes have been made. Please have another look :smile:

app/config.py
```diff
@@ -5,18 +5,23 @@
...
+
     def __init__(self):
         self.logger = get_logger(__name__)
 
         self._db_user = os.getenv("INVENTORY_DB_USER", "insights")
         self._db_password = os.getenv("INVENTORY_DB_PASS", "insights")
         self._db_host = os.getenv("INVENTORY_DB_HOST", "localhost")
         self._db_name = os.getenv("INVENTORY_DB_NAME", "insights")
+        self._db_ssl_mode = os.getenv("INVENTORY_DB_SSL_MODE", "")
+        self._db_cert = os.getenv("INVENTORY_DB_CERT", "")
```
dehort: Can you change INVENTORY_DB_CERT to INVENTORY_DB_SSL_CERT just to match the other env var?

app/config.py
```diff
@@ -50,3 +61,8 @@ def log_configuration(self, config_name):
             self.logger.info("Management URL Path Prefix: %s" % self.mgmt_url_path_prefix)
             self.logger.info("DB Host: %s" % self._db_host)
             self.logger.info("DB Name: %s" % self._db_name)
+            self.logger.info("DB Connection URI: %s" % self.db_uri)
```
dehort: Can you change this to where it calls the _build_db_uri() method and modify the method to where when called from here it prints out XXXX for the password?

I think seeing the db uri that is getting used would be good, but we do not want to log the db password.

app/config.py
```diff
@@ -47,10 +47,12 @@ def _build_api_path(self):
         api_path = f"{base_url_path}/{version}"
         return api_path
 
-    def _build_db_uri(self, ssl_mode):
+    def _build_db_uri(self, ssl_mode, hide_password=False):
```
catastrophe-brandon: I'm just trying to better acquaint myself with some of the service code. I noticed that we're defaulting to `False` here. Is this to facilitate debugging? Normally I would expect passwords to be hidden by default and some action to be taken to override the behavior.

Jason-RH: @catastrophe-brandon Yeah we're hiding the password for the log message that gets printed out here: https://github.com/RedHatInsights/insights-host-inventory/pull/219/files/e2a9ac71736665379e89f298657697103daca627#diff-4b1fc3357c304348032d92a3c42978d0R66

By default that uri needs the actual password so it can connect to postgres. When logging the password is just replaced by X's

PR #233, opened 2019-04-24
Search endpoint

fijshion:


fijshion:
@Glutexo ready for a review.

dehort:
@Glutexo has this been reviewed?

Glutexo:
Started reviewing today, but left it for maintenance of the pagination pull request. Will have to do it on Monday.

Glutexo:
From a quick glance it looks fine, but wanted to thoroughly test it.

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+
+    def test_query_with_invalid_host_id_list(self):
+        invalid_query_docs = [None,
+                              {},
+                              {"host_id_list": []},
+                              {"host_id_list": {}}
+                              ]
+
+        for invalid_query_doc in invalid_query_docs:
+            with self.subTest(invalid_query_doc=invalid_query_doc):
```
Glutexo: Nice usage of subTests for parametrization! 👍🏻

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+                          invalid_query_doc, 400)
+
+    def test_query_with_exclude_facts(self):
+        host_list = self.added_hosts
+        host_id_list = [host.id for host in host_list]
+        query_doc = {"host_id_list": host_id_list, "exclude_fields": ["facts"]}
+        response = self.post(SEARCH_URL, query_doc, 200)
+
+        for host in response["results"]:
+            assert host["id"] in host_id_list
```
```suggestion
            self.assertIn(host["id"], host_id_list)
```

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+            assert "facts" not in host
```
```suggestion
            self.assertIn("facts", host)
```

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+        self._post_paging_test(SEARCH_URL, query_doc, len(self.added_hosts))
+
+    def test_query_with_exclude_system_profile(self):
+        host_list = self.added_hosts
+        host_id_list = [host.id for host in host_list]
+        query_doc = {"host_id_list": host_id_list, "exclude_fields": ["system_profile_facts"]}
+        response = self.post(SEARCH_URL, query_doc, 200)
+
+        for host in response["results"]:
+            assert host["id"] in host_id_list
```
```suggestion
            self.assertIn(host["id"], host_id_list)
```

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+            assert "facts" in host
```
```suggestion
            self.assertIn("facts", host)
```

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+            assert "system_profile_facts" not in host
```
```suggestion
            self.assertNotIn("system_profile_facts", host)
```

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+        query_doc = {"host_id_list": host_id_list, "exclude_fields": ["facts"]}
+        response = self.post(SEARCH_URL, query_doc, 200)
+
+        for host in response["results"]:
+            assert host["id"] in host_id_list
+            assert "facts" not in host
+
+        self._post_paging_test(SEARCH_URL, query_doc, len(self.added_hosts))
+
+    def test_query_with_exclude_system_profile(self):
```
Glutexo: Although this test essentially works, the _system_profile_facts_ is not present in the result, it doesn’t verify any part of code. The _system_profile_facts_ is **never** part of the response and if you change the _excluded_fields_ list to `[]`, the test still passes.

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+        response = self.post(SEARCH_URL, query_doc, 200)
+
+        for host in response["results"]:
+            assert host["id"] in host_id_list
+            assert "facts" in host
+            assert "system_profile_facts" not in host
+
+        self._post_paging_test(SEARCH_URL, query_doc, len(self.added_hosts))
+
+    def test_query_with_exclude_facts_and_system_profile(self):
```
Glutexo: The same applies to this test too. The _system_profile_facts_ field is never present in the output, regardless of the _excluded_fields_ value.

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+        host_list = self.added_hosts
+        host_id_list = [host.id for host in host_list]
+        query_doc = {"host_id_list": host_id_list, "exclude_fields": ["facts", "system_profile_facts"]}
+        response = self.post(SEARCH_URL, query_doc, 200)
+
+        for host in response["results"]:
+            assert host["id"] in host_id_list
```
```suggestion
            self.assertIn(host["id"], host_id_list)
```

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+            assert "facts" not in host
```
```suggestion
            self.assertIn("facts", host)
```

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+            assert "system_profile_facts" not in host
```
```suggestion
            self.assertNotIn("system_profile_facts", host)
```

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+
+        self._post_paging_test(SEARCH_URL, query_doc, len(self.added_hosts))
+
+    def test_query_with_invalid_exclude(self):
+        host_list = self.added_hosts
+        host_id_list = [host.id for host in host_list]
+        query_doc = {"host_id_list": host_id_list, "exclude_fields": ["invalid"]}
+        response = self.post(SEARCH_URL, query_doc, 400)
+        assert "Invalid exclude_fields. Valid values are: ['facts', 'system_profile_facts']" in response["detail"]["exclude_fields"][0]
```
Glutexo: This line is too long.
```suggestion
        self.assertIn("Invalid exclude_fields. Valid values are: ['facts', 'system_profile_facts']", response["detail"]["exclude_fields"][0])
```

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+            assert host["id"] in host_id_list
+            assert "facts" in host
+            assert "system_profile_facts" not in host
+
+        self._post_paging_test(SEARCH_URL, query_doc, len(self.added_hosts))
+
+    def test_query_with_exclude_facts_and_system_profile(self):
+        host_list = self.added_hosts
+        host_id_list = [host.id for host in host_list]
+        query_doc = {"host_id_list": host_id_list, "exclude_fields": ["facts", "system_profile_facts"]}
```
Glutexo: This line is too long.

test_api.py
```diff
@@ -1467,6 +1480,83 @@ def test_query_with_no_matching_insights_id(self):
...
+        for host in response["results"]:
+            assert host["id"] in host_id_list
+            assert "facts" not in host
+
+        self._post_paging_test(SEARCH_URL, query_doc, len(self.added_hosts))
+
+    def test_query_with_exclude_system_profile(self):
+        host_list = self.added_hosts
+        host_id_list = [host.id for host in host_list]
+        query_doc = {"host_id_list": host_id_list, "exclude_fields": ["system_profile_facts"]}
```
Glutexo: This line is too long.

api/search.py
```diff
@@ -0,0 +1,24 @@
...
+
+from .host import build_paginated_host_list_response, get_host_list_by_id_list
+
+logger = logging.getLogger(__name__)
+
+
+@api_operation
+@metrics.api_request_time.time()
+def post(post_body, page=1, per_page=100):
+    host_list_field_name = "host_id_list"
```
Glutexo: I don’t know… this variable looks a bit redundant to me.
```suggestion
```

api/search.py
```diff
@@ -0,0 +1,24 @@
...
+    validated_input = SearchSchema(strict=True).load(post_body)
+    exclude = validated_input.data["exclude_fields"]
+    query = get_host_list_by_id_list(current_identity.account_number,
+                                     validated_input.data[host_list_field_name],
```
Glutexo:
```suggestion
                                     validated_input.data["host_id_list"],
```

api/host.py
```diff
@@ -186,13 +186,13 @@ def get_host_list(display_name=None, fqdn=None,
...
 
-    return _build_paginated_host_list_response(
+    return build_paginated_host_list_response(
         query_results.total, page, per_page, query_results.items
     )
 
 
-def _build_paginated_host_list_response(total, page, per_page, host_list):
-    json_host_list = [host.to_json() for host in host_list]
+def build_paginated_host_list_response(total, page, per_page, host_list, exclude=[]):
```
Glutexo: Removing the leading underscore to make the function public is essentially correct, it however smears the distinction between the API operation functions and helper functions. I think it’s ok to keep it like this, since even the already present method names are not consistent. Just mentioning for the record that it’d be nice to either have all helper methods prefixed, or to extract them to a separate module.

api/search.py
```diff
@@ -0,0 +1,24 @@
+import logging
```
Glutexo: I don’t like much the fact that the search is out of the _hosts_ namespace and is not presented at _/hosts/search_. Like this, it’s logical to have this separate module, but still it shares many functions with the [_api.host_](https://github.com/RedHatInsights/insights-host-inventory/pull/233/files#diff-a5c55b46cc9c536070cb56798c97adfa) module resulting in a spaghetti code. Also it requires (although softly) renaming the shared methods, which increases the confusion rate in that module.

If the search operation will be moved to the _/hosts/search_ endpoint, then I’d move this only operation to the [_api.host_](https://github.com/RedHatInsights/insights-host-inventory/pull/233/files#diff-a5c55b46cc9c536070cb56798c97adfa) module. Refactoring that module can be then easier and done in one place.

api/host.py
```diff
@@ -247,29 +247,31 @@ def find_hosts_by_hostname_or_id(account_number, hostname):
...
+    return build_paginated_host_list_response(
         query_results.total, page, per_page, query_results.items
     )
 
 
-def _get_host_list_by_id_list(account_number, host_id_list):
-    return Host.query.filter(
-        (Host.account == account_number)
-        & Host.id.in_(host_id_list)
+def get_host_list_by_id_list(account_number, host_id_list, exclude=[]):
```
Glutexo: Again, a mutable object as a default value. Even here it doesn’t do any harm, but it’s not good practice.

app/__init__.py
```diff
@@ -12,17 +12,14 @@
 from app.logging import configure_logging, threadctx
 from app.validators import verify_uuid_format  # noqa: 401
 from tasks import start_consumer
+from marshmallow import ValidationError
+
+from .error_handlers import render_exception, validation_error_handler
```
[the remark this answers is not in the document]
fijshion: IMO creating separate PRs for such small tasks creates more overhead and will slow down development. What's the harm in a small, incremental refactor as part of another task?

test_api.py
fijshion: Since there is no linter defined on this project, what is the accepted line length?

api/host.py
fijshion: I agree these should eventually be broken out into a separate module.

fijshion:
IMO /hosts/search is not RESTful (although REST can be subjective at times). I don't like using /hosts/search because it overloads the already defined /hosts/{id} path. Am I asking for a host resource with ID search?

POST to retrieve objects has become fairly common in REST interfaces when doing a "search" operation. Elastic Search actually does a GET with a body for it's search operations. It's a hotly debated topic which has no single best answer. 

https://stackoverflow.com/questions/5020704/how-to-design-restful-search-filtering
https://softwareengineering.stackexchange.com/questions/233164/how-do-searches-fit-into-a-restful-interface

I agree needing this endpoint is not ideal. The reason we need this is because we do not yet have a good way to join data across the platform. One use case I'm aware of is that advisor has a long list of host ids it needs to retrieve details for to join with it's host specific data. The list is too large to put in a query parameter. Regardless of pagination, they only want to retrieve those specific hosts.

I'd rather not create separate PRs for this (extra overhead with minimal benefit). I think it all fits into the "search endpoint" task. I will fix the other comments in a new commit.

app/__init__.py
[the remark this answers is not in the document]
fijshion: renamed the variable

app/__init__.py
```diff
@@ -55,6 +52,7 @@ def create_app(config_name):
     # Add an error handler that will convert our top level exceptions
     # into error responses
     connexion_app.add_error_handler(InventoryException, render_exception)
+    connexion_app.add_error_handler(ValidationError, validation_error_handler)
```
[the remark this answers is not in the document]
fijshion: renamed the variable

app/error_handlers.py
```diff
@@ -0,0 +1,25 @@
+import logging
+
+from flask import jsonify
+
```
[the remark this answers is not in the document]
fijshion: removed. Although since this is purely a stylistic thing and there is no defined style for the project, I'm not sure it's relevant.

app/error_handlers.py
```diff
@@ -0,0 +1,25 @@
...
+from flask_api import status
+
+logger = logging.getLogger(__name__)
+
+
+# Handles marshmallow schema validation errors
```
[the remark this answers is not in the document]
fijshion: Renamed the function and removed the comment.

app/models.py
```diff
@@ -83,15 +83,18 @@ def from_json(cls, d):
             d.get("system_profile", {}),
         )
 
-    def to_json(self):
+    def to_json(self, exclude=[]):
```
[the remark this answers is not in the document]
fijshion: Fixed, thanks for the link.

test_api.py
fijshion: Shortened the line

api/search.py
fijshion: Removed this var

api/search.py
fijshion: Fixed

api/host.py
fijshion: Fixed

app/__init__.py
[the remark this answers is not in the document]
Glutexo: I agree that there indeed is overhead – more commits, descriptions, merge conflicts and in general more work. As a result, there are better, more atomic commits, cleaner history, sharper focus on a single problem etc. The best advantage of having those small refactors in a separate PR is better reviewability. A small PR with an incremental refactor can easily get approved and merged and doesn’t hang on another one with functionality change.

test_api.py
Glutexo: I’d stick to PEP 8 which says:

> Limit all lines to a maximum of 79 characters.
> 
> For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.

Glutexo:
I was thinking the same with _/hosts/search_; had the same issue with _/hosts/download_. However since the host ID is always a UUID, it’s at least unambiguous. Rails even [uses](https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions) _/resource/new_ as a default for new resource page. But agreed. If it’s not going to be a completely generic search, but a search only for hosts, I’d at least rename it to _host_search_ and then rename the _host_id_list_ argument to _id_list_, or even better, just _id_.

app/__init__.py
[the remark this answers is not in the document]
fijshion: I agree being able to merge refactors quickly can be nice. However, we're a small team on a small project trying to move fast. I would rather move quickly than slow ourselves down with extra process.

As for the commit log, a single PR can have multiple commits. I think we should also disable merge commits on this project as they significantly muddy the log and make rebases more cumbersome.

fijshion:
What's the disadvantage to creating a more generic /search endpoint which could be enhanced later? I'd rather start with a foundation to improve upon than create a specific endpoint for each use case. I also don't want to design a public API simply to fit into a single code file.

I think the discussion on using GET/POST is out of scope. I've been told (see RHCLOUD-144) the requirement is to use POST to retrieve a list of hosts by id. At this point, I don't think we can (or should) put more work on the consumers of inventory to retrieve host details. We will likely deprecate this /search endpoint when the real search solution is complete next release.

dehort:
I haven't had a chance to look at this in detail.  I did notice that you have to explicitly exclude the facts and system_profile fields.  Can you modify this so that they are excluded from the response by default and only included when explicitly requested?  Our main performance issues have been due to the large json document processing.

fijshion:
I can do that if you think it's necessary. I'd rather not because I think it muddies the API. 

As a user of an API, I would intuitively expect all fields to be returned unless I specify to exclude fields (or am required to explicitly specify all the fields I want). It seems odd to exclude arbitrary (at least from a users perspective) fields unless specified to include them.

Couldn't the client (only advisor in this case?) specify the exclude in their query?

dehort:
Unfortunately, I feel like its necessary at this point.  Processing large amounts of json has proven to hurt performance.  So I'd like to return as little data as possible by default and make the client explicitly ask for the larger data sets.

fijshion:
ok. Pushed a new commit switching to include_fields.

test_api.py
```diff
@@ -15,7 +15,9 @@
 from datetime import datetime, timezone
 from urllib.parse import urlsplit, urlencode, parse_qs, urlunsplit
 
-HOST_URL = "/api/inventory/v1/hosts"
+BASE_URL = "/api/inventory/v1"
+HOST_URL = BASE_URL + "/hosts"
+SEARCH_URL = BASE_URL + "/search"
```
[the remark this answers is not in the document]
Glutexo: If the endpoint is renamed, it’d have to be changed here to..

```suggestion
SEARCH_URL = BASE_URL + "/host_search"
```

test_api.py
Glutexo: The _system_profile_facts_ field is never present in the response, regardless of the _included fields_ value.

test_api.py
```diff
@@ -1488,6 +1501,71 @@ def test_query_with_no_matching_insights_id(self):
...
+        query_doc = {"host_id_list": host_id_list, "include_fields": ["facts"]}
+        response = self.post(SEARCH_URL, query_doc, 200)
+
+        for host in response["results"]:
+            assert host["id"] in host_id_list
+            assert "facts" in host
+            assert "system_profile_facts" not in host
+
+        self._post_paging_test(SEARCH_URL, query_doc, len(self.added_hosts))
+
```
Glutexo: Either test the inclusion of the _system_profile_facts_ field and see it fails, or remove the hints that this field can be included.

test_api.py
Glutexo: Now, regardless of the _inlcuded_fields_ value.

api/search.py
Glutexo: Even if just renamed to _hosts_search_, I’d still put it to the bulky [_api.host_](https://github.com/fijshion/insights-host-inventory/blob/626e62d1a3611be6d090ebd7dbd82435e6073beb/api/host.py) module. Some extraction can be done later as a separate step.

app/__init__.py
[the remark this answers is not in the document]
Glutexo: I agree that without merge commits, the history would be cleaner. Rebasing as a way to go would be good. 

With the process overhead slowing down the development I am not in 100% agreement with you, but I understand the pros and thus the motivation. I won’t block this PR because it contains minor refactorings.

test_api.py
```diff
@@ -1488,6 +1501,71 @@ def test_query_with_no_matching_insights_id(self):
...
+    def test_query_with_invalid_exclude(self):
+        host_list = self.added_hosts
+        host_id_list = [host.id for host in host_list]
+        query_doc = {"host_id_list": host_id_list, "include_fields": ["invalid"]}
+        response = self.post(SEARCH_URL, query_doc, 400)
+        assert "Invalid include_fields. Valid values are: ['facts', 'system_profile_facts']" in response["detail"]["include_fields"][0]
```
```suggestion
        self.assertIn("Invalid include_fields. Valid values are: ['facts', 'system_profile_facts']", response["detail"]["include_fields"][0])
```

test_api.py
```diff
@@ -1488,6 +1501,71 @@ def test_query_with_no_matching_insights_id(self):
         self._base_query_test(uuid_that_does_not_exist_in_db, 0)
 
 
+class SearchTests(PreCreatedHostsBaseTestCase):
```
Glutexo: If the endpoint gets renamed to _/hosts_search_, I’d rename this class to _HostsSearchTests_ for symmetry.

Glutexo:
Talked a bit with @karelhala and finally understood, where does the need for the search by many host IDs comes from. The main use case for this – having a list of hosts filtered by conditions not known to the Inventory – can’t be viably solved just by increasing the number of requestable host IDs. The applications possibly need to pass hundreds of IDs to the query and then request separate pages of the result set, grab the results in different order, use additional filters etc. I think that the only way to this is to use this POST to create a search query object in the Inventory, add host IDs to it and then use a generated ID of this object to request the hosts.

api/search.py
dehort: I don't see an issue with having a /search endpoint.

Eventually I would like to move the host related logic (including searching) out of the api.host module and into a separate module so that logic can be shared.

api/search.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@api_operation
+@metrics.api_request_time.time()
+def post(post_body, page=1, per_page=100):
+    validated_input = SearchSchema(strict=True).load(post_body)
+    include = validated_input.data["include_fields"]
+    exclude = []
+
+    if "facts" not in include:
```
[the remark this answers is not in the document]
dehort: I would like to remove the Host.to_json() methods and replace them with an export() method that simply returns a "complete" host dict.  We could then have an exporter object that is responsible for filtering out the unneeded fields.

Here is an example that shows what I was thinking:
https://github.com/dehort/insights-host-inventory/tree/add_exporter_object_to_render_host_query_output

app/models.py
```diff
@@ -372,3 +375,18 @@ def validate_mac_addresses(self, mac_address_list):
 class PatchHostSchema(Schema):
     ansible_host = fields.Str(validate=validate.Length(min=0, max=255))
     display_name = fields.Str(validate=validate.Length(min=1, max=200))
+
+
+class SearchSchema(Schema):
+    host_id_list = fields.List(fields.Str(validate=verify_uuid_format),
+                               validate=validate.Length(min=1, max=5000),
```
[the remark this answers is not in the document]
dehort: Good idea.

app/models.py
```diff
@@ -372,3 +378,18 @@ def validate_mac_addresses(self, mac_address_list):
...
+                               required=True)
+    include_fields = fields.List(fields.Str(validate=validate.Length(min=1, max=255)),
```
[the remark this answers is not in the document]
dehort: It would be nice to list out  to list out the valid values. I think an enum makes sense here.

app/models.py
```diff
@@ -372,3 +375,18 @@ def validate_mac_addresses(self, mac_address_list):
...
+    host_id_list = fields.List(fields.Str(validate=verify_uuid_format),
+                               validate=validate.Length(min=1, max=5000),
+                               required=True)
+    exclude_fields = fields.List(fields.Str(validate=validate.Length(min=1, max=255)),
+                                 required=False,
+                                 missing=[])
+
+    @validates("exclude_fields")
+    def validate_exclude_fields(self, fields):
+        valid_fields = ["facts", "system_profile_facts"]
```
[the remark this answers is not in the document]
dehort: I was thinking the search by post would allow for including / excluding the facts and system_profile.

PR #236, opened 2019-04-25
Adding delete

jhjaggars:


tasks/__init__.py
```diff
@@ -15,6 +16,13 @@
 TOPIC = os.environ.get("KAFKA_TOPIC", "platform.system-profile")
 KAFKA_GROUP = os.environ.get("KAFKA_GROUP", "inventory")
 BOOTSTRAP_SERVERS = os.environ.get("KAFKA_BOOTSTRAP_SERVERS", "kafka:29092")
+EVENT_TOPIC = os.environ.get("KAFKA_EVENT_TOPIC", "platform.inventory.events")
```
Glutexo: I’d prefer to have the environment variable names prefixed to not clash with other apps in development. But since other ones are not prefixed too, it’s not necessary to do it here and make this pull request inconsistent. Created a separate issue for that instead. #271

tasks/__init__.py
```diff
@@ -15,6 +16,13 @@
...
+
+producer = KafkaProducer(bootstrap_servers=BOOTSTRAP_SERVERS)
+
+
+def emit_event(e):
+    producer.send(EVENT_TOPIC, value=e.encode("utf-8"))
```
Glutexo: [According](https://kafka-python.readthedocs.io/en/1.3.5/usage.html#kafkaproducer) to the documentation, the messages are sent asynchronously. If the application crashes before the messages is actually sent to the queue, it’s lost. This can be solved by calling _producer.flush_, which with synchronously block, until all messages are sent.

Glutexo:
Created a little demo on how to test the emitted events: see the [_delete-hosts_](https://github.com/Glutexo/insights-host-inventory/commits/delete-hosts) branch in my [fork](https://github.com/Glutexo/insights-host-inventory/). It swaps the message producer in the tests, replacing the _KafkaProducer_ with a class that mocks the used part of its interface, storing the messages in the memory. Those can be then inspected by the tests. I added one simple [test](https://github.com/Glutexo/insights-host-inventory/blob/65d3d38f3d1b05d66d721c72e3ca127100e08bcc/test_api.py#L251) that checks the message.

Glutexo:
I added one more [commit](https://github.com/Glutexo/insights-host-inventory/commit/f4f5ace34bf5192deb8b1a6904be2593518cc3e2) to my fixes demo [branch](https://github.com/Glutexo/insights-host-inventory/tree/delete-hosts). It uses Flask’s [App Context](http://flask.pocoo.org/docs/1.0/appcontext/) teardown to flush the message queue instead of doing so right after the emit. This is how it should ][work](http://flask.pocoo.org/docs/1.0/appcontext/#storing-data). The messages are flushed even if an exception and _500 Internal Server Error_ occurs.

PR #240, opened 2019-04-25
Add verbose message to bad offset error

Glutexo:
If a query is run with bad _LIMIT/OFFSET_ parameters, the 404 error message is not explanatory. It doesn’t say “Page Not Found” any more.

[Suggested](https://github.com/RedHatInsights/insights-host-inventory/pull/216#issuecomment-486225182) by @fijshion at #216.


Glutexo:
Thanks, @fijshion. Merge, @dehort?

PR #253, opened 2019-05-04
This PR abandons pipenv for setup.py and pip

jhjaggars:


Glutexo:
Why?

Glutexo:
Just wondering, why is the [_setup-py_](https://github.com/RedHatInsights/insights-host-inventory/tree/setup-py) branch in the [main repository](https://github.com/RedHatInsights/insights-host-inventory)? 🤔 Having a write access is nice, but shouldn’t the branches in there be restricted to real project branches like [_stable_](https://github.com/RedHatInsights/insights-host-inventory/tree/stable), versions and possibly epics (like my [_pagination_](https://github.com/RedHatInsights/insights-host-inventory/tree/pagination))?

dehort:
The reason to abandon Pipenv is to gain some reliablity.  Pipenv has broken on me 2 times in the past month or so.  I tried to resolve the issues but in the end I did not have time to waste debugging a Pipenv issue.  I ended up checking out new copies of the inventory source code, going through the pipenv process again (without changing anything) and the issue was resolved.

At this point, I am not looking to start any flamewars.  I am simply looking for a more reliable mechanism for dependency management.

Glutexo:
Oh. Never found Pipenv that unreliable. 😯 Thanks for making this clearer! Good news is that even without “officialy” using _Pipenv_, it’s still possible to use it with plain old _requirements.txt_ only at the developer’s machine.

It looks that _pyproject.toml_ is the current **standard** way to go (see [PEP 518](https://www.python.org/dev/peps/pep-0518/)) and _poetry_, which uses it, currently gains a lot of popularity. Not advocating that tool, neither having anything against _pip/setuptools_. Just mentioning for the record. 🙂

PR #277, opened 2019-05-22
Change base64 sample

SteveHNH:
We're about to change some things that will make org_id a required field. Want to get ahead of it in these examples.


dehort:
Looks good.

PR #285, opened 2019-05-30
Modify the db migration scripts so that the indices match what is in the production db

dehort:
Modify the db migration scripts to create indices that match what 
exists in the production db.

Add indices to the model so that the alembic "autogenerate"
functionality does not try to delete them each time a new
migration is created.


Glutexo:
How can there be any database structure in the production without doing so by a migration? 😱

Issue #102, opened 2019-01-14
Make the tests structured


Glutexo:
All tests are currently stuffed in two files: [_test_api.py_](
https://github.com/RedHatInsights/insights-host-inventory/blob/142588853c96554443b5782a4b9db42d90d8f84a/test_api.py) and [_test_unit.py_](
https://github.com/RedHatInsights/insights-host-inventory/blob/142588853c96554443b5782a4b9db42d90d8f84a/test_unit.py). As new tests and cases are added with every pull request, it’d be nice to have them more structured.

I suggest putting them in a _/tests_ folder and then possibly into _api_ and _unit_ subfolders. Then I’d put every test case in a single file, or at least keep only the most closely related together.

Any further improvements like splitting into finer categories or utilizing _Pytest_ etc. can be done subsequently as new issues.

dehort:
I agree.  We should certain start moving the tests into a sub-directory to help organize them better.

Glutexo:
#288 is a step towards this.

PR #288, opened 2019-06-03
Reorganizing the source code

dehort:
Reorganizing the source code.  This should allow us to start
reorganizing the tests (splitting the test_api.py file apart, etc).

Do not explicitly add "inventory" prefix to the logger names.  This is
no longer needed due to the reorganization of the code.


dehort:
@Glutexo I think we should start looking into reorganizing the code.  This would be a first step.  Let me know what you think.

Glutexo:
This is a step towards #102.

Glutexo:
This is now tracked in [RHCLOUD-8501](https://projects.engineering.redhat.com/browse/RHCLOUD-8501). Closing to clean up the queue.

This pull request is heavily obsolete. To revive it, it would be necessary to do the work from the ground up. The original idea is still relevant though and it would be a good practice to implement it.

PR #290, opened 2019-06-05
Extract common query parameters

Glutexo:
The OpenAPI [specification](https://github.com/RedHatInsights/insights-host-inventory/blob/1dab76aaaab5f77aeaf866efbd075b31022ad738/swagger/api.spec.yaml) describes the same parameter multiple times for different operations. Extracted those the parameters sections and replaced the in-place definition with a reference.

This also solves an UUID validation bug (#289) for facts operations, as there was missing the string format key.

### DIFF
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -39,12 +39,7 @@ paths:
             format: uuid
           description: Search for a host by insights_id
           required: false
-        - in: query
-          name: branch_id
-          schema:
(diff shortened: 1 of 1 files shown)


Glutexo:
Used the test from #290 to confirm that the bug is fixed by this pull request too.

Glutexo:
This required a rebase, because #289 has been merged. Done. Tests still passing. 🍏

Issue #147, opened 2019-03-18
Unify the query functions


Glutexo:
Many of the query functions called by [_get_host_list_](https://github.com/RedHatInsights/insights-host-inventory/blob/581d9fea1cd40de0c88072fb9b131e74c659b932/api/host.py#L145) are essentially one-liners, or will became ones if #145 gets merged. At the same time, the [_else_ branch](https://github.com/RedHatInsights/insights-host-inventory/blob/581d9fea1cd40de0c88072fb9b131e74c659b932/api/host.py#L162) doesn’t have its own function. It would be nice to unify this a bit, doesn’t really matter which way.

Glutexo:
Related to #172.

PR #294, opened 2019-06-12
Add support for custom ordering ✨

Glutexo:
Added ordering query parameters ([_order_by_](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/swagger/api.spec.yaml#L357) and [_order_how_](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/swagger/api.spec.yaml#L367)) to all three query operations ([_/hosts_](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/swagger/api.spec.yaml#L7), [_/hosts/{host_id_list}_](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/swagger/api.spec.yaml#L105), [_/hosts/{host_id_list}/system_profile_](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/swagger/api.spec.yaml#L209)). This allows to request the items ordered by one of the two fields ([_display_name_](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/swagger/api.spec.yaml#L364), [_updated_](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/swagger/api.spec.yaml#L365)) available in the UI.

Default ordering is still _updated DESC_. Default direction for _updated_ is _DESC_, for _display_name_ it’s _ASC_. When ordering by _display_name_, an _ordering_ by _updated DESC_ is added to the end. It’s not possible to specify only an ordering direction.

There is some repeating code around handling the query parameters. I didn’t want to DRY this single bit as it’s a part of a bigger functionality that is common to the queries – pagination, ordering, response formatting. There is already an issue for that: #147. At least the [__get_host_list_by_id_list_](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/api/host.py#L338) method is now a bit [simpler](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:custom_ordering?expand=1#diff-a5c55b46cc9c536070cb56798c97adfaL300).

### DIFF
--- a/test_unit.py
+++ b/test_unit.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 from api import api_operation
+from api.host import _params_to_order_by, _order_how
 from app.config import Config
 from app.auth.identity import (Identity,
                                validate,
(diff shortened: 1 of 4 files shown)


Glutexo:
cc @eduardocerqueira, @catastrophe-brandon

test_unit.py
```diff
@@ -246,5 +247,86 @@ def test_config_development_settings(self):
...
+        self.assertEquals(actual, expected)
+        order_how.assert_called_once_with(display_name, "ASC")
+
+    @patch("api.host.Host.display_name")
+    def test_order_by_display_name_desc(self, display_name, modified_on, order_how):
+        actual = _params_to_order_by("display_name", "DESC")
+        expected = (order_how.return_value, modified_on.desc.return_value)
+        self.assertEquals(actual, expected)
+        order_how.assert_called_once_with(display_name, "DESC")
+
```
dehort: I do not understand why these tests (which are testing the top level method) need to have any knowledge of the the methods which compose the function under test.  The tests now depend on the internal implementation of the _params_to_order_by method.  This seems brittle to me.

Glutexo: It’s true, the internal calls are mocked and tested and the tests are thus brittle. It’s the point that if you change [__order_how_](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/api/host.py#L201), the tests still pass. Like that it’s a real unit test testing only that single method. That all works well together is tested by the new API [tests](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/test_api.py#L1616).

If [__order_how_](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/api/host.py#L201) were not mocked, anything that would break its [tests](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/test_unit.py#L250) would also break __params_to_order_by_ [tests](https://github.com/Glutexo/insights-host-inventory/blob/64d83729ff4ff1120e089fca4bbd975313e6a953/test_unit.py#L272).

There can be an aggregate test suite that would test these two as a single unit. This suite can augment or replace the current ones.

I don’t have strong opinions on how we should do unit testing in the app. It would be though great to agree on some way – e.g. that we‘d not test functions internal workings and use patch only to mock out a library or a service.

Glutexo:
I created one more branch [_custom_ordering_tests_](https://github.com/Glutexo/insights-host-inventory/blob/custom_ordering_tests/), where I updated the tests so they don’t patch the methods internals. Like that the [tests](https://github.com/Glutexo/insights-host-inventory/blob/custom_ordering_tests/test_unit.py#L251) are less unit tests, have more dependencies and will fail together. On the other hand they are less brittle and a bit more expressive. Please take a [look](https://github.com/Glutexo/insights-host-inventory/blob/custom_ordering_tests/test_unit.py#L251), so we can decide which approach is better.

Glutexo:
Rebased to resolve conflicts. Also changed the deprecated _assertEquals_ to _assertEqual_.

PR #302, opened 2019-06-20
Add ORDER BY id DESC to queries

Glutexo:
To ensure unambiguous record ordering, order by _modified_on_ is not enough. Order by primary key must be there so the same order is really guaranteed. [Added](https://github.com/RedHatInsights/insights-host-inventory/pull/302/files#diff-a5c55b46cc9c536070cb56798c97adfaR232) this _ORDER BY_ clause to the host queries.

The [tests](https://github.com/Glutexo/insights-host-inventory/blob/f0cdd77923091168df59883fd2e0d0d609f4dd8a/test_api.py#L1714) are quite complicated. I could not find a way how to do it better. Tampering with a primary key and with the autofilled timestamps is not a very easy things to do. Hopefully the comments alleviate it a bit.

one-does-not-simply-test-ordering-by-automatic-timestamps-and-a-primary-key

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -241,7 +241,7 @@ def _params_to_order_by(order_by=None, order_how=None):
             "Provide order_by={updated,display_name}."
         )
 
-    return ordering + modified_on_ordering
+    return ordering + modified_on_ordering + (Host.id.desc(),)
 
 
 def _build_paginated_host_list_response(total, page, per_page, host_list):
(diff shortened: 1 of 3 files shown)


Glutexo:
Rebased to resolve the formatting conflicts.

Glutexo:
Rebased to resolve conflicts. Also found out that there are failing tests just because I didn’t amend them after adding the feature. Fixed those then.

Glutexo:
Rebased to resolve formatting conflicts.

PR #306, opened 2019-06-24
Normalize insights_id when searching

Glutexo:
If a host is queried by an insights ID, it is now found regardless of the UUID formats used. It doesn’t matter whether it is stored or queried with or without hyphens.

The _or_ [approach](https://github.com/Glutexo/insights-host-inventory/blob/0e8dabb3470f4391533e2342a1e90680d0886334/api/host.py#L142) used would not be viable if there were more values requiring normalization. But as this affects only _insights_id_, we can keep it this simple. Otherwise we’d need to store the normalized values along the original ones.

On my machine with about one million hosts in the database I didn’t see any difference in performance.

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -20,6 +20,8 @@ from app import create_app
 from app import db
 from app.auth.identity import Identity
 from app.utils import HostWrapper
+from app.utils import uuid_with_hyphens
+from app.utils import uuid_without_hyphens
 from tasks import msg_handler
 from test_utils import rename_host_table_and_indexes
 from test_utils import set_environment
@@ -38,8 +40,12 @@ ACCOUNT = "000501"
 SHARED_SECRET = "SuperSecretStuff"
 
 
-def generate_uuid():
-    return str(uuid.uuid4())
+def generate_uuid(with_hyphens=True):
+    generated_uuid = uuid.uuid4()
+    if with_hyphens:
+        return uuid_with_hyphens(generated_uuid)
+    else:
+        return uuid_without_hyphens(generated_uuid)
 
 
 def test_data(display_name="hi", facts=None):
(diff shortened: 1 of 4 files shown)


test_api.py
```diff
@@ -34,8 +34,12 @@
 SHARED_SECRET = "SuperSecretStuff"
 
 
-def generate_uuid():
-    return str(uuid.uuid4())
+def generate_uuid(with_hyphens=True):
```
jhjaggars: bare boolean arguments tend to age poorly.  Consider something like this:

```
def generate_uuid(format_func=uuid_with_hyphens):
    return format_func(uuid.uuid4())
```

One more nitpick would be to rename those functions so your calls would look like:
```
generate_uuid(with_hyphens)
```
or
```
generate_uuid(without_hyphens)
```

Glutexo: Good point, I like the functional style! I kept the function names in the [declaration](https://github.com/Glutexo/insights-host-inventory/blob/8c08d867154722db89b9c33022773cc5fc92cb96/app/utils.py#L4) to be still descriptive, bud changed them on [import](https://github.com/Glutexo/insights-host-inventory/blob/8c08d867154722db89b9c33022773cc5fc92cb96/test_api.py#L14) in the [test suite](https://github.com/Glutexo/insights-host-inventory/blob/8c08d867154722db89b9c33022773cc5fc92cb96/test_api.py).

Glutexo:
Removed the boolean argument as @jhjaggars suggested? Was that only a random comment, or a complete review with just one complain?

Glutexo:
Rebased to resolve the formatting conflicts.

Glutexo:
Rebased.

Glutexo:
Rebased.

Glutexo:
Fixed all new rule violations.

Glutexo:
Tracked by [RHCLOUD-8503](https://projects.engineering.redhat.com/browse/RHCLOUD-8503) to clean up the queue. Reopen if relevant.

PR #310, opened 2019-06-25
Dedup hosts using subscription manager

dehort:
This change uses subscription_manager_id as well as insights_id as an "elevated" id for deduplication. 

RHCLOUD-149

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -89,18 +89,40 @@ def _add_host(host):
 @metrics.host_dedup_processing_time.time()
 def find_existing_host(account_number, canonical_facts):
     existing_host = None
-    insights_id = canonical_facts.get("insights_id", None)
 
-    if insights_id:
-        # The insights_id is the most important canonical fact.  If there
-        # is a matching insights_id, then update that host.
-        existing_host = find_host_by_insights_id(account_number, insights_id)
+    elevated_canonical_fact_fields = ("insights_id",
+                                      "subscription_manager_id",
+                                      "external_id",
+                                      )
+
+    id_dict = {}
+    for cf in elevated_canonical_fact_fields:
+        cf_value = canonical_facts.get(cf)
+        if cf_value:
+            id_dict[cf] = cf_value
+
+    print("id_dict:", id_dict)
+
+    if id_dict:
+        # There is at least one "elevated" cf passed in
+        existing_host = _find_host_by_elevated_ids(account_number, **id_dict)
 
     if not existing_host:
         existing_host = find_host_by_canonical_facts(account_number,
                                                      canonical_facts)
 
     return existing_host
+
+
+@metrics.find_host_using_elevated_ids.time()
+def _find_host_by_elevated_ids(account_number, **kwargs):
+    filter_list = [Host.canonical_facts[k].astext == v
+                   for k, v in kwargs.items()]
+
+    return Host.query.filter(sqlalchemy.and_(
+        *[Host.account == account_number,
+          sqlalchemy.or_(*filter_list)]
+        )).order_by(Host.created_on).first()
 
 
 def find_host_by_insights_id(account_number, insights_id):
--- a/api/host.py
+++ b/api/host.py
@@ -90,9 +90,26 @@ def _add_host(host):
 def find_existing_host(account_number, canonical_facts):
     existing_host = None
 
+    input_elevated_canonical_facts = _pluck_elevated_canonical_facts(
+            canonical_facts
+            )
+
+    if input_elevated_canonical_facts:
+        # There is at least one "elevated" canonical fact passed in
+        # Search for an existing host using the "elevated" canonical facts
+        existing_host = _find_host_by_elevated_ids(account_number,
+                **input_elevated_canonical_facts)
+
+    if not existing_host:
+        existing_host = find_host_by_canonical_facts(account_number,
+                                                     canonical_facts)
+
+    return existing_host
+
+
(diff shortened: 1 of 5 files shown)


test_host_dedup_logic.py
```diff
@@ -66,3 +66,16 @@ def test_find_host_using_insights_id_match(flask_app_fixture):
                               }
 
     basic_host_dedup_test(canonical_facts, search_canonical_facts)
+
```
Glutexo: I am missing tests that would search by both _insights_id_ and _subscription_manager_id_.

- One that would just test that having both of them doesn’t break the search.
- One that would test that if there are more hosts, one matching by _insights_id_ and one matching by _subscription_manager_id_, that the one matching by _insights_id_ is found if if it has higher _created_on_ timestamp.

The second one would currently fail, because only a simple _or_ condition is used. Maybe that’s ok, but the specification in the ticket describes a different logic. Here you can see a failing [test](https://github.com/Glutexo/insights-host-inventory/blob/4e149ed475ad679cb4b568f26572fa8011be753d/test_host_dedup_logic.py#L83).

test_host_dedup_logic.py
```diff
@@ -66,3 +66,16 @@ def test_find_host_using_insights_id_match(flask_app_fixture):
...
+def test_find_host_using_subscription_manager_id_match(flask_app_fixture):
+    sub_mgr_id_key = "subscription_manager_id"
+    canonical_facts = {"fqdn": "fred",
+                       "bios_uuid": generate_uuid(),
+                       sub_mgr_id_key: generate_uuid(),
+                       }
+
+    # Change the bios_uuid so that falling back to subset match will fail
```
Glutexo: - Why are these three lines so complicated?

```python
    search_canonical_facts = {
        "bios_uuid": generate_uuid(),
        sub_mgr_id_key: canonical_facts[sub_mgr_id_key]
    }
```

- What about not setting the _bios_uuid_ at all. Then there would be no risk of falling back to generic canonical fact matching.

app/models.py
```diff
@@ -33,7 +33,10 @@ class Host(db.Model):
     # alembic autogenerate functionality does not try to remove the indexes
     __table_args__ = (Index("idxinsightsid", text("(canonical_facts ->> 'insights_id')")),
                       Index("idxgincanonicalfacts", "canonical_facts"),
-                      Index("idxaccount", "account"),)
+                      Index("idxaccount", "account"),
+                      Index("hosts_subscription_manager_id_index",
```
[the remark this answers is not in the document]
dehort: At this point, I'm trying to figure out what the naming convention should be.  I like the easier to read / more specific naming convention that this change uses.  Let's start with this one and renamed the others later on.

api/host.py
```diff
@@ -103,6 +107,30 @@ def find_existing_host(account_number, canonical_facts):
...
+    id_dict = {}
+    for cf in elevated_canonical_fact_fields:
+        cf_value = canonical_facts.get(cf)
+        if cf_value:
+            id_dict[cf] = cf_value
+    return id_dict
+
+
+@metrics.find_host_using_elevated_ids.time()
+def _find_host_by_elevated_ids(account_number, **kwargs):
```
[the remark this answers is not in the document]
dehort: I made changes to simplify this change.  Take a look now.  I think the name more closely matches the logic now.

api/host.py
```diff
@@ -103,6 +107,30 @@ def find_existing_host(account_number, canonical_facts):
...
+    filter_list = [Host.canonical_facts[k].astext == v
+                   for k, v in kwargs.items()]
+
+    return Host.query.filter(sqlalchemy.and_(
+        *[Host.account == account_number,
+          sqlalchemy.or_(*filter_list)]
```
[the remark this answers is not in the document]
dehort: That is a good point.  I took your suggestion (and the test).  Take a look at it now.

The performance of this logic concerns me a bit as we move forward though.  I'll ask our QA folks to run some performance tests on this.

test_host_dedup_logic.py
dehort: If you take out the explicit sub_mgr_id matching, then this test fails when the bios_uuid is explicitly changed.  Without explicitly changing the bios_uuid, the test would not fail when the explicit sub_mgr_id matching is removed.

app/models.py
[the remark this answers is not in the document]
Glutexo: I can probably live with that.

test_host_dedup_logic.py
Glutexo: Oh! I didn’t realize that these elevated canonical facts are still used for the generic CF matching. This looks to me is a bit unintuitive behavior of the code, although not necessarily illogical.

test_host_dedup_logic.py
```diff
@@ -66,3 +66,28 @@ def test_find_host_using_insights_id_match(flask_app_fixture):
...
     basic_host_dedup_test(canonical_facts, search_canonical_facts)
+
+def test_find_host_using_subscription_manager_id_match(flask_app_fixture):
+    canonical_facts = {"fqdn": "fred",
+                       "bios_uuid": generate_uuid(),
+                       "subscription_manager_id": generate_uuid(),
+                       }
+
+    # Change the bios_uuid so that falling back to subset match will fail
+    search_canonical_facts = {"bios_uuid": generate_uuid(), }
```
Glutexo: Do we want that trailing comma when the dict is just on one line?

Glutexo:
I can confirm this runs in CI.

PR #311, opened 2019-06-25
Getting app.logging wired into the new inv_mq_service

Jason-RH:
@dehort I confirmed that threadctx was working as expected and could see the request_id and account_number in the logstash message locally.

### DIFF
--- a/inv_mq_service.py
+++ b/inv_mq_service.py
@@ -1,5 +1,4 @@
 import json
-import logging
 import os
 
 from kafka import KafkaConsumer
@@ -10,14 +9,21 @@ from marshmallow import ValidationError
 from app import create_app
 from app.config import Config
 from app.exceptions import InventoryException
+from app.logging import get_logger, threadctx
 from lib.host import add_host
 
 
-logger = logging.getLogger("inventory_mq_service")
+logger = get_logger("mq_service")
 
 
 def handle_message(host):
     try:
+        threadctx.account_number = host["account"]
+        if "request_id" in host:
+            threadctx.request_id = host["request_id"]
+        else:
+            threadctx.request_id = "-1"
+        logger.info("Attempting to add host...") # Need to add log message once complete
         add_host(host)
     except InventoryException as e:
         logger.exception("Error adding host", extra={"host": host})
@@ -32,9 +38,9 @@ def handle_message(host):
 def event_loop(consumer, flask_app, handler=handle_message):
     with flask_app.app_context():
         while True:
-            print("waiting for msg")
+            logger.info("Waiting for message")
             for msg in consumer:
-                print("got a msg")
+                logger.info("Message received")
                 try:
                     data = json.loads(msg.value)
                 except Exception:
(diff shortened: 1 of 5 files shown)


inv_mq_service.py
```diff
@@ -32,9 +38,9 @@ def handle_message(host):
 def event_loop(consumer, flask_app, handler=handle_message):
     with flask_app.app_context():
         while True:
-            print("waiting for msg")
+            logger.info("Waiting for message")
```
dehort: Let's change this to a debug message.

inv_mq_service.py
```diff
@@ -32,9 +38,9 @@ def handle_message(host):
...
             for msg in consumer:
-                print("got a msg")
+                logger.info("Message received")
```
dehort: Let's change this to a debug message.

inv_mq_service.py
```diff
@@ -10,14 +9,21 @@
...
+logger = get_logger("mq_service")
 
 
 def handle_message(host):
     try:
+        threadctx.account_number = host["account"]
+        if "request_id" in host:
+            threadctx.request_id = host["request_id"]
+        else:
+            threadctx.request_id = "-1"
```
dehort: What do you think about moving the threadctx setup into a method?  Something like "initialize_thread_local_storage" or something similar.  My function names end up being kind of verbose, but I use the method names as a replacement for comments.  

Also, instead of the explicitly checking for "request_id" in host, how about the following:
  threadctx.request_id = host.get("request_id", "-1")

inv_mq_service.py
```diff
@@ -44,21 +50,20 @@ def event_loop(consumer, flask_app, handler=handle_message):
...
 
     config = Config()
-    config.log_configuration(config_name)
 
-    print("host_ingress_topic:", config.host_ingress_topic)
-    print("consumer_group:", config.consumer_group)
-    print("bootstrap_servers:", config.bootstrap_servers)
+    logger.info("Host Ingress Topic: %s" % config.host_ingress_topic)
+    logger.info("Consumer Group: %s" % config.consumer_group)
+    logger.info("Bootstrap Servers: %s" % config.bootstrap_servers)
```
dehort: What do you think about moving these log messages into the Config.log_configuration() method and calling that method here?  It is going to contain some information that is specific to the rest api service but maybe that is ok for now.

inv_mq_service.py
Jason-RH: :+1:

inv_mq_service.py
Jason-RH: :+1:

inv_mq_service.py
Jason-RH: Definitely a good idea. I'll get this updated in a bit.

inv_mq_service.py
Jason-RH: Yeah I'm fine putting these in configs log_configuration. Just wasn't sure how badly we'd want to see "inventory.mq_service" reporting it rather than "inventory.app.config"

Jason-RH: @dehort just to avoid confusion you mentioned calling Config.log_configuration here, but that's already being called via create_app. I'll come back around to it once that's moved out of create_app to see what needs to be done here.

PR #316, opened 2019-06-26
Add a mq service_specific kafka consumer group

dehort:


Makefile
```diff
@@ -16,4 +16,4 @@ run_inv_mq_service:
 	INVENTORY_LOGGING_CONFIG_FILE=logconfig.ini INVENTORY_LOG_LEVEL=DEBUG python inv_mq_service.py
 
 run_inv_mq_service_test:
-	KAFKA_TOPIC="platform.host-ingress" python utils/kafka_producer.py
+	KAFKA_GROUP="inventory-mq" KAFKA_TOPIC="platform.host-ingress" python utils/kafka_producer.py
```
Jason-RH: I noticed in config we have a "self.host_ingress_topic" that is obtained from "KAFKA_HOST_INGRESS_TOPIC". Should we be using that env variable rather than "KAFKA_TOPIC" here?

dehort: Basically, that kafka_producer.py util is a hack at this point.  It is not "aligned" with the service.  It should be cleaned up and made to share the same env vars, etc.

Issue #335, opened 2019-07-10
Fix all Flake8 rules


Glutexo:
Related to #189. This is a list of small subtasks and respective pull requests that would lead to a nicely formatted code. This code would

* stay unchanged when run through Black,
* not break any Flake8 rule that is not ignored with its default settings,
* be generally nice and consistent.

The concrete steps are:

1. Fix all Flake8 violations in reasonably small chunks.

PR #331, opened 2019-07-05
Pre-commit - code style/quality

diegorafrh:
Added *pre-commit* to help with the code quality/style.

Pre-commit runs a set of code analysis packages/hooks, such as:

- black
- flake8
- reorder-python-imports
- trailing-whitespace
- end-of-file-fixer
- check-yaml
- debug-statements
- pyupgrade

You can check the hooks available here: https://github.com/pre-commit/pre-commit-hooks

As you can see a lot of files has been automatically modified by **black** and I had to adjust manually a few others to fix all **flake8** warnings.

This PR comes from the **poc/ci** , I decided to put it apart so we can still merge the **poc/ci** without all this stuff and let this open to discuss and adjust the **pre-commit** till it fits the project needs.

Install **pre-commit**: `pipenv install pre-commit`
Running: `pre-commit run --all` 

It is worth mention that this effort is tied with the CI process to improve our Quality Gate

Issue #189, opened 2019-04-12
Ultimately fix code formatting


Glutexo:
The code is a formatting mess. Virtually every Python file is formatted completely randomly, only remotely adhering to [PEP 8](https://www.python.org/dev/peps/pep-0008/).

@jhjaggars suggested using [_Black_](https://github.com/ambv/black), which is a reasonable choice. It’d be nice to use a linter to block pull requests from being merged, but as first step, we can at least:

* Reformat existing code.
* Don’t push unformatted code any more.
* Don’t approve unformatted pull requests.

The first bullet point is the only one doable immediately to close this issue. I am aware that this would break all pull requests. Let’s take advantage of the fact that there are not many of them ending at the moment.

Glutexo:
Side note: automate all the things. Run the linter and do not let through unformatted code. That sounds nice, but this should be a part of a whole CI bot running tests and stuff on every PR. But that doesn’t mean, we can’t spin up a simple linter before the bigger thing happens.

Glutexo:
One more side note: Reformatting all the code may not be that catastrophic for the existing pull requests. If the same reformatting would be done on them, virtually all the conflicts should be resolved.

PR #333, opened 2019-07-09
Formatting: Add Flake8 to dev dependencies

Glutexo:
Added [Flake8](https://github.com/Glutexo/insights-host-inventory/blob/ce029c7422a3e3bc22c4bacbbf7ffee99b80f990/Pipfile#L36) to [Pipenv](https://github.com/Glutexo/insights-host-inventory/blob/ce029c7422a3e3bc22c4bacbbf7ffee99b80f990/Pipfile) [development dependencies](https://github.com/Glutexo/insights-host-inventory/blob/ce029c7422a3e3bc22c4bacbbf7ffee99b80f990/Pipfile#L32). Created a simple [configuration file](https://github.com/Glutexo/insights-host-inventory/blob/ce029c7422a3e3bc22c4bacbbf7ffee99b80f990/.flake8) that:

- [excludes](https://github.com/Glutexo/insights-host-inventory/blob/ce029c7422a3e3bc22c4bacbbf7ffee99b80f990/.flake8#L2) all non-Python files in the repository,
- [ignores](https://github.com/Glutexo/insights-host-inventory/blob/ce029c7422a3e3bc22c4bacbbf7ffee99b80f990/.flake8#L16) all errors currently present in the code base.

This may serve as the very first step towards a clean code style. #189 The ignored rules can be then removed in chunks to keep further pull requests small, see #335.

PR #345, opened 2019-07-11
Formatting: fix long lines

Glutexo:
Fixed the _E501_ Flake8 rule targetting too long lines. In most places, the code has been only reformatted. Sometimes a variable name has been shortened or another code-aware change has been made. I am not sure whether these changes don’t trigger another Flake8 rule. That should not be a big of an issue, since such cases should be rare and can be easily fixed.

I used a line length of 88 character, which is what Black uses by default.

In the [_app.config_](https://github.com/Glutexo/insights-host-inventory/blob/31ac6a483f9ab2cc33b69ce1e7fa948899d096f9/app/config.py) module I changed the [_import_](https://github.com/Glutexo/insights-host-inventory/blob/31ac6a483f9ab2cc33b69ce1e7fa948899d096f9/app/config.py#L1) statement and replaced _os.environ.get_ with an equivalent _os.getenv_.

In the [_api.metrics_](https://github.com/Glutexo/insights-host-inventory/blob/31ac6a483f9ab2cc33b69ce1e7fa948899d096f9/api/metrics.py) module I was targeting only the long lines, which results in not-unified code style. Improving the style and making it consistent would be a subject of another pull request.

This is a part of the first step towards a nicely formatted codebase. See #189, #335 and #331. It is currently the biggest pull request, but containing only trivial changes not affecting logic.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -21,10 +21,9 @@ extend-ignore =
     E303,
     E305,
     E402,
-    E501,
     F401,
     F811,
     F841,
     W291,
     W601
-max-line-length = 100
+max-line-length = 119
--- a/api/host.py
+++ b/api/host.py
@@ -447,7 +447,12 @@ def update_facts_by_namespace(operation, host_id_list, namespace, fact_dict):
     logger.debug("hosts_to_update:%s" % hosts_to_update)
 
     if len(hosts_to_update) != len(host_id_list):
-        error_msg = "ERROR: The number of hosts requested does not match the " "number of hosts found in the host database.  This could " " happen if the namespace " "does not exist or the account number associated with the " "call does not match the account number associated with " "one or more the hosts.  Rejecting the fact change request."
+        error_msg = (
+            "ERROR: The number of hosts requested does not match the number of hosts found in the "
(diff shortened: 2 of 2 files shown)


Glutexo:
Oh, thanks for clarifying. I was not sure whether @diegorafrh didn’t pick 100 just arbitrarily. It would be nice to document this in the [_README_](https://github.com/RedHatInsights/insights-host-inventory/blob/master/README.md).

Glutexo:
Extending the line length to 100 reduced the number of violating lines. Updated this pull request to use this value instead.

diegorafrh:
@catastrophe-brandon is right, I did follow what IQE have implemented but I also thought a little bit more about it: https://github.com/RedHatInsights/insights-host-inventory/pull/333#pullrequestreview-261837441

Glutexo:
Based on @catastrophe-brandon ’s comment I increased the maximum line length to 119 characters. With the limit being this high, there was only one line that violates the rule. Moreover this line was formatted like that probably by mistake, because it was a really long chopped into pieces, separated by spaces `"like " "this"`, but kept on a single line.

PR #351, opened 2019-07-15
Fix some tests that broke due to moving code around

dehort:
I'm not sure what happened here.  I must have merged some code into the branch in the wrong order and broke something.

@Glutexo please take a look and make sure I didn't do anymore damage.  Also, there are some more tests failing.  I'd appreciate it if you can look at those.

### DIFF
--- a/test_db_model.py
+++ b/test_db_model.py
@@ -2,6 +2,7 @@ import uuid
 
 from app import db
 from app.models import Host
+from app.serialization import Host as SerializationHost
 from test_utils import flask_app_fixture
 
 """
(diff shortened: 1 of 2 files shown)


Glutexo:
Roger! 🦸🏻‍♂️

Glutexo:
I added a new commit to this pull request. It fixes the broken _Host_ deserialization unit tests.

Originally, the (de)serialization methods were defined in the [_app.models.Hosts_](https://github.com/RedHatInsights/insights-host-inventory/blob/16e731c9502a9e7b1fd215fe16bd2b552ac2359d/app/models.py#L30) class, the same class that served as a database model. It uses its _classmethod_ _cls_ argument as a constructor. After the extraction to the [_app.serialization_](https://github.com/RedHatInsights/insights-host-inventory/blob/16e731c9502a9e7b1fd215fe16bd2b552ac2359d/app/serialization.py) module, the deserialization method now creates an instance of a completely different class, not using the argument any more.

Because of that I could get rid of the constructor [_shenanigans_](https://github.com/RedHatInsights/insights-host-inventory/blob/16e731c9502a9e7b1fd215fe16bd2b552ac2359d/test_unit.py#L392) and simplify the tests. 👍🏻

Glutexo:
Also created one more pull request that renames the test cases #352.

PR #353, opened 2019-07-16
Adding prometheus metrics to inv_mq_service

Jason-RH:
Initial commit. We can discuss other metrics we would like to add.

### DIFF
--- a/inv_mq_service.py
+++ b/inv_mq_service.py
@@ -8,7 +8,9 @@ from app.config import Config
 from app.exceptions import InventoryException, ValidationException
 from app.logging import get_logger, threadctx
 from app.queue.ingress import parse_operation_message
-from lib import host, metrics
+from app.queue import metrics
+from lib import host
+from prometheus_client import start_http_server
 
 
 logger = get_logger("mq_service")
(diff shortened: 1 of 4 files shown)


inv_mq_service.py
```diff
@@ -59,6 +59,7 @@ def initialize_thread_local_storage(operation_message):
 def main():
     config_name = os.getenv('APP_SETTINGS', "development")
     application = create_app(config_name, start_tasks=False)
+    metrics.start_http_server(9126)
```
dehort: I think we should import the start_http_server method directly in the inv_mq_service.py file.

lib/metrics.py
```diff
@@ -1,11 +1,13 @@
-from prometheus_client import Counter, Summary
+from prometheus_client import start_http_server, Counter, Summary
 
 host_dedup_processing_time = Summary("inventory_dedup_processing_seconds",
                                      "Time spent looking for existing host (dedup logic)")
 new_host_commit_processing_time = Summary("inventory_new_host_commit_seconds",
                                           "Time spent committing a new host to the database")
 update_host_commit_processing_time = Summary("inventory_update_host_commit_seconds",
                                              "Time spent committing a update host to the database")
+ingress_message_parsing_time = Summary("ingress_message_parsing_seconds",
```
dehort: I'd like to prefix the name with "inventory_".

What do you think about moving this metric into app/queue/metrics.py?  I think it'd be cleaner to move the metrics closer to where they are used.  I don't think we've done a great job with this in the past, but I think its a good idea.  What do you think?

Jason-RH: Yeah I'm fine with this! I was also considering having a separate metrics.py

inv_mq_service.py
Jason-RH: :+1:

PR #356, opened 2019-07-18
Formatting: use Pyupgrade to remove deprecated code

Glutexo:
Added [_Pyupgrade_](https://github.com/asottile/pyupgrade/) to development dependencies. This tool search Python scripts for pieces of deprecated code and replaces them with more modern variants. In our codebase, this targetted only the [API tests](https://github.com/Glutexo/insights-host-inventory/blob/pyupgrade/test_api.py), replacing the `%` string-formatting operator with the _str.format_ method. The tests are still passing 🍏 after the automatic fix.

This is a part of the way towards a better formatted code. #189 #331

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -35,6 +35,7 @@ pytest = "*"
 pytest-cov = "*"
 coverage = "*"
 flake8 = "*"
+pyupgrade = "*"
 
 [requires]
 python_version = "3.6"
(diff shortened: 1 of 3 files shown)


Glutexo:
[Fixed](https://github.com/RedHatInsights/insights-host-inventory/pull/356/commits/b866653edab1f36e902240895f486be6169997ce) the [_Pipfile_](https://github.com/Glutexo/insights-host-inventory/blob/pyupgrade/Pipfile) as I accidentally added the dependency as a regular one and not as a development one.

PR #359, opened 2019-07-19
Formatting: unify quotes by Black’s standards

Glutexo:
Use double quotes almost everywhere. Single quotes are used only for strings that contains inside more double quotes than single quotes, so the escaping is not too excessive. This happened only in some error messages.

This quoting style is how [Black](https://github.com/python/black) does it. It is possible to disable it, because there are some other possible approaches. This pull request assumes we can stick to the default.

This is a part to fix formatting in the whole codebase to be able to run [Black](https://github.com/python/black), [Flake8](https://gitlab.com/pycqa/flake8) and other tools automatically. #335


Glutexo:

PR #360, opened 2019-07-19
Formatting: unify newlines by Black’s standards

Glutexo:
Add newlines between class and function definitions and other pieces of code that need to be visually separated. Also remove excessive newlines. Used [Black](https://github.com/python/black) to fix these cases.

This is a part to fix formatting in the whole codebase to be able to run [Black](https://github.com/python/black), [Flake8](https://gitlab.com/pycqa/flake8) and other tools automatically. #335

### DIFF
--- a/app/exceptions.py
+++ b/app/exceptions.py
@@ -1,5 +1,3 @@
-
-
 class InventoryException(Exception):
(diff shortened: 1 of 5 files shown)


Glutexo:

PR #361, opened 2019-07-19
Formatting: contract multi-line expressions

Glutexo:
Used [Black](https://github.com/python/black) to contract expression spanning more than one line into a single line, if it fits into the 119 line-length limit. If the expression is a collection of comma-separated items, the trailing comma has been removed. No other changes to the code has been made.

Expressions, whose line count could be reduced, but would need changing the indentation style, have not been touched. That will be subject of another pull request.

This is a part of an ultimate formatting fix so [Black](https://github.com/python/black) can be run automatically for every commit and pull request. #335

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -27,9 +27,7 @@ FactOperations = Enum("FactOperations", ["merge", "replace"])
 # given priority in the host deduplication process.
 # NOTE: The order of this tuple is important.  The order defines
 # the priority.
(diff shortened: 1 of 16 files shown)


Glutexo:
Contracted one more line consisting only of a string split to multiple lines.

Glutexo:

Glutexo:
Rebased, mergeable again.

PR #365, opened 2019-07-22
Formatting: string interpolation

Glutexo:
Unified all string interpolations.

- Log messages use the logging module’s native % interpolation, receiving the values as position arguments.
- String literals use f-strings instead of the % operator and the _str.format_ method.

This is a part of making the code clean and nicely formatted. #335


Glutexo:

PR #368, opened 2019-07-23
Export version

Jason-RH:
Small PR to add build version into the prometheus export. I'm using the get_build_version() method, but can modify it to grab/create something else if OPENSHIFT_BUILD_COMMIT isn't what we want.


inv_mq_service.py
```diff
@@ -57,6 +58,7 @@ def initialize_thread_local_storage(operation_message):
 def main():
     config_name = os.getenv('APP_SETTINGS', "development")
     application = create_app(config_name, start_tasks=False)
+    metrics.version.info({"version": get_build_version()})
```
dehort: What do you think about moving the version.info() call into the metrics.py?  I think that will work??  Is that cleaner?  Let me know what you think.

Jason-RH: Yeah that works just fine! I originally had it that way but was unsure which would be best. I agree adding it to metrics.py looks cleaner.

PR #371, opened 2019-07-24
Formatting: Simplify super calls

Glutexo:
In newer versions of Python, it’s possible to call just super() instead of super(ParentClass, self). Replaced these obsolete calls with their more concise variant. This is what `pyupgrade --py36-plus` would do.

See the [documentation](https://docs.python.org/3.6/library/functions.html?highlight=super#super) for more info.

This is a part of ultimately fixing the code so it can be checked and fixed automatically on commit.

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -150,7 +150,7 @@ class DBAPITestCase(APIBaseTestCase):
         """
         Initializes the database by creating all tables.
         """
-        super(DBAPITestCase, self).setUp()
+        super().setUp()
 
         # binds the app to the current context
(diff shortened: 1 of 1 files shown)


Glutexo:

catastrophe-brandon:
Assuming we can get the conflicts resolved, I'm in favor of merging it.

Glutexo:
Rebased again. I suppose that now, when there are two approvals, I can merge it without asking @dehort to do so. If I am mistaken, please let me know.

PR #385, opened 2019-08-02
Lock packages to specific versions (stable)

Glutexo:
Copied the exact versions from [_Pipfile.lock_](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_stable/Pipfile.lock) to [_Pipfile_](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_stable/Pipfile). This replaces the very broad _*_ for almost all packages and should help to prevent wild upgrades to backwards incompatible versions of the packages.

Locking the directly required packages versions still doesn’t lock the second-level dependencies. Because of that, it’s still necessary to install the packages using `pipenv sync` (using [_Pipfile.lock_](https://github.com/RedHatInsights/insights-host-inventory/blob/stable/Pipfile.lock)) instead of `pipenv install` (using [_Pipfile_](https://github.com/RedHatInsights/insights-host-inventory/blob/stable/Pipfile)). The former installs the exact locked version for the whole dependency tree. The latter installs the most recent packages and updates the [lock](https://github.com/RedHatInsights/insights-host-inventory/blob/stable/Pipfile.lock) file accordingly. Please make sure that the right command is used on deploy. Even in case it weren’t, this pull request still lessens the damage risk that an unexpected upgrade would impose.

This pull request requests merge directly to the [_stable_](https://github.com/RedHatInsights/insights-host-inventory/tree/stable) branch, because [_master_](https://github.com/Glutexo/insights-host-inventory/tree/master) already has more dependencies and can’t be thus cherry-picked without conflicts.

I am aware that locking to the patch versions of the packages is suboptimal, but I think it’s better than _*_ everywhere. Later on, we should consider what version to lock to for every single level-one dependency. This applies to [development](https://github.com/RedHatInsights/insights-host-inventory/blob/stable/Pipfile#L32) requirements too, because tests depend on those and broken tests are just as bad as broken executive code. In the end, we’d probably end up with `~=` for most of the packages, but some are not using semantic versioning though.


j00bar:
I'm confused - isn't Pipfile supposed to be prescriptive of intent and Pipfile.lock to be a descriptive & explicit set of packages+versions? If there's a backward incompatible version change in Pipfile.lock, isn't the solution to ensure we're testing thoroughly enough to detect the breakage?

Glutexo:
It is just like you say. I am aware that just copying the exact version numbers does not describe intent well. I just thought it’s better than having no versions at all, making the rules more loose in a next pull request, ending up with `"==1.2.*"` in most cases.

But you’re right that with good tests, a package upgrade should not pass the pipeline. Still, I’d find locking to a userspace compatible reasonable.

Does at least this “second step” make sense to you?

Glutexo:
The CI is failing because of a bug in Pipenv. See @catastrophe-brandon’s new PR https://github.com/RedHatInsights/insights-pipeline-lib/pull/37.

Glutexo:
I read and watched a bit and I am now even more sure that we should lock the versions loosely, instead of keeping the *’s.

Pure [_pip_](https://pypi.org/project/pip/) doesn’t have lock files, so it [suggests](https://pip.readthedocs.io/en/1.1/requirements.html) using `pip freeze` to (duh) freeze the specific versions together with all dependencies. [_Pipenv_](https://pypi.org/project/pipenv/) uses more [Bundler](https://rubygems.org/gems/bundler)-like [approach](https://bundler.io/v1.12/rationale.html) with a lock file.

That allows us to specify the required packages [semantically](https://semver.org/) – describing an intention, just as you said, @j00bar. We mostly want to have a package version _a.b.*_, because we want bugfixes and security patches, but we don’t want incompatible changes. We use some features and expect some behavior. When updating, we want to make sure that our tests still pass, possibly remove some hacks, but don’t want to rewrite any code because of changed behavior.

It’s ironical that [Pipenv](https://pypi.org/project/pipenv/) itself [doesn’t use](https://pypi.org/project/pipenv/#history) [semantic versioning](https://semver.org/). 🤦🏻‍♂️

Still I’d keep this freeze pull request here as a first step and add a more deliberate one as a next step. I at least checked that the [pipeline](https://github.com/RedHatInsights/insights-pipeline-lib) correctly [uses](https://github.com/RedHatInsights/insights-pipeline-lib/blob/20ae5942d25e66fab2625a961deb0625956bcb95/vars/runPipenvInstall.groovy#L20) Pipenv’s _--deploy_ flag that makes it use solely [_Pipfile.lock_](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_stable/Pipfile.lock).

Glutexo:
I just found out that deploying this would fail even without the aforementioned bug. It’s because the although the versions in [_Pipfile_](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_stable/Pipfile) and [_Pipfile.lock_](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_stable/Pipfile.lock) are same the hash isn’t and the [lock file](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_stable/Pipfile.lock) is then considered out of date.

```
$ pipenv install --deploy                                                       
Your Pipfile.lock (caeff6) is out of date. Expected: (7510a6).
Aborting deploy.
```

That means that this can’t be deployed without upgrading the second-level dependencies. Thus closing this, because anything like that can’t go directly to [_stable_](https://github.com/RedHatInsights/insights-host-inventory/tree/stable).

Glutexo:
Continuing work in #386

Issue #367, opened 2019-07-22
Lock dependencies to specific versions


Glutexo:
As this is not a library, but a deployable application, it should have its dependencies stated more clearly. Currently all packages are [listed](https://github.com/RedHatInsights/insights-host-inventory/blob/599bc48cdc69d772ade2040a94488a65b2af70f5/Pipfile) without a version. Although this is not a problem when deploying, thanks to the [lock](https://github.com/RedHatInsights/insights-host-inventory/blob/599bc48cdc69d772ade2040a94488a65b2af70f5/Pipfile.lock) file, it brings some mild problems:

* Special caution is required when dealing with the packages: rebuilding the [lock](https://github.com/RedHatInsights/insights-host-inventory/blob/599bc48cdc69d772ade2040a94488a65b2af70f5/Pipfile.lock) incautiously file may result in bumping up a package to a too high version.
* Code is not written for specific versions of the libraries, the exact version used can be determined only by looking into the [lock](https://github.com/RedHatInsights/insights-host-inventory/blob/599bc48cdc69d772ade2040a94488a65b2af70f5/Pipfile.lock) file.

This is not how it should work and is not semantic at all. The [_Pipfile_](https://github.com/RedHatInsights/insights-host-inventory/blob/599bc48cdc69d772ade2040a94488a65b2af70f5/Pipfile) itself should contain a list of packages and their version constrains. The [lock](https://github.com/RedHatInsights/insights-host-inventory/blob/599bc48cdc69d772ade2040a94488a65b2af70f5/Pipfile.lock) file is there to replicate the same environment.

Also please see a notoriously known [article](https://caremad.io/posts/2013/07/setup-vs-requirement/), slightly related to this.

Glutexo:
One more reason to do so is that without specific version, it’s very easy to accidentally upgrade the packs in the [lock](https://github.com/RedHatInsights/insights-host-inventory/blob/599bc48cdc69d772ade2040a94488a65b2af70f5/Pipfile.lock) file. That can happen when adding a new dependency. Then, anything can break unexpectedly.

Glutexo:
There are some pull requests for this:

* #385 (

PR #386, opened 2019-08-02
Lock packages to specific versions

Glutexo:
Copied the exact versions from [Pipfile.lock](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_master/Pipfile.lock) to [Pipfile](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_master/Pipfile). This replaces the very broad _*_ for almost all packages and should help to prevent wild upgrades to backwards incompatible versions of the packages.

This is a version of #367 based on the [_master_](https://github.com/RedHatInsights/insights-host-inventory/tree/master) branch. It has locked version for the new development [dependencies](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_master/Pipfile#L36) too.

I am aware that locking to the patch versions of the packages is suboptimal, but I think it’s better than _*_ everywhere. Later on, we should consider what version to lock to for every single level-one dependency. This applies to [development](https://github.com/RedHatInsights/insights-host-inventory/blob/stable/Pipfile#L32) requirements too, because tests depend on those and broken tests are just as bad as broken executive code. In the end, we’d probably end up with `~=` for most of the packages, but some are not using semantic versioning though.


Glutexo:
There is a discussion going on at #367 on whether this is actually a good idea. Locking to the very specific version may not be a good idea even though it’s intended to be replaced by a more thought-through variant.

Glutexo:
I found out that even though the versions are directly taken from [_Pipfile.lock_](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_master/Pipfile.lock), it’s still considered out of date. That means that without re-locking, the app would become undeployable.

```
$ pipenv install --deploy                                                       
Your Pipfile.lock (caeff6) is out of date. Expected: (7510a6).
Aborting deploy.
```

Updating the [lock file](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_master/Pipfile.lock) though means upgrading _level 2_ dependencies. There should be no harm in doing that as only very few packages upgrade and those are only bugfix releases. But it **must** go through QE to verify nothing broke.

Glutexo:
Updated the [lock file](https://github.com/Glutexo/insights-host-inventory/blob/lock_versions_master/Pipfile.lock). The tests are still green. That means that if something breaks, it’s something we don’t have covered by tests.

PR #388, opened 2019-08-05
Remove quotes from package names

Glutexo:
Removed double quotes from the package names that contains digits. It is not necessary for them to be there. The TOML file is still valid and the formatting is more unified like that.

This is just a cosmetic change. It does not require recreating the [lock](https://github.com/Glutexo/insights-host-inventory/blob/unify_pipfile_formatting/Pipfile.lock) file.


Glutexo:
Thanks, @dehort! Rebased on the current master, resolved conflicts. Mergeable again.

PR #392, opened 2019-08-07
Added new elements to delete event

aleccohan:
Needed to add insights_id to delete event in order to ping legacy API to duplicate the deletion.  Also added instructions to the README to reflect the difficulty I had with standing up the host-inventory platform locally.


README.md
```diff
@@ -92,6 +92,14 @@ _prometheus_multiproc_dir_ environment variable. This is done automatically.
 python run_gunicorn.py
 ```
 
+Should issues occur with the server's connection to the docker service, it might
```
Glutexo: This is actually not relevant to the subject of this pull request. I would extract that to another pull request. But it should be no longer relevant as we have #394 almost merged in.

run.py
```diff
@@ -3,7 +3,7 @@
 
 from app import create_app
 
-config_name = os.getenv("APP_SETTINGS", "development")
+config_name = os.getenv("APP_SETTINGS", "testing")
```
Glutexo: This has been probably added by accident.

api/host.py
```diff
@@ -286,7 +286,7 @@ def delete_by_id(host_id_list):
     host_ids_to_delete = []
     for host in query.all():
         try:
-            host_ids_to_delete.append(host.id)
+            host_ids_to_delete.append({"id": host.id, **host.canonical_facts, "account": host.account})
```
Glutexo: The name of the variable doesn’t reflect its contents any more. Now it’s more _hosts_to_delete_ than _host_ids_. I wonder however, can’t we just pass the whole _Host_ object to the _delete_ event [factory](https://github.com/aleccohan/insights-host-inventory/blob/a789023e26b5491ceae5ebab72fd61cc35214915/app/events.py#L27)? Like this we are building an arbitrary host-like dictionary for this single purpose.

We can argue whether the events module should know about the models, but I don’t see a problem in that. It’s its _output_ what has to be independent, but a most natural input for a host is a… _Host_.

app/events.py
```diff
@@ -11,7 +12,21 @@ class HostEvent(Schema):
...
+    bios_uuid = fields.Str(validate=verify_uuid_format)
+    ip_addresses = fields.List(fields.Str(validate=validate.Length(min=1, max=255)))
+    mac_addresses = fields.List(fields.Str(validate=validate.Length(min=1, max=255)))
+    external_id = fields.Str(validate=validate.Length(min=1, max=500))
 
 
-def delete(id):
-    return HostEvent(strict=True).dumps({"id": id, "timestamp": datetime.utcnow(), "type": "delete"}).data
+def delete(host):
+    return HostEvent(strict=True).dumps({"id": host['id'], "insights_id": host['insights_id'], "rhel_machine_id": host['rhel_machine_id'],
```
Glutexo: I see quite an overlap with existing _Host_ object [serialization](https://github.com/aleccohan/insights-host-inventory/blob/a789023e26b5491ceae5ebab72fd61cc35214915/app/models.py#L100). Which this actually is, although the format is a bit different.

I‘d ask whether it’s not possible just to dump a whole serialized host in there, only adding the _timestamp_ and _type_ fields to that. If not (probably because of the size of the payload, e.g. the _facts_ field), then I’d rather add a _to_event_json_ method to the _Host_ module to keep the logic at the same place. You also have a [list](https://github.com/aleccohan/insights-host-inventory/blob/a789023e26b5491ceae5ebab72fd61cc35214915/app/models.py#L214) of Canonical Facts there, which you can use.

The [_CanonicalFacts.to_json_](https://github.com/aleccohan/insights-host-inventory/blob/a789023e26b5491ceae5ebab72fd61cc35214915/app/models.py#L236) method already populates all the Canonical Facts, filling _None_ for missing values.

app/events.py
```diff
@@ -11,7 +12,21 @@ class HostEvent(Schema):
     id = fields.UUID()
     timestamp = fields.DateTime(format="iso8601")
     type = fields.Str()
+    account = fields.Str(validate=validate.Length(min=1, max=255))
```
Glutexo: I’m thinking whether it wouldn’t be possible to DRY this to some extend with the existing [_HostSchema_](https://github.com/aleccohan/insights-host-inventory/blob/a789023e26b5491ceae5ebab72fd61cc35214915/app/models.py#L350) using inheritance. I don’t insist on doing it in this pull request though.

app/events.py
Glutexo: 💥 This doesn’t work if the host doesn’t have all canonical facts. If you don’t pass along all canonical facts when creating/updating a host, they key is not there. It is even removed if it contains _None_ or an empty string. It happens [here](https://github.com/aleccohan/insights-host-inventory/blob/a789023e26b5491ceae5ebab72fd61cc35214915/app/models.py#L230).

You can bypass it by using host.get("insights_id"), but if you use the existing Canonical Facts [serialization method](https://github.com/aleccohan/insights-host-inventory/blob/new_delete_conditions/app/models.py#L236) as suggested, you’ll get all the _None_’s for free.

Side note: I was thinking about changing this to actually always store all fields, although with a _None_ if not filled in, but it has some consequences. We insert the internal representation of the canonical facts directly to the database and it’s necessary not to have the empty fields there for a native PostgreSQL deduplication to work correctly. That happens [here](https://github.com/aleccohan/insights-host-inventory/blob/new_delete_conditions/api/host.py#L125).

api/host.py
dehort: Good idea!

app/events.py
dehort: I like the idea of not repeating the field definitions, but it might take some work to remove the duplication.  I think we'd have to break the HostSchema down into smaller chunks (CanonicalFactsSchema, for example, which only contains the canonical facts) and then we'd have to construct the larger schemas (HostSchema, HostEventSchema, etc) out of the smaller schemas using inheritance/composition (if that's possible using marshmallow...i've not looked into marshmallow in that much detail).

api/host.py
```diff
@@ -283,28 +283,28 @@ def find_hosts_by_hostname_or_id(account_number, hostname):
 def delete_by_id(host_id_list):
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
-    host_ids_to_delete = []
+    hosts_to_delete = []
     for host in query.all():
         try:
-            host_ids_to_delete.append(host.id)
+            hosts_to_delete.append(host.to_json())
```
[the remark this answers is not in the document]
dehort: Brandon ran a test that was causing the accessing of the host object to raise that exception.  I cannot remember the specifics.  I do like the idea of passing the host object down to the event module.  Agreed.  The event code should be responsible for producing the external representaion.

api/host.py
```diff
@@ -283,28 +283,28 @@ def find_hosts_by_hostname_or_id(account_number, hostname):
...
     db.session.commit()
 
-    metrics.delete_host_count.inc(len(host_ids_to_delete))
+    metrics.delete_host_count.inc(len(hosts_to_delete))
 
-    logger.debug("Deleted hosts: %s", host_ids_to_delete)
+    logger.debug("Deleted hosts: %s", hosts_to_delete)
 
-    for deleted_host_id in host_ids_to_delete:
+    for deleted_host_id in hosts_to_delete:
```
dehort: I'm being a bit picky here, but we should likely change deleted_host_id to deleted_host or something similar.

app/events.py
```diff
@@ -11,7 +15,28 @@ class HostEvent(Schema):
...
+    external_id = fields.Str(validate=validate.Length(min=1, max=500))
 
 
-def delete(id):
-    return HostEvent(strict=True).dumps({"id": id, "timestamp": datetime.utcnow(), "type": "delete"}).data
+def delete(host):
+    return (
+        HostEvent()
+        .dumps(
+            {
```
dehort: This appears to be missing the id field.

app/events.py
Glutexo: This proposed solution sounds great! I created an issue for that, so it’s not forgotten. https://github.com/RedHatInsights/insights-host-inventory/issues/398

app/events.py
Glutexo: Really! It probably weren’t in my proposed code snippet and nobody realized. Good catch!

Glutexo: Good thing is that the current [tests](https://github.com/aleccohan/insights-host-inventory/blob/1bbf69c66a22a49ee796702b145754f589371ef3/test_api.py#L1239) catch this. 🙂

app/events.py
```diff
@@ -11,7 +15,28 @@ class HostEvent(Schema):
...
+                "timestamp": datetime.utcnow(),
+                **CanonicalFacts.to_json(host.canonical_facts),
+                "account": host.account,
```
Glutexo:
```suggestion
                "id": host.id,
                "account": host.account,
```

app/events.py
```diff
@@ -11,7 +15,28 @@ class HostEvent(Schema):
...
+    ip_addresses = fields.List(fields.Str(validate=validate.Length(min=1, max=255)))
+    mac_addresses = fields.List(fields.Str(validate=validate.Length(min=1, max=255)))
+    external_id = fields.Str(validate=validate.Length(min=1, max=500))
 
 
-def delete(id):
-    return HostEvent(strict=True).dumps({"id": id, "timestamp": datetime.utcnow(), "type": "delete"}).data
+def delete(host):
+    return (
+        HostEvent()
```
Glutexo: Do we actually need to use the _Schema_ to format the message? It’s our code that composes it, so it is inherently trusted. The format can (and should) be tested by our tests. Using _Schema_ for validation makes sense for data from the outside.

[_Host_ serialization](https://github.com/aleccohan/insights-host-inventory/blob/b2299f0b2214f1da2e477020fddbbb8d5d585c8e/app/models.py#L100) currently doesn’t use Schema and it results in a discrepancy in how DateTime is formatted. In the HTTP output, the [_created_](https://github.com/aleccohan/insights-host-inventory/blob/b2299f0b2214f1da2e477020fddbbb8d5d585c8e/app/models.py#L107) and [_updated_](https://github.com/aleccohan/insights-host-inventory/blob/b2299f0b2214f1da2e477020fddbbb8d5d585c8e/app/models.py#L108) are formatted manually as _2019-08-15T11:36:28.105816Z_. Marshmallow formats it as _2019-08-15T11:36:28.105816+00:00_. Both are valid ISO 8601 formats, but they are different.

But maybe a good solution would be to use a Schema for the _Host_ output too instead of removing it from here. This looks like a new issue.

api/host.py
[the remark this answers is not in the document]
Glutexo: If that’s really so, then the exception may not be raised when not accessing the _id_ property.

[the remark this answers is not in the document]
Glutexo: Created an issue: #400

aleccohan:
These are the updated changes to PR #392 which includes all of the changes made to PR #396.

aleccohan:
I ran the tests and I seem to be getting this error back every time
`sqlalchemy.orm.exc.ObjectDeletedError: Instance '<Host at 0x10f23a630>' has been deleted, or its row is otherwise not present.`
with only appending _host_, so as far as I can tell it seems that we haven't broken anything.  I also tested deleting the row with an access to an element of _host_ before the append and it gives me back the same error

app/events.py
dehort: I like the idea of using the Schemas for defining what our input and output should look like.  I think it makes it easy to point someone to the schema so that they can see what is expected for input/output.  Otherwise, it can be kind of difficult to determine what the expected input/output is.

I did just think of something that we need to watch out for though.  I think we should remove the validation checks from this schema.  As @Glutexo mentioned, we are the ones producing the data...so there shouldn't be a need to validate it on the way out.  Plus if someone tries to delete an "old" host that has data that doesn't match our current rules, then it would fail to produce a delete event for that object.

aleccohan:
Oh!  No the app works fine when I test it, what I was talking about were the tests I did to see if we broke the ObjectDeletedError test and from the results I'm pretty sure that it's working just as it's supposed to.

dehort:
@Glutexo do you see any issues with this PR?

Glutexo:
But beware – I approved the changes as the code changes look OK to me. We have still unanswered questions though, so please don’t merge before those are resolved:

* Is the [_ObjectDeletedError_ race condition guard](https://github.com/aleccohan/insights-host-inventory/blob/355ba890da6cea01a06e14dd220c3c17e5f1a958/api/host.py#L290) still working? If not, please fix.
* There is possibly a bug that hitting that race condition causes an uncaught error 500. Does it happen even before this PR? If so, don’t introduce any fix unless it naturally develops from fixing the aforementioned _guard_. On the other hand, if the error has been introduced by this PR, fix it.

Also don’t deploy this anywhere, not even CI (than means don’t merge) unless a check against the existing database is run. If the check fails, either consider loosening the Validation or merge in a migration that sanitizes the existing data.

aleccohan:
This change in the placement of the ObjectDeletedError test removes the triggering of the 500 error and also fixes the race condition guard.

api/host.py
```diff
@@ -305,7 +299,13 @@ def delete_by_id(host_id_list):
     logger.debug("Deleted hosts: %s", hosts_to_delete)
 
     for deleted_host in hosts_to_delete:
-        emit_event(events.delete(deleted_host))
+        try:
```
Glutexo: This change actually modifies the behavior. Before the change, if a host has been already deleted, it is not added to the _hosts_to_delete_ list. That caused aborting with 404 if no hosts were actually deleted, and also those not actually deleted hosts not to be logged. Now, all hosts to delete are logged even though they have been already deleted and raise the _ObjectDeletedError_ here.

api/host.py
```diff
@@ -285,13 +285,7 @@ def delete_by_id(host_id_list):
...
     hosts_to_delete = []
     for host in query.all():
-        try:
-            hosts_to_delete.append(host)
-        except sqlalchemy.orm.exc.ObjectDeletedError:
-            logger.exception(
-                "Encountered sqlalchemy.orm.exc.ObjectDeletedError exception during delete_by_id operation.  Host was "
-                "already deleted."
-            )
+        hosts_to_delete.append(host)
```
Glutexo: `query.all()` actually returns a pure list, so we can do `hosts_to_delete = query.all()` directly. But as I noted in the next comment, we are not checking now in any way whether the host has been already deleted or not.

I guess we need to figure out how to eagerly load the host here, essentially doing manually what happens on the attribute access. I would not eagerly load all of them at once, just do something like object identity check here.

api/host.py
dehort: This might change the behavior a bit but i'm not sure how big of a change it really is.

With this change, the only difference I see is that we might not send back a 404 in the case where all the hosts that were requested to be deleted, were already deleted.   (This situation is likely pretty rare anyhow.)  I suspect with the try/catch added around the emit_event(), then we shouldn't produce a delete event in the case where a host was deleted already (that would be the correct behavior in my opinion).

In either case (200 return code vs 404 return code), the host would be deleted and on event would be logged.  I'm not sure that's a big difference.

lphiri:
> With regards to the validation of the outgoing data, we are currently not performing a full validation of the data that the REST service returns. I do not feel like validating the fields in the delete event are required.
> 
> What is the worst case scenario? If we do not validate the data for the event, then I think the worst case scenario is that we send out an event that might contain canonical fact fields that contain bad data. If we validate the data, then we might not send out an event for a host that is deleted...thus potentially orphaning a host in the apps.

Hey guys, really appreciate the discussion here, but I would like us to move ahead with the merge to stay within the original scope : currently clients are only interested in account_number, host_id and insights_id and are not going to be broken if we send extraneous invalid data outside these fields.

Currently we do not validate our data output when requested via REST, so I don't think that its essential that we do that when emitting events.   Blocking a delete event because of a validation failure of extraneous facts would be really bad.

We do need to clean our database and purge invalid data, but lets do that a separate task.

Glutexo:
We‘re already validating the event. It is [a simple, safe validation](https://github.com/RedHatInsights/insights-host-inventory/blob/aa5481588cc76459a08f2de78946c0ec125f9aea/app/events.py#L10), because it contained only [_id_](https://github.com/RedHatInsights/insights-host-inventory/blob/aa5481588cc76459a08f2de78946c0ec125f9aea/app/events.py#L11), [_timestamp_](https://github.com/RedHatInsights/insights-host-inventory/blob/aa5481588cc76459a08f2de78946c0ec125f9aea/app/events.py#L12) and [_type_](https://github.com/RedHatInsights/insights-host-inventory/blob/aa5481588cc76459a08f2de78946c0ec125f9aea/app/events.py#L13). After the changes, even without the _length_ and _required_ rules, the validation is [still there](https://github.com/RedHatInsights/insights-host-inventory/pull/392/commits/a2b0c4d876cf76b9d521b3086c87a150024522ab), listing the allowed fields.

What about listing only the fields we really want to output – _account_ and _insights_id_? _account_ would be without a length limit as we don’t have sanitized data in the database yet. Like that it would be safe and at the same time it would well communicate the intended message format.

```python
class HostEvent(Schema):
    id = fields.UUID()
    timestamp = fields.DateTime(format="iso8601")
    type = fields.Str()
    account = fields.Str(required=True)
    insights_id = fields.Str(validate=verify_uuid_format)
```

The `**CanonicalFacts.to_json(host.canonical_facts)` there in the [_dumps_](https://github.com/aleccohan/insights-host-inventory/blob/a2b0c4d876cf76b9d521b3086c87a150024522ab/app/events.py#L33) can remain there or can be replaced by explicit listing of those two fields. Both would be OK I think.

As for the changed behavior, I can understand that it’s not a problem. The exception catching has been added probably because it was the first place where it occurred, without a clear definition of what should be logged, what the response code should be etc. in such case. So even though it’s not very good practice, I agree that we can let it go like that. As long as I’m right that there is no defined _correct_ behavior that we’d be breaking.

aleccohan:
This is the most recent push with only the fields that we intend to validate at this time.  I also included the request_id since it is requested on the JIRA ticket.

api/host.py
```diff
@@ -283,29 +283,29 @@ def find_hosts_by_hostname_or_id(account_number, hostname):
...
 
     with metrics.delete_host_processing_time.time():
         query.delete(synchronize_session="fetch")
     db.session.commit()
 
-    metrics.delete_host_count.inc(len(host_ids_to_delete))
+    metrics.delete_host_count.inc(len(hosts_to_delete))
 
-    logger.debug("Deleted hosts: %s", host_ids_to_delete)
+    logger.debug("Deleted hosts: %s", hosts_to_delete)
```
Glutexo: This works only thanks to the fact that _logger_ has some own exception handling mechanism. This logging call actually calls [_Host.__repr_](https://github.com/aleccohan/insights-host-inventory/blob/b586b700ef74917a77d00b26adf145a966fe13ec/app/models.py#L197) for every host, even for those that are already deleted. This function internally [calls](https://github.com/aleccohan/insights-host-inventory/blob/b586b700ef74917a77d00b26adf145a966fe13ec/app/models.py#L199) _Host.id_ that triggers the lazy loading and bang, an uncaught exception occurs.

Hopefully, this is internally caught by the logger output to _stderr_ and the execution continues normally. This is probably for proper logging calls not to cause unexpected termination. But if _logger_’s interpolation weren’t used (`"Deleted hosts: {hosts_to_delete}"`), this would explode spectacularly.

So, sorry, @dehort. I understand that you don’t want to have this pending any longer, but we’re really on a thin ice here. The logging doesn’t crash just because it behaves unexpectedly wise, and catching the error there when emitting an event, that’s just randomly putting a _try-expect_ on a first side-effect that triggers the lazy loading.

If you really need to have this merged in the current state. Please, at least, let @aleccohan to fix this in a subsequent PR, having that properly in a Sprint. This is not something we want in our codebase. It can go well with writing automatic test for this.

api/host.py
Glutexo: If we’re not going to eagerly load in this PR, please abbreviate this to

```suggestion
    hosts_to_delete = query.all()
```

Glutexo:
This has been merged although there are no tests for request ID other than default.

PR #396, opened 2019-08-13
Added delete tests

aleccohan:
Added new delete tests to match with the changes made to the delete method.


test_api.py
```diff
@@ -1227,7 +1244,21 @@ def __call__(self, e):
...
+            assert insights_id in m.events[0]
+            assert rhel_machine_id in m.events[0]
+            assert subscription_manager_id in m.events[0]
+            assert satellite_id in m.events[0]
+            assert bios_uuid in m.events[0]
+            assert fqdn in m.events[0]
+            assert external_id in m.events[0]
+            assert account in m.events[0]
+
+            for ip in ip_addresses:
```
Glutexo: Not only this has the same problem as mentioned in my previous comment, but it’s lacking the negative test too. The test verifies whether all the expected IP addresses are there, but not whether there isn’t another IP that shouldn’t be there. This applies to the MAC address test below too.

Since the lists are encoded as ordered all the way around, it’s safe to plainly compare them.

```python
self.assertEqual(ip_addresses, event["ip_addresses"])
```

If you’d want to compare them regardless of their order, you could do it like this:

```python
from collections import Counter
self.assertEqual(Counter(ip_addresses), Counter(event["ip_addresses"]))
```

But that’s an unnecessary complexity here as these are always lists and never sets.

test_api.py
```diff
@@ -1227,7 +1244,21 @@ def __call__(self, e):
...
+                assert ip in m.events[0]
+            
+            for mac in mac_addresses:
+                assert mac in m.events[0]
+
```
Glutexo: You’re testing for all canonical facts and the account number. What fields are actually required in the message though? In the tickets I saw only requests for account number (RHCLOUD-1561) and machine ID (RHCLOUD-1995). But it makes sense to include all the canonical facts.

There is no check though to the _type_ and _timestamp_ fields.

aleccohan: Yeah, The only fields required in the message are the account number and machine ID. Derek suggested that I add all of the canonical facts in case someone else in the future needed them for something.

test_api.py
```diff
@@ -1226,7 +1233,39 @@ def __call__(self, e):
...
+                "satellite_id",
+                "bios_uuid",
+                "ip_addresses",
+                "fqdn",
+                "mac_addresses",
+                "external_id",
+            }
+            self.assertEqual(set(event.keys()), expected_keys)
+
+            self.assertEqual(self.added_hosts[0].timestamp, event["timestamp"])
```
dehort: The HostWrapper does not have a timestamp field.  I think its ok if you check for the existence of the timestamp field.

test_api.py
```diff
@@ -1226,7 +1233,39 @@ def __call__(self, e):
...
+            self.assertEqual(self.added_hosts[0].type, event["type"])
```
dehort: The HostWrapper does not have a type field.  You can verify that the "type" field is set to "delete".

test_api.py
Glutexo: That makes perfect sense! It’d be nice to have this solution documented somewhere, at least in a JIRA ticket comment.

test_api.py
Glutexo:
```
>           self.assertEqual(self.added_hosts[0].timestamp, event["timestamp"])
E           AttributeError: 'HostWrapper' object has no attribute 'timestamp'
```

Glutexo: It’s also possible to patch the _datetime.utcnow_ method. Like that it’s even possible to check its format.

```python
@unittest.mock.patch("app.events.datetime", **{"utcnow.return_value": datetime.utcnow()})
def test_create_then_delete(self, datetime_mock):
    …
    with unittest.mock.patch("api.host.emit_event", new=MockEmitEvent()) as m:
        …
        timestamp_iso = datetime_mock.utcnow.return_value.isoformat()
        self.assertEqual(f"{timestamp_iso}+00:00", event["timestamp"])
        …
```

test_api.py
Glutexo:
```
>           self.assertEqual(self.added_hosts[0].type, event["type"])
E           AttributeError: 'HostWrapper' object has no attribute 'type'
```

Glutexo:
```suggestion
            self.assertEqual("delete", event["type"])
```

Glutexo:
This can be closed, now when it’s merged into #392, right?

aleccohan:
Yeah, this can be closed!  I merged it in.

PR #397, opened 2019-08-13
Synchronized Pipfile.lock with the removal of pytest

dehort:
Removing pytest from the Pipfile broke the build in the buildfactory environment.  I uninstalled pytest using pipenv uninstall and added it back to just the dev section.  This appears to have resolved the issue in my testing.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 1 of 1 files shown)


Glutexo:
Oh, I see. I really failed to add update [_Pipfile.lock_](https://github.com/RedHatInsights/insights-host-inventory/blob/master/Pipfile.lock). 🙁 Of course the build didn’t work without it. Thanks for fixing this.

PR #402, opened 2019-08-16
Replace rpm call with a literal

Glutexo:
Replaced a call to _rpm_ with a literal list with ~300 packages. This removed dependency on an external tool and thus adds support for non-RHEL platforms. It also prevents leaking of private data.

I am aware that listing this much literal data in the script is not an optimal solution. I’ll extract the literals in another pull request.


Glutexo:
Is this OK, @Jason-RH? Can it be merged?a

Jason-RH:
I don't have any issues with this, but I'm unaware of why subprocess was being used in the first place. @dehort You cool with this change?

Glutexo:
I am only guessing, but I think that the only reason was to easily grab a reasonable list of RPM packages from somewhere. This is eventually going to be replaced with a generated list with an optional mid-step of extracting that at least to a text file.

dehort:
This change looks fine to me.

Glutexo:
Great! Merging. Thanks!

PR #405, opened 2019-08-19
Unify producer tool env with config

Glutexo:
Made environment variable names and default values in the producer tool with those in the [_Config_](https://github.com/Glutexo/insights-host-inventory/blob/717cd6e88a5549dc6dcd802c5a149030cb49202f/app/config.py#L7) class. This makes it possible to run the tool with the same environment configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ run_inv_mq_service:
 	INVENTORY_LOGGING_CONFIG_FILE=logconfig.ini INVENTORY_LOG_LEVEL=DEBUG python inv_mq_service.py
 
 run_inv_mq_service_test_producer:
-	KAFKA_GROUP="inventory-mq" KAFKA_TOPIC="platform.inventory.host-ingress" python utils/kafka_producer.py
+	python utils/kafka_producer.py
 
 run_inv_mq_service_test_consumer:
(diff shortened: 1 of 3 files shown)


Jason-RH:
@Glutexo with this change you might also want to make a small update to the Makefile here: https://github.com/RedHatInsights/insights-host-inventory/blob/performance_input/Makefile#L24

Glutexo:
Good catch, @Jason-RH! I removed the variables from the [_Makefile_](https://github.com/Glutexo/insights-host-inventory/blob/performance_input_env/Makefile), because there are now working defaults. I extended the changes to the [consumer](https://github.com/Glutexo/insights-host-inventory/blob/performance_input_env/utils/kafka_consumer.py) as well.

If _KAFKA_BOOTSTRAP_SERVERS_ is pointing to a working Kafka server, then running the pipeline [producer](https://github.com/Glutexo/insights-host-inventory/blob/performance_input_env/utils/kafka_producer.py) → [MQ service](https://github.com/Glutexo/insights-host-inventory/blob/performance_input_env/inv_mq_service.py) → [consumer](https://github.com/Glutexo/insights-host-inventory/blob/performance_input_env/utils/kafka_consumer.py) works without any additional settings. 🎉

PR #409, opened 2019-08-21
Reorder imports

Glutexo:
Reordered imports so they are listed in a standard way. Split multiple imports to single lines.


Glutexo:
I don’t expect this to be any problem. Can we merge it, @Jason-RH?

PR #415, opened 2019-08-21
Pass payload builder function

Glutexo:
Remove the [_build_data_](https://github.com/RedHatInsights/insights-host-inventory/blob/169ef29ee8bb2e86ccfeef020b0a5c4d01d381f4/utils/payloads.py#L169) function that decides what kind of payload is built. Made [_build_payloads_](https://github.com/Glutexo/insights-host-inventory/blob/8b2f7fac010235f35f4168afad18e415d75f8c40/utils/payloads.py#L171) function to directly accept the builder function as an argument. This allows to easier expansion, also prevents dict mutation.

### DIFF
--- a/utils/payloads.py
+++ b/utils/payloads.py
@@ -67,75 +67,77 @@ def create_system_profile():
     }
 
 
-rhsm_payload = {
-    "account": "939054",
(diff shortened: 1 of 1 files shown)


Glutexo:
Is this OK, @Jason-RH? Can we merge it?

PR #416, opened 2019-08-22
Fix a little whitespace violation

Glutexo:
Black complains about [one file](https://github.com/Glutexo/insights-host-inventory/blob/d435af4c58fcb31d7de2eb3dc4c58e47c7436839/app/models.py) being invalid because of whitespace. Fixed.

Please don't commit anything without having _pre-commit_ hooks installed. Don’t accept any pull requests that don’t pass the _pre-commit_ checks. Thank you!

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -73,7 +73,7 @@ class Host(db.Model):
             )
 
         self.canonical_facts = canonical_facts
-        
+
         if display_name:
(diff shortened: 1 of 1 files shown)


Glutexo:
This is a trivial formatting change. Merging without approval. If this is a problem, please tell me @dehort and I’ll refrain from that in the future.

PR #421, opened 2019-08-28
added validation script to test db in cluster.

aleccohan:
Created Validation script to test all hosts in the db and see if the current validations would suffice for ALL hosts currently stored there.  In reference to the `threadctx.request_id = "-1"` I decided to use a global in order to temporarily assign a value since no actual HTTP requests are being called in this script.

### DIFF
--- a/utils/validate_hosts_for_delete.py
+++ b/utils/validate_hosts_for_delete.py
@@ -1,0 +1,38 @@
+from app import create_app
+from app import events
+from app import UNKNOWN_REQUEST_ID_VALUE
+from app.events import HostEvent
+from app.logging import get_logger
+from app.logging import threadctx
+from app.models import Host
+
+logger = get_logger("utils")
+
+
+def test_validations(host):
+    schema = HostEvent()
+    event = events.delete(host)
+    deserialized = schema.loads(event)
+    return deserialized.errors
+
+
+def main():
+    flask_app = create_app(config_name="validation_script")
+    with flask_app.app_context() as ctx:
+        threadctx.request_id = UNKNOWN_REQUEST_ID_VALUE
+        ctx.push()
+    query = Host.query
+    logger.info(f"Validating delete event for hosts.")
+    logger.info(f"Total number of hosts: %i", query.count())
+
+    number_of_errors = 0
+    for host in query.yield_per(1000):
+        host_validation_errors = test_validations(host)
+        if host_validation_errors:
+            number_of_errors += 1
+            logger.info("Output validation error host ID %s, error %s", host.id, host_validation_errors)
+    logger.info("Number of Host Validation Errors: %i", number_of_errors)
+
+
+if __name__ == "__main__":
+    main()
--- a/logconfig.ini
+++ b/logconfig.ini
@@ -1,5 +1,5 @@
 [loggers]
-keys=root, gunicorn.error, gunicorn.access, sqlalchemy.engine, app, api, tasks, mq_service, lib
+keys=root, gunicorn.error, gunicorn.access, sqlalchemy.engine, app, api, tasks, mq_service, lib, utils
 
 [handlers]
 keys=logstash
@@ -58,6 +58,12 @@ level=DEBUG
 handlers=logstash
 propagate=0
 qualname=inventory.lib
+
+[logger_utils]
+level=INFO
+handlers=logstash
+propagate=0
+qualname=inventory.utils
 
 [handler_logstash]
 class=StreamHandler


validation_script.py
```diff
@@ -0,0 +1,49 @@
...
+logger = get_logger(__name__)
+
+UNKNOWN_REQUEST_ID_VALUE = "-1"
+
+app = create_app(config_name="validation_script")
+with app.app_context() as ctx:
+    db.create_all()
+    threadctx.request_id = UNKNOWN_REQUEST_ID_VALUE
+    ctx.push()
+
```
dehort: I would move this with statement into main or closer to where its needed.  Out here its kind of difficult to track down visually.

I do not think its necessary to do the db.create_all(), set the request_id or ctx.push().  You might try removing those and see if it still works.

validation_script.py
```diff
@@ -0,0 +1,49 @@
...
+
+def TestValidations(host_list):
```
dehort: Can you change the name of this test_validations?  Or test_event_validations?  We try to follow the pep8 standards.

validation_script.py
```diff
@@ -0,0 +1,49 @@
...
+        try:
+            emit_event(events.delete(host))
+        except ValidationError as e:
+            number_of_errors += 1
+            logger.exception("Output validation error", extra={"host": host})
+            response_host_list.append(
+                {"status": 400, "title": "Bad Request", "detail": str(e.messages), "type": "unknown", "host": host}
+            )
+
+    response = {"total": len(response_host_list), "errors": number_of_errors, "data": response_host_list}
```
dehort: I would not be worried about building up a response dict like that.  We did that in the api/host.py code because we are returning a chunk of json.  Here you can likely just keep up with the number of errors and return that.

validation_script.py
aleccohan: So I tried removing the 3 statements that listed and it spit out this big old error saying how no flask application could be found.  As far as I found, the `db.create_all()` statement and `ctx.push()` were needed to get a flask application up and running.  

As for the request ID, I was running into some problems when I was first running this script that since no actual HTTP request is called the `request_id` called in `delete()` was nowhere to be found so I had to declare it there in order to get around it.

validation_script.py
```diff
@@ -0,0 +1,46 @@
...
+
+app = create_app(config_name="validation_script")
+
+
+def test_validations(host_list):
+    number_of_errors = 0
+    for host in host_list:
+        try:
+            emit_event(events.delete(host))
+        except ValidationError:
```
Glutexo: This actually doesn’t happen if a host has an invalid Insights ID, which is why we do the validation in the first place. The reason is that [_Schema.dumps_](https://marshmallow.readthedocs.io/en/stable/api_reference.html#marshmallow.Schema.dumps) doesn’t run custom validations.

It’s possible though to deserialize it back here (slow, but that’s not an issue) using [Schema.loads](https://marshmallow.readthedocs.io/en/stable/api_reference.html#marshmallow.Schema.loads). This doesn’t raise en exception though, it just populates the _errors_ property.

validation_script.py
```diff
@@ -0,0 +1,46 @@
...
+logger = get_logger(__name__)
+
+UNKNOWN_REQUEST_ID_VALUE = "-1"
+
+app = create_app(config_name="validation_script")
+
+
+def test_validations(host_list):
+    number_of_errors = 0
+    for host in host_list:
```
Glutexo:
```suggestion
    schema = HostEvent()
    for host in host_list:
        event = events.delete(host)
        deserialized = schema.loads(event)
        if deserialized.errors:
             number_of_errors += 1
             logger.info("Output validation error: host ID %s; errors %s", host.id, deserialized.errors)
```

utils/validate_hosts_for_delete.py
```diff
@@ -0,0 +1,44 @@
...
+from app.logging import get_logger
+from app.logging import threadctx
+from app.models import Host
+
+logger = get_logger(__name__)
+
+app = create_app(config_name="validation_script")
+
+
+def test_validations(host, number_of_errors):
```
Glutexo: I don’t like much the passing of the _number_of_errors_ variable just to increment it. Even more when it can be only incremented always just by one.

```python
def test_validations(host):
    schema = HostEvent()
    event = events.delete(host)
    deserialized = schema.loads(event)
    return deserialized.errors

def main():
    …
    number_of_errors = 0
    …
    for host in query.yield_per(1000):
        host_validation_errors = test_validations(host)
        if host_validation_errors:
            number_of_errors += 1
            logger.info("Output validation error: host ID %s, errors %s", host.id, host_validation_errors)
    …
```

jharting:
@aleccohan will you be able to resolve the conflict and finish this PR?

aleccohan:
@jharting Yes absolutely!  I will rebase the request immediately.

aleccohan:
@jharting merge has been completed!  Looks like tests are passing.

Glutexo:
Merged.

PR #423, opened 2019-08-29
Move delete log message to avoid uncaught exception if the host has already been deleted

dehort:


api/host.py
```diff
@@ -296,10 +296,9 @@ def delete_by_id(host_id_list):
 
     metrics.delete_host_count.inc(len(hosts_to_delete))
 
-    logger.debug("Deleted hosts: %s", hosts_to_delete)
-
     for deleted_host in hosts_to_delete:
         try:
+            logger.debug("Deleted host: %s", deleted_host)
```
Glutexo: This still doesn't go through the _except_ clause, because _logger_ has its own exception handling. If you want to avoid that, it’s necessary to move this after the _emit_event_ call.

PR #430, opened 2019-09-09
Created tests for status codes

aleccohan:
Created more test cases for the automatic race condition tests focused on the status codes returned in a series of scenarios.  See https://projects.engineering.redhat.com/browse/RHCLOUD-2222 for more details.

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -9,6 +9,7 @@ from datetime import datetime
 from datetime import timezone
 from itertools import chain
 from json import dumps
+from unittest.mock import patch
 from urllib.parse import parse_qs
 from urllib.parse import urlencode
 from urllib.parse import urlsplit
@@ -16,9 +17,11 @@ from urllib.parse import urlunsplit
 
 import dateutil.parser
 
+from api.host import _get_host_list_by_id_list
 from app import create_app
 from app import db
 from app.auth.identity import Identity
+from app.models import Host
 from app.utils import HostWrapper
 from tasks import msg_handler
 from test_utils import rename_host_table_and_indexes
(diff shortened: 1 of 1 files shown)


Glutexo:
This pull request heavily relies on the code from #427. I’ll try to review at least to some extent.

test_api.py
```diff
@@ -1264,6 +1278,30 @@ def test_delete_with_invalid_host_id(self):
...
+            # deletion should give back 200 status
+            self.delete(url, 200, return_response_as_json=False)
+
+    def test_delete_when_all_hosts_are_deleted(self):
+        url = HOST_URL + "/" + self.added_hosts[0].id + "/" + self.added_hosts[1].id
+        with patch("api.host._get_host_list_by_id_list", self.RaceCondition):
+            # deletion should give back 404 status
+            self.delete(url, 404, return_response_as_json=False)
+
+    def test_delete_when_one_host_is_deleted(self):
```
Glutexo: How does this test differ from the one above? I guess you wanted to test a situation when you are deleting two hosts, but one of them has been deleted in a different process, resulting in _200 OK_.

test_api.py
```diff
@@ -1264,6 +1278,30 @@ def test_delete_with_invalid_host_id(self):
...
+        url = HOST_URL + "/" + self.added_hosts[0].id + "/" + self.added_hosts[1].id
+        with patch("api.host._get_host_list_by_id_list", self.RaceCondition):
+            # deletion should give back 404 status
+            self.delete(url, 404, return_response_as_json=False)
+
+    def test_delete_when_no_hosts_are_found(self):
+        url = HOST_URL + "/" + generate_uuid() + "," + generate_uuid()
```
Glutexo: I think one random UUID is enough here to find nothing. But… what is this test testing anyway? If it doesn’t find any hosts, there is no way the race condition to happen.

test_api.py
```diff
@@ -1264,6 +1278,30 @@ def test_delete_with_invalid_host_id(self):
...
         self.delete(url, 400)
 
+    def test_delete_when_host_is_deleted(self):
+        url = HOST_URL + "/" + self.added_hosts[0].id
+        with patch("api.host._get_host_list_by_id_list", self.RaceCondition):
+            # deletion should give back 200 status
+            self.delete(url, 200, return_response_as_json=False)
+
+    def test_delete_when_all_hosts_are_deleted(self):
+        url = HOST_URL + "/" + self.added_hosts[0].id + "/" + self.added_hosts[1].id
```
Glutexo: There should be a comma (,) instead of a slash (/).

```suggestion
        url = HOST_URL + "/" + self.added_hosts[0].id + "," + self.added_hosts[1].id
```

test_api.py
```diff
@@ -1264,6 +1278,30 @@ def test_delete_with_invalid_host_id(self):
...
+        with patch("api.host._get_host_list_by_id_list", self.RaceCondition):
+            # deletion should give back 404 status
+            self.delete(url, 404, return_response_as_json=False)
+
+    def test_delete_when_one_host_is_deleted(self):
+        url = HOST_URL + "/" + self.added_hosts[0].id + "/" + self.added_hosts[1].id
```
```suggestion
        url = HOST_URL + "/" + self.added_hosts[0].id + "," + self.added_hosts[1].id
```

test_api.py
```diff
@@ -1264,6 +1278,30 @@ def test_delete_with_invalid_host_id(self):
...
+        url = HOST_URL + "/" + self.added_hosts[0].id
+        with patch("api.host._get_host_list_by_id_list", self.RaceCondition):
+            # deletion should give back 200 status
+            self.delete(url, 200, return_response_as_json=False)
+
+    def test_delete_when_all_hosts_are_deleted(self):
+        url = HOST_URL + "/" + self.added_hosts[0].id + "/" + self.added_hosts[1].id
+        with patch("api.host._get_host_list_by_id_list", self.RaceCondition):
+            # deletion should give back 404 status
+            self.delete(url, 404, return_response_as_json=False)
```
Glutexo: Why is 404 the right answer? By these tests we are actually defining an otherwise undocumented behavior: what should happen if all hosts to be deleted have been deleted by a different process. It doesn’t matter if it’s only one or many. From our previous discussions we concluded that both _200 OK_ and _404 Not Found_ are valid answers. But we should be at least pick one and be consistent with it. The current state of the code returns _200 OK_ and does not emit any event. That means that we should check for 200 here.

Glutexo:
Offering a [patch](https://github.com/Glutexo/insights-host-inventory/tree/delete_status_codes):

I modified the tests so they are really testing what happens in various situations. That required modifying the RaceCondition mock class, so it can delete different hosts than the executive code. Also mocked out the emit_event function, because if not all hosts are deleted by the parallel process, it is actually called causing the tests to fail.

Glutexo:
There are still some conflicts. 🙁

Glutexo:
So we only need a green light ☘️ from @dehort to merge this.

PR #431, opened 2019-09-10
Updated race condition catch

aleccohan:
Updated the race condition catch and added additional comments to explain the operation.  See RHCLOUD-2219 for additional information on the changes made.

(https://projects.engineering.redhat.com/browse/RHCLOUD-2219)

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -6,6 +6,7 @@ import sqlalchemy
 import ujson
 from flask_api import status
 from marshmallow import ValidationError
+from sqlalchemy.orm.base import instance_state
 
 from api import api_operation
 from api import metrics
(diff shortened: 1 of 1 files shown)


api/host.py
```diff
@@ -314,19 +315,23 @@ def delete_by_id(host_id_list):
...
+        # on checking the session after it has been updated by the commit()
+        # function and marked the deleted hosts as expired.  It is after this
+        # change that the host is called by a new query and, if deleted by a
+        # different process, triggers the ObjectDeletedError and is not emited.
         for deleted_host in hosts_to_delete:
-            try:
+            # Prevents ObjectDeletedError from being raised.
+            if instance_state(deleted_host).expired:
+                logger.info("Host already deleted. Delete event not emitted.")
+                pass
```
dehort: This pass looks unnecessary.  Perhaps I am missing something?

I'm not going to hold up the PR just for that though.

api/host.py
```diff
@@ -314,19 +315,22 @@ def delete_by_id(host_id_list):
...
 
+        # This process of checking for an already deleted host relies
+        # on checking the session after it has been updated by the commit()
+        # function and marked the deleted hosts as expired.  It is after this
+        # change that the host is called by a new query and, if deleted by a
+        # different process, triggers the ObjectDeletedError and is not emited.
         for deleted_host in hosts_to_delete:
-            try:
+            if instance_state(deleted_host).expired:
+                # host has been deleted do not emit
```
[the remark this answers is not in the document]
dehort: A metric would be nice.

Perhaps a comment in the if statement explaining that the attributes on the host should not be accessed would be good.

api/host.py
Glutexo: Oh. Right. 😃

Glutexo: Created a pull request for that. https://github.com/RedHatInsights/insights-host-inventory/pull/439

api/host.py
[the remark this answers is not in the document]
Glutexo: Created a pull request for that. https://github.com/RedHatInsights/insights-host-inventory/pull/440

PR #439, opened 2019-09-26
Remove unnecessary pass

Glutexo:
There was a forgotten [_pass_](https://github.com/RedHatInsights/insights-host-inventory/blob/18e3af13fe3477784682887d586ffdd914654c31/api/host.py#L327) clause, not removed after refactoring. Removed to reduce clutter.


Glutexo:
My permissions got botched somehow, so I can’t request a review. 😢 @aleccohan @dehort

PR #440, opened 2019-09-26
Add an explanatory comment for ObjectDeletedError

Glutexo:
If a race condition occurs during host delete, we [log](https://github.com/RedHatInsights/insights-host-inventory/blob/18e3af13fe3477784682887d586ffdd914654c31/api/host.py#L326) it, but we can’t log the ID of the host that was not deleted. [Added](https://github.com/RedHatInsights/insights-host-inventory/compare/master...Glutexo:host_delete_comment?expand=1#diff-a5c55b46cc9c536070cb56798c97adfaR326) a comment explaining why.


Glutexo:
@aleccohan @dehort

PR #441, opened 2019-09-27
Make examples in API spec compliant

Glutexo:
Some of the value examples in the OpenAPI [specification](https://github.com/RedHatInsights/insights-host-inventory/blob/47469ae589c2ccfba2ceba58b9cc3575c6091c71/swagger/api.spec.yaml) file didn’t adhere to the [specification](https://swagger.io/specification/). A [Schema Object](https://swagger.io/specification/#schema-object) allows only one unnamed example without metadata, its value being a literal. Simplified examples that attempted to mimic the multi-example schema of [Request Body](https://swagger.io/specification/#request-body-object) or [Media Type](https://swagger.io/specification/#media-type-object).

This caused confusion on the API consumers side, namely @karelhala from the UI. This fix should make it more clear on what values are actually produced.

### DIFF
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -670,10 +670,7 @@ components:
       description: Representation of one mounted device
       properties:
         device:
-          example:
-            device:
-              summary: device example
-              value: "/dev/fdd0"
+          example: "/dev/fdd0"
           type: string
         label:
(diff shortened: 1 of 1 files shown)


Glutexo:
May I ask for a review from the inside? @dehort

PR #443, opened 2019-10-01
Explicitly manage the db.session / transaction for the system profile consumer thread.

dehort:
This fixes an issue where the db.session/transaction for the system profile consumer thread does not get cleaned up if the database goes down during processing.  When this happens, the sql operations for that thread always fail.

I'm not overly happy with where the context manager code lives currently.  Perhaps we need a app/db_utils.py for util methods like this.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1,4 +1,5 @@
 import uuid
+from contextlib import contextmanager
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


Jason-RH:
@dehort Tried this for 2113 and it seems to work there too.

PR #444, opened 2019-10-01
updating the event handler to handle db disconnects

Jason-RH:
Adding db_context_manager to handle db disconnects


Jason-RH:
@dehort took the context manager from https://github.com/RedHatInsights/insights-host-inventory/pull/443#pullrequestreview-295764556 and copied it over to this branch. Note that this does not include retries for the mq service so any messages read from the queue while the database is down will be lost.

PR #445, opened 2019-10-02
Added branch_id and unit tests to each API method

aleccohan:
Added branch_id and unit tests to Post, Delete, Patch, Put and Get.  

See:  https://projects.engineering.redhat.com/browse/RHCLOUD-1611


test_api.py
```diff
@@ -1666,6 +1719,23 @@ def test_add_facts_to_multiple_hosts(self):
 
         self._basic_fact_test(facts_to_add, expected_facts, False)
 
+    def test_replace_and_add_facts_to_multiple_hosts_using_branch_id(self):
```
dehort: Minor nitpick.  Can the "using" be changed to "with" in the method name?  That will more closely match the other methods (i think).

PR #447, opened 2019-10-03
[WIP] Changes to add minimal support for tags

Jason-RH:
@dehort Just putting this WIP pr up. Trying to get all the initial pieces in to support tags. Feel free to suggest anything else that needs to be included/changed before merging.

### DIFF
--- a/app/serialization.py
+++ b/app/serialization.py
@@ -42,7 +42,6 @@ def serialize_host(host):
     json_dict["display_name"] = host.display_name
     json_dict["ansible_host"] = host.ansible_host
     json_dict["facts"] = _serialize_facts(host.facts)
-    json_dict["tags"] = _serialize_tags(host.tags)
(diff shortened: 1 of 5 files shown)


app/serialization.py
```diff
@@ -84,6 +89,25 @@ def _deserialize_facts(data):
...
+    tag_dict = {}
+    for tag in data:
+        tag_info = re.split("[/]", tag)
+        if tag_info[0] not in tag_dict:
+            tag_dict[tag_info[0]] = [tag_info[1]]
+        else:
+            tag_dict[tag_info[0]].append(tag_info[1])
+    return tag_dict
+
+
```
dehort: I think its possible to end up with duplicate tags.  We probably want to remove the duplicates before handing the tags to the db layer.

Jason-RH:
@dehort I'm gonna close this PR and open one against master

PR #419, opened 2019-08-28
Added request_id test

aleccohan:
Added the additional request_id test and query.all() in host.py.

PR #449, opened 2019-10-04
Generate random request ID

Glutexo:
Used the [_generate_uuid_](https://github.com/Glutexo/insights-host-inventory/blob/5f5028c359b3e1d1f18b5a707223eda7c4afc672/test_api.py#L41) function to create a random request ID for the non-empty ID test. This tests the feature a bit better than a static value.


Glutexo:

PR #450, opened 2019-10-04
Pass around a request ID instead of a full header

Glutexo:
The header composition is a common thing both request ID [tests](https://github.com/Glutexo/insights-host-inventory/blob/7a1bde96d287a3233e74100d8ae42b3ab54060f8/test_api.py#L1270). Moved it
to the common [method](https://github.com/Glutexo/insights-host-inventory/blob/7a1bde96d287a3233e74100d8ae42b3ab54060f8/test_api.py#L1230). This will also allow to compare the passed [expected](https://github.com/Glutexo/insights-host-inventory/blob/7a1bde96d287a3233e74100d8ae42b3ab54060f8/test_api.py#L1258) request ID with the actual one without having to extract it from an already composed header.

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -1227,7 +1227,7 @@ class PatchHostTestCase(PreCreatedHostsBaseTestCase):
 
 
 class DeleteHostsTestCase(PreCreatedHostsBaseTestCase):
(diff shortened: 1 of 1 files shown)


Glutexo:

PR #452, opened 2019-10-04
Use passed request ID in the test

Glutexo:
Do not use the [REQUEST_ID](https://github.com/Glutexo/insights-host-inventory/blob/3de5be6363854e88d4b1ec0135f34eeffbc888a4/test_api.py#L39) constant in the [assert](https://github.com/Glutexo/insights-host-inventory/blob/3de5be6363854e88d4b1ec0135f34eeffbc888a4/test_api.py#L1255). Like that we are essentially checking for a different value than is [passed in](https://github.com/Glutexo/insights-host-inventory/blob/3de5be6363854e88d4b1ec0135f34eeffbc888a4/test_api.py#L1269), although in fact it’s the same.


Glutexo:

PR #459, opened 2019-10-10
Delete tests refactoring: Replace Mock.new with new_callable

Glutexo:
_unittest.Mock_ can use a callable (including a class constructor) to create the mock. This is actually better than passing a manually created class instance as we are sure that the instance is not shared anywhere. Use as much from the _Mock_ capabilities as possible.

This doesn’t change any actual behavior of the tests. [_DeleteHostsTestCase_](https://github.com/Glutexo/insights-host-inventory/blob/b753df49432e7a7e44de59b9e1fe28f86d24764b/test_api.py#L1261) is still passing. 🍏

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -1309,7 +1309,7 @@ class DeleteHostsTestCase(PreCreatedHostsBaseTestCase):
         self.assertEqual(before_response["total"], 1)
 
         # Delete the host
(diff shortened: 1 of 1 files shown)


dehort:
Looks good to me.  Please resolve the conflict and we'll get it merged.

Glutexo:
@dehort Cool! Rebased.

PR #463, opened 2019-10-10
Delete tests refactoring: Split the delete tests class into three

Glutexo:
The [_DeleteHostsTestCase_](https://github.com/RedHatInsights/insights-host-inventory/blob/8d9206814a12418d14922c77e1c0baa6ef763046/test_api.py#L1261) class contains sets of rather unrelated tests.

- The race condition [tests](https://github.com/Glutexo/insights-host-inventory/blob/1dbcc9247f7026030d97778020b1bb7d4ed04571/test_api.py#L1321) with a special interceptor class.
- The event [tests](https://github.com/Glutexo/insights-host-inventory/blob/1dbcc9247f7026030d97778020b1bb7d4ed04571/test_api.py#L1274) with a special mock class.
- The invalid query [tests](https://github.com/Glutexo/insights-host-inventory/blob/1dbcc9247f7026030d97778020b1bb7d4ed04571/test_api.py#L1261) that don’t require pre-created hosts.

Split into three separate classes.

This doesn’t change the tests behavior. All tests still pass. 🍏

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -1258,7 +1258,20 @@ class PatchHostTestCase(PreCreatedHostsBaseTestCase):
                 self.patch(f"{HOST_URL}/{host_id_list}", patch_doc, 400)
 
 
-class DeleteHostsTestCase(PreCreatedHostsBaseTestCase):
+class DeleteHostsErrorTestCase(DBAPITestCase):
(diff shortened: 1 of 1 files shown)


Glutexo:
Rebased on current master. Split into _three_ classes.

PR #467, opened 2019-10-11
Basic changes needed to get tags back into the db.

Jason-RH:
Moving this PR from master to tagging branch

### DIFF
--- a/test_db_model.py
+++ b/test_db_model.py
@@ -1,7 +1,12 @@
 import uuid
 
+import pytest
+from marshmallow import ValidationError
+
 from app import db
+from app.models import _deserialize_tags
 from app.models import Host
+from app.models import HostSchema
 
 """
 These tests are for testing the db model classes outside of the api.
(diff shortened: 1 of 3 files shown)


test_db_model.py
```diff
@@ -116,3 +121,60 @@ def test_create_host_with_system_profile(flask_app_fixture):
...
+    validated_host = HostSchema(strict=True).load(host)
+
+    assert validated_host.data["tags"] == tags
+
+    tags = ["Sat/env", "AWS/env"]
+    host = {"fqdn": "fred.flintstone.com", "display_name": "display_name", "account": "00102", "tags": tags}
+    validated_host = HostSchema(strict=True).load(host)
+
+    assert validated_host.data["tags"] == tags
+
```
dehort: I think you can make this test parameterized by using the following approach:

```python
@pytest.mark.parametrize("tags", [ ["Sat/env=prod", "AWS/env=ci"], ["Sat/env", "AWS/env"] ])
def test_host_schema_valid_tags(flask_app_fixture, tags):
    host = {"fqdn": "fred.flintstone.com", "display_name": "display_name", "account": "00102", "tags": tags}
    validated_host = HostSchema(strict=True).load(host)
    assert validated_host.data["tags"] == tags
```

I don't think this test depends on the db so I think you can remove the flask_app_fixture (this should speed up the test).

test_db_model.py
```diff
@@ -116,3 +121,60 @@ def test_create_host_with_system_profile(flask_app_fixture):
...
+    assert "Sat/" in str(excinfo.value)
+
+    tags = ["Sat/env=prod", "bad_tag"]
+    host = {"fqdn": "fred.flintstone.com", "display_name": "display_name", "account": "00102", "tags": tags}
+    with pytest.raises(ValidationError) as excinfo:
+        _ = HostSchema(strict=True).load(host)
+
+    assert "bad_tag" in str(excinfo.value)
+
```
dehort: I think you can parameterize this test as well.

I don't think this test depends on the db so I think you can remove the flask_app_fixture (this should speed up the test).

test_db_model.py
```diff
@@ -116,3 +121,60 @@ def test_create_host_with_system_profile(flask_app_fixture):
...
+def test_tag_deserialization(flask_app_fixture):
+    tags = ["Sat/env=prod", "Sat/env=test", "Sat/geo=somewhere", "AWS/env=ci", "AWS/env"]
+    expected_tags = {"Sat": {"env": ["prod", "test"], "geo": ["somewhere"]}, "AWS": {"env": ["", "ci"]}}
+    deserialized_tags = _deserialize_tags(tags)
+
+    assert len(deserialized_tags["Sat"]["env"]) == len(expected_tags["Sat"]["env"])
+    assert len(deserialized_tags["Sat"]["geo"]) == len(expected_tags["Sat"]["geo"])
+    assert len(deserialized_tags["AWS"]["env"]) == len(expected_tags["AWS"]["env"])
+
```
dehort: I think its possible for these asserts to pass if the contents of the deserialized lists differs from the contents of the expected tag list. It might be nice to be a bit more strict here.

Jason-RH:
@dehort One clarification in regards to tag updating.. Are we just replacing everything in the tag column for the hosts on update and disregarding what was already there?

Jason-RH:
I can spin up an additional PR after this one to work out the update behavior (and tests).

Jason-RH:
@dehort Understood! I'll start working on that in a separate PR

PR #478, opened 2019-10-22
Request ID tests: compose request ID header in the common test

Glutexo:
Move the request ID header [composition](https://github.com/RedHatInsights/insights-host-inventory/blob/b0308b020de7c3f5b046a5f971f7aac7280f98cb/test_api.py#L1290) from the DELETE [method](https://github.com/RedHatInsights/insights-host-inventory/blob/b0308b020de7c3f5b046a5f971f7aac7280f98cb/test_api.py#L1288) to the request ID [tests](https://github.com/Glutexo/insights-host-inventory/blob/252af41c0fdd07674b50468cd5b05b8480759151/test_api.py#L1341). Like this, the logic is at a more appropriate [place].

This is a follow-up to #475.

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -1285,9 +1285,8 @@ class DeleteHostsEventTestCase(PreCreatedHostsBaseTestCase):
         def __call__(self, e):
             self.events.append(e)
 
-    def _delete(self, url, request_id):
+    def _delete(self, url, header):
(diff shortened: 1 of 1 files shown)


Glutexo:
Resolved by #476. Closing.

PR #481, opened 2019-10-22
Request ID tests: extract datetime patching to setUp

Glutexo:
Moved _datetime_ [return value](https://github.com/Glutexo/insights-host-inventory/blob/53f215846fe93eb46aaf6bcea8d8d10f3abb001b/app/events.py#L27) [patch](https://github.com/RedHatInsights/insights-host-inventory/blob/b0308b020de7c3f5b046a5f971f7aac7280f98cb/test_api.py#L1279) to the shared [__delete_](https://github.com/Glutexo/insights-host-inventory/blob/53f215846fe93eb46aaf6bcea8d8d10f3abb001b/test_api.py#L1291) method and the [timestamp](https://github.com/Glutexo/insights-host-inventory/blob/53f215846fe93eb46aaf6bcea8d8d10f3abb001b/test_api.py#L1289) to [_setUp_](https://github.com/Glutexo/insights-host-inventory/blob/53f215846fe93eb46aaf6bcea8d8d10f3abb001b/test_api.py#L1287). That gets us rid of passing around the timestamp and digging it out of the mock.

This is a follow-up to #419.

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -1276,7 +1276,6 @@ class DeleteHostsErrorTestCase(DBAPITestCase):
         self.delete(url, 400)
 
 
-@unittest.mock.patch("app.events.datetime", **{"utcnow.return_value": datetime.utcnow()})
 class DeleteHostsEventTestCase(PreCreatedHostsBaseTestCase):
     class MockEmitEvent:
         def __init__(self):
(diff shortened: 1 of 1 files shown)


Glutexo:
So many LGTM’s… sweet! ❤️

PR #482, opened 2019-10-22
Clean up the unittest imports

Glutexo:
There were some _unittest_ modules and submodules [imports](https://github.com/RedHatInsights/insights-host-inventory/blob/b0308b020de7c3f5b046a5f971f7aac7280f98cb/test_api.py#L5) in [_test_api_](https://github.com/RedHatInsights/insights-host-inventory/blob/b0308b020de7c3f5b046a5f971f7aac7280f98cb/test_api.py), but they were used quite chaotically. Cleaned that up by always importing and using the specific methods.

Induced by #419.

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -2,13 +2,14 @@
 import copy
 import json
 import tempfile
-import unittest.mock
 import uuid
 from base64 import b64encode
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


Glutexo:
Rebased and ready for review.

PR #489, opened 2019-10-31
migrate timestamp columns to use timestamptz

jharting:
https://projects.engineering.redhat.com/browse/RHCLOUD-2872

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1,6 +1,7 @@
 import uuid
 from contextlib import contextmanager
 from datetime import datetime
+from datetime import timezone
 
 from flask_sqlalchemy import SQLAlchemy
 from marshmallow import fields
@@ -48,6 +49,10 @@ def _set_display_name_on_save(context):
         return params["canonical_facts"].get("fqdn") or params["id"]
 
 
+def _time_now():
+    return datetime.now(timezone.utc)
+
+
 class Host(db.Model):
     __tablename__ = "hosts"
     # These Index entries are essentially place holders so that the
@@ -63,8 +68,8 @@ class Host(db.Model):
     account = db.Column(db.String(10))
     display_name = db.Column(db.String(200), default=_set_display_name_on_save)
     ansible_host = db.Column(db.String(255))
-    created_on = db.Column(db.DateTime, default=datetime.utcnow)
-    modified_on = db.Column(db.DateTime, default=datetime.utcnow, onupdate=datetime.utcnow)
+    created_on = db.Column(db.DateTime(timezone=True), default=_time_now)
+    modified_on = db.Column(db.DateTime(timezone=True), default=_time_now, onupdate=_time_now)
     facts = db.Column(JSONB)
     tags = db.Column(JSONB)
     canonical_facts = db.Column(JSONB)
--- a/app/serialization.py
+++ b/app/serialization.py
@@ -1,3 +1,5 @@
+from datetime import timezone
+
 from app.exceptions import InputFormatException
 from app.models import Host as Host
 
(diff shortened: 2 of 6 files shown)


migrations/alembic.ini
```diff
@@ -25,7 +24,7 @@ handlers = console
 qualname =
 
 [logger_sqlalchemy]
-level = WARN
+level = INFO
```
Glutexo: I don’t think this belongs in this PR.

app/serialization.py
```diff
@@ -38,8 +40,10 @@ def serialize_host(host):
     json_dict["display_name"] = host.display_name
     json_dict["ansible_host"] = host.ansible_host
     json_dict["facts"] = _serialize_facts(host.facts)
-    json_dict["created"] = host.created_on.isoformat() + "Z"
-    json_dict["updated"] = host.modified_on.isoformat() + "Z"
+    # without astimezone(timezone.utc) the isoformat() method does not include timezone offset even though iso-8601
+    # requires it
+    json_dict["created"] = host.created_on.astimezone(timezone.utc).isoformat()
```
Glutexo: This slightly changes the format from _Z_ to _+00:00_. It’s still a valid and equivalent _ISO 8601_ date, yet it may be a breaking change for dumb clients and tests.

jharting: @Glutexo I am aware. Hopefully there are not a lot of dumb clients out there.

migrations/alembic.ini
jharting: @Glutexo this is intentional in order to be able to monitor the progress of the migration as it runs. As it is not directly related to the schema change it is committed in a separate commit.

jharting:
@Glutexo even though this works as is internally this change is necessary to allow for streaming of data to other systems (xjoin).

Glutexo:
I neither think that _timestamp with timezone_ is appropriate here, nor it makes any assumptions in other systems.  However, if you believe it makes things easier to change the data type, then ok.

I’d just try to figure out where the one hour shift comes from. If we want to make this change, then let’s at least be sure it works correctly.

migrations/alembic.ini
Glutexo: To me, the _INFO_ level seems maybe a bit too verbose, but no problem. Let’s have it like that.

jharting:
@Glutexo if we store it without timezone then other systems have to assume the timestamp they received is in UTC. If we embed the timezone information then they do not have to.

I am going to verify the 1 hour offset but you identified.

app/models.py
```diff
@@ -63,8 +63,8 @@ class Host(db.Model):
     account = db.Column(db.String(10))
     display_name = db.Column(db.String(200), default=_set_display_name_on_save)
     ansible_host = db.Column(db.String(255))
-    created_on = db.Column(db.DateTime, default=datetime.utcnow)
-    modified_on = db.Column(db.DateTime, default=datetime.utcnow, onupdate=datetime.utcnow)
+    created_on = db.Column(db.DateTime(timezone=True), default=datetime.utcnow)
+    modified_on = db.Column(db.DateTime(timezone=True), default=datetime.utcnow, onupdate=datetime.utcnow)
```
Glutexo: This is the culprit. `datetime.utcnow` is time zone naïve – the _datetime_ object doesn’t conain the time zone information. SQLAlchemy then treats is as a local time and translates that to UTC. It’s necessary to

- either pass a time zone aware _datetime_ object, be it UTC or local
- or use _now_ instead of _utcnow_ to pass a naïve local time.

app/models.py
```diff
@@ -63,8 +63,8 @@ class Host(db.Model):
     account = db.Column(db.String(10))
     display_name = db.Column(db.String(200), default=_set_display_name_on_save)
     ansible_host = db.Column(db.String(255))
-    created_on = db.Column(db.DateTime, default=datetime.utcnow)
-    modified_on = db.Column(db.DateTime, default=datetime.utcnow, onupdate=datetime.utcnow)
+    created_on = db.Column(db.DateTime(timezone=True), default=datetime.utcnow)
```
Glutexo:
```suggestion
    created_on = db.Column(db.DateTime(timezone=True), default=datetime.now)
```

app/models.py
Glutexo:
```suggestion
    modified_on = db.Column(db.DateTime(timezone=True), default=datetime.now, onupdate=datetime.now)
```

PR #497, opened 2019-11-01
host tag support

jharting:
https://projects.engineering.redhat.com/browse/RHCLOUD-2434


api/host.py
```diff
@@ -196,6 +225,8 @@ def get_host_list(
         query = find_hosts_by_hostname_or_id(current_identity.account_number, hostname_or_id)
     elif insights_id:
         query = find_hosts_by_canonical_facts(current_identity.account_number, {"insights_id": insights_id})
+    elif tags:
+        query = find_hosts_by_tag(current_identity.account_number, tags)
```
jharting: This should not be exclusive. It must be possible to combine `tags` filter with e.g. `hostname_or_id` filter

swagger/api.spec.yaml
```diff
@@ -591,6 +694,10 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/FactSet'
+        tags:
+          type: array
+          items:
+            type: string
```
jharting: This should be structured tag representation, not string

swagger/api.spec.yaml
```diff
@@ -332,6 +393,48 @@ components:
...
+        per_page:
+          description: A page size – a number of entries per single page.
+          type: integer
+        results:
+          description: The list of tags on the systems
+          type: object
+          additionalProperties:
+            type: array
+            items:
+              type: string
```
jharting: This should be structured tag representation, not string

test_unit.py
```diff
@@ -319,5 +320,67 @@ def test_order_by_only_how_raises_error(self):
             _params_to_order_by(Mock(), order_how="ASC")
 
 
+class TagUtilsTestCase(TestCase):
```
jharting: Let's add at least all the variants listed in the tag representation document (e.g. tags with special characters) to have decent coverage

app/models.py
```diff
@@ -49,6 +50,32 @@ def _set_display_name_on_save(context):
...
+def _split_tag(tag):
+    try:
+        namespace, t_key, t_value = re.split("[/ =]", tag)
+    except ValueError:
+        namespace, t_key = re.split("[/]", tag)
+        t_value = ""
+    return (namespace, t_key, t_value)
+
+
+def _deserialize_tags(data):
```
jharting: Instead of this we should be using the Tag util methods, in particular `create_nested_from_tags`

swagger/api.spec.yaml
```diff
@@ -250,6 +259,58 @@ paths:
           description: Invalid request.
         '404':
           description: Host not found.
+  '/hosts/{host_id_list}/tags':
```
jharting: Does it make sense to support a list of hosts here?

jharting:
This PR has conflicts that should be resolved

jharting:
Moreover, `utils/payloads.py` should be extended to have some tag information

swagger/api.spec.yaml
Glutexo: Probably not much, but at least all those host ID endpoints behave the same.

app/models.py
ryandillinfelton: create_nested_from_tags works a little differently. I'll add a new static method create_nested_from_string_tags that can handle it

jharting: @ryandillinfelton I don't think we should be using the string representation of tags here

swagger/api.spec.yaml
ryandillinfelton: You mean like with the pattern? 
```
          items:
              type: string
              pattern: '^\w+\/\w+=\w+$|^\w+\/\w+$|^\w+$|^\w+=\w+$'
```

jharting: No, an array of objects each of them having a namespace, key and value properties

ryandillinfelton: Okay, I see what you mean now

swagger/api.spec.yaml
ryandillinfelton: Changed

swagger/api.spec.yaml
ryandillinfelton: Changed

swagger/api.spec.yaml
```diff
@@ -332,6 +389,57 @@ components:
...
+          type: object
+          additionalProperties:
+            type: array
+            items: 
+              $ref: '#/components/schemas/StructuredTag'
+    StructuredTag:
+      type: object
+      properties:
+        namespace:
+          type: string
```
jharting: namespace and value should be `nullable: true`, otherwise looks good

ryandillinfelton: Since we can't query for hosts with a tag without a namespace, and from my conversations with you and Shaun it looks like we don't want to add that functionality, does it make sense to allow users to try to? Maybe I'm missing something, but I feel like I'm getting conflicting information about what kind of tags we should be able to filter by.

app/models.py
ryandillinfelton: Changed to use structured tag format for creating tags on a host

api/host.py
ryandillinfelton: New commit makes this work and adds some tests for it

api/host.py
```diff
@@ -372,3 +419,74 @@ def update_facts_by_namespace(operation, host_id_list, namespace, fact_dict):
...
+
+# returns counts in format [{id: count}, {id: count}]
+def _count_tags(host_list):
+    counts = {}
+
+    for host in host_list:
+        host_tag_count = 0
+        for namespace in host.tags:
+            for tag in host.tags[namespace]:
+                host_tag_count += len(host.tags[namespace][tag])
```
jharting: if a tag does not have a value it should still be counted as 1

app/models.py
```diff
@@ -53,6 +54,15 @@ def _time_now():
     return datetime.now(timezone.utc)
 
 
+def _split_tag(tag):
```
jharting: does not seem to be used

app/models.py
```diff
@@ -100,8 +111,41 @@ def __init__(
...
+    #     json_dict["display_name"] = self.display_name
+    #     json_dict["ansible_host"] = self.ansible_host
+    #     json_dict["facts"] = Facts.to_json(self.facts)
+    #     json_dict["created"] = self.created_on.isoformat() + "Z"
+    #     json_dict["updated"] = self.modified_on.isoformat() + "Z"
+    #     return json_dict
+
+    # def to_system_profile_json(self):
+    #     json_dict = {"id": str(self.id), "system_profile": self.system_profile_facts or {}}
+    #     return json_dict
```
jharting: ??

app/models.py
```diff
@@ -263,6 +320,12 @@ class FactsSchema(Schema):
     facts = fields.Dict()
 
 
+class TagsSchema(Schema):
+    namespace = fields.Str()
+    key = fields.Str()
+    value = fields.Str()
```
jharting: value should be nullable

app/utils.py
```diff
@@ -132,3 +142,183 @@ def to_json(self):
...
+    @value.setter
+    def value(self, value):
+        self.__data["value"] = value
+
+    def _split_string_tag(self, string_tag):
+        namespace = None
+        key = None
+        value = None
+
+        if re.match(r"\S+\/\S+=\S+", string_tag):  # NS/key=value
```
jharting: This should be consistent with the regex we use in schema definition. In addition, it would be less error-prone to use regex groups here rather than splitting again on the next line

swagger/api.spec.yaml
jharting: Good point. We will need to support no-namespace tags in the future but we do not need to now. As long as we don't do anything that would make introducing no-namespace tags hard in the future I am fine with that.

app/models.py
ryandillinfelton: Must've missed removing this. I kept it around during the merge to reference

api/host.py
```diff
@@ -98,13 +99,55 @@ def _add_host(input_host):
     return add_host(input_host, update_system_profile=False)
 
 
+def find_host_by_canonical_facts(account_number, canonical_facts):
```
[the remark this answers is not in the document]
ryandillinfelton: Removed it from host.py. Good catch

api/host.py
```diff
@@ -98,13 +99,55 @@ def _add_host(input_host):
...
+def find_hosts_by_tag(account_number, tags, query):
+    """
+    Returns all of the hosts with the tag/tags
+    """
+    logger.debug("find_host_by_tag(%s)", tags)
+
+    hosts = _tags_host_query(account_number, tags, query)
+
+    if hosts:
+        logger.debug("found the host with those ids: %s", hosts)
```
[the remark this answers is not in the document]
ryandillinfelton: I've updated the function to make that more clear

app/models.py
ryandillinfelton: removed

api/host.py
ryandillinfelton: Fixed, and added a test for this situation

test_api.py
```diff
@@ -1193,9 +1194,37 @@ def setUp(self):
...
-            ("host3", generate_uuid(), "host2.domain.test"),
+            (
+                "host1",
+                generate_uuid(),
+                "host1.domain.test",
+                [
+                    {"namespace": "NS1", "key": "key1", "value": "val1"},
+                    {"namespace": "NS1", "key": "key2", "value": "val1"},
+                    {"namespace": "SPECIAL", "key": "tag", "value": "ToFind"},
+                    {"namespace": "no", "key": "value"},
```
jharting: `value` should be specified here with its value being null/None

test_api.py
```diff
@@ -2144,5 +2345,122 @@ def test_version(self):
...
+        for host, tags in zip(host_list, self.tags_list):
+            expected_response[str(host.id)] = [tag.data() for tag in tags]
+
+        url_host_id_list = self._build_host_id_list_for_url(host_list)
+
+        test_url = f"{HOST_URL}/{url_host_id_list}/tags?order_by=updated&order_how=ASC"
+        response = self.get(test_url, 200)
+
+        # TODO: remove
+        print("expected results")
```
jharting: print statements

test_api.py
```diff
@@ -2144,5 +2345,122 @@ def test_version(self):
         assert response["version"] is not None
 
 
+class TagTestCase(TagsPreCreatedHostsBaseTestCase, PaginationBaseTestCase):
```
jharting: this should be extended to also test pagination (`page`, `per_page` attributes)

app/serialization.py
```diff
@@ -32,12 +33,15 @@ def deserialize_host(raw_data):
...
     facts = _deserialize_facts(validated_data.get("facts"))
+    tags = _deserialize_tags(validated_data.get("tags"))
     return Host(
         canonical_facts,
         validated_data.get("display_name", None),
         validated_data.get("ansible_host"),
         validated_data.get("account"),
         facts,
+        tags,
+        # tags,
```
jharting: ?

jharting:
Here's a summary of what's missing to get this merged:
* [ ] fix pagination for `/tags` and `/tags/count` and add tests
* [ ] use equivalent regex for tags query param validation and actual parsing - simplify the parsing logic by using regex groups
* [ ] expand the payload in `utils/payloads.py` to include sample tags

PR #506, opened 2019-11-06
collect and export HTTP metrics from flask

jharting:
https://projects.engineering.redhat.com/browse/RHCLOUD-2880

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -17,7 +17,7 @@ flask-migrate = "~=2.5.2"
 itsdangerous = "~=1.1.0"
 request = "==2019.4.13"  # Not semver
 gunicorn = "~=19.9.0"
-connexion = {version = "~=2.3.0", extras = ["swagger-ui"]}
+connexion = {version = "~=2.3.0",extras = ["swagger-ui"]}
 pyyaml = "~=5.1.2"
 prometheus-client = "~=0.7.1"
 logstash-formatter = "~=0.5.17"
@@ -27,6 +27,7 @@ ujson = "==1.35"  # Not semver
 watchtower = "~=0.6.0"
 boto3 = "~=1.9.183"
 kafka-python = "~=1.4.6"
+prometheus-flask-exporter = "~=0.11.0"
 
 [dev-packages]
 pytest = "~=5.0.1"
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -3,6 +3,7 @@ import yaml
 from connexion.resolver import RestyResolver
 from flask import jsonify
 from flask import request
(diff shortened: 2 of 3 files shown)


app/__init__.py
```diff
@@ -96,4 +97,13 @@ def set_request_id():
 
     payload_tracker.init_payload_tracker(app_config, producer=payload_tracker_producer)
 
+    # HTTP request metrics
+    if config_name != "testing":
+        PrometheusMetrics(
+            flask_app,
```
Glutexo: Maybe add `path=None` here? The _/metrics_ endpoint is already [registered](https://github.com/jharting/insights-host-inventory/blob/913ebce7125596e9b78563ae7e20a2c36c206c3c/app/__init__.py#L73), so there is no need to do it twice.

app/__init__.py
```diff
@@ -96,4 +97,13 @@ def set_request_id():
...
+            defaults_prefix="inventory",
+            group_by="url_rule",
+            excluded_paths=["^/metrics$", "^/health$", "^/version$", "^/favicon.ico$"],
```
Glutexo: This is a regex, so the period in `^/favicon.ico$` should be escaped.
```suggestion
            excluded_paths=["^/metrics$", "^/health$", "^/version$", "^/favicon\\.ico$"],
```

app/__init__.py
```diff
@@ -96,4 +97,13 @@ def set_request_id():
 
     payload_tracker.init_payload_tracker(app_config, producer=payload_tracker_producer)
 
+    # HTTP request metrics
+    if config_name != "testing":
```
Glutexo: These Flask metrics are disabled in testing, but the [other ones](https://github.com/jharting/insights-host-inventory/blob/913ebce7125596e9b78563ae7e20a2c36c206c3c/app/__init__.py#L73) are not. Is there a reason for that?

jharting: Yes, the metrics registry is global but the app is not. Therefore, if we register them in testing they are re-registered for every test run which causes problems

Pipfile
```diff
@@ -17,7 +17,7 @@ flask-migrate = "~=2.5.2"
 itsdangerous = "~=1.1.0"
 request = "==2019.4.13"  # Not semver
 gunicorn = "~=19.9.0"
-connexion = {version = "~=2.3.0", extras = ["swagger-ui"]}
+connexion = {version = "~=2.3.0",extras = ["swagger-ui"]}
```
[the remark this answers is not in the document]
jharting: This is the formatting that pipenv prefers ¯\_(ツ)_/¯

PR #508, opened 2019-11-07
Validate hosts when deserializing

Glutexo:
Move the schema from the repository host add logic layer to the deserialization one. As a result, every work with an input host dictionary is done after it’s been validated. That includes an account number comparison. Ergo, if the account number is invalid, it’s not being even compared to the one in the identity header.

Fixes [RHCLOUD-3018](https://projects.engineering.redhat.com/browse/RHCLOUD-3018).

Added some tests verifying that the validation actually happens and that it happens in the right place. That required amending all the currently invalid host inputs in tests.

Steps to reproduce:

1. Try to add a host with an invalid account number.

```json
[
    {
        "account": "",
        "fqdn": "test"
    }
]
```

2. See that a Marshmallow validation exception (_Bad Request) is presented instead of our logic one (Invalid request).

```json
{
    "total": 1,
    "errors": 1,
    "data": [
        {
            "detail": "{'account': ['Length must be between 1 and 10.']}",
            "status": 400,
            "title": "Bad Request",
            "type": "about:blank",
            "host": {
                "account": "",
                "fqdn": "test"
            }
        }
    ]
}
```

not

```json
{
    "total": 1,
    "errors": 1,
    "data": [
        {
            "detail": "The account number associated with the user does not match the account number associated with the host",
            "status": 400,
            "title": "Invalid request",
            "type": "about:blank",
            "host": {
                "account": "abc124",
                "fqdn": "test"
            }
        }
    ]
}
```

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -22,6 +22,7 @@ from app.models import PatchHostSchema
 from app.payload_tracker import get_payload_tracker
 from app.payload_tracker import PayloadTrackerContext
 from app.payload_tracker import PayloadTrackerProcessingContext
+from app.serialization import deserialize_host
 from app.serialization import serialize_host
 from app.serialization import serialize_host_system_profile
 from lib.host_repository import _canonical_facts_host_query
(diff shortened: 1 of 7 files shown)


Glutexo:
Fixed that failing test. Removed it, because it was no longer relevant.

Glutexo:
One more review? @ryandillinfelton ?

PR #510, opened 2019-11-08
Add culling columns

Glutexo:
Added _stale_timestamp_ (timestamp with time zone) and _reporter_ (varchar 255) to the [database](https://github.com/Glutexo/insights-host-inventory/blob/08abebd01fe8a2b15c6081020322644442545637/migrations/versions/c6bfa7ebeaee_add_stale_timestamp_and_reporter_columns.py) and to the SQLAlchemy [model](https://github.com/Glutexo/insights-host-inventory/blob/08abebd01fe8a2b15c6081020322644442545637/app/models.py#L56). Both are nullable, because they are not filled for the existing records.

The _stale_timestamp_ includes time zone information for it to be aligned with other timestamp columns. Although this is not necessary, it makes it somehow easier to pump the data to the cross join database.

The new fields are not in the [HostWrapper](https://github.com/Glutexo/insights-host-inventory/blob/08abebd01fe8a2b15c6081020322644442545637/app/utils.py#L4) class, because it is only used for the [API tests](https://github.com/Glutexo/insights-host-inventory/blob/08abebd01fe8a2b15c6081020322644442545637/test_api.py). The API doesn’t use those columns yet.

No tests as there is nothing to test. There are even no pure model tests, no nothing to add there.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -74,6 +74,8 @@ class Host(db.Model):
     tags = db.Column(JSONB)
     canonical_facts = db.Column(JSONB)
     system_profile_facts = db.Column(JSONB)
+    stale_timestamp = db.Column(db.DateTime(timezone=True))
+    reporter = db.Column(db.String(255))
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


Glutexo:
Assigned @dehort as he may want to look at the database migration. Requested by @shdunning.

Glutexo:
Cool! Merging. Thanks!

PR #511, opened 2019-11-08
Test basic model properties

Glutexo:
Added a few database model tests that verify column presence, type and default values.

Triggered by #510 a #489. There are no tests checking that there are correct columns in the database and the model is correctly set.


Glutexo:
Added the new fields from #510.

PR #515, opened 2019-11-12
Culling

Glutexo:


Glutexo:
Fixed failing unit [test](https://github.com/RedHatInsights/insights-host-inventory/blob/9b962fc24db7a2e18b2003c36a965f9d48a773ae/test_unit.py#L593). Added [a new one](https://github.com/RedHatInsights/insights-host-inventory/blob/9b962fc24db7a2e18b2003c36a965f9d48a773ae/test_unit.py#L781) for the default values.

Glutexo:
Added the computed timestamps to the responses. Added explicit tests.

ryandillinfelton:
If we're not trying to add testing for MQ in this PR I think we're ready for review on this. I've got MQ accepting `stale_timestamp` and outputting `stale_timestamp`, `stale_warning_timestamp`, and `culled_timestamp`. It holds up in my limited manual testing

Glutexo:
[Added](https://github.com/RedHatInsights/insights-host-inventory/pull/515/commits/c210b6538279e5847a8dd6e0f9761efb2a0e05b5) filtering by staleness state to the REST API. It uses states suggested by @jharting: _fresh_, _stale_, _stale_warning_ and _culled_ instead of the originally intended _stale_ boolean parameter. Added an _unknown_ state to get hosts that don’t have a _stale_timestamp_ filled. Default filtering is _fresh, stale, unknown_. The state is not passed as a field and clients must compute it manually from the timestamps.

api/host.py
```diff
@@ -112,6 +116,21 @@ def find_hosts_by_tag(account_number, string_tags, query):
     return query.filter(Host.tags.contains(tags_to_find))
 
 
+def find_hosts_by_staleness(states, query):
+    now = datetime.now(timezone.utc)
+    stale_warning_timestamp = Host.stale_timestamp + timedelta(weeks=1)
+    culled_timestamp = Host.stale_timestamp + timedelta(weeks=2)
```
jharting: let's make these configurable with env variables

swagger/api.spec.yaml
```diff
@@ -388,6 +390,21 @@ components:
...
+      required: false
+      schema:
+        type: array
+        items:
+          type: string
+          enum:
+            - fresh
+            - stale
+            - stale_warning
+            - culled
```
jharting: Should we be exposing the culled state? My understanding is that systems in this state are effectively gone and we should not returning anything about them unless they are brought back to some other state.

swagger/api.spec.yaml
```diff
@@ -388,6 +390,21 @@ components:
       description: >-
         Direction of the ordering, defaults to ASC for display_name and to DESC for
         updated
+    stalenessParam:
+      name: staleness
+      in: query
+      required: false
+      schema:
+        type: array
```
jharting: can we declare the default value here using openapi `default`?

api/host.py
Glutexo: Any idea on format or at least granularity? Is days enough?

swagger/api.spec.yaml
Glutexo: Is there any harm in exposing such hosts? I have no problem in hiding this state, just asking.

swagger/api.spec.yaml
Glutexo: I’ll try.

api/host.py
jharting: I think days as the basic unit will work fine for the real use-cases. For testing I suppose it might be useful to support something more granular but I may be wrong.

Glutexo: For the sake of simplicity I just used days. Tests use one hour more to enter the _stale_warning_ or _culled_ state.

Glutexo:
Made the culling offsets configurable. This required some deeper changes to make the configuration available wherever necessary.

Glutexo:
Removed the exposition of the _culled_ state.

ryandillinfelton:
I filled in two of the three empty tests. Per a conversation with @dehort I believe the final unfinished test should be finished later. It's testing a relatively benign bug that we haven't fixed yet. So I think the tests should be added along with the fix for that bug. https://projects.engineering.redhat.com/browse/RHCLOUD-3503

Glutexo:
Extracted the culling timestamp computation logic to a separate [module](https://github.com/RedHatInsights/insights-host-inventory/blob/1c4e0bfd50c33523e3b8b05899c447ee733cc1e6/app/culling.py). As a result, the  serialization code no longer depends on the whole [Config](https://github.com/RedHatInsights/insights-host-inventory/blob/1c4e0bfd50c33523e3b8b05899c447ee733cc1e6/app/config.py#L7) object.

PR #516, opened 2019-11-13
Add host to docs to match code in egress.py

pilhuhn:
Signed-off-by: Heiko W. Rupp <hrupp@redhat.com>


pilhuhn:
That is actually not right. Will re-open another one

PR #526, opened 2019-11-22
RHCLOUD-2462 Tag listing API

jharting:
https://projects.engineering.redhat.com/browse/RHCLOUD-2462


jharting:
There are still a few TODO which are waiting for https://github.com/RedHatInsights/insights-host-inventory/pull/532 Other than those the PR is ready for review

Glutexo:
Broke it! 🎉

```
  File "/Users/stomsa/Vývoj/Insights/Repozitáře/insights-host-inventory/api/tag.py", line 82, in get_tags
    "name": f".*{re.escape(parse.quote_plus(tag_name, safe=[]))}.*"
  File "/Users/stomsa/.pyenv/versions/3.6.7/lib/python3.6/urllib/parse.py", line 805, in quote_plus
    string = quote(string, safe + space, encoding, errors)
```

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+                "x-rh-insights-request-id": "353b230b-5607-4454-90a1-589fbd61fde9",
+            },
+        )
+
+    def test_query_variables_default(self, graphql_query):
+        self.get(TAGS_URL, 200)
+
+        graphql_query.assert_called_once()
+        variables = graphql_query.call_args[0][1]["variables"]
+        self.assertEqual(variables, {"order_by": "tag", "order_how": "ASC", "limit": 50, "offset": 0})
```
Glutexo: It’s also possible to use _unittest.mock.Mock.assert_called_once_with_ together with its _ANY_ object.

```python
        from unittest.mock import ANY
        variables = {"order_by": "tag", "order_how": "ASC", "limit": 50, "offset": 0}
        graphql_query.assert_called_once_with(ANY, {"query": ANY, "variables": variables}, ANY)
```

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+
+        graphql_query.assert_called_once()
+        variables = graphql_query.call_args[0][1]["variables"]
+        self.assertEqual(
+            variables,
+            {
+                "order_by": "tag",
+                "order_how": "ASC",
+                "limit": 50,
+                "offset": 0,
```
Glutexo: If we are testing only variables, we can limit that further to only the relevant keys.

```python
                "order_by": ANY,
                "order_how": ANY,
                "limit": ANY,
                "offset": ANY,
```

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+                "hostFilter": {"AND": [{"tag": {"namespace": "insights-client", "key": "os", "value": "fedora"}}]},
+            },
+        )
+
+    def test_query_variables_tags_complex(self, graphql_query):
+        self.get(
+            f"{TAGS_URL}?tags=Sat/env=prod&tags=insights-client%2Fspecial%252Fkey%25CE%2594with%25C4%258Dhars"
+            "%3Dspecial%252Fvalue%25CE%2594with%25C4%258Dhars%2521",
+            200,
+        )
```
Glutexo: Would it be possible to encode this programmatically so the query string is more readable? Especially as double URL encoding takes place here, it’d be nice to see what’s actually going on here.

```python
        from urllib.parse import quote
        complex_tag_namespace = "insights_client"
        complex_tag_key = "special/keyΔwithčhars"
        complex_tag_value = "special/valueΔwithčhars!"
        complex_tag_composed = (
            quote(complex_tag_namespace, safe="") + "/"
            + quote(complex_tag_key, safe="") + "="
            + quote(complex_tag_value, safe="")
        )
        complex_tag_encoded = quote(complex_tag_composed, safe="")
        self.get(f"{TAGS_URL}?tags=Sat/env=prod&tags={complex_tag_encoded}", 200)
```

The variables can be then used below in the expected call assert.

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+
+        graphql_query.assert_called_once()
+        variables = graphql_query.call_args[0][1]["variables"]
+        self.assertEqual(
+            variables,
+            {
+                "order_by": "tag",
+                "order_how": "ASC",
+                "limit": 50,
+                "offset": 0,
```
Glutexo: Dtto _ANY_.

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+                "filter": {"name": ".*\\%CE\\%94with\\%C4\\%8Dhar\\%21\\%2F\\%7E\\%7C\\%2B.*"},
```
Glutexo: I wonder whether not reuse the escape/quote machinery here. We want to test those Python function are used correctly, not necessarily their inner workings.

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+                                "key": "special/keyΔwithčhars",
+                                "value": "special/valueΔwithčhars!",
+                            }
+                        },
+                    ]
+                },
+            },
+        )
+
+    def test_query_variables_tag_name(self, graphql_query):
```
Glutexo: This test doesn’t catch a failure case. If you query `f"{TAGS_URL}?tag_name=.+"`, 500 happens. See my comment in the code.

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+            200,
+            extra_headers={"x-rh-insights-request-id": "353b230b-5607-4454-90a1-589fbd61fde9", "foo": "bar"},
+        )
+
+        graphql_query.assert_called_once()
+        headers = graphql_query.call_args[0][2]
+        self.assertEqual(
+            headers,
+            {
+                "x-rh-identity": "eyJpZGVudGl0eSI6IHsiYWNjb3VudF9udW1iZXIiOiAiMDAwNTAxIn19",
```
Glutexo: I‘d rather reuse somehow the already computed [value](https://github.com/jharting/insights-host-inventory/blob/3f93ad0ed717f11a078f97ba4e0361b12a639235/test_api.py#L102) than having it as a literal here.

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+                                "value": "special/valueΔwithčhars!",
+                            }
+                        },
+                    ]
+                },
+            },
+        )
+
+    def test_query_variables_tag_name(self, graphql_query):
+        self.get(f"{TAGS_URL}?tag_name=%CE%94with%C4%8Dhar%21%2F%7E%7C%2B", 200)
```
Glutexo: Dtto encoding.

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+        )
+
+    def test_query_variables_ordering_dir(self, graphql_query):
+        self.get(f"{TAGS_URL}?order_how=DESC", 200)
+
+        graphql_query.assert_called_once()
+        variables = graphql_query.call_args[0][1]["variables"]
+        self.assertEqual(variables, {"order_by": "tag", "order_how": "DESC", "limit": 50, "offset": 0})
+
+    def test_query_variables_ordering_by(self, graphql_query):
```
Glutexo: Parametrize for both _count_ and _tag_.

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+            {
+                "order_by": "tag",
+                "order_how": "ASC",
+                "limit": 50,
+                "offset": 0,
+                "filter": {"name": ".*\\%CE\\%94with\\%C4\\%8Dhar\\%21\\%2F\\%7E\\%7C\\%2B.*"},
+            },
+        )
+
+    def test_query_variables_ordering_dir(self, graphql_query):
```
Glutexo: Parametrize for both _DESC_ and _ASC_? 🤷🏻‍♂️

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+        self.get(f"{TAGS_URL}?order_how=DESC", 200)
+
+        graphql_query.assert_called_once()
+        variables = graphql_query.call_args[0][1]["variables"]
+        self.assertEqual(variables, {"order_by": "tag", "order_how": "DESC", "limit": 50, "offset": 0})
```
Glutexo: Dtto _ANY_. We’re testing only the _order_how_ key here.

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+
+    def test_query_variables_ordering_by(self, graphql_query):
+        self.get(f"{TAGS_URL}?order_by=count", 200)
+
+        graphql_query.assert_called_once()
+        variables = graphql_query.call_args[0][1]["variables"]
+        self.assertEqual(variables, {"order_by": "count", "order_how": "ASC", "limit": 50, "offset": 0})
```
Glutexo: Dtto _ANY_. We’re testing only the _order_by_ key here.

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+
+    def test_query_variables_pagination(self, graphql_query):
```
Glutexo: - Some parametrization?
- More complex case, where some real _page/per_page_ → _limit/offset_ translation happens.
- Test for invalid cases?
- Test for 404 if page number too high.

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+        self.get(f"{TAGS_URL}?per_page=2&page=2", 200)
+
+        graphql_query.assert_called_once()
+        variables = graphql_query.call_args[0][1]["variables"]
+        self.assertEqual(variables, {"order_by": "tag", "order_how": "ASC", "limit": 2, "offset": 2})
```
Glutexo: Dtto _ANY_. We’re testing only the _limit_ and _offset_ here.

api/tag.py
```diff
@@ -0,0 +1,107 @@
...
+    logger.debug(f"variables {variables}")
+    payload = {"query": TAGS_QUERY, "variables": variables}
+
+    headers = _get_forwarded_headers(request.headers)
+
+    # TODO: reuse graphql client code
+    # TODO: actual URL
+    response = graphql_query("http://localhost:4000/graphql", payload, headers)
+    data = response["data"]["hostTags"]
+
```
Glutexo: There is no check whether we don’t have an empty result because we provided too high page number. In the Inventory it is customary to abort with 404 in such case.

test_api.py
```diff
@@ -3045,5 +3048,155 @@ def test_tags_count_pagination(self):
...
+                "total": 3,
+                "count": 3,
+                "page": 1,
+                "per_page": 50,
+                "results": [
+                    {"count": 3, "tag": {"key": "env", "namespace": "Sat", "value": "prod"}},
+                    {"count": 2, "tag": {"key": "database", "namespace": "insights-client", "value": None}},
+                    {"count": 2, "tag": {"key": "os", "namespace": "insights-client", "value": "fedora"}},
+                ],
+            },
```
Glutexo: Can’t we reuse the _self.RESPONSE_ object?

```python
        response = self.RESPONSE["data"]["hostTags"]
        self.assertEqual(
            result,
            {
                "total": response["meta"]["total"],
                "count": response["meta"]["count"],
                "page": 1,
                "per_page": 50,
                "results": response["data"],
            },
        )
```

api/tag.py
```diff
@@ -0,0 +1,107 @@
...
+        "per_page": per_page,
+        "results": data["data"],
+    }
+
+
+@api_operation
+@metrics.api_request_time.time()
+def get_tags(tag_name=None, tags=None, order_by=None, order_how=None, page=None, per_page=None):
+    if False:  # TODO: check if ES configured
+        flask.abort(503)
```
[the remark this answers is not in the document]
jharting: Yes, the comment is confusing. What it's supposed to do is check if XJOIN_GRAPHQL_URL is configured.

api/tag.py
```diff
@@ -0,0 +1,107 @@
...
+
+    limit, offset = _pagination_params(page, per_page)
+
+    variables = {"order_by": order_by, "order_how": order_how, "limit": limit, "offset": offset}
+
+    if tag_name:
+        variables["filter"] = {
+            # Escaped to prevent ReDoS
+            "name": f".*{re.escape(parse.quote_plus(tag_name, safe=[]))}.*"
```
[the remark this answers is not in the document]
jharting: Good catch!. quote_plus because on the ES side we're using https://docs.oracle.com/javase/8/docs/api/index.html?java/net/URLEncoder.html

api/tag.py
```diff
@@ -0,0 +1,107 @@
...
+        }
```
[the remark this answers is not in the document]
jharting: It does a substring match of all three segments (namespace, key, value) at once. Added a better explanation to API spec.

swagger/api.spec.yaml
```diff
@@ -309,6 +301,38 @@ paths:
...
+  /tags:
+    get:
+      tags:
+        - tags
+      summary: Get the active host tags for a given account
+      operationId: api.tag.get_tags
+      security:
+        - ApiKeyAuth: []
+      parameters:
+        - $ref: '#/components/parameters/tagsParam'
```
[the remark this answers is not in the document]
jharting: Yes. You are going to be able to combine multiple tags. Initially, when no tag is selected the counts will match the entire account. Once you select the first tag we want to provide updated counts.

api/tag.py
```diff
@@ -0,0 +1,107 @@
...
+    if False:  # TODO: check if ES configured
+        flask.abort(503)
+
+    limit, offset = _pagination_params(page, per_page)
+
+    variables = {"order_by": order_by, "order_how": order_how, "limit": limit, "offset": offset}
+
+    if tag_name:
+        variables["filter"] = {
+            # Escaped to prevent ReDoS
```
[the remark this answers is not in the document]
jharting: While I agree it should be obvious I don't think an extra comment is going to hurt anyone.

test_api.py
jharting: I find it useful to test that regardless of ordering we always default to the expected values for limit/offset

test_api.py
jharting: I find it useful to test that regardless of filtering/ordering we always default to the expected values for limit/offset

api/common.py
```diff
@@ -0,0 +1,10 @@
+import flask
+import ujson
```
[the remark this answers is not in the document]
jharting: No clue. I discovered it used in code so I added it here for consistency. I don't know what the history is neither how it performs compared to standard json. Would be curious to hear more about it.

api/tag.py
```diff
@@ -0,0 +1,84 @@
+import re
+from urllib.parse import quote_plus as url_quote
```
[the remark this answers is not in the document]
jharting: Because we need the result to match how the data is encoded in ES and that's done by https://docs.oracle.com/javase/8/docs/api/index.html?java/net/URLEncoder.html

api/tag.py
```diff
@@ -0,0 +1,84 @@
...
+
+def is_enabled():
+    return current_app.config["INVENTORY_CONFIG"].bulk_query_source == BulkQuerySource.xjoin
+
+
+@api_operation
+@metrics.api_request_time.time()
+def get_tags(search=None, tags=None, order_by=None, order_how=None, page=None, per_page=None):
+    if not is_enabled():
+        flask.abort(503)
```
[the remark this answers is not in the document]
jharting: Decorator? This statement is only used once - not sure what decorator buys us.

api/tag.py
```diff
@@ -0,0 +1,84 @@
...
+
+    if tags:
+        variables["hostFilter"] = {"AND": [{"tag": Tag().from_string(tag).data()} for tag in tags]}
+
+    logger.debug("executing TAGS_QUERY, variables: %s", variables)
+
+    response = graphql_query(TAGS_QUERY, variables)
+    data = response["hostTags"]
+
+    check_pagination(offset, data["meta"]["total"])
```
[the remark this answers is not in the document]
jharting: What?

test_api.py
```diff
@@ -3093,5 +3103,173 @@ def test_tags_count_pagination(self):
         self._per_page_test(2, len(host_list), int((len(host_list) + 1) / 2), test_url, expected_responses_2_per_page)
 
 
+@patch("api.tag.is_enabled", return_value=True)
```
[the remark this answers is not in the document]
jharting: How do I do that?

Glutexo:
Added one little [patch](https://github.com/Glutexo/insights-host-inventory/commit/30f56ee20fd5c2b7e153baf6d57723e611500f39) to my fork. It reverts to @jharting’s original header forwarding solution and adds a test.

test_api.py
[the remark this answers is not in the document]
Glutexo: [_test_utils.set_environment_](https://github.com/RedHatInsights/insights-host-inventory/blob/7b97f5a49e01690a54c3ac4229021abf734303d2/test_utils.py#L9) should be able to do that.

PR #328, opened 2019-07-03
Serialization: Host serialization cleanup

Glutexo:
Did a little cleanup of the (de)serialization methods. This consists of two changes, which I didn’t want to break up as they would be uncomfortably little:

* Made the code more DRY and concise. Replaced some rather cryptic lines, leverage dict comprehension etc.
* Made the deserialized objects trusted ones. Concentrated the validation/defaults logic into deserialization.

Both should make the code more durable and understandable. Although there is a change in the internal logic of the serialization, there are no changes for the code using the methods. No changes in the test nor in the overall behavior.

There are no unit tests in this pull request as the original methods were already untested. There is a separate PR (#325) with the unit tests. Some of them would have to be slightly changed to be green after merging this. Refer my rebased branch [_host_serialization_cleanup_with_tests_](https://github.com/Glutexo/insights-host-inventory/tree/host_serialization_cleanup_with_tests) to see the changes.

PR #529, opened 2019-11-22
Serialization: Host serialization cleanup

Glutexo:
Change the (de)serialization logic, so the serialization expects that the object is valid. All default values and validation happens now upon deserialization. That concentrates the logic into a single place.

Rebased version of #328.


Glutexo:
Rebased.

Glutexo:
This is tracked in [RHCLOUD-3521](https://projects.engineering.redhat.com/browse/RHCLOUD-3521). Closing to clean up the queue.

PR #530, opened 2019-11-22
Fix/create host tag no namespace

ryandillinfelton:
This allows for hosts to be created with tags with no specified namespace. These tags are stored in the database under the key "null" instead of a namespace name. It also adds two tests to cover creating hosts with tags with all the information, and tags without namespace specified in two different forms.


test_api.py
```diff
@@ -711,6 +710,72 @@ def test_create_host_with_invalid_ansible_host(self):
...
+
+        host_lookup_results = self.get(f"{HOST_URL}/{original_id}", 200)
+
+        self._validate_host(host_lookup_results["results"][0], host_data, expected_id=original_id)
+
+        host_tags = self.get(f"{HOST_URL}/{original_id}/tags", 200)["results"][original_id]
+
+        expected_tags = [
+            {"namespace": "Sat", "key": "prod", "value": None},
+            {"namespace": "null", "key": "key2", "value": "val2"},
```
jharting: While it is correct to store this as "null" in the nested representation in the database the namespace should be converted into `null` when translated back into structured representation. https://docs.google.com/document/d/1Gfkc3QRWezyHXr8jnrHNYSUsBsetG8bZyx9uCmcRxEg/edit

PR #531, opened 2019-11-25
Fix string tag validation

ryandillinfelton:
Addresses bug found in [RHCLOUD-3372](https://projects.engineering.redhat.com/browse/RHCLOUD-3372). Adds validation to input tags in host payload. Prevents hosts from being created with fields that are too short, too long, or empty strings.


jharting:
@ryandillinfelton needs to be rebased

ryandillinfelton:
Fixed the confict

app/models.py
```diff
@@ -307,9 +307,15 @@ class FactsSchema(Schema):
 
 
 class TagsSchema(Schema):
-    namespace = fields.Str(required=False, allow_none=True)
-    key = fields.Str(required=True)
-    value = fields.Str(required=False, allow_none=True)
+    namespace = fields.Str(
+        required=False, allow_none=True, validate=[validate.Length(min=1, max=255), validate.Regexp(r"[\w.\-%~]+")]
```
jharting: This regex looks overly strict. Unlike with the REST query params where the tags we receive are in the string representation thus percent encoded here we're working with the structured representation which is not percent encoded therefore we need to allow for any utf-8 characters.

jharting: We should also have a test that ensures tha tags with wild characters make it into the db and out.

PR #532, opened 2019-11-26
Host query GraphQL integration

Glutexo:
Add [_xjoin-search_](https://github.com/RedHatInsights/xjoin-search) as an alternative data source for the host search query. This is configured by [an environment variable](https://github.com/Glutexo/insights-host-inventory/blob/bd014f0faed3bf7c0d4eef15534ef303c754b984/app/config.py#L67). Extract the _xjoin-search_ GraphQL code from the tags queries. Extract common parts from the database queries.

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -1,10 +1,6 @@
-import uuid
-from datetime import datetime
-from datetime import timezone
 from enum import Enum
 
 import flask
-import sqlalchemy
 from flask_api import status
 from marshmallow import ValidationError
 from sqlalchemy.orm.base import instance_state
@@ -13,10 +9,16 @@ from api import api_operation
 from api import build_collection_response
 from api import flask_json_response
 from api import metrics
+from api.host_query import build_paginated_host_list_response
+from api.host_query import staleness_timestamps
+from api.host_query_db import get_host_list as get_host_list_db
+from api.host_query_db import params_to_order_by
+from api.host_query_xjoin import get_host_list as get_host_list_xjoin
 from app import db
 from app import events
-from app import staleness_offset
+from app import inventory_config
 from app.auth import current_identity
+from app.config import BulkQuerySource
 from app.exceptions import InventoryException
 from app.exceptions import ValidationException
 from app.logging import get_logger
@@ -27,17 +29,18 @@ from app.payload_tracker import get_payload_tracker
 from app.payload_tracker import PayloadTrackerContext
 from app.payload_tracker import PayloadTrackerProcessingContext
 from app.serialization import deserialize_host
-from app.serialization import serialize_host
 from app.serialization import serialize_host_system_profile
 from app.utils import Tag
-from lib.host_repository import _canonical_facts_host_query
 from lib.host_repository import add_host
 from lib.host_repository import AddHostResults
 from tasks import emit_event
 
+# from api.host_query_db import find_hosts_by_staleness
 
+
+FactOperations = Enum("FactOperations", ("merge", "replace"))
 TAG_OPERATIONS = ("apply", "remove")
-FactOperations = Enum("FactOperations", ["merge", "replace"])
+GET_HOST_LIST_FUNCTIONS = {BulkQuerySource.db: get_host_list_db, BulkQuerySource.xjoin: get_host_list_xjoin}
 
 logger = get_logger(__name__)
 
@@ -102,36 +105,7 @@ def _add_host(input_host):
             "host",
         )
 
-    return add_host(input_host, staleness_offset(), update_system_profile=False)
-
-
-def find_hosts_by_tag(account_number, string_tags, query):
-    tags = []
-
-    for string_tag in string_tags:
-        tags.append(Tag().from_string(string_tag))
-
-    tags_to_find = Tag.create_nested_from_tags(tags)
-
-    return query.filter(Host.tags.contains(tags_to_find))
-
-
-def find_hosts_by_staleness(states, query):
-    staleness_offset_ = staleness_offset()
-    stale_timestamp = staleness_offset_.stale_timestamp(Host.stale_timestamp)
-    stale_warning_timestamp = staleness_offset_.stale_warning_timestamp(Host.stale_timestamp)
-    culled_timestamp = staleness_offset_.culled_timestamp(Host.stale_timestamp)
-
-    null = None
-    now = datetime.now(timezone.utc)
-
-    condition_map = {
-        "fresh": stale_timestamp > now,
-        "stale": sqlalchemy.and_(Host.stale_timestamp <= now, stale_warning_timestamp > now),
-        "stale_warning": sqlalchemy.and_(stale_warning_timestamp <= now, culled_timestamp > now),
-        "unknown": Host.stale_timestamp == null,
-    }
-    return query.filter(sqlalchemy.or_(condition_map[state] for state in states))
+    return add_host(input_host, staleness_timestamps(), update_system_profile=False)
 
 
 @api_operation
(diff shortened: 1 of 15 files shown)


Glutexo:
Squashed and rebased on the [current](https://github.com/RedHatInsights/insights-host-inventory/tree/45851897d8ebcc3d71870ea89867ccfc79679d1c) [_master_](https://github.com/RedHatInsights/insights-host-inventory). Tests are failing. Plan is:

1. Finish the executive part.
2. Fix the existing tests.
3. Write new tests.

api/query_db.py
```diff
@@ -0,0 +1,146 @@
+from datetime import datetime
```
jharting: I wonder if `query_db` is a good name. Right now we only have a single controller (host) that take care of a single entity type (host). When this gets expanded to cover multiple entities there is going to be a need for multiple query files. Therefore, I think the filename should reflect the entity (e.g. host_query_db)

api/query_xjoin.py
```diff
@@ -0,0 +1,88 @@
...
+    parameters = [
+        Parameter("limit", "$limit"),
+        Parameter("offset", "$offset"),
+        Parameter("order_by", "$order_by"),
+        Parameter("order_how", "$order_how"),
+    ]
+    items = (
+        simple_object("meta", ("total",)),
+        simple_object("data", ("id", "account", "display_name", "created_on", "modified_on", "canonical_facts")),
+    )
```
jharting: This looks overly complicated. I don't think it is necessary to build the query object and the parameters from scratch every time. You can define the query once (with variable placeholders) and save it as a constant. Each request will then need to build a list of parameters only. These will be used together with the constant query.

Glutexo: Some parts of the query still need to be dynamic. I don’t like an idea of having the query partially pre-composed and partially dynamic much. But if you insist on that, I can use Jinja instead. 🤷🏻‍♂️

jharting: > Some parts of the query still need to be dynamic.
What parts need to be dynamic?

Glutexo: Filtering.

jharting: They don't. You can define all the variables in the static template and then only pass in the parameters that you need.

Glutexo: It looks like I was assuming the GraphQL server is dumber than it actually is. [Your code](https://github.com/RedHatInsights/insights-host-inventory/pull/526/files#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2R3092) shows that a need for a smart composition is not that strong as it is possible to pass a complex object as a variable. That’s great news, means you’re right and yes – it’s possible to simplify this! 🎉

jharting: perfect! :tada:

Glutexo:
Added:

- Filtering by Insights ID. It’s the last exclusive condition. Tags and staleness remain. https://github.com/RedHatInsights/insights-host-inventory/pull/532/commits/1e4c2e0c4332db838cf7f2735170e5f84b37a949
- _ansible_host_, _reporter_ and _stale_timestamp_ to the results. https://github.com/RedHatInsights/insights-host-inventory/pull/532/commits/068fa4ab354c0cf6d4a44a090946fecf28b626f5

_tags_ are not present as they are not in the cross-join service. Falsely returned as an empty array.

jharting:
> tags are not present as they are not in the cross-join service.
Not sure what you mean. You should be able to obtain tags from xjoin-search https://github.com/RedHatInsights/xjoin-search/blob/master/src/schema/schema.graphql#L116

Glutexo:
Working querying by tags and by staleness. The logic of translation of the _staleness_ param to the _stale_timestamp_ comparison logic is shared between the database and cross-join subsystem. The cross-join part can be shared between the host query and the tag query. Based on the code from #548.

Glutexo:
First version with passing tests! 🎉

Also, merged with current master and that means with #548. Refactored that a little, so the logic is DRY. I am aware that some [bits](https://github.com/Glutexo/insights-host-inventory/blob/96270e6394814e530bc53290765b594d6a217ef5/app/culling.py#L67) are now rather cryptic. I preferred not to delay further work by subsequent refactoring for simplicity.

Going to add some tests for the cross-join stuff. That should be the last thing to this pull request.

Glutexo:
There are now tests. Ready for review.

Glutexo:
Okay, ignoring a _Push failed_ message was not that great idea. 😅 Fixed the push.

Glutexo:
Also rebased on current master.

api/host.py
```diff
@@ -148,105 +122,26 @@ def get_host_list(
...
+                display_name, fqdn, hostname_or_id, insights_id, tags, page, per_page, order_by, order_how, staleness
+            )
+        except ValueError as e:
+            flask.abort(400, str(e))
+    elif config.bulk_query_source == BulkQuerySource.xjoin:
+        host_list, total = get_host_list_xjoin(
+            display_name, fqdn, hostname_or_id, insights_id, tags, page, per_page, order_by, order_how, staleness
         )
+    else:
+        raise RuntimeError("Invalid bulk query source.")
```
jharting: This looks unnecessary - an invalid bulk query source would make the app fail at startup in config.py - no need for runtime check

api/host.py
```diff
@@ -148,105 +122,26 @@ def get_host_list(
...
+    if config.bulk_query_source == BulkQuerySource.db:
+        try:
+            host_list, total = get_host_list_db(
+                display_name, fqdn, hostname_or_id, insights_id, tags, page, per_page, order_by, order_how, staleness
+            )
+        except ValueError as e:
+            flask.abort(400, str(e))
+    elif config.bulk_query_source == BulkQuerySource.xjoin:
+        host_list, total = get_host_list_xjoin(
+            display_name, fqdn, hostname_or_id, insights_id, tags, page, per_page, order_by, order_how, staleness
```
jharting: given that the function signatures are identical there's no need to duplicate function invocation

```suggestion
    get_host_list = get_host_list_db if config.bulk_query_source == BulkQuerySource.db else get_host_list_xjoin

    try:
        host_list, total = get_host_list(
            display_name, fqdn, hostname_or_id, insights_id, tags, page, per_page, order_by, order_how, staleness
        )
    except ValueError as e:
        flask.abort(400, str(e))
```

api/host_query_xjoin.py
```diff
@@ -0,0 +1,129 @@
...
+    $order_how: ORDER_DIR,
+    $filter: HostFilter!
+) {
+    hosts(
+        limit: $limit,
+        offset: $offset,
+        order_by: $order_by,
+        order_how: $order_how,
+        filter: {
+            OR: [
```
jharting: why is this top level OR operation needed?

app/auth/__init__.py
```diff
@@ -1,15 +1,13 @@
...
 
 from api.metrics import login_failure_count
-from app import IDENTITY_HEADER
 from app.auth.identity import from_auth_header
 from app.auth.identity import from_bearer_token
 from app.auth.identity import validate
 from app.logging import get_logger
 
-__all__ = ["current_identity", "bearer_token_handler", "authentication_header_handler"]
+__all__ = ("authentication_header_handler", "bearer_token_handler", "authenticated_request")
```
jharting: is `"authenticated_request"` a leftover?

app/serialization.py
```diff
@@ -48,11 +48,29 @@ def deserialize_host(raw_data):
...
-def serialize_host(host, staleness_offset):
+def deserialize_host_xjoin(data):
+    host = Host(
+        canonical_facts=data["canonical_facts"],
+        display_name=data["display_name"],
+        ansible_host=data["ansible_host"],
+        account=data["account"],
+        facts={},
+        tags={},
+        system_profile_facts={},
```
jharting: These can be fetched from xjoin-search (disregard the schema comment). https://github.com/RedHatInsights/xjoin-search/blob/master/src/schema/schema.graphql#L114

app/xjoin.py
```diff
@@ -23,9 +25,8 @@ def graphql_query(query_string, variables):
     url_ = url()
     logger.info("QUERY: URL %s; query %s, variables %s", url_, query_string, variables)
```
jharting: these should be debug level

run.py
```diff
@@ -4,7 +4,7 @@
 from app import create_app
 
 config_name = os.getenv("APP_SETTINGS", "development")
-application = create_app(config_name, start_tasks=True, start_payload_tracker=True)
```
jharting: is this intentional?

logconfig.ini
```diff
@@ -24,7 +24,7 @@ propagate=1
 qualname=gunicorn.access
 
 [logger_sqlalchemy.engine]
-level=WARNING
```
jharting: is this intentional?

api/tag.py
```diff
@@ -48,17 +49,9 @@
...
-        filter["gte"] = gte.isoformat()
-    if lte:
-        filter["lte"] = lte.isoformat()
-    return {"stale_timestamp": filter}
-
-
 def is_enabled():
-    return flask.current_app.config["INVENTORY_CONFIG"].bulk_query_source == BulkQuerySource.xjoin
+    config = inventory_config()
+    return config == BulkQuerySource.xjoin
```
jharting:
```suggestion
    return config.bulk_query_source == BulkQuerySource.xjoin
```

api/host.py
Glutexo: This objection is valid, my personal taste however warns be before non-exhaustive if/elif/else statements. (Some programing languages, e.g. Swift, even fail to compile in such case.) I‘ll refactor a bit.

api/host.py
Glutexo: You mean like, inline the contents of that function? I slightly disagree here. There is quite a lot of logic in both function and having them extracted improves testability. Also there is a nice symmetry between the _db_ and _xjoin_ variants.

I‘d rather leave it as it is for now, but I agree that there is a room for improvement. Let’s rather leave that for an explicit refactoring. There is more of the code shuffling and unspaghetizing that would this code utilize.

api/host_query_xjoin.py
Glutexo: Oh. Really. It doesn’t do any harm, but is redundant. Nice catch!

api/host.py
jharting: @Glutexo no, inlining is not at all what I meant. Take a look at the code suggestion.

app/xjoin.py
Glutexo: SQLAlchemy [doesn’t think](https://docs.sqlalchemy.org/en/13/core/engines.html?highlight=logging#configuring-logging) so. I tried to mimic its logging level behavior, so its consistent between the two data sources.

> sqlalchemy.engine - controls SQL echoing. set to logging.INFO for SQL query output, logging.DEBUG for query + result set output. These settings are equivalent to echo=True and echo="debug" on create_engine.echo, respectively.

jharting: Looks like we also log the result on INFO level

api/host.py
Glutexo: Oh, I see. The signature is not exactly same. Unfortunatly they differ in what exceptions they can trigger, which in my eyes is a part of the signature. I was thinking about refactoring that to introduce the exact change you suggest. I decided to keep that for some further refactoring cycle though.

If you insist on the refactoring, I can include it in this pull request. I wouldn’t want to do it the simplest way though: having two _except_ blocks for the same call would result in one of them always not being reachable. That seems like an antipattern to me.

app/serialization.py
Glutexo: Actually, system profile facts [are not included](https://github.com/RedHatInsights/insights-host-inventory/blob/e81e7621fd0ce0f3e4452004eb41ba3685358ac6/swagger/api.spec.yaml#L682) in the _get_host_list_ output. Thus there is no need to even load them. Facts and tags should be there though, you’re right there.

api/host.py
jharting: How do they differ in exceptions? From what I can see they both can raise `ValueError`. Not handling it for `get_host_list_xjoin` seems like an oversight not me.

app/xjoin.py
Glutexo: That should not be. Result should be DEBUG.

api/host.py
Glutexo: 😯 You’re right. I’d swear I had a reason for that. This is worth adding a regression test.

Glutexo:
Added facts.

Glutexo:
Rebased.

test_api.py
```diff
@@ -3103,246 +3112,712 @@ def test_tags_count_pagination(self):
...
+                        )
+
+        def test_response_invalid_pagination(self, is_enabled):
+            for page, per_page in [(0, 10), (-1, 10), (1, 0), (1, -5), (1, 101)]:
+                with self.subTest(page=page):
+                    with self.patch_with_empty_response():
+                        self.get(f"{TAGS_URL}?per_page={per_page}&page={page}", 400)
+
+    @patch("api.tag.is_enabled", return_value=True)
+    class TagsResponseTestCase(APIBaseTestCase):
```
jharting: why is this nested i.e. not top-level class?

test_api.py
```diff
@@ -3103,246 +3112,712 @@ def test_tags_count_pagination(self):
...
-        self.get(f"{TAGS_URL}?per_page=2&page=3", 404)
+    @patch_with_response()
+    def test_response_pagination_index_error(self, graphql_query):
+        self.get(f"{HOST_URL}?per_page=2&page=3", 404)
+        graphql_query.assert_called_once()
 
-        graphql_query.assert_called_once_with(
-            TAGS_QUERY, {"order_by": "tag", "order_how": "ASC", "limit": 2, "offset": 4, "hostFilter": {"OR": ANY}}
+    @patch("api.tag.is_enabled", return_value=True)
+    class TagsRequestTestCase(XjoinRequestBaseTestCase):
```
jharting: why is this nested i.e. not top-level class?

test_api.py
Glutexo: Argh! My humble guess is Black tried to guess its best when I forgot some closing bracket somewhere. Thanks for spotting! Fixed. 😇

jharting: @Glutexo don't forget to git push

PR #534, opened 2019-12-02
Tagging fix 500

ryandillinfelton:
Fixes 500 error when sending string formatted tag instead of structured tag while creating host. Now returns 400


Glutexo:
Why has this been closed without merging?

ryandillinfelton:
I opened it with a different branch instead to make the history cleaner.

PR #535, opened 2019-12-02
Fix 500 error when adding host with tags in string format (RHCLOUD-3342)

ryandillinfelton:
Fixes 500 error when sending string formatted tag instead of structured tag while creating host. Now returns 400

### DIFF
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -442,8 +442,6 @@ components:
         value:
           type: string
           nullable: true
-    MarshmallowError:
-      type: string
     TagCountOut:
       type: object
       properties:
@@ -606,6 +604,11 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/FactSet'
+        tags:
+          description: The tags on a host
+          type: array
+          items:
+            $ref: '#/components/schemas/StructuredTag'
         system_profile:
           $ref: '#/components/schemas/SystemProfileIn'
         stale_timestamp:
@@ -734,9 +737,7 @@ components:
           description: An array of the tags on the host
           type: array
           items:
-            oneOf:
-              - $ref: '#/components/schemas/StructuredTag'
-              - $ref: '#/components/schemas/MarshmallowError'
+            - $ref: '#/components/schemas/StructuredTag'
         stale_timestamp:
           description: Timestamp from which the host is considered stale.
           type: string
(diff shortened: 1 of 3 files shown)


swagger/api.spec.yaml
```diff
@@ -729,9 +731,12 @@ components:
           items:
             $ref: '#/components/schemas/FactSet'
         tags:
+          description: An array of the tags on the host
           type: array
           items:
-            $ref: '#/components/schemas/StructuredTag'
+            oneOf:
+              - $ref: '#/components/schemas/StructuredTag'
```
jharting: We're getting a 500 because of the mismatch of validation constraints on the REST layer. The output format is stricter (only allows structured tag representation) than the input format (allows tags field of any format). As a result, an invalid representation can get in but cannot get out (does not pass output validation).

Instead of relaxing the output validation (by adding MarshmallowError) we should make input validation more strict (validate the tags field)

ryandillinfelton:
Re-review request was an accident

swagger/api.spec.yaml
ryandillinfelton: Is there somewhere else I ought to be doing input validation here? I have Marshmallow validation for tags which seems to be catching tags that are in the wrong format, but breaking the response schema with the error response which is why I thought relaxing the response was the way to go.

I have noticed that some of the validation errors that marshmallow creates for tags create a proper response without relaxing the output validation which is strange, I haven't been able to pin down why though,

ryandillinfelton: Actually, i think I've figured it out now

swagger/api.spec.yaml
```diff
@@ -729,9 +734,10 @@ components:
           items:
             $ref: '#/components/schemas/FactSet'
         tags:
+          description: An array of the tags on the host
           type: array
           items:
-            $ref: '#/components/schemas/StructuredTag'
+            - $ref: '#/components/schemas/StructuredTag'
```
jharting: Why this change? It does not look correct. (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#response-object-examples) Other than this the PR looks good.

jharting:
Also please remember to put JIRA id to PR and commit messages.

swagger/api.spec.yaml
ryandillinfelton: Ahh, I missed removing the dash when I removed the "oneOf" part and the marshmallow error schema

PR #536, opened 2019-12-03
Implement repository changes to system_profile

semtexzv:
Implement schema changes, which correspond to `system_profile` changes from https://github.com/RedHatInsights/insights-puptoo/pull/36

- Added local repo-id into `yum_repos` objects
- Added the `dnf_modules` key containing information about modules.
- Fixed swagger doc, schemas and test to reflect these changes.

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -31,6 +31,7 @@ from app.utils import Tag
 from tasks import msg_handler
 from test_utils import rename_host_table_and_indexes
 from test_utils import set_environment
+from test_utils import valid_system_profile
 
 HOST_URL = "/api/inventory/v1/hosts"
 HEALTH_URL = "/health"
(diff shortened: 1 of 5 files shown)


test_api.py
```diff
@@ -1216,14 +1216,23 @@ def _valid_system_profile(self):
...
+            "yum_repos": [
+                {
+                    "id": "repo1",
+                    "name": "repo1",
+                    "gpgcheck": True,
+                    "enabled": True,
+                    "base_url": "http://rpms.redhat.com",
+                }
+            ],
+            "dnf_modules": [],
```
jharting: please add some sample values here so that schema validation is actually tested

semtexzv: Fixed

test_mq_service.py
```diff
@@ -220,6 +219,32 @@ def test_add_host_simple(self, datetime_mock):
 
         self._base_add_host_test(host_data, expected_results, host_keys_to_check)
 
+    @patch("app.queue.egress.datetime", **{"utcnow.return_value": datetime.utcnow()})
+    def test_host_add_with_system_profile(self, datetime_mock):
```
Glutexo: This test checks that the new information is accepted. It doesn’t however check whether it’s correctly stored in the database and loaded. 🤷🏻‍♂️

PR #537, opened 2019-12-03
Dissable staleness check when hosts are requested by id

ryandillinfelton:
"To help with rollout of system culling we should relax the enforcement on GETs for a single system so they return the system even if it's culled."

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -299,11 +299,14 @@ def delete_by_id(host_id_list):
 
 @api_operation
 @metrics.api_request_time.time()
-def get_host_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None, staleness=None):
+def get_host_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None, staleness=False):
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
-    if staleness:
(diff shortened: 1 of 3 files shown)


api/host.py
```diff
@@ -299,11 +299,14 @@ def delete_by_id(host_id_list):
...
 @metrics.api_request_time.time()
-def get_host_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None, staleness=None):
+def get_host_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None, staleness=False):
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
-    if staleness:
-        query = find_hosts_by_staleness(staleness, query)
+    # The staleness check is currently disabled when getting a single host by id
+    # to ease onboarding (RHCLOUD-3562)
+    if len(host_id_list) > 1:
```
jharting: Apologies upfront for confusion @ryandillinfelton . What https://projects.engineering.redhat.com/browse/RHCLOUD-3562 is supposed to say (and I changed it already) is "relaxed enforcement on system GET by id" i.e. this entire api operation. In other words, remove stalenessness/culling from this function entirely for now.

ryandillinfelton: Got it!

PR #538, opened 2019-12-04
allow 20byte MAC addresses for infiniband devices

ryandillinfelton:


ryandillinfelton:
Do we want to add a limit for the mac_addresses field also? right now it allows anything between 1-255

PR #539, opened 2019-12-05
check incoming messages for invalid unicode sequences

jharting:
Due to RHCLOUD-3610 we're receiving messages with invalid unicode code points (invalid surrogate pairs).
Python pretty much ignores that but it is not possible to store such strings in the database (db INSERTS blow up)
I added validation logic that looks for such invalid sequences with the intention of marking such messages as invalid (i.e. detect validation error rather than failing later in the db layer)

### DIFF
--- a/app/queue/ingress.py
+++ b/app/queue/ingress.py
@@ -1,5 +1,4 @@
 import json
-import re
 
 from marshmallow import fields
 from marshmallow import Schema
@@ -17,7 +16,6 @@ from app.queue.egress import build_event
 from app.serialization import deserialize_host
 from lib import host_repository
 
-INVALID_UNICODE_PATTERN = re.compile("[\uDC00-\uDFFF][\uDC00-\uDFFF]")
 logger = get_logger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


app/queue/ingress.py
```diff
@@ -26,15 +27,26 @@ class OperationSchema(Schema):
     data = fields.Dict(required=True)
 
 
+# Due to RHCLOUD-3610 we're receiving messages with invalid unicode code points (invalid surrogate pairs)
+# Python pretty much ignores that but it is not possible to store such strings in the database (db INSERTS blow up)
+# This functions looks for such invalid sequences with the intention of marking such messages as invalid
+def _validate_unicode_code_points(message):
+    match = re.search(INVALID_UNICODE_PATTERN, message.encode().decode("unicode-escape"))
+    if match:
+        raise UnicodeError("utf-8", f"invalid unicode sequence {match.group(0)}", message, match.start(), match.end())
```
Glutexo: `message.encode()` actually itself raises a _UnicodeEncodeError_ if the JSON contains an unescaped invalid Unicode sequence. That means that the proposed solution fixes the actual problem, because the buggy client sends us **escaped** data, but it doesn’t save from possible further problems if invalid **unescaped** data is sent.

This can be however quite easily normalized. Python’s _json_ module loads/dumps the string without checking Unicode validity. It however escapes all Unicode characters including the invalid ones. We already parse the JSON string using [_loads_](https://github.com/jharting/insights-host-inventory/blob/0ab0849976eec4d8fd4bc9c2c816c46dea1b1d70/app/queue/ingress.py#L43). With that in memory, UTF-8 validation is as simple as calling _parsed_message.dumps().encode()_ and checking for _UnicodeEncodeError_. Voilà! 🎉

inv_mq_service.py
```diff
@@ -24,6 +24,7 @@ def main():
         group_id=config.host_ingress_consumer_group,
         bootstrap_servers=config.bootstrap_servers,
         api_version=(0, 10),
+        value_deserializer=lambda m: m.decode(),
```
Glutexo: Can we even put JSON decode here?

Glutexo:
Just found out that str/bytes encode/decode "unicode-escape" is not 100% compatible with JSON dumps/loads. 😭

```python
def main():
    utf = "🧜🏿‍♂️"
    print(utf.encode("unicode-escape"))

    dumped = dumps(utf)
    print(dumped)

    encoded = dumped.encode("ascii")
    print(encoded)
    print(encoded.decode("unicode-escape"))
```
```
  File "/Users/stomsa/Vývoj/Insights/Repozitáře/insights-host-inventory/pokus.py", line 29, in <module>
    main()
  File "/Users/stomsa/Vývoj/Insights/Repozitáře/insights-host-inventory/pokus.py", line 13, in main
    print(encoded.decode("unicode-escape"))
UnicodeEncodeError: 'utf-8' codec can't encode characters in position 1-4: surrogates not allowed
b'\\U0001f9dc\\U0001f3ff\\u200d\\u2642\\ufe0f'
"\ud83e\udddc\ud83c\udfff\u200d\u2642\ufe0f"
b'"\\ud83e\\udddc\\ud83c\\udfff\\u200d\\u2642\\ufe0f"'
```

inv_mq_service.py
Glutexo: I‘d even consider put the whole validation logic here. It seems that it logically belongs here. Not necessarily in this pull request.

Glutexo:
Created a [patch](https://github.com/Glutexo/insights-host-inventory/commits/unicode-errors-validate) for this pull request. It removes the manual surrogate check, relying solely on _str.encode_.

The good:

- Removes arbitrary, non-standard and possibly buggy Unicode handling. Relies on the well-proven Standard Library instead.
- Unifies handling of raw and escaped invalid Unicode entities.
- Doesn’t potentially break characters that are encoded differently by _json_ and _str_ modules.
- Fixes catching surrogates not matched by the original solution. E.g. `"\ud83e\udddc\ud83c\udfff\u200d"`

The bad:
- Traverses the loaded object instead of parsing the whole message as a string. Doesn’t use regexes though.

If performance impact may become a problem, comparison should be done properly. It’s hard to estimate difference between iterating and regex matching over a big string.

Glutexo:
Accidentally pushed to the source branch. Reverted. Sorry.

Added [a fix and unit tests](https://github.com/Glutexo/insights-host-inventory/commit/e7d68cffc297bda7c2cf7e0de607053012821fbc) to my fork.

PR #545, opened 2019-12-11
Cross-join common code

Glutexo:
Common cross join code for #532 and #526. Final touches and tests to be added.

### DIFF
--- a/app/xjoin.py
+++ b/app/xjoin.py
@@ -1,0 +1,41 @@
+from logging import getLogger
+
+from flask import abort
+from flask import current_app
+from requests import post
+
+from app.auth import authenticated_request
+
+__all__ = ("graphql_query", "pagination_params", "string_contains", "url")
+
+logger = getLogger("graphql")
+
+
+def check_pagination(offset, total):
+    if total and offset >= total:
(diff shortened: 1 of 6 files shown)


Glutexo:
That was better than expected. Cool! Going to add some tests and stuff in a new PR.

app/xjoin.py
```diff
@@ -0,0 +1,41 @@
...
+        abort(404)  # Analogous to flask_sqlalchemy.BaseQuery.paginate
+
+
+def graphql_query(query_string, variables):
+    url_ = url()
+    logger.info("QUERY: URL %s; query %s, variables %s", url_, query_string, variables)
+    payload = {"query": query_string, "variables": variables}
+
+    response = authenticated_request(post, url_, json=payload)
+    logger.info("QUERY: response %s", response.text)
```
jharting: should these be debug level logs rather than info?

Glutexo: I mimicked SQLAlchemy’s logging configuration.

> sqlalchemy.engine - controls SQL echoing. set to logging.INFO for SQL query output, logging.DEBUG for query + result set output. These settings are equivalent to echo=True and echo="debug" on create_engine.echo, respectively.

https://docs.sqlalchemy.org/en/13/core/engines.html#configuring-logging

PR #546, opened 2019-12-11
Make all REST operations respect culling (RHCLOUD-3515)

ryandillinfelton:
This PR makes it so each of the REST endpoints will by default ignore culled hosts. For get operation this behaviour can be overridden/customized with a parameter.

It also adds test to check that these changes work

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -302,8 +302,10 @@ def delete_by_id(host_id_list):
 def get_host_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None, staleness=None):
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
-    if staleness:
-        query = find_hosts_by_staleness(staleness, query)
+    # The staleness check is currently disabled when getting a single host by id
+    # to ease onboarding (RHCLOUD-3562)
+    # if staleness:
+    #     query = find_hosts_by_staleness(staleness, query)
 
     try:
         order_by = _params_to_order_by(order_by, order_how)
--- a/api/host.py
+++ b/api/host.py
@@ -234,8 +234,6 @@ def patch_by_id(host_id_list, host_data):
 
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
-    query = find_hosts_by_staleness(ALL_STALENESS_STATES, query)
-
     hosts_to_update = query.all()
 
     if not hosts_to_update:
@@ -301,9 +299,7 @@ def update_facts_by_namespace(operation, host_id_list, namespace, fact_dict):
 @api_operation
 @metrics.api_request_time.time()
 def get_host_tag_count(host_id_list, page=1, per_page=100, order_by=None, order_how=None):
-    query = Host.query.filter((Host.account == current_identity.account_number) & Host.id.in_(host_id_list))
-
-    query = find_hosts_by_staleness(ALL_STALENESS_STATES, query)
+    query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
     try:
         order_by = params_to_order_by(order_by, order_how)
@@ -339,9 +335,7 @@ def _count_tags(host_list):
 @api_operation
 @metrics.api_request_time.time()
 def get_host_tags(host_id_list, page=1, per_page=100, order_by=None, order_how=None, search=None):
-    query = Host.query.filter((Host.account == current_identity.account_number) & Host.id.in_(host_id_list))
-
-    query = find_hosts_by_staleness(ALL_STALENESS_STATES, query)
+    query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
     try:
         order_by = params_to_order_by(order_by, order_how)
--- a/api/host.py
+++ b/api/host.py
@@ -160,8 +160,6 @@ def delete_by_id(host_id_list):
 
     with PayloadTrackerContext(payload_tracker, received_status_message="delete operation"):
         query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
-        #  TODO: probably remove
-        # query = find_hosts_by_staleness(ALL_STALENESS_STATES, query)
 
         if not query.count():
             flask.abort(status.HTTP_404_NOT_FOUND)
--- a/api/host.py
+++ b/api/host.py
@@ -186,8 +186,6 @@ def delete_by_id(host_id_list):
 @metrics.api_request_time.time()
 def get_host_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None):
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
-    # TODO: probably remove
-    # query = find_hosts_by_staleness(["fresh", "stale"], query)
 
     try:
         order_by = params_to_order_by(order_by, order_how)
@@ -220,7 +218,6 @@ def get_host_system_profile_by_id(host_id_list, page=1, per_page=100, order_by=N
         flask.abort(400, str(e))
     else:
         query = query.order_by(*order_by)
-
     query_results = query.paginate(page, per_page, True)
 
     response_list = [serialize_host_system_profile(host) for host in query_results.items]
@@ -308,7 +305,7 @@ def update_facts_by_namespace(operation, host_id_list, namespace, fact_dict):
 def get_host_tag_count(host_id_list, page=1, per_page=100, order_by=None, order_how=None):
     query = Host.query.filter((Host.account == current_identity.account_number) & Host.id.in_(host_id_list))
 
-    query = find_hosts_by_staleness(["fresh", "stale"], query)
+    query = find_hosts_by_staleness(ALL_STALENESS_STATES, query)
 
     try:
         order_by = params_to_order_by(order_by, order_how)
@@ -346,7 +343,7 @@ def _count_tags(host_list):
 def get_host_tags(host_id_list, page=1, per_page=100, order_by=None, order_how=None, search=None):
     query = Host.query.filter((Host.account == current_identity.account_number) & Host.id.in_(host_id_list))
 
-    query = find_hosts_by_staleness(["fresh", "stale"], query)
+    query = find_hosts_by_staleness(ALL_STALENESS_STATES, query)
 
     try:
         order_by = params_to_order_by(order_by, order_how)
--- a/api/host.py
+++ b/api/host.py
@@ -302,10 +302,8 @@ def delete_by_id(host_id_list):
 def get_host_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None, staleness=None):
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
-    # The staleness check is currently disabled when getting a single host by id
-    # to ease onboarding (RHCLOUD-3562)
-    # if staleness:
-    #     query = find_hosts_by_staleness(staleness, query)
+    if staleness:
+        query = find_hosts_by_staleness(staleness, query)
 
     try:
         order_by = _params_to_order_by(order_by, order_how)
@@ -461,8 +459,11 @@ def _count_tags(host_list):
 
 @api_operation
 @metrics.api_request_time.time()
-def get_host_tags(host_id_list, page=1, per_page=100, order_by=None, order_how=None):
+def get_host_tags(host_id_list, page=1, per_page=100, order_by=None, order_how=None, staleness=None):
     query = Host.query.filter((Host.account == current_identity.account_number) & Host.id.in_(host_id_list))
+
+    if staleness:
+        query = find_hosts_by_staleness(staleness, query)
 
     try:
         order_by = _params_to_order_by(order_by, order_how)
--- a/api/host.py
+++ b/api/host.py
@@ -115,7 +115,7 @@ def find_hosts_by_tag(account_number, string_tags, query):
     return query.filter(Host.tags.contains(tags_to_find))
 
 
-def find_hosts_by_staleness(states=["fresh", "stale", "unknown"], query=None):
+def find_hosts_by_staleness(query, states=["fresh", "stale", "unknown"]):
     staleness_offset_ = staleness_offset()
     stale_timestamp = staleness_offset_.stale_timestamp(Host.stale_timestamp)
     stale_warning_timestamp = staleness_offset_.stale_warning_timestamp(Host.stale_timestamp)
@@ -163,7 +163,7 @@ def get_host_list(
         query = find_hosts_by_tag(current_identity.account_number, tags, query)
 
     if staleness:
-        query = find_hosts_by_staleness(staleness, query)
+        query = find_hosts_by_staleness(query, staleness)
 
     try:
         order_by = _params_to_order_by(order_by, order_how)
@@ -266,7 +266,7 @@ def delete_by_id(host_id_list):
     with PayloadTrackerContext(payload_tracker, received_status_message="delete operation"):
         query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
-        query = find_hosts_by_staleness(query=query)
+        query = find_hosts_by_staleness(query)
 
         hosts_to_delete = query.all()
 
@@ -304,7 +304,7 @@ def get_host_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
     if staleness:
-        query = find_hosts_by_staleness(staleness, query)
+        query = find_hosts_by_staleness(query, staleness)
 
     try:
         order_by = _params_to_order_by(order_by, order_how)
@@ -328,7 +328,7 @@ def _get_host_list_by_id_list(account_number, host_id_list):
 def get_host_system_profile_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None):
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
-    query = find_hosts_by_staleness(query=query)
+    query = find_hosts_by_staleness(query)
 
     try:
         order_by = _params_to_order_by(order_by, order_how)
@@ -363,7 +363,7 @@ def patch_by_id(host_id_list, host_data):
 
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
-    query = find_hosts_by_staleness(query=query)
+    query = find_hosts_by_staleness(query)
 
     hosts_to_update = query.all()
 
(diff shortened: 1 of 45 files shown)


api/host.py
```diff
@@ -264,9 +264,11 @@ def delete_by_id(host_id_list):
     payload_tracker = get_payload_tracker(account=current_identity.account_number, payload_id=threadctx.request_id)
 
     with PayloadTrackerContext(payload_tracker, received_status_message="delete operation"):
-
         query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
+        staleness = ["fresh", "stale", "unknown"]
```
jharting: We should not be repeating this. Either define a constant or perhaps make this the default value of the states param in `find_hosts_by_staleness`?

api/host.py
```diff
@@ -115,7 +115,7 @@ def find_hosts_by_tag(account_number, string_tags, query):
     return query.filter(Host.tags.contains(tags_to_find))
 
 
-def find_hosts_by_staleness(states, query):
+def find_hosts_by_staleness(query, states=["fresh", "stale", "unknown"]):
```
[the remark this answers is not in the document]
Glutexo: Mutable default values values are dangerous. Even though you’re not mutating the object, I’d rather use a _tuple_ or _None_ there.

Glutexo:
Are we still going to merge this? Is it worth rebasing and reviewing?

jharting:
@Glutexo @ryandillinfelton we are but not before https://projects.engineering.redhat.com/browse/RHCLOUD-3755 is implemented and released

Glutexo:
The Host Reaper is there, released. What about this one? Or are we waiting for the culling field enforcing?

Glutexo:
With #582 now merged, this should be unblocked.

test_api.py
```diff
@@ -2696,6 +2700,46 @@ def test_dont_get_only_culled(self):
...
+    def test_patch_facts_ignores_culled(self):
+        url = HOST_URL + "/" + self.culled_host["id"] + "/rhsm"
+
+        self.patch(url, {"ARCHITECTURE": "patched"}, 404)
+
+    def test_put_facts_ignores_culled(self):
+        url = HOST_URL + "/" + self.culled_host["id"] + "/otherNS"
+
+        self.put(url, {"ARCHITECTURE": "patched"}, 404)
+
```
Glutexo: I’d like to:

- Put these tests in a separate _TestCase_ class.
- Extract __get_created_hosts_by_id_ that is only used in these new tests.
- Add tests for get hosts by ID.
- Tests that fresh, stale etc. hosts are _not_ ignored. (It’s important to have both positive _and_ negative tests.)

Also I see that there are some tests missing for get host list operation:

- The default _staleness_ param value verification.
- Verification that regardless of the specified parameters, the culled hosts are still omitted.

test_api.py
```diff
@@ -2622,7 +2622,11 @@ def _assert_values(response_host):
 
 class QueryStalenessBaseTestCase(DBAPITestCase):
     def _create_host(self, stale_timestamp):
-        data = {"account": ACCOUNT, "insights_id": str(generate_uuid())}
+        data = {
+            "account": ACCOUNT,
+            "insights_id": str(generate_uuid()),
+            "facts": [{"facts": {"ARCHITECTURE": "x86_64"}, "namespace": "rhsm"}],
```
Glutexo: Is this intentional?

test_api.py
```diff
@@ -2670,7 +2674,7 @@ def _get_created_hosts_by_id_url(self, query):
 
     def _get_created_hosts_by_id(self, query):
         hosts = self._get_hosts_by_id(query, self._created_hosts())
-        return tuple(host["id"] for host in hosts)
+        return tuple(host for host in hosts)
```
Glutexo: Is there any reason why the _id_ is not sufficent?

test_api.py
```diff
@@ -1108,7 +1108,7 @@ def test_dont_update_stale_timestamp_from_different_reporter(self):
 class DeleteHostsBaseTestCase(DBAPITestCase):
```
Glutexo: We are running into a problem here: culled hosts are no longer distinguishable from hard-deleted ones. That means that for the Host Reaper tests, we can no longer use the REST API to verify the host presence and deletion. You need to check manually in the database (_Host.query.get_). For the REST DELETE, it’s still possible to use the API GET to check that the host got deleted as long as you verify it’s no longer retrievable. If you want to check that it really got deleted from the database, you’d need to use the direct approach there as well. Maybe both, why not.

Once you do that, please pay attention to the test case structure. It’d be nice if the base test case classes still had semantically meaningful names and that the test cases would share only as few helper methods as necessary.

api/host.py
```diff
@@ -40,6 +41,7 @@
 GET_HOST_LIST_FUNCTIONS = {BulkQuerySource.db: get_host_list_db, BulkQuerySource.xjoin: get_host_list_xjoin}
 XJOIN_HEADER = "x-rh-cloud-bulk-query-source"  # will be xjoin or db
 REFERAL_HEADER = "referer"
+DEFAULT_STALENESS_PARAM = ["fresh", "stale", "stale_warning"]
```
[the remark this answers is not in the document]
ryandillinfelton: I went with ALL_STALENESS_STATES. Does that seems appropriately descriptive?

test_api.py
ryandillinfelton: The problem I run into with returning the ID is that when the query is "/tags" there is no named id field in the response from the API, so host["id"] creates an error in that case.

test_api.py
ryandillinfelton: I guess not. Looks like the facts are unnecessary.

api/host.py
```diff
@@ -211,12 +212,15 @@ def _get_host_list_by_id_list(account_number, host_id_list):
 def get_host_system_profile_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None):
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
+    query = find_hosts_by_staleness(["fresh", "stale"], query)
```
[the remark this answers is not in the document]
ryandillinfelton: :+1:

test_api.py
```diff
@@ -2696,41 +2700,80 @@ def test_dont_get_only_culled(self):
...
+                stale_timestamp = datetime.now(timezone.utc) + timedelta(hours=1)
+                host = self._create_and_get_host(stale_timestamp)
 
-#                 culled_timestamp = stale_timestamp + timedelta(days=culling_culled_offset_days)
-#                 self.assertEqual(culled_timestamp.isoformat(), host["culled_timestamp"])
+                culled_timestamp = stale_timestamp + timedelta(days=culling_culled_offset_days)
+                self.assertEqual(culled_timestamp.isoformat(), host["culled_timestamp"])
 
 
+# TODO: Maybe remove?
```
ryandillinfelton: I'm removing these test since we don't want a staleness parameter for this endpoint.

api/host.py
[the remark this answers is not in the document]
Glutexo: 👍🏻

Glutexo:
All checks have passed. 😮

@ryandillinfelton Is this now ready for a new review, or are you going to push more changes before that?

ryandillinfelton:
@Glutexo Still some more changes coming before I think it's ready

Glutexo:
@ryandillinfelton Thanks! Please re-request a review once ready.

api/host.py
```diff
@@ -183,10 +186,8 @@ def delete_by_id(host_id_list):
 def get_host_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None, staleness=None):
```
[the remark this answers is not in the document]
ryandillinfelton: Removed the staleness parameter :+1:

api/host.py
```diff
@@ -183,10 +186,8 @@ def delete_by_id(host_id_list):
...
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
-    # The staleness check is currently disabled when getting a single host by id
-    # to ease onboarding (RHCLOUD-3562)
-    # if staleness:
-    #     query = find_hosts_by_staleness(staleness, query)
+    if staleness:
+        query = find_hosts_by_staleness(staleness, query)
```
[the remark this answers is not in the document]
ryandillinfelton: I went with the latter approach :+1:

test_api.py
ryandillinfelton: I think the current tests intrinsically verify that stale, stale_warning, and fresh hosts are successfully retrieved by default since they compare a list containing the expected host IDs to verify that the culled host is not returned.

jharting:
@Glutexo @ryandillinfelton hey folks, let's make sure https://projects.engineering.redhat.com/browse/RHCLOUD-4825 is addressed by this PR also. QE is testing db/xjoin modes side-by-side therefore we want to merge a change that is consistent across both of them.

Glutexo:
@jharting Okay. Will review once the cross-join part is there.

ryandillinfelton:
I believe the xjoin part is already in place for the endpoints that support xjoin as a data_source. There are tests [here](https://github.com/RedHatInsights/insights-host-inventory/blob/master/test_api.py#L3640). Unless there's a difference in the default staleness between xjoin and db queries

test_api.py
```diff
@@ -2690,82 +2692,136 @@ def test_get_only_stale_warning(self):
...
+            return tuple(hosts.keys())
+        else:
+            return tuple(host["id"] for host in hosts)
+
     def test_dont_get_only_culled(self):
         for get_hosts_url_method in (self._get_all_hosts_url, self._get_created_hosts_by_id_url):
             with self.subTest(get_hosts_url_method=get_hosts_url_method):
-                url = get_hosts_url_method("?staleness=culled")
+                url = get_hosts_url_method(query="?staleness=culled")
                 self.get(url, 400)
```
Glutexo: There is a test missing that would get all hosts (_get host list_ operation, not by ID) and verify that the culled hosts are ignored. I’d say the same about the _add host_ operation, but although related, we have a separate ticket for that, so let’s leave it as it is for now.

test_api.py
```diff
@@ -2690,82 +2692,136 @@ def test_get_only_stale_warning(self):
...
+        self.assertEqual(expected_host_ids, retrieved_host_ids)
+
+    def test_patch_ignores_culled(self):
+        url = HOST_URL + "/" + self.culled_host["id"]
+
+        self.patch(url, {"display_name": "patched"}, 404)
+
+    def test_patch_works_on_non_culled(self):
+        host_id_url_list = ",".join([self.fresh_host["id"]])
+        url = HOST_URL + "/" + host_id_url_list
```
Glutexo: You’re joining a hard-coded list with just one item. You can just go without the joining and use the value directly.

```suggestion
        url = HOST_URL + "/" + self.fresh_host["id"]
```

test_api.py
```diff
@@ -2690,82 +2692,136 @@ def test_get_only_stale_warning(self):
...
 
-#                 stale_timestamp = datetime.now(timezone.utc) + timedelta(hours=1)
-#                 host = self._create_and_get_host(stale_timestamp)
+        self.patch(url, {"ARCHITECTURE": "patched"}, 404)
 
-#                 stale_warning_timestamp = stale_timestamp + timedelta(days=culling_stale_warning_offset_days)
-#                 self.assertEqual(stale_warning_timestamp.isoformat(), host["stale_warning_timestamp"])
+    def test_patch_facts_works_on_non_culled(self):
+        host_id_url_list = ",".join([self.fresh_host["id"]])
+        url = HOST_URL + "/" + host_id_url_list + "/rhsm"
```
```suggestion
        url = HOST_URL + "/" + self.fresh_host["id"] + "/rhsm"
```

test_api.py
```diff
@@ -2690,82 +2692,136 @@ def test_get_only_stale_warning(self):
...
+        url = HOST_URL + "/" + self.culled_host["id"]
 
-#     def _assert_host_not_in_state(self, host, state):
-#         self._assert_host_state(host, state, self.assertNotIn)
+        self.delete(url, 404)
+
+    def test_delete_works_on_non_culled(self):
+        with patch("lib.host_delete.emit_event", new_callable=MockEmitEvent):
+            host_id_url_list = ",".join([self.fresh_host["id"]])
+            url = HOST_URL + "/" + host_id_url_list
```
```suggestion
            url = HOST_URL + "/" + self.fresh_host["id"]
```

test_api.py
```diff
@@ -2690,82 +2692,136 @@ def test_get_only_stale_warning(self):
...
+                config.culling_culled_offset_days = culling_culled_offset_days
 
-#                 config = self.app.config["INVENTORY_CONFIG"]
-#                 config.culling_culled_offset_days = culling_culled_offset_days
+                stale_timestamp = datetime.now(timezone.utc) + timedelta(hours=1)
+                host = self._create_and_get_host(stale_timestamp)
 
-#                 self._assert_host_not_in_state(host, "fresh,stale,stale_warning,unknown")
+                culled_timestamp = stale_timestamp + timedelta(days=culling_culled_offset_days)
+                self.assertEqual(culled_timestamp.isoformat(), host["culled_timestamp"])
```
Glutexo: I’d like to ask you for one little improvement of existing Cross-Join tests. There are tests verifying the staleness filter parameter passed to the Cross-Join GraphQL query. These tests only test one single staleness values. It’d be great to test more that one value (comma-separated) too.

- [Hosts](https://github.com/RedHatInsights/insights-host-inventory/blob/813b9997687e5e48d464f501a8e475133bf8fbb4/test_api.py#L3726)
- [Tags](https://github.com/RedHatInsights/insights-host-inventory/blob/813b9997687e5e48d464f501a8e475133bf8fbb4/test_api.py#L3931)

api/host.py
```diff
@@ -235,6 +234,8 @@ def patch_by_id(host_id_list, host_data):
 
     query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
 
+    query = find_hosts_by_staleness(ALL_STALENESS_STATES, query)
```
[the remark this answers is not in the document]
ryandillinfelton: Removed

api/host.py
```diff
@@ -302,6 +303,8 @@ def update_facts_by_namespace(operation, host_id_list, namespace, fact_dict):
 def get_host_tag_count(host_id_list, page=1, per_page=100, order_by=None, order_how=None):
     query = Host.query.filter((Host.account == current_identity.account_number) & Host.id.in_(host_id_list))
 
+    query = find_hosts_by_staleness(ALL_STALENESS_STATES, query)
+
```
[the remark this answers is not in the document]
ryandillinfelton: Nope

api/host.py
```diff
@@ -338,6 +341,8 @@ def _count_tags(host_list):
 def get_host_tags(host_id_list, page=1, per_page=100, order_by=None, order_how=None, search=None):
     query = Host.query.filter((Host.account == current_identity.account_number) & Host.id.in_(host_id_list))
 
+    query = find_hosts_by_staleness(ALL_STALENESS_STATES, query)
```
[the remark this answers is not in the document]
ryandillinfelton: Also nope. Switched both

test_api.py
```diff
@@ -2632,17 +2636,17 @@ def _create_host(self, stale_timestamp):
...
-    def _get_hosts_by_id(self, query, host_id_list):
-        url = self._get_hosts_by_id_url(query, host_id_list)
+    def _get_hosts_by_id(self, host_id_list, endpoint="", query=""):
+        url = self._get_hosts_by_id_url(host_id_list, endpoint, query)
         response = self.get(url, 200)
         return response["results"]
 
 
-class QueryStalenessGetHostsTestCase(QueryStalenessBaseTestCase, CullingBaseTestCase):
+class QueryStalenessGetHostsBaseCase(QueryStalenessBaseTestCase, CullingBaseTestCase):
```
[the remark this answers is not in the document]
ryandillinfelton: :+1:

test_api.py
```diff
@@ -1126,16 +1126,21 @@ def _assert_events_are_valid(self, events, hosts, timestamp):
...
+            for host_id in host_ids:
+                if Host.query.filter(Host.id == host_id).first():
+                    retrieved_ids.append(host_id)
+        return retrieved_ids
+
     def _check_hosts_are_present(self, host_ids):
-        response = self._get_hosts(host_ids)
-        self.assertEqual(response["total"], len(host_ids))
+        retrieved_ids = self._get_hosts_from_db(host_ids)
+        self.assertEqual(len(retrieved_ids), len(host_ids))
```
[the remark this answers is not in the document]
ryandillinfelton: :+1:

test_api.py
ryandillinfelton: I think test_dont_get_only_culled does that. It checks both getting all hosts, and getting them by ID.

test_api.py
ryandillinfelton: Added

test_api.py
```diff
@@ -2780,82 +2786,125 @@ def test_get_only_stale_warning(self):
...
-# class QueryStalenessConfigFilterTestCase(QueryStalenessBaseTestCase):
-#     def _assert_host_state(self, host, state, assert_method):
-#         hosts = self._get_hosts_by_id(f"?staleness={state}", (host["id"],))
-#         assert_method(host["id"], tuple(host["id"] for host in hosts))
+    def test_patch_facts_works_on_non_culled(self):
+        url = HOST_URL + "/" + self.fresh_host["id"] + "/rhsm"
 
-#     def _assert_host_in_state(self, host, state):
-#         self._assert_host_state(host, state, self.assertIn)
+        self.patch(url, {"ARCHITECTURE": "patched"}, 404)
```
jharting: Hmm, how does it work if it returns 404?

test_api.py
```diff
@@ -2780,82 +2786,125 @@ def test_get_only_stale_warning(self):
...
 
-#     def test_stale_warning_timestamp(self):
-#         for culling_stale_warning_offset_days in (1, 7, 12):
-#             with self.subTest(culling_stale_warning_offset_days=culling_stale_warning_offset_days):
-#                 config = self.app.config["INVENTORY_CONFIG"]
-#                 config.culling_stale_warning_offset_days = culling_stale_warning_offset_days
 
-#                 stale_timestamp = datetime.now(timezone.utc) + timedelta(hours=1)
-#                 host = self._create_and_get_host(stale_timestamp)
+class QueryStalenessGetHostsIgnoresCulledTestCase(QueryStalenessGetHostsBaseTestCase, DeleteHostsBaseTestCase):
```
jharting: These tests seem to be all missing the `facts` segment in the URL. See https://cloud.redhat.com/docs/api/inventory#operations-hosts-api\.host\.replace_facts

test_api.py
ryandillinfelton: It doesn't, good catch. Now it does

test_api.py
ryandillinfelton: Fixed!

PR #547, opened 2019-12-12
release

jharting:


api/__init__.py
```diff
@@ -55,3 +56,11 @@ def _get_status_code(results):
         return results.status_code
     else:
         return -1
+
+
+def flask_json_response(json_data, status=200):
```
Glutexo: Awww

api/tag.py
```diff
@@ -0,0 +1,78 @@
...
+
+def is_enabled():
+    return flask.current_app.config["INVENTORY_CONFIG"].bulk_query_source == BulkQuerySource.xjoin
+
+
+@api_operation
+@metrics.api_request_time.time()
+def get_tags(search=None, tags=None, order_by=None, order_how=None, page=None, per_page=None):
+    if not is_enabled():
+        flask.abort(503)
```
Glutexo: I mentioned a decorator, because this is not the only place where this would be used. The host query would need this too. But that can be extracted later, no stress. 😀

api/tag.py
```diff
@@ -0,0 +1,78 @@
+import re
+from urllib.parse import quote_plus as url_quote
```
Glutexo: The argument with matching Elastic’s encoding rules is rather invalid. It’s the app’s responsibility to decode what it gets from REST and re-encode it for ES. Like this, xjoin-search exposes internal workings of itself and its dependencies. Let’s keep it though not to add unnecessary friction.

PR #548, opened 2019-12-16
add staleness param to /tags (RHCLOUD-2462)

jharting:


test_api.py
```diff
@@ -3132,7 +3132,91 @@ def test_query_variables_default(self, graphql_query, is_enabled):
...
+        graphql_query.assert_called_once_with(
+            TAGS_QUERY,
+            {
+                "order_by": "tag",
+                "order_how": "ASC",
+                "limit": 50,
+                "offset": 0,
+                "hostFilter": {"OR": [{"stale_timestamp": {"gte": ANY, "lte": ANY}}]},
+            },
         )
```
Glutexo: What is the intended meaning of these tests?

test_api.py
```diff
@@ -3132,7 +3132,91 @@ def test_query_variables_default(self, graphql_query, is_enabled):
...
+
+    @patch("api.tag.datetime")
+    def test_query_variables_staleness(self, datetime_mock, is_enabled):
+        now = datetime(2019, 12, 16, 10, 10, 6, 754201, tzinfo=timezone.utc)
+        datetime_mock.now = mock.Mock(return_value=now)
+
+        for staleness, expected in (
+            ("fresh", {"gte": "2019-12-16T10:10:06.754201+00:00"}),
+            ("stale", {"gte": "2019-12-09T10:10:06.754201+00:00", "lte": "2019-12-16T10:10:06.754201+00:00"}),
+            ("stale_warning", {"gte": "2019-12-02T10:10:06.754201+00:00", "lte": "2019-12-09T10:10:06.754201+00:00"}),
```
Glutexo: I’m missing tests that would provide multiple staleness values.

api/tag.py
```diff
@@ -68,7 +101,7 @@ def get_tags(search=None, tags=None, order_by=None, order_how=None, page=None, p
         }
 
     if tags:
-        variables["hostFilter"] = {"AND": [{"tag": Tag().from_string(tag).data()} for tag in tags]}
+        variables["hostFilter"]["AND"] = [{"tag": Tag().from_string(tag).data()} for tag in tags]
```
Glutexo: The resulting _hostFilter_ dictionary looks strange, when it has the _OR_ and _AND_ keys:

```python
{"OR": […], "AND": […]}
```

Although it is correct and understandable, still, it looks strange. Having all in an extra _AND_ list may be more expressive:

```python
{"AND": [{"OR": […]}, {"AND": […]}]}
```

To be honest, I don’t like the dictionary-drive _xjoin-search_ filtering interface much. List-driven filters may be more readable? 🤷🏻‍♂️

```python
(
    "AND",
    (
        ("tag", "first tag"),
        ("tag", "second tag"),
        (
            "OR",
            (
                ("staleness", "stale"),
                ("staleness", "fresh")
            )
        )
    )
)
```

api/tag.py
```diff
@@ -47,19 +51,48 @@
...
+    staleness_offset_ = staleness_offset()
 
-    variables = {"order_by": order_by, "order_how": order_how, "limit": limit, "offset": offset}
+    variables = {
+        "order_by": order_by,
+        "order_how": order_how,
+        "limit": limit,
+        "offset": offset,
+        "hostFilter": {
+            # we're not indexing null timestamps in ES
```
Glutexo: Does this mean that it’s not possible to get a host with an empty stale timestamp from Elastic? Or that such records are not stored there at all?

Glutexo:
Wait.

Glutexo:
Ok, never mind. Had to check the intervals once again. 🙂

api/tag.py
```diff
@@ -47,19 +51,48 @@
 """
 
 
+def staleness_filter(gte=None, lte=None):
+    filter = {}
+    if gte:
+        filter["gte"] = gte.isoformat()
+    if lte:
+        filter["lte"] = lte.isoformat()
+    return {"stale_timestamp": filter}
```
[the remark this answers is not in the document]
jharting: Actually I was wrong. We can use lt, gt operators. Just need to expose them via xjoin-search.

api/tag.py
jharting: The former - we cannot query for these records.

api/tag.py
jharting: With that approach you would lose type safety

test_api.py
```diff
@@ -3132,7 +3132,91 @@ def test_query_variables_default(self, graphql_query, is_enabled):
...
+                "offset": 0,
+                "hostFilter": {
+                    "OR": [
+                        {"stale_timestamp": {"gte": "2019-12-16T10:10:06.754201+00:00"}},
+                        {
+                            "stale_timestamp": {
+                                "gte": "2019-12-09T10:10:06.754201+00:00",
+                                "lte": "2019-12-16T10:10:06.754201+00:00",
+                            }
+                        },
```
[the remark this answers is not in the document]
jharting: This way it is easier to assure it if I am making a programming mistake writing the implementation I am not repeating the same mistake in the tests.

api/tag.py
```diff
@@ -47,19 +51,48 @@
 """
 
 
+def staleness_filter(gte=None, lte=None):
+    filter = {}
```
Glutexo: _filter_ is a Python built-in name.

test_api.py
```diff
@@ -3132,7 +3132,91 @@ def test_query_variables_default(self, graphql_query, is_enabled):
...
         graphql_query.assert_called_once_with(
-            TAGS_QUERY, {"order_by": "tag", "order_how": "ASC", "limit": 50, "offset": 0}
+            TAGS_QUERY, {"order_by": "tag", "order_how": "ASC", "limit": 50, "offset": 0, "hostFilter": {"OR": ANY}}
+        )
+
+    @patch_with_empty_response()
+    @patch("api.tag.datetime")
+    def test_query_variables_staleness_default(self, datetime_mock, graphql_query, is_enabled):
+        now = datetime(2019, 12, 16, 10, 10, 6, 754201, tzinfo=timezone.utc)
+        datetime_mock.now = mock.Mock(return_value=now)
```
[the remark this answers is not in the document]
jharting: Thanks!

test_api.py
jharting: My bad, these were replaced by `test_query_variables_staleness`

api/tag.py
Glutexo: Would I? How? The _dicts_ are not much typed. And even if so, does it actually matter? Python is not type-safe at all.

jharting: GraphQL is. The xjoin-search interface is not driven by dictionaries but by exact type definitions. Since python does not have types using dictionaries that match the GraphQL types is the natural choice. You can do list filters but you'll need to write a layer that translates those lists to dictionaries anyway.

Also, I am not convinced that the list option is more readable. Given that JSON is ubiquitous nowadays I think people are more used to interpreting dictionaries than nested lists.

PR #553, opened 2020-01-06
Use int64 format for system_memory_bytes

semtexzv:
Added int64 format specifier for system_memory_bytes. By default, openapi assumes int32, which means, that system profiles with more than 3gb memory weren't parsed by this client: https://github.com/RedHatInsights/patchman-clients, and could not be processed by SPM application

### DIFF
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -1010,6 +1010,7 @@ components:
           type: integer
         system_memory_bytes:
           type: integer
+          format: int64
         infrastructure_type:
           type: string
         infrastructure_vendor:


Glutexo:
I’d refrain from changing the specification because of a quirk of some library. But in this case, even the OpenAPI specification [says](https://swagger.io/specification/#dataTypes) that _integer_ is by default “signed 32 bits”. Sad, but true. So now, 64bit integer ought to be enough for everybody.

It may be worth consideration to check whether Connexion restricts integers the same way and write tests for a big memory amount number. Is this the only place where such a big number can appear in the system profile?

semtexzv:
I believe so. Only other places I found integers were paging counters, which should be fine with int32.

PR #554, opened 2020-01-07
intial commit - filter host tags

yashnalla:


swagger/api.spec.yaml
```diff
@@ -268,6 +268,12 @@ paths:
         - $ref: '#/components/parameters/pageParam'
         - $ref: '#/components/parameters/orderByParam'
         - $ref: '#/components/parameters/orderHowParam'
+        - in: query
```
jharting: Since this parameter is now going to be used more than once we should move the definition to the `components/parameters` section and import it where it is going to be used (similarly to how e.g. orderByParam is imported above)

app/utils.py
```diff
@@ -340,7 +340,7 @@ def _if_null(cls, value):
             return value
 
     @staticmethod
-    def create_tags_from_nested(nested_tags):
+    def create_tags_from_nested(nested_tags, search):
```
jharting: These util functions are supposed to be only doing one thing and doing it well. It is not right to add additional logic to them. It would be better do define a new searching function that filters a given list of tags down based on the search param. That way we can combine the functions when needed but also use/test in isolation.

app/utils.py
```diff
@@ -339,6 +339,37 @@ def _if_null(cls, value):
         else:
             return value
 
+    @staticmethod
+    def create_tags_from_nested_search(nested_tags, search):
```
jharting: This function duplicates a lot of logic from `create_tags_from_nested` plus adds search. That's wrong because:

1. if we need to tweak the tag conversion logic there are now two places to do it
1. instead of doing one thing the function now combines multiple independent aspects

A better way would be to create a function that is solely focused on tag filtering. It would take a list of tags as its input (the format that `create_tags_from_nested` produces) and filter that list based on the search term. The two functions could then be composed together
```
tags = filter_tags(create_tags_from_nested(tags), search)
```

That way we get the desired logic while keeping each function minimal and avoiding code duplication.

test_api.py
```diff
@@ -2990,6 +2990,34 @@ def test_get_tags_of_hosts_that_doesnt_exist(self):
 
         self.assertEqual(expected_response, host_tag_results["results"])
 
+    def test_get_filtered_by_search_tags_of_multiple_hosts(self):
+        """
+        send a request for tags to one host with the search='a'
```
jharting: can this test be parameterized to also test other more complex search terms than 'a'?

test_api.py
```diff
@@ -2990,6 +2990,34 @@ def test_get_tags_of_hosts_that_doesnt_exist(self):
...
+        for host, tags in zip(host_list, self.tags_list):
+            expected_response[str(host.id)] = [tag.data() for tag in tags]
+            filtered_responses[str(host.id)] = []
+
+        for key in expected_response.keys():
+            for elem in expected_response[key]:
+                values = list(filter(None, elem.values()))
+
+                if any("a" in value for value in values):
+                    filtered_responses[key].append(elem)
```
jharting: This can be simplified to
```
        filtered_responses = {}

        for host, tags in zip(host_list, self.tags_list):
            filtered_responses[str(host.id)] = [
                tag.data() for tag in tags if any(filter(lambda x: x is not None and "a" in x, tag.data().values()))
            ]
```

test_api.py
```diff
@@ -2990,6 +2990,29 @@ def test_get_tags_of_hosts_that_doesnt_exist(self):
...
+        """
+        host_list = self.added_hosts
+
+        filtered_responses = {}
+
+        for host, tags in zip(host_list, self.tags_list):
+            filtered_responses[str(host.id)] = [
+                tag.data()
+                for tag in tags
+                if any(filter(lambda x: x is not None and searchTerm in x, tag.data().values()))
```
jharting: This uses the same logic in the implementation and in the test i.e. if there is a problem in the logic the test will not reveal it. It would be better to define the expected values as a constant in the test. Even better would be to parameterize and define several search values alongside with expected responses. Take a look e.g. here https://github.com/RedHatInsights/insights-host-inventory/blob/96466403793f93c8b06a8b75ea7ba1225d287b0a/test_api.py#L2389

test_unit.py
```diff
@@ -452,6 +452,31 @@ def test_structured_to_nested_no_value(self):
 
         self._base_structured_to_nested_test(structured_tag, expected_nested_tag)
 
+    """
+    structure to filtered structured test
+    """
+
+    def _base_structured_to_filtered_test(self, structured_tags, expected_filtered_tags):
```
jharting: Can we add more testcases?
* search string matches namespace
* search string matches tag key
* search string matches x in for more than 1 tag
* search string is an empty string
* search string is a prefix of x (namespace, value, key)
* search string is a suffix of x (namespace, value, key)
...

api/host.py
```diff
@@ -345,16 +345,19 @@ def get_host_tags(host_id_list, page=1, per_page=100, order_by=None, order_how=N
...
     query = query.paginate(page, per_page, True)
 
-    tags = _build_serialized_tags(query.items)
+    tags = _build_serialized_tags(query.items, search)
 
     return _build_paginated_host_tags_response(query.total, page, per_page, tags)
 
 
-def _build_serialized_tags(host_list):
+def _build_serialized_tags(host_list, search):
```
ryandillinfelton: I'd prefer if the filtering wasn't baked into this function to keep it simpler. But if that's unreasonable give search a default value of None so that it can be called with just the host_list parameter specified. Otherwise I think this PR looks bueno

test_unit.py
```diff
@@ -457,6 +457,109 @@ def test_structured_to_nested_no_value(self):
...
+        structured_tags = [Tag("NS1", "key1", "val"), Tag(None), Tag("NS2", "key2"), Tag("NS3", "key3", "value3")]
+        expected_filtered_tags = [Tag("NS1", "key1", "val")]
+
+        self._base_structured_to_filtered_test(structured_tags, expected_filtered_tags, "key1")
+
+    def test_complex_filter(self):
+        structured_tags = [Tag("NS1", "key1", "val"), Tag(None), Tag("NS2", "key2"), Tag("NS3", "key3", "value3")]
+        expected_filtered_tags = [Tag("NS1", "key1", "val"), Tag("NS3", "key3", "value3")]
+
+        self._base_structured_to_filtered_test(structured_tags, expected_filtered_tags, "val")
```
Glutexo: How is this “complex”? To me it looks the same as the previous test.

test_unit.py
```diff
@@ -457,6 +457,109 @@ def test_structured_to_nested_no_value(self):
...
+        structured_tags = [Tag("NS1", "key1", "val"), Tag(None), Tag("NS2", "key2"), Tag("NS3", "key3", "value3")]
+        expected_filtered_tags = [Tag("NS1", "key1", "val"), Tag("NS2", "key2"), Tag("NS3", "key3", "value3")]
+
+        self._base_structured_to_filtered_test(structured_tags, expected_filtered_tags, "")
+
+    def test_space(self):
+        structured_tags = [Tag("NS1", "key1", "val"), Tag(None), Tag("NS2", "key2"), Tag("NS3", "key3", "value3")]
+        expected_filtered_tags = []
+
+        self._base_structured_to_filtered_test(structured_tags, expected_filtered_tags, " ")
```
Glutexo: I’d ask, is this defined behavior? I mean, we can’t receive an empty search string from the API. For unit tests, this is probably OK though.

test_unit.py
```diff
@@ -457,6 +457,109 @@ def test_structured_to_nested_no_value(self):
...
+        # value
+        structured_tags = [
+            Tag("NS1", "key1", "val"),
+            Tag(None),
+            Tag("NS2", "Key2", "something"),
+            Tag("NS3", "key3", "value3"),
+        ]
+        expected_filtered_tags = [Tag("NS1", "key1", "val"), Tag("NS3", "key3", "value3")]
+
+        self._base_structured_to_filtered_test(structured_tags, expected_filtered_tags, "val")
```
Glutexo: Can this be split into own test methods?

test_unit.py
```diff
@@ -457,6 +457,109 @@ def test_structured_to_nested_no_value(self):
 
         self._base_structured_to_nested_test(structured_tag, expected_nested_tag)
 
+    """
+    structure to filtered structured test
+    """
+
+    def _base_structured_to_filtered_test(self, structured_tags, expected_filtered_tags, searchTerm):
```
Glutexo: The name doesn’t explain much, what the method does. Also, snake case.

```suggestion
    def _base_structured_to_filtered_test(self, structured_tags, expected_filtered_tags, search_term):
```

test_unit.py
```diff
@@ -457,6 +457,109 @@ def test_structured_to_nested_no_value(self):
...
+        filtered_tags = Tag.filter_tags(structured_tags, searchTerm)
+        self.assertEqual(len(filtered_tags), len(expected_filtered_tags))
+
+        for i in range(len(filtered_tags)):
+            self.assertEqual(filtered_tags[i].namespace, expected_filtered_tags[i].namespace)
+            self.assertEqual(filtered_tags[i].key, expected_filtered_tags[i].key)
+            self.assertEqual(filtered_tags[i].value, expected_filtered_tags[i].value)
```
Glutexo: This should be possible to do as a simple comparison. Here, the order is preserved, so even without converting the lists to sets. Then, it may not even be necessary to have this in a common helper function.

test_unit.py
```diff
@@ -457,6 +457,109 @@ def test_structured_to_nested_no_value(self):
...
+        # value
+        structured_tags = [
+            Tag("NS1", "key1", "val"),
+            Tag(None),
+            Tag("NS2", "Key2", "something"),
+            Tag("NS3", "key3", "value3"),
+        ]
+        expected_filtered_tags = [Tag("NS3", "key3", "value3")]
+
+        self._base_structured_to_filtered_test(structured_tags, expected_filtered_tags, "ue3")
```
Glutexo: Dtto split.

api/host.py
yashnalla: done

Glutexo:
@yashnalla I know you are working on the suggested changes. I‘m going to merge this pull request, because your code already works and the PR itself is getting a bit bulky. I‘d like to ask you to create a new pull request once you have your patch ready and reference this one for context. Thanks!

test_unit.py
Glutexo: 🤨

PR #556, opened 2020-01-14
make CloudWatch log_stream configurable (and default to hostname)

jharting:
https://projects.engineering.redhat.com/browse/RHCLOUD-3864

### DIFF
--- a/app/logging.py
+++ b/app/logging.py
@@ -42,7 +42,7 @@ def _configure_watchtower_logging_handler():
     aws_secret_access_key = os.getenv("AWS_SECRET_ACCESS_KEY", None)
     aws_region_name = os.getenv("AWS_REGION_NAME", None)
     log_group = os.getenv("AWS_LOG_GROUP", "platform")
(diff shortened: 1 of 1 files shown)


jharting:
@Glutexo good feedback! I updated the PR.

app/logging.py
```diff
@@ -62,14 +62,8 @@ def _configure_watchtower_logging_handler():
...
-def _get_aws_logging_stream_name(namespace_filename):
-    try:
-        with open(namespace_filename) as namespace_fh:
-            return namespace_fh.read()
-    except FileNotFoundError:
-        namespace = DEFAULT_AWS_LOGGING_NAMESPACE
-        print(f"Error reading the OpenShift namepsace file.  Using {namespace} as aws logging stream name")
-        return namespace
+def _get_hostname():
+    return os.uname()[1]
```
Glutexo: Does _uname_ always have at least two parts? 🤔 If not, this would raise an _IndexError_.

Glutexo: Checked the documentation. _uname_ returns an object and it always has all the elements. That means that `uname()[1]` is safe. Using it as an iterable is considered legacy. `uname().nodename` would be more semantic.

PR #561, opened 2020-01-27
Allow user to enable xjoin with request header (RHCLOUD-3758)

ryandillinfelton:
We need a way to enable xjoin for only specific users, or in a specific environment. To facilitate this I'm adding a new header:x-rh-cloud-bulk-query-source. I am also adding a behaviour that checks the forwarded referer header and switches the bulk-query-source to xjoin when the URL contains /beta.

When set the header will take priority will override both the BulkQuerySource set by the OpenShift environment variable, and the URL based behaviour.

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -107,7 +107,7 @@ def _add_host(input_host):
     return add_host(input_host, staleness_timestamps(), update_system_profile=False)
 
 
-def _get_bulk_query_source():
+def get_bulk_query_source():
     if XJOIN_HEADER in connexion.request.headers:
         if connexion.request.headers[XJOIN_HEADER].lower() == "xjoin":
             return BulkQuerySource.xjoin
@@ -116,10 +116,7 @@ def _get_bulk_query_source():
     if REFERAL_HEADER in connexion.request.headers:
         if "/beta" in connexion.request.headers[REFERAL_HEADER]:
             return inventory_config().bulk_query_source_beta
-        else:
-            return inventory_config().bulk_query_source
-    else:
-        return inventory_config().bulk_query_source
+    return inventory_config().bulk_query_source
 
 
 @api_operation
@@ -139,7 +136,7 @@ def get_host_list(
     total = 0
     host_list = ()
 
-    bulk_query_source = _get_bulk_query_source()
+    bulk_query_source = get_bulk_query_source()
 
     get_host_list = GET_HOST_LIST_FUNCTIONS[bulk_query_source]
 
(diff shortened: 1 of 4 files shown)


api/host.py
```diff
@@ -104,6 +107,19 @@ def _add_host(input_host):
...
 
+def _get_bulk_query_source():
+    if XJOIN_HEADER in connexion.request.headers:
+        if connexion.request.headers[XJOIN_HEADER].lower() == "xjoin":
+            return BulkQuerySource.xjoin
+        elif connexion.request.headers[XJOIN_HEADER].lower() == "db":
+            return BulkQuerySource.db
+    if REFERAL_HEADER in connexion.request.headers:
+        if "/beta" in connexion.request.headers[REFERAL_HEADER]:
+            return BulkQuerySource.xjoin
```
jharting: `/beta` should not automatically return "xjoin" - it should return whathever the value of `BULK_QUERY_SOURCE_BETA` env variable is i.e. a configuration option analogous to BULK_QUERY_SOURCE.

It's needed so that we can switch even the /beta client back to db in case we need to without releasing code. E.g. right now there is no xjoin in prod so we could not release this code as it would break `/beta` clients.

api/tag.py
```diff
@@ -49,15 +49,14 @@
 """
 
 
-def is_enabled():
-    config = inventory_config()
-    return config.bulk_query_source == BulkQuerySource.xjoin
+def xjoin_enabled():
+    return _get_bulk_query_source == BulkQuerySource.xjoin
```
jharting: Does this work? Looks like you're not calling the function at all.

api/host.py
```diff
@@ -104,6 +107,21 @@ def _add_host(input_host):
...
+        elif connexion.request.headers[XJOIN_HEADER].lower() == "db":
+            return BulkQuerySource.db
+    if REFERAL_HEADER in connexion.request.headers:
+        if "/beta" in connexion.request.headers[REFERAL_HEADER]:
+            return inventory_config().bulk_query_source_beta
+        else:
+            return inventory_config().bulk_query_source
+    else:
+        return inventory_config().bulk_query_source
```
jharting: \<nitpick\/\> if you shift this last return statement 4 spaces to the left and completely remove lines 119, 120, 121 you'll get identical behavior

api/host.py
```diff
@@ -104,6 +107,21 @@ def _add_host(input_host):
     return add_host(input_host, staleness_timestamps(), update_system_profile=False)
 
 
+def _get_bulk_query_source():
```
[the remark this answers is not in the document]
ryandillinfelton: Is there a more appropriate place to put it that `api/hosts` now that it's used in multiple files?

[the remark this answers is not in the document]
jharting: @ryandillinfelton perhaps here https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/__init__.py
@Glutexo WDYT?

PR #565, opened 2020-01-29
Push Host Reaper metrics

Glutexo:
Use a Prometheus Pushgateway to report metrics from the [Host Reaper](https://github.com/Glutexo/insights-host-inventory/blob/2fc6fd481bd3841d6fa40149f231146a0ae76192/host_reaper.py). A new [configuration value / environment variable](https://github.com/Glutexo/insights-host-inventory/blob/2fc6fd481bd3841d6fa40149f231146a0ae76192/app/config.py#L46) is used for the target.

### DIFF
--- a/host_reaper.py
+++ b/host_reaper.py
@@ -23,7 +23,7 @@ from tasks import init_tasks
 
 __all__ = ("main", "run")
 
-PROMETHEUS_JOB = "host_reaper"
+PROMETHEUS_JOB = "inventory-reaper"
 LOGGER_NAME = "host_reaper"
 COLLECTED_METRICS = (delete_host_count, delete_host_processing_time)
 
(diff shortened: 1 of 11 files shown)


host_reaper.py
```diff
@@ -30,33 +39,45 @@ def _init_db(config):
...
+    init_tasks(config)
+
+    registry = CollectorRegistry()
+    for metric in COLLECTED_METRICS:
+        registry.register(metric)
+
+    Session = _init_db(config)
+    session = Session()
+
+    run(config, session)
```
jharting: Can we add a metric for when this fails?

host_reaper.py
```diff
@@ -16,11 +21,15 @@
 from tasks import init_tasks
 
 
-__all__ = ("main",)
+__all__ = ("main", "run")
+
+PROMETHEUS_JOB = "host_reaper"
```
jharting: Make sure to update this to reflect the kubernetes namespace (`NAMESPACE` env var)

host_reaper.py
Glutexo: Sure.

Glutexo:
Rebased. I see some failed tests. Gonna fix.

Glutexo:
Tests are fixed. The PR should be now mergeable.

PR #566, opened 2020-01-30
Iterate over hosts to delete

Glutexo:
Instead of loading all hosts into memory on delete, use SQLAlchemy native iteration. This prevents the app/job getting killed for using too many resources.

Additional advantages include:

- The message queue events are emitted right after the host is deleted from the database.
- Host ID is logged and payload-tracked even if the host is not deleted because of a race condition.

Refactored the race condition tests and the rather cryptic iterator. Modified the Reaper tests so they use the same database connection. That should prevent database locking.

### DIFF
--- a/host_reaper.py
+++ b/host_reaper.py
@@ -17,6 +17,7 @@ from tasks import init_tasks
 
 
 __all__ = ("main",)
+SELECT_CHUNK_SIZE = 1000
 
 
 def _init_config(config_name):
@@ -35,7 +36,7 @@ def run(config, session):
 
     conditions = Conditions.from_config(config)
     query_filter = stale_timestamp_filter(*conditions.culled())
-    query = session.query(Host).filter(query_filter)
+    query = session.query(Host).filter(query_filter).yield_per(SELECT_CHUNK_SIZE)
 
     events = delete_hosts(query)
     for host_id, deleted in events:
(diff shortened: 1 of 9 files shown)


Glutexo:
Rebased on _master_.

host_reaper.py
```diff
@@ -30,29 +30,31 @@ def _init_db(config):
...
+
+
+def main(config_name):
+    config = _init_config(config_name)
+    init_tasks(config)
+
+    Session = _init_db(config)
+    session = Session()
+
+    run(config, session)
```
jharting: should this be wrapped with try/finally for the session to be closed properly in case something breaks?

Glutexo:
Added result chunking – manual, as native [_yield_per_](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.yield_per) has its [issues](https://stackoverflow.com/a/12233167/1307676). Also wrapped the Reaper with the existing database session [guard](https://github.com/Glutexo/insights-host-inventory/blob/7c9868da53828fca1b7907589b52efd7414ee1fd/lib/db.py#L5) to protect the session in case of an unexpected failure.

Glutexo:
Fix the unintentionally commited stuff.

jharting:
This was merged

PR #567, opened 2020-01-30
Add tag data to host output on egress kafka topic (RHCLOUD-4202, RHCLOUD-3835)

ryandillinfelton:
This PR adds tags and system_profile to egress messages. This covers changes to the information we put on egress requested by the Advisor team. I've merged both changes into the one PR since they are very similar changes. The relevant tickets are below:

https://projects.engineering.redhat.com/browse/RHCLOUD-4202
https://projects.engineering.redhat.com/browse/RHCLOUD-3835

### DIFF
--- a/app/queue/egress.py
+++ b/app/queue/egress.py
@@ -5,6 +5,8 @@ from marshmallow import fields
 from marshmallow import Schema
 
 from app.logging import get_logger
+from app.models import SystemProfileSchema
+from app.models import TagsSchema
 from app.queue import metrics
 
 
@@ -79,6 +81,8 @@ class HostSchema(Schema):
     stale_warning_timestamp = fields.Str()
     culled_timestamp = fields.Str()
     reporter = fields.Str()
+    tags = fields.List(fields.Nested(TagsSchema))
+    system_profile = fields.Nested(SystemProfileSchema)
 
 
 class HostEvent(Schema):
--- a/app/queue/ingress.py
+++ b/app/queue/ingress.py
@@ -14,10 +14,13 @@ from app.payload_tracker import PayloadTrackerContext
 from app.payload_tracker import PayloadTrackerProcessingContext
 from app.queue import metrics
 from app.queue.egress import build_event
+from app.serialization import DEFAULT_FIELDS
 from app.serialization import deserialize_host
 from lib import host_repository
 
 logger = get_logger(__name__)
+
+EGRESS_HOST_FIELDS = DEFAULT_FIELDS + ("tags", "system_profile")
 
 
 class OperationSchema(Schema):
@@ -93,7 +96,9 @@ def add_host(host_data):
             logger.info("Attempting to add host...")
             input_host = deserialize_host(host_data)
             staleness_timestamps = Timestamps.from_config(inventory_config())
-            (output_host, add_results) = host_repository.add_host(input_host, staleness_timestamps)
+            (output_host, add_results) = host_repository.add_host(
+                input_host, staleness_timestamps, fields=EGRESS_HOST_FIELDS
+            )
             metrics.add_host_success.labels(
                 add_results.name, host_data.get("reporter", "null")
             ).inc()  # created vs updated
--- a/app/serialization.py
+++ b/app/serialization.py
@@ -25,6 +25,20 @@ _CANONICAL_FACTS_FIELDS = (
     "external_id",
 )
 
+DEFAULT_FIELDS = (
+    "id",
+    "account",
+    "display_name",
+    "ansible_host",
+    "facts",
+    "reporter",
+    "stale_timestamp",
+    "stale_warning_timestamp",
+    "culled_timestamp",
+    "created",
+    "updated",
+)
+
 
 def deserialize_host(raw_data):
     try:
@@ -66,7 +80,7 @@ def deserialize_host_xjoin(data):
     return host
 
 
-def serialize_host(host, staleness_timestamps):
+def serialize_host(host, staleness_timestamps, fields=DEFAULT_FIELDS):
     if host.stale_timestamp:
         stale_timestamp = staleness_timestamps.stale_timestamp(host.stale_timestamp)
         stale_warning_timestamp = staleness_timestamps.stale_warning_timestamp(host.stale_timestamp)
@@ -76,22 +90,38 @@ def serialize_host(host, staleness_timestamps):
         stale_warning_timestamp = None
         culled_timestamp = None
 
-    return {
-        **serialize_canonical_facts(host.canonical_facts),
-        "id": _serialize_uuid(host.id),
-        "account": host.account,
-        "display_name": host.display_name,
-        "ansible_host": host.ansible_host,
-        "facts": _serialize_facts(host.facts),
-        "reporter": host.reporter,
-        "stale_timestamp": stale_timestamp and _serialize_datetime(stale_timestamp),
-        "stale_warning_timestamp": stale_timestamp and _serialize_datetime(stale_warning_timestamp),
-        "culled_timestamp": stale_timestamp and _serialize_datetime(culled_timestamp),
+    serialized_host = {**serialize_canonical_facts(host.canonical_facts)}
+
+    if "id" in fields:
+        serialized_host["id"] = _serialize_uuid(host.id)
+    if "account" in fields:
+        serialized_host["account"] = host.account
+    if "display_name" in fields:
+        serialized_host["display_name"] = host.display_name
+    if "ansible_host" in fields:
+        serialized_host["ansible_host"] = host.ansible_host
+    if "facts" in fields:
+        serialized_host["facts"] = _serialize_facts(host.facts)
+    if "reporter" in fields:
+        serialized_host["reporter"] = host.reporter
+    if "stale_timestamp" in fields:
+        serialized_host["stale_timestamp"] = stale_timestamp and _serialize_datetime(stale_timestamp)
+    if "stale_warning_timestamp" in fields:
+        serialized_host["stale_warning_timestamp"] = stale_timestamp and _serialize_datetime(stale_warning_timestamp)
+    if "culled_timestamp" in fields:
+        serialized_host["culled_timestamp"] = stale_timestamp and _serialize_datetime(culled_timestamp)
         # without astimezone(timezone.utc) the isoformat() method does not include timezone offset even though iso-8601
         # requires it
-        "created": _serialize_datetime(host.created_on),
-        "updated": _serialize_datetime(host.modified_on),
-    }
+    if "created" in fields:
+        serialized_host["created"] = _serialize_datetime(host.created_on)
+    if "updated" in fields:
+        serialized_host["updated"] = _serialize_datetime(host.modified_on)
+    if "tags" in fields:
+        serialized_host["tags"] = _serialize_tags(host.tags)
+    if "system_profile" in fields:
+        serialized_host["system_profile"] = host.system_profile_facts or {}
+
+    return serialized_host
 
 
 def serialize_host_system_profile(host):
@@ -142,3 +172,7 @@ def _serialize_uuid(u):
 def _deserialize_tags(tags):
     # TODO: Move the deserialization logic to this method.
     return Tag.create_nested_from_tags(Tag.create_structered_tags_from_tag_data_list(tags))
+
+
+def _serialize_tags(tags):
+    return [tag.data() for tag in Tag.create_tags_from_nested(tags)]
--- a/lib/host_repository.py
+++ b/lib/host_repository.py
@@ -5,6 +5,7 @@ from sqlalchemy import and_
 from app.logging import get_logger
 from app.models import db
 from app.models import Host
+from app.serialization import DEFAULT_FIELDS
 from app.serialization import serialize_host
 from lib import metrics
 from lib.db import session_guard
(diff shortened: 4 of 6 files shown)


ryandillinfelton:
To avoid changed the responses of the API in other situations I also added a parameter to the host serializer to specify additional fields to add to the serialized host. I figure we can expand the use of the parameter if we ever decide to make the response more configurable

app/serialization.py
```diff
@@ -66,7 +66,7 @@ def deserialize_host_xjoin(data):
     return host
 
 
-def serialize_host(host, staleness_timestamps):
+def serialize_host(host, staleness_timestamps, fields=[]):
```
Glutexo: It’s not a good practice to use mutable objects (here a _list_) as default function arguments. Use an empty tuple `()` or `None` instead. [Reference](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument)

lib/host_repository.py
```diff
@@ -31,7 +31,7 @@
 logger = get_logger(__name__)
 
 
-def add_host(input_host, staleness_offset, update_system_profile=True):
+def add_host(input_host, staleness_offset, update_system_profile=True, fields=[]):
```
Glutexo: Dtto mutable defaults.

app/serialization.py
```diff
@@ -93,6 +93,14 @@ def serialize_host(host, staleness_timestamps):
...
     }
 
+    # optional fields
+    if "tags" in fields:
+        serialized_host["tags"] = _serialize_tags(Tag.create_tags_from_nested(host.tags))
+    if "system_profile" in fields:
+        serialized_host["system_profile"] = host.system_profile_facts or {}
+
+    return serialized_host
+
```
Glutexo: I don’t know. I somehow considered the (de)serialize functions more use-case oriented than fully dynamic. There are already some: [_deserialize_host_xjoin_](https://github.com/RedHatInsights/insights-host-inventory/blob/60c19cdbb2116dd16cd544baa6ecd0df2c85dbab/app/serialization.py#L51), [_serialize_host_system_profile_](https://github.com/RedHatInsights/insights-host-inventory/blob/60c19cdbb2116dd16cd544baa6ecd0df2c85dbab/app/serialization.py#L97)… Moreover, if some fields are to be specified in the argument, then that should be true for all of them. I‘d rather add s new _serialize_host_egress_ function and extract the common part of this original function.

app/serialization.py
```diff
@@ -142,3 +150,12 @@ def _serialize_uuid(u):
 def _deserialize_tags(tags):
     # TODO: Move the deserialization logic to this method.
     return Tag.create_nested_from_tags(Tag.create_structered_tags_from_tag_data_list(tags))
+
+
+def _serialize_tags(tags):
+    json_tags = []
+
+    for tag in tags:
+        json_tags.append({"namespace": tag.namespace, "key": tag.key, "value": tag.value})
```
Glutexo: This is essentially [`tag.data()`](https://github.com/RedHatInsights/insights-host-inventory/blob/60c19cdbb2116dd16cd544baa6ecd0df2c85dbab/app/utils.py#L177).

app/serialization.py
```diff
@@ -142,3 +150,12 @@ def _serialize_uuid(u):
...
+
+    return json_tags
```
Glutexo: This can be done as a list comprehension.

```python
    return [tag.data() for tag in tags]
```

ryandillinfelton: Gotcha! that's slick

app/serialization.py
ryandillinfelton: I'd like to go fully dynamic with the function rather than having multiple versions of it. I see what you mean about having it be situational and that may well fit in better with the rest of the code, but I think having one flexible host serialization function might pay off in the future in case we get more request like this, or if we want to allow user to specify what data they want returned back (e.g. when Karel wanted tags in host response, but we didn't want to make all responses contain tags for performance reasons). I'd be happy to talk to you more about it in the future before we commit to one way or the other. In the meantime I'll switch to fully dynamic since that seems closer to what I've already written.

app/queue/egress.py
```diff
@@ -79,6 +79,8 @@ class HostSchema(Schema):
     stale_warning_timestamp = fields.Str()
     culled_timestamp = fields.Str()
     reporter = fields.Str()
+    tags = fields.List(fields.Dict(fields.Str()))
```
jharting: Can we reuse the existing schema definitions from here? https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/models.py#L302

app/serialization.py
Glutexo: These two approaches don’t necessarily need to be in contradiction. There can be a fully dynamic serialization method/class. This can be then called by situational methods, or using some kind of presets. That may also help with concert separation, because the serializer then wouldn’t have to know about the scenarios. 👍🏻

app/queue/egress.py
Glutexo: Vielleicht `fields.List(fields.Nested(TagsSchema))`?

app/serialization.py
```diff
@@ -25,6 +25,20 @@
...
+    "display_name",
+    "ansible_host",
+    "facts",
+    "reporter",
+    "stale_timestamp",
+    "stale_warning_timestamp",
+    "culled_timestamp",
+    "created",
+    "updated",
+)
```
Glutexo: Oh, a tuple! 💚

app/serialization.py
```diff
@@ -76,24 +90,32 @@ def serialize_host(host, staleness_timestamps, fields=[]):
...
-        "id": _serialize_uuid(host.id),
-        "account": host.account,
-        "display_name": host.display_name,
-        "ansible_host": host.ansible_host,
-        "facts": _serialize_facts(host.facts),
-        "reporter": host.reporter,
-        "stale_timestamp": stale_timestamp and _serialize_datetime(stale_timestamp),
-        "stale_warning_timestamp": stale_timestamp and _serialize_datetime(stale_warning_timestamp),
-        "culled_timestamp": stale_timestamp and _serialize_datetime(culled_timestamp),
+    serialized_host = {**serialize_canonical_facts(host.canonical_facts)}
```
Glutexo: The canonical facts fields are still not dynamic. Neither the exact fields, nor the whole set. 😁 But I don’t really mind.

app/serialization.py
```diff
@@ -76,24 +90,32 @@ def serialize_host(host, staleness_timestamps, fields=[]):
...
-    }
-
-    # optional fields
+    if "created" in fields:
+        serialized_host["created"] = _serialize_datetime(host.created_on)
+    if "updated" in fields:
+        serialized_host["updated"] = _serialize_datetime(host.modified_on)
     if "tags" in fields:
         serialized_host["tags"] = _serialize_tags(Tag.create_tags_from_nested(host.tags))
     if "system_profile" in fields:
```
Glutexo: I can imagine some key to function mapping to simplify this large block of _ifs_. Maybe next time.

```python
# lib/serialization/host.py
def id(host):
    return _serialize_uuid(host.id)

# app/serialization.py
def serialize_host(host, fields=("id",)):
    return {field: getattr("app.serialization.host", field)(host) for field in fields}

File: app/serialization.py [hunk 10]
```diff
@@ -153,9 +175,4 @@ def _deserialize_tags(tags):
 
 
 def _serialize_tags(tags):
-    json_tags = []
-
-    for tag in tags:
-        json_tags.append({"namespace": tag.namespace, "key": tag.key, "value": tag.value})
-
-    return json_tags
+    return [tag.data() for tag in tags]
```
Glutexo: 💖

app/queue/egress.py
Glutexo: `TagsSchema` confirmed working.

app/queue/ingress.py
```diff
@@ -93,7 +94,9 @@ def add_host(host_data):
             logger.info("Attempting to add host...")
             input_host = deserialize_host(host_data)
             staleness_timestamps = Timestamps.from_config(inventory_config())
-            (output_host, add_results) = host_repository.add_host(input_host, staleness_timestamps)
+            (output_host, add_results) = host_repository.add_host(
+                input_host, staleness_timestamps, fields=DEFAULT_FIELDS + ("tags", "system_profile")
```
Glutexo: I wouldn’t be afraid of naming the _fields_ value.
```python
EGRESS_HOST_FIELDS = DEFAULT_FIELDS + ("tags", "system_profile")
```
```suggestion
                input_host, staleness_timestamps, fields= EGRESS_HOST_FIELDS)
```

app/serialization.py
```diff
@@ -76,22 +90,38 @@ def serialize_host(host, staleness_timestamps):
...
         # requires it
-        "created": _serialize_datetime(host.created_on),
-        "updated": _serialize_datetime(host.modified_on),
-    }
+    if "created" in fields:
+        serialized_host["created"] = _serialize_datetime(host.created_on)
+    if "updated" in fields:
+        serialized_host["updated"] = _serialize_datetime(host.modified_on)
+    if "tags" in fields:
+        serialized_host["tags"] = _serialize_tags(Tag.create_tags_from_nested(host.tags))
```
Glutexo: The `Tag.create_tags_from_nested` call probably belongs to the __serialize_tags_ method.

test_mq_service.py
```diff
@@ -278,7 +278,37 @@ def test_host_add_with_system_profile(self, datetime_mock):
...
+        timestamp_iso = datetime_mock.utcnow.return_value.isoformat() + "+00:00"
+
+        host_data = {
+            "display_name": "test_host",
+            "insights_id": expected_insights_id,
+            "account": "0000001",
+            "tags": [
+                {"namespace": "NS1", "key": "key3", "value": "val3"},
+                {"namespace": "NS3", "key": "key2", "value": "val2"},
+                {"namespace": "Sat", "key": "prod", "value": None},
```
Glutexo: I‘m missing a missing namespace.

test_mq_service.py
```diff
@@ -278,7 +278,37 @@ def test_host_add_with_system_profile(self, datetime_mock):
             "type": "created",
         }
 
-        host_keys_to_check = ["display_name", "insights_id", "account"]
+        host_keys_to_check = ["display_name", "insights_id", "account", "system_profile"]
+
+        self._base_add_host_test(host_data, expected_results, host_keys_to_check)
+
+    @patch("app.queue.egress.datetime", **{"utcnow.return_value": datetime.utcnow()})
+    def test_add_host_with_tags(self, datetime_mock):
```
Glutexo: We can add tests for the default values. Without them, I can replace `serialized_host["system_profile"] = host.system_profile_facts or {}` with `serialized_host["system_profile"] = host.system_profile_facts` and the tests still pass.

```python
        expected_results = {
            "host": {**host_data},
            "platform_metadata": {},
            "timestamp": timestamp_iso,
            "type": "created",
            "system_profile": {}
        }
        host_keys_to_check = ["display_name", "insights_id", "account", "tags", "system_profile"]
```

```python
        expected_results = {
            "host": {**host_data},
            "platform_metadata": {},
            "timestamp": timestamp_iso,
            "type": "created",
            "tags": []
        }
        host_keys_to_check = ["display_name", "insights_id", "account", "system_profile", "tags"]
```

app/serialization.py
ryandillinfelton: Oh I see. So like one dynamic serialization function, and some situational functions that have a set selection of fields for that situation? That definitely solves for the issue I was having of sending the same tuple of fields in almost every call to the dynamic serializer, and the issue of tracking down every instance if we ever change what fields we want in one situation.

app/serialization.py
ryandillinfelton: :+1:

app/queue/ingress.py
ryandillinfelton: :+1:

test_mq_service.py
ryandillinfelton: Have you spotted any other shortcomings of the current MQ tests? If there are many more I might make a separate PR of just MQ tests improvements to not hold this PR up

ryandillinfelton:
I made some of the simple suggested improvement. Nothing should be all that different from the last review. 

I'm holding off on adding better mq add host tests to not hold up the PR more. I think the testing improvements can be their own PR since I imagine there are more general shortcomings in the MQ testing since it's still pretty new.

app/queue/egress.py
```diff
@@ -55,6 +56,18 @@ def build_event(event_type, host, metadata):
         raise ValueError(f"Invalid event type ({event_type})")
 
 
+class TagsSchema(Schema):
```
jharting: is there a reason to redefine this instead of reusing the same thing from app/models.py?

app/queue/egress.py
```diff
@@ -79,6 +92,8 @@ class HostSchema(Schema):
     stale_warning_timestamp = fields.Str()
     culled_timestamp = fields.Str()
     reporter = fields.Str()
+    tags = fields.List(fields.Nested(TagsSchema))
+    system_profile = fields.Dict()
```
jharting: same here - we already have a detailed definition in SystemProfileSchema

app/queue/egress.py
ryandillinfelton: Nope, definitely just a silly mistake. Just fixed it

PR #570, opened 2020-02-05
Check xjoin-search response (RHCLOUD-4272)

ryandillinfelton:
When xjoin-search returns a non 200 status code insights-host-inventory still tries to process it as if the request was successful. This behaviour is part of what cause the bug mentioned in the title of the PR.

This PR adds a check to make sure that the request to xjoin was successful, and if not forwards the status code from xjoin.

### DIFF
--- a/app/xjoin.py
+++ b/app/xjoin.py
@@ -29,7 +29,7 @@ def graphql_query(query_string, variables):
     response = post(url_, json=payload, headers=_forwarded_headers())
     status = response.status_code
     if status != 200:
-        logger.debug("xjoin-search returned status: %s", status)
+        logger.debug("xjoin-search returned status: %s", 500)
         abort(status, "xjoin-search error")
 
     logger.debug("QUERY: response %s", response.text)
--- a/app/xjoin.py
+++ b/app/xjoin.py
@@ -29,8 +29,8 @@ def graphql_query(query_string, variables):
     response = post(url_, json=payload, headers=_forwarded_headers())
     status = response.status_code
     if status != 200:
-        logger.debug("xjoin-search returned status: %s", 500)
-        abort(status, "xjoin-search error")
+        logger.error("xjoin-search returned status: %s", status)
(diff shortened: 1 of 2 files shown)


app/xjoin.py
```diff
@@ -27,6 +27,11 @@ def graphql_query(query_string, variables):
     payload = {"query": query_string, "variables": variables}
 
     response = post(url_, json=payload, headers=_forwarded_headers())
+    status = response.status_code
+    if status != 200:
+        logger.debug("xjoin-search returned status: %s", status)
+        abort(status, "xjoin-search error")
```
fijshion: I'm not totally sure how this abort function works. I'm assuming this will forward the status code received from xjoin on to the client. Instead I think this should always send a 500 regardless of the invalid code received from xjoin.

ryandillinfelton: You're correct, right now It forwards the status from xjoin-search. I'll swap it to a 500

fijshion: Looks like that only changes the log statement. The xjoin-search status code should be logged, then a 500 should be sent to the client via `abort(500)`. Another nit-picky thing is we probably don't need to tell the client "xjoin-search error". It's generally not good to leak internal stuff out to the client. Last thing - I think this should be an error level log statement.

app/xjoin.py
```diff
@@ -27,6 +27,11 @@ def graphql_query(query_string, variables):
     payload = {"query": query_string, "variables": variables}
 
     response = post(url_, json=payload, headers=_forwarded_headers())
+    status = response.status_code
+    if status != 200:
+        logger.debug("xjoin-search returned status: %s", 500)
```
jharting: let's make this an error rather than debug level

app/xjoin.py
ryandillinfelton: Whoops, changed the wrong line. Everything should be sorted out now and I've added some tests

PR #572, opened 2020-02-07
Cross-join host query Marshmallow schema deserialization

Glutexo:
Marshmallow can be used to deserialize timestamps. That fixes a bug with parsing timestamps from Cross-Join responses. Currently, a [specific format](https://github.com/RedHatInsights/insights-host-inventory/blob/313d1c0f2b432a3299a161f69faa6b47d7a111a0/app/serialization.py#L133) is required, but the service sometimes sends timestamps that don’t adhere to this subset. Marshmallow provides full ISO 8601 support. In addition this creates a contract between Inventory and Cross-Join, properly validating the responses. That can prevent processing invalid messages and running into later in the process.

### DIFF
--- a/api/host_query_xjoin.py
+++ b/api/host_query_xjoin.py
@@ -62,15 +62,15 @@ ORDER_HOW_MAPPING = {"modified_on": "DESC", "display_name": "ASC"}
 
 
 class HostQueryResponseHostDataCanonicalFactsSchema(Schema):
-    insights_id = String(validate=verify_uuid_format)
-    rhel_machine_id = String(validate=verify_uuid_format)
-    subscription_manager_id = String(validate=verify_uuid_format)
-    satellite_id = String(validate=verify_uuid_format)
-    bios_uuid = String(validate=verify_uuid_format)
-    ip_addresses = List(String(validate=Length(min=1, max=255)), validate=Length(min=1))
-    fqdn = String(validate=Length(min=1, max=255))
-    mac_addresses = List(String(validate=Length(min=1, max=59)), validate=Length(min=1))
-    external_id = String(validate=Length(min=1, max=500))
+    insights_id = String(allow_none=True, validate=verify_uuid_format)
+    rhel_machine_id = String(allow_none=True, validate=verify_uuid_format)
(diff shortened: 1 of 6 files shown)


Glutexo:
Added an extensive [test suite](https://github.com/Glutexo/insights-host-inventory/blob/4a99df9081d7cb3f5ea74e077abc325f3eaa5952/test_api.py#L3906) verifying that the whole schema is validated properly. All the validation rules are copied from [_HostSchema_](https://github.com/Glutexo/insights-host-inventory/blob/4a99df9081d7cb3f5ea74e077abc325f3eaa5952/app/models.py#L314), which is used to validate both message queue and REST API input. In an ideal world, that would mean that everything that is in the database is in accordance with those rules.

I didn’t allow _None_’s in [_display_name_](https://github.com/Glutexo/insights-host-inventory/blob/4a99df9081d7cb3f5ea74e077abc325f3eaa5952/api/host_query_xjoin.py#L79). For quite a long time, this value is automatically filled from _fqdn_ or _id_. Are we sure that all hosts in the database has this field filled in?

[_stale_timestamp_](https://github.com/Glutexo/insights-host-inventory/blob/4a99df9081d7cb3f5ea74e077abc325f3eaa5952/api/host_query_xjoin.py#L85) and [_reporter_](https://github.com/Glutexo/insights-host-inventory/blob/4a99df9081d7cb3f5ea74e077abc325f3eaa5952/api/host_query_xjoin.py#L86) can be _None_ as they are currently not required on input. Although such hosts are currently not copied to the cross-join database, its interface itself doesn’t prevent them to be returned.

Glutexo:
Allowed empty display name.

Glutexo:
Added a validation [script](https://github.com/Glutexo/insights-host-inventory/blob/52fe2279e9e5b3b623c83f476ce8f212cbe30585/utils/validate_hosts_for_xjoin.py) that runs the cross-join schema validation against all hosts in the database. Make sure to run with loglevel _INFO_ (not _DEBUG_) to avoid information leak.

Glutexo:
Tracked by [RHCLOUD-5489](https://projects.engineering.redhat.com/browse/RHCLOUD-5489). Closing to clean up the queue.

PR #574, opened 2020-02-12
Don't allow empty key names

fijshion:
This is a potential fix for ES not allowing empty string keys. I'm not sure if we have a way to test the performance of this. While it will need to look through the entire body of each request, the comparison is very simple.


jharting:
superseded by https://github.com/RedHatInsights/insights-host-inventory/pull/576

PR #577, opened 2020-02-14
Use dateutil.isoparse to parse timestamps

Glutexo:
Use _dateutil_ to parse ISO 8601 timestamps from the Cross-Join responses. This applies to _created_on_, _modified_on_ and _stale_timestamp_ fields. Timestamps without decimal part and with offset timezones are now accepted.

### DIFF
--- a/app/serialization.py
+++ b/app/serialization.py
@@ -1,6 +1,6 @@
-from datetime import datetime
 from datetime import timezone
 
+from dateutil.parser import isoparse
 from marshmallow import ValidationError
 
 from app.exceptions import InputFormatException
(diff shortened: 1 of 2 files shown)


Glutexo:
Added [_python-dateutil_](https://github.com/Glutexo/insights-host-inventory/blob/b0bfefa55c9861091ebb01c4b7fab4a3a3f3bbae/Pipfile#L31) as a direct dependency. It has been already [used](https://github.com/Glutexo/insights-host-inventory/blob/b0bfefa55c9861091ebb01c4b7fab4a3a3f3bbae/test_api.py#L25) in tests without being declared. Now it is used in the executive code. Fixed.

Glutexo:
Pipenv doesn’t allow to upgrade a single library. Upgrading everything in the current boundaries broke Connexion+Werkzeug. Upgraded Connexion to ~2.6.0 to resolve this. This version brought some breaking changes and some tests are now failing. Working on fixing this.

Glutexo:
Reverted to a state when dateutils were not in Pipfile. Ugly, but working. Dependencies will be resolved in a separate PR.

PR #578, opened 2020-02-17
add "captured_date" to system profile

beav:
This commit adds a "captured_date" field to the system profile. This
date records when the system profile was captured on the client
system.

For example, if a system upload was created on a Monday but uploaded
on Tuesday, this date would be Monday. PUP will populate this from the
system time in the uploaded tarball.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -289,7 +289,7 @@ class SystemProfileSchema(Schema):
     installed_products = fields.List(fields.Nested(InstalledProductSchema()))
     insights_client_version = fields.Str(validate=validate.Length(max=50))
     insights_egg_version = fields.Str(validate=validate.Length(max=50))
-    captured_date = fields.Str(validate=validate.Length(max=30))
+    captured_date = fields.Str(validate=validate.Length(max=32))
     installed_packages = fields.List(fields.Str(validate=validate.Length(max=512)))
     installed_services = fields.List(fields.Str(validate=validate.Length(max=512)))
     enabled_services = fields.List(fields.Str(validate=validate.Length(max=512)))
--- a/app/models.py
+++ b/app/models.py
@@ -289,6 +289,7 @@ class SystemProfileSchema(Schema):
     installed_products = fields.List(fields.Nested(InstalledProductSchema()))
     insights_client_version = fields.Str(validate=validate.Length(max=50))
     insights_egg_version = fields.Str(validate=validate.Length(max=50))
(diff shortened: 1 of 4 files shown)


beav:
@Glutexo @jharting is there a way to get the tests to re-run? it looks like they are stuck

Glutexo:
Just for the record, rebasing is a way to make the tests run again.

app/models.py
```diff
@@ -289,6 +289,7 @@ class SystemProfileSchema(Schema):
     installed_products = fields.List(fields.Nested(InstalledProductSchema()))
     insights_client_version = fields.Str(validate=validate.Length(max=50))
     insights_egg_version = fields.Str(validate=validate.Length(max=50))
+    captured_date = fields.Str(validate=validate.Length(max=50))
```
Glutexo: This is the first timestamp field we have in the system profile. I wonder whether we want to keep is as _String_ type, or whether to deserialize it as _DateTime_.

Since we don’t manipulate the System Profile data any way, only validating it’s structure, it should be OK to keep it as string – it makes things simple. That means the timestamps will be normalized neither in format, nor in timezone and precision.

Is it however still necessary to limit the length to 50? ISO 8601 dates are usually much shorter, even with a seconds decimal part.

beav: I prefer string as well. We had problems in the past where we'd do validation in the system_profile, but clients would send in unusual data. For example, one system had yum repo URLs that were not in a valid URL format, and we would reject the whole record because of that. By the time we validate the data, the user has already uploaded and received a 202, which makes it impossible to notify the user that there was an issue.

I don't have a strong preference on the length of the string. It looks like an iso8601 without any precision is 25 char. I'll change it to 30 char which should be plenty.

beav: i updated it to 30 just now

Glutexo: PostgreSQL stores timestamps with decimal precision of 6. That, with a long timezone format, makes the whole timestamp string 32 characters long. I’d use that as a safe value. _2020-02-05T11:56:07.337557+00:00_

Glutexo: Added a commit with that tiny change so we don’t have to wait. I didn’t modify the payloads so the maximum length is not tested. But we are not even expecting that long timestamps, so that shouldn’t be an issue. The standard one we expect is there and thus tested.

Glutexo:
@jharting Is this ok from your side? Merge?

PR #579, opened 2020-02-17
Add dateutil to Pipfile

Glutexo:
[_python-dateutil_](https://pypi.org/project/python-dateutil/) is used to parse the ISO 8601 timestamps. This library is already present as a deep dependency. There are some explicit [uses](https://github.com/Glutexo/insights-host-inventory/blob/db6fa3bc90cadcc403382e4c9f38be1d60cb41aa/test_api.py#L25) of _dateutil_ in the [tests](https://github.com/Glutexo/insights-host-inventory/blob/db6fa3bc90cadcc403382e4c9f38be1d60cb41aa/test_api.py). That could be potentially a problem as n-th level dependency can disappear at any time without notice. Fixed.

There is a bug making _Connexion_ and _Werkzeug_ fail together. [Upgrading](https://github.com/Glutexo/insights-host-inventory/blob/db6fa3bc90cadcc403382e4c9f38be1d60cb41aa/Pipfile#L20) _Connexion_ to 2.6.0 fixes the issue, but brings some breaking changes. Some tests are now failing.

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -491,7 +491,7 @@ class CreateHostsTestCase(DBAPITestCase):
         response_data = self.post(HOST_URL, [host_data.data()], 400)
 
         self.verify_error_response(
-            response_data, expected_title="Bad Request", expected_detail="'account' is a required property"
+            response_data, expected_title="Bad Request", expected_detail="'account' is a required property - '0'"
         )
 
     def test_create_host_with_invalid_account(self):
--- a/Pipfile
+++ b/Pipfile
@@ -17,7 +17,7 @@ flask-migrate = "~=2.5.2"
 itsdangerous = "~=1.1.0"
 request = "==2019.4.13"  # Not semver
 gunicorn = "~=19.9.0"
-connexion = {version = "~=2.3.0",extras = ["swagger-ui"]}
+connexion = {version = "~=2.6.0",extras = ["swagger-ui"]}
 pyyaml = "~=5.1.2"
 prometheus-client = "~=0.7.1"
 logstash-formatter = "~=0.5.17"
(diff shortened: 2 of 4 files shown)


Glutexo:
Fixed the tests.

- Manually [set](https://github.com/Glutexo/insights-host-inventory/blob/bda8430548c1d806a2e71dbe339d3e5adf0ddb4f/api/host.py#L178) DELETE HTTP response code to _200 OK_. Connexion new default is _204 No Content_. 
- Fix an exception message [expectation](https://github.com/Glutexo/insights-host-inventory/blob/bda8430548c1d806a2e71dbe339d3e5adf0ddb4f/test_api.py#L494) to reflect new format Connexion uses when validation fails.

Glutexo:
Here is the source for the different validation error message format: https://github.com/zalando/connexion/pull/1063

test_api.py
```diff
@@ -491,7 +491,7 @@ def test_create_host_without_account(self):
         response_data = self.post(HOST_URL, [host_data.data()], 400)
 
         self.verify_error_response(
-            response_data, expected_title="Bad Request", expected_detail="'account' is a required property"
+            response_data, expected_title="Bad Request", expected_detail="'account' is a required property - '0'"
```
ryandillinfelton: what's the - '0' thing about?

Glutexo: Perfect question, my friend! I was wondering too and did some detective work.

It is a path to the invalid field. See this patch https://github.com/zalando/connexion/pull/1063 in Connexion. The posted data is a list and the missing field is in its first element (indexed 0), so the path (a list of keys/indexes down the structure) is _0_. It could be _0.system_profile_ or something longer if the invalid value was deeper down the rabbit hole.

Glutexo:
Merged. I hope that this big upgrade didn’t break anything not covered by our tests. Let’s see what QE says.

PR #582, opened 2020-02-19
Require culling fields in API specification

Glutexo:
Make _stale_timestamp_ and _reporter_ required fields for new hosts. That makes all hosts without either of the field rejected. Both null values and empty strings are rejected. This applies both to the HTTP REST and to the Message Queue API.

- For the HTTP API, the fields are [required](https://github.com/Glutexo/insights-host-inventory/blob/8aca13f864858a73b08b256258a219fdd1f7b486/swagger/api.spec.yaml#L589) on the [OpenAPI Specification](https://github.com/Glutexo/insights-host-inventory/blob/8aca13f864858a73b08b256258a219fdd1f7b486/swagger/api.spec.yaml) level. If the field is not present, the whole add host request fails with _400 Bad Request_. If it’s there, but is invalid, the request itself succeeds wth _207 Multi-Status_, but the host is not saved _400 Bad Request_.
- For the MQ API, a validation exception is [raised and logged](https://github.com/Glutexo/insights-host-inventory/blob/8aca13f864858a73b08b256258a219fdd1f7b486/app/queue/ingress.py#L73).

### DIFF
--- a/test_mq_service.py
+++ b/test_mq_service.py
@@ -398,7 +398,7 @@ class MQCullingTests(MQAddHostBaseClass):
                     with self.assertRaises(InventoryException):
                         handle_message(json.dumps(message), mock_event_producer)
 
-    def test_add_host_stale_timestamp_invalud_culling_fields(self):
+    def test_add_host_stale_timestamp_invalid_culling_fields(self):
         """
         tests to check the API will reject a host if it doesn’t have both
         culling fields. This should raise InventoryException.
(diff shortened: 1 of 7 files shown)


Glutexo:
Fixed the failing db model [test](https://github.com/Glutexo/insights-host-inventory/blob/7a1be977c8291b85559860b19807ac9439a61595/test_db_model.py#L130).

test_mq_service.py
```diff
@@ -348,44 +376,53 @@ def test_add_host_stale_timestamp(self, datetime_mock):
...
         """
+        mock_event_producer = mockEventProducer()
 
-        additional_host_data = {"reporter": "puptoo"}
-
-        self._base_incomplete_staleness_data_test(additional_host_data)
+        additional_host_data = (
+            {"stale_timestamp": "2019-12-16T10:10:06.754201+00:00", "reporter": ""},
+            {"stale_timestamp": "not a timestamp", "reporter": "puptoo"},
+            {"stale_timestamp": "", "reporter": "puptoo"},
```
ryandillinfelton: Seems like we should throw in null values here to make sure they're also being rejected

Glutexo: Good point! The distinction between a missing key and a missing value is important. Fixed. Thanks!

PR #583, opened 2020-02-19
RHCLOUD-4449 - Add message key to kafka event

tahmidefaz:
Jira: https://projects.engineering.redhat.com/browse/RHCLOUD-4449

### DIFF
--- a/app/queue/egress.py
+++ b/app/queue/egress.py
@@ -26,11 +26,13 @@ class KafkaEventProducer:
         self._kafka_producer = KafkaProducer(bootstrap_servers=config.bootstrap_servers)
         self._topic = config.host_egress_topic
 
-    def write_event(self, event):
-        logger.debug(f"Topic: {self._topic} => {event}")
+    def write_event(self, event, key):
+        logger.debug("Topic: %s, key: %s, event: %s", self._topic, key, event)
 
         try:
-            self._kafka_producer.send(self._topic, value=event.encode("utf-8"))
+            k = key.encode("utf-8") if key else None
+            v = event.encode("utf-8")
+            self._kafka_producer.send(self._topic, key=k, value=v)
             metrics.egress_message_handler_success.inc()
         except Exception:
             logger.exception("Failed to send event")
@@ -41,8 +43,8 @@ class NullEventProducer:
     def __init__(self):
         logger.info("Starting NullEventProducer()")
 
-    def write_event(self, event):
-        logger.debug("NullEventProducer - logging event: %s" % event)
+    def write_event(self, event, key=None):
+        logger.debug("NullEventProducer - logging key: %s, event: %s", key, event)
 
 
 @metrics.egress_event_serialization_time.time()
--- a/app/queue/ingress.py
+++ b/app/queue/ingress.py
@@ -135,7 +135,7 @@ def handle_message(message, event_producer):
 
         event = build_event(event_type, output_host, metadata)
 
-        event_producer.write_event(event)
+        event_producer.write_event(event, output_host["id"])
 
 
 def event_loop(consumer, flask_app, event_producer, handler=handle_message):
--- a/lib/host_delete.py
+++ b/lib/host_delete.py
@@ -42,4 +42,5 @@ def _deleted_by_this_query(host):
 
 def _emit_event(host):
     event = delete_event(host)
-    emit_event(event)
+    key = str(host.id)
+    emit_event(event, key)
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -16,8 +16,8 @@ logger = get_logger(__name__)
 
 
 class NullProducer:
-    def send(self, topic, value=None):
-        logger.debug("NullProducer - logging message:  topic (%s) - message: %s", topic, value)
+    def send(self, topic, value=None, key=None):
+        logger.debug("NullProducer - logging message:  topic (%s) - message: %s - message_id: %s", topic, value, key)
 
     def flush(self):
         logger.debug("NullProducer – flushing")
@@ -47,8 +47,8 @@ def _init_event_producer(config):
         return NullProducer()
 
 
-def emit_event(e):
-    producer.send(cfg.event_topic, value=e.encode("utf-8"))
+def emit_event(e, key):
+    producer.send(cfg.event_topic, key=key.encode("utf-8") if key else None, value=e.encode("utf-8"))
 
 
 def flush():
--- a/test_api.py
+++ b/test_api.py
@@ -158,8 +158,8 @@ class MockEmitEvent:
     def __init__(self):
         self.events = []
 
-    def __call__(self, e):
-        self.events.append(e)
+    def __call__(self, e, key=None):
+        self.events.append((e, key))
 
 
 class APIBaseTestCase(TestCase):
@@ -1168,21 +1168,17 @@ class DeleteHostsBaseTestCase(DBAPITestCase):
         url_part = ",".join(host_ids)
         return self.get(f"{HOST_URL}/{url_part}", 200)
 
-    def _assert_events_are_valid(self, events, hosts, timestamp):
-        self.assertEqual(len(events), len(hosts))
-        hosts_by_ids = {host.id: host for host in hosts}
+    def _assert_event_is_valid(self, event, key, host, timestamp):
+        self.assertIsInstance(event, dict)
+        expected_keys = {"timestamp", "type", "id", "account", "insights_id", "request_id"}
+        self.assertEqual(set(event.keys()), expected_keys)
 
-        for event in events:
-            self.assertIsInstance(event, dict)
-            expected_keys = {"timestamp", "type", "id", "account", "insights_id", "request_id"}
-            self.assertEqual(set(event.keys()), expected_keys)
+        self.assertEqual(timestamp.replace(tzinfo=timezone.utc).isoformat(), event["timestamp"])
+        self.assertEqual("delete", event["type"])
 
-            self.assertEqual(timestamp.replace(tzinfo=timezone.utc).isoformat(), event["timestamp"])
-            self.assertEqual("delete", event["type"])
+        self.assertEqual(host.insights_id, event["insights_id"])
 
-            self.assertIn(event["id"], hosts_by_ids)
-            deleted_host = hosts_by_ids[event["id"]]
-            self.assertEqual(deleted_host.insights_id, event["insights_id"])
+        self.assertEqual(key, host.id)
 
     def _check_hosts_are_present(self, host_ids):
         response = self._get_hosts(host_ids)
@@ -1245,17 +1241,18 @@ class HostReaperTestCase(DeleteHostsBaseTestCase, CullingBaseTestCase):
         return self.get(f"{HOST_URL}/{url_part}")
 
     def test_culled_host_is_removed(self, emit_event):
-        added_hosts = self._add_hosts(
+        added_host = self._add_hosts(
             ({"stale_timestamp": self.staleness_timestamps["culled"].isoformat(), "reporter": "some reporter"},)
-        )
-        added_host_id = added_hosts[0].id
-        self._check_hosts_are_present((added_host_id,))
+        )[0]
+        self._check_hosts_are_present((added_host.id,))
 
         self._run_host_reaper()
-        self._check_hosts_are_deleted((added_host_id,))
+        self._check_hosts_are_deleted((added_host.id,))
 
-        events = tuple(json.loads(event) for event in emit_event.events)
-        self._assert_events_are_valid(events, added_hosts, self.now_timestamp)
+        self.assertEqual(len(emit_event.events), 1)
+
+        event, key = emit_event.events[0]
+        self._assert_event_is_valid(json.loads(event), key, added_host, self.now_timestamp)
 
     def test_non_culled_host_is_not_removed(self, emit_event):
         hosts_to_add = []
@@ -1887,34 +1884,32 @@ class DeleteHostsEventTestCase(PreCreatedHostsBaseTestCase, DeleteHostsBaseTestC
             with patch("app.events.datetime", **{"utcnow.return_value": self.timestamp}):
                 url = f"{self.delete_url}{url_query}"
                 self.delete(url, 200, header, return_response_as_json=False)
-                return json.loads(m.events[0])
-
-    def _assert_event_is_valid(self, event):
-        self._assert_events_are_valid((event,), (self.host_to_delete,), self.timestamp)
+                event, key = m.events[0]
+                return json.loads(event), key
 
     def test_create_then_delete(self):
         self._check_hosts_are_present((self.host_to_delete.id,))
-        event = self._delete()
-        self._assert_event_is_valid(event)
+        event, key = self._delete()
+        self._assert_event_is_valid(event, key, self.host_to_delete, self.timestamp)
         self._check_hosts_are_deleted((self.host_to_delete.id,))
 
     def test_create_then_delete_with_branch_id(self):
         self._check_hosts_are_present((self.host_to_delete.id,))
-        event = self._delete(url_query="?branch_id=1234")
-        self._assert_event_is_valid(event)
+        event, key = self._delete(url_query="?branch_id=1234")
+        self._assert_event_is_valid(event, key, self.host_to_delete, self.timestamp)
         self._check_hosts_are_deleted((self.host_to_delete.id,))
 
     def test_create_then_delete_with_request_id(self):
         request_id = generate_uuid()
         header = {"x-rh-insights-request-id": request_id}
-        event = self._delete(header=header)
-        self._assert_event_is_valid(event)
+        event, key = self._delete(header=header)
+        self._assert_event_is_valid(event, key, self.host_to_delete, self.timestamp)
         self.assertEqual(request_id, event["request_id"])
 
     def test_create_then_delete_without_request_id(self):
         self._check_hosts_are_present((self.host_to_delete.id,))
-        event = self._delete(header=None)
-        self._assert_event_is_valid(event)
+        event, key = self._delete(header=None)
+        self._assert_event_is_valid(event, key, self.host_to_delete, self.timestamp)
         self.assertEqual("-1", event["request_id"])
 
 
--- a/test_mq_service.py
+++ b/test_mq_service.py
@@ -7,6 +7,7 @@ from datetime import timezone
 from unittest import main
 from unittest import TestCase
 from unittest.mock import call
+from unittest.mock import MagicMock
 from unittest.mock import Mock
 from unittest.mock import patch
 
@@ -31,15 +32,14 @@ class FakeKafkaMessage:
         self.value = message
 
 
-class mockEventProducer:
+class MockEventProducer:
     def __init__(self):
-        self.__data = {}
+        self.event = None
+        self.key = None
 
-    def write_event(self, val):
-        self.__data["write_event"] = val
-
-    def get_write_event(self):
-        return self.__data["write_event"]
+    def write_event(self, event, key):
+        self.event = event
+        self.key = key
 
 
 class MQServiceBaseTestCase(TestCase):
@@ -126,7 +126,7 @@ class MQServiceTestCase(MQServiceBaseTestCase):
 
 
 @patch("app.queue.ingress.build_event")
-@patch("app.queue.ingress.add_host", return_value=(None, None))
+@patch("app.queue.ingress.add_host", return_value=(MagicMock(), None))
 class MQServiceParseMessageTestCase(MQServiceBaseTestCase):
     def _message(self, display_name):
         return f'{{"operation": "", "data": {{"display_name": "hello{display_name}"}}}}'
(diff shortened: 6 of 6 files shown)


tahmidefaz:
Not sure if this the correct way to address the issue, but let me know if things need to be changed or improved! :slightly_smiling_face:

app/queue/egress.py
```diff
@@ -26,11 +26,11 @@ def __init__(self, config):
         self._kafka_producer = KafkaProducer(bootstrap_servers=config.bootstrap_servers)
         self._topic = config.host_egress_topic
 
-    def write_event(self, event):
+    def write_event(self, event, host):
```
ryandillinfelton: It looks like you only use the host ID, so can you make it only pass that instead of the entire host?

tahmidefaz: I initially had it set that way, but the tests started throwing a `NoneType is not subscriptable` error [here](https://github.com/RedHatInsights/insights-host-inventory/pull/583/files#diff-6328f2b569f25842a2178bfd1aaafde3R138). This was the easiest fix, but I will give another try! :+1:

tahmidefaz:
I just added the test, not sure if I did it correctly though. Let me know if you see anything :slightly_smiling_face:

app/queue/egress.py
```diff
@@ -41,8 +41,8 @@ class NullEventProducer:
     def __init__(self):
         logger.info("Starting NullEventProducer()")
 
-    def write_event(self, event):
-        logger.debug("NullEventProducer - logging event: %s" % event)
+    def write_event(self, event, host):
+        logger.debug("NullEventProducer - logging event: %s - message_key: %s" % event, host["id"])
```
ryandillinfelton: I'd like this to just be passed the ID instead of the whole object

ryandillinfelton: Also, does this one need a host_id as a key since it's a Null event?

test_api.py
```diff
@@ -155,8 +155,8 @@ class MockEmitEvent:
     def __init__(self):
         self.events = []
 
-    def __call__(self, e):
-        self.events.append(e)
+    def __call__(self, e, id):
+        self.events.append((e, id))
```
ryandillinfelton: Shouldn't id be part of the event instead of in a tuple with the event? Seems like it'd avoid all of the extra subscripting and be more true to how events are actually returned unless I'm missing something here

app/queue/egress.py
tahmidefaz: Yes, you are right! We don't need that here. I was going for consistency, I guess I thought too much!

test_api.py
tahmidefaz: Can you elaborate a little bit here? Key isn't part of the event right now ([here](https://github.com/RedHatInsights/insights-host-inventory/pull/583/files#diff-53f0a5ac005ca800cc1e581e9e303bd4R33)). Making `write_event` take only `event` instead of `event` and `host_id` will require parsing it back from a JSON obj to a dictionary. I am sure if that would be optimal if I did it... :thinking: 
`event` gets converted to a JSON in `build_event()` of `egress.py`

ryandillinfelton: I see. Disregard then, I thought the event included the key

tahmidefaz: Cool! :+1: :+1:

jharting:
@tahmidefaz the changes look good to me. There seems to be a failing test that will need to be investigated.

tahmidefaz:
@jharting that patch update for the `add_host` method gets the test to pass right now

Glutexo:
🤨 A review is coming.

app/queue/egress.py
```diff
@@ -26,11 +26,11 @@ def __init__(self, config):
         self._kafka_producer = KafkaProducer(bootstrap_servers=config.bootstrap_servers)
         self._topic = config.host_egress_topic
 
-    def write_event(self, event):
-        logger.debug(f"Topic: {self._topic} => {event}")
+    def write_event(self, event, host_id):
```
Glutexo: [_KafkaEventProducer_](https://github.com/tahmidefaz/insights-host-inventory/blob/65a96083e98f9cf05042ad1e73165f02a4773bff/app/queue/egress.py#L23) looks like a rather generic class. Let’s not bind it to the host logic. I’d rather call this argument _key_ just as _KafkaProducer_ does. This argument can be optional. It’s the caller’s responsibility to provide the key.

```suggestion
    def write_event(self, event, key=None):
```

Also the [_NullEventProducer.write_event_](https://github.com/tahmidefaz/insights-host-inventory/blob/65a96083e98f9cf05042ad1e73165f02a4773bff/app/queue/egress.py#L44) method interface should be compatible with this.

```python
    def write_event(self, event, key=None):
        logger.debug("NullEventProducer - logging event: %s, key: %s", event, key)
```

app/queue/egress.py
```diff
@@ -26,11 +26,11 @@ def __init__(self, config):
...
+        logger.debug(f"Topic: {self._topic}, Key: {host_id} => {event}")
```
Glutexo: With _logger_, please use it’s own string interpolation instead of Python’s native formatting. In that case, objects are not stringified if the logging does not take place. Also, the message format is a bit strange – the key has a prefix, but the event itself doesn’t.

```suggestion
        logger.debug(f"Topic: %s, key: %s, event: %s", self._topic, key, event)
```

app/queue/ingress.py
```diff
@@ -135,7 +135,8 @@ def handle_message(message, event_producer):
 
         event = build_event(event_type, output_host, metadata)
 
-        event_producer.write_event(event)
+        host_id = output_host["id"]
+        event_producer.write_event(event, host_id)
```
Glutexo: Just making a bit more concise.

```suggestion
        event_producer.write_event(event, output_host["id"])
```

tasks/__init__.py
```diff
@@ -47,8 +47,8 @@ def _init_event_producer(config):
         return NullProducer()
 
 
-def emit_event(e):
-    producer.send(cfg.event_topic, value=e.encode("utf-8"))
+def emit_event(e, id):
```
Glutexo: Same here, I’d name it _key_ and make it optional.

```suggestion
def emit_event(e, key=None):
```

app/queue/egress.py
```diff
@@ -26,11 +26,11 @@ def __init__(self, config):
...
 
         try:
-            self._kafka_producer.send(self._topic, value=event.encode("utf-8"))
+            self._kafka_producer.send(self._topic, key=host_id.encode("utf-8"), value=event.encode("utf-8"))
```
Glutexo:
```suggestion
            k = key.encode("utf-8") if key else None
            v = event.encode("utf-8")
            self._kafka_producer.send(self._topic, key=k, value=v)
```

tasks/__init__.py
```diff
@@ -47,8 +47,8 @@ def _init_event_producer(config):
...
+    producer.send(cfg.event_topic, value=e.encode("utf-8"), key=str(id).encode("utf-8"))
```
Glutexo: Assuming that this is a generic method that expects a message key string, I’d just treat the value as a string. It’s the caller’s responsibility to pass a string.

```suggestion
    k = key.encode("utf-8") if key else None
    v = e.encode("utf-8")
    producer.send(cfg.event_topic, key=k, value=v)
```

lib/host_delete.py
```diff
@@ -42,4 +42,4 @@ def _deleted_by_this_query(host):
 
 def _emit_event(host):
     event = delete_event(host)
-    emit_event(event)
+    emit_event(event, host.id)
```
Glutexo: Let’s convert the host ID to _str_ here.

```suggestion
    key = str(host.id)
    emit_event(event, key)
```

test_api.py
Glutexo: The event is only the actual payload. It may make sense though to make this an Event struct (namedtuple) with _payload_ and _key_ properties. A plain tuple should be fine here though.

```suggestion
    def __call__(self, e, id):
        self.events.append((e, key))
```

test_api.py
```diff
@@ -155,8 +155,8 @@ class MockEmitEvent:
     def __init__(self):
         self.events = []
 
-    def __call__(self, e):
-        self.events.append(e)
+    def __call__(self, e, id):
```
```suggestion
    def __call__(self, e, key=None):
```

test_api.py
```diff
@@ -1196,7 +1196,7 @@ def test_culled_host_is_removed(self, emit_event):
         self._run_host_reaper()
         self._check_hosts_are_deleted((added_host_id,))
 
-        events = tuple(json.loads(event) for event in emit_event.events)
+        events = tuple(json.loads(event[0]) for event in emit_event.events)
```
Glutexo: We can get rid of the magic number by unpacking the tuple.

```suggestion
        events = tuple(json.loads(event) for event, key in emit_event.events)
```

test_api.py
```diff
@@ -1829,7 +1829,7 @@ def _delete(self, url_query="", header=None):
             with patch("app.events.datetime", **{"utcnow.return_value": self.timestamp}):
                 url = f"{self.delete_url}{url_query}"
                 self.delete(url, 200, header, return_response_as_json=False)
-                return json.loads(m.events[0])
+                return json.loads(m.events[0][0])
```
Glutexo: This starts to be a bit cryptic.

```suggestion
                event, key = m.events[0]
                return json.loads(event)
```

test_mq_service.py
```diff
@@ -34,12 +34,16 @@ class mockEventProducer:
     def __init__(self):
         self.__data = {}
 
-    def write_event(self, val):
+    def write_event(self, val, host_id):
```
Glutexo:
```suggestion
    def write_event(self, val, key=None):
```

test_mq_service.py
```diff
@@ -34,12 +34,16 @@ class mockEventProducer:
...
         self.__data["write_event"] = val
+        self.__data["key"] = host_id
```
Glutexo:
```suggestion
        self.__data["key"] = key
```

test_mq_service.py
```diff
@@ -125,7 +129,7 @@ def test_handle_message_happy_path(self, datetime_mock):
 
 
 @patch("app.queue.ingress.build_event")
-@patch("app.queue.ingress.add_host", return_value=(None, None))
+@patch("app.queue.ingress.add_host", return_value=({"id": "fAkE-iD"}, None))
```
Glutexo: - We can even generate a random UUID here.
```suggestion
@patch("app.queue.ingress.add_host", return_value=({"id": str(uuid.uuid4())}, None))
```
- Or create _MagicMock_ instance.
```suggestion
@patch("app.queue.ingress.add_host", return_value=(MagicMock(), None))
```

test_mq_service.py
```diff
@@ -251,6 +255,27 @@ def test_handle_message_verify_metadata_is_not_required(self):
...
+        message = {"operation": "add_host", "data": host_data}
+
+        with self.app.app_context():
+            with unittest.mock.patch("app.queue.ingress.host_repository.add_host") as m:
+                m.return_value = ({"id": host_id, "insights_id": None}, AddHostResults.created)
+                mock_event_producer = mockEventProducer()
+                handle_message(json.dumps(message), mock_event_producer)
+
+                self.assertEqual(mock_event_producer.get_message_key(), m.return_value[0]["id"])
+
```
Glutexo: Maybe it’s already time to DRY the test case a bit. This is the third one already, repeating almost everything.

test_mq_service.py
```diff
@@ -251,6 +255,27 @@ def test_handle_message_verify_metadata_is_not_required(self):
                 json.loads(mock_event_producer.get_write_event())["host"][key], expected_results["host"][key]
             )
 
+    def test_handle_message_verify_message_key(self):
+        expected_insights_id = str(uuid.uuid4())
+        host_id = uuid.uuid4()
+
+        host_data = {
+            "display_name": "test_host",
+            "insights_id": expected_insights_id,
```
Glutexo:
```suggestion
            "insights_id": str(uuid.uuid4()),
```

test_mq_service.py
```diff
@@ -251,6 +255,27 @@ def test_handle_message_verify_metadata_is_not_required(self):
                 json.loads(mock_event_producer.get_write_event())["host"][key], expected_results["host"][key]
             )
 
+    def test_handle_message_verify_message_key(self):
+        expected_insights_id = str(uuid.uuid4())
```
Glutexo: This is used only in one place. But I see the symmetry with the other test, so ok.
```suggestion
```

test_mq_service.py
```diff
@@ -251,6 +255,27 @@ def test_handle_message_verify_metadata_is_not_required(self):
...
+        host_id = uuid.uuid4()
```
Glutexo: - Why is this so far from the actual place of use?
- This should be a string. It works with a UUID only because the  event producer is mocked and doesn’t call the _str.encode_ method.

```suggestion
```

test_mq_service.py
```diff
@@ -251,6 +255,27 @@ def test_handle_message_verify_metadata_is_not_required(self):
...
+            "insights_id": expected_insights_id,
+            "account": "0000001",
+            "stale_timestamp": "2019-12-16T10:10:06.754201+00:00",
+            "reporter": "test",
+        }
+        message = {"operation": "add_host", "data": host_data}
+
+        with self.app.app_context():
+            with unittest.mock.patch("app.queue.ingress.host_repository.add_host") as m:
+                m.return_value = ({"id": host_id, "insights_id": None}, AddHostResults.created)
```
Glutexo:
```suggestion
                host_id = str(uuid.uuid4())
                m.return_value = ({"id": host_id, "insights_id": None}, AddHostResults.created)
```

test_mq_service.py
```diff
@@ -251,6 +255,27 @@ def test_handle_message_verify_metadata_is_not_required(self):
...
+                mock_event_producer = mockEventProducer()
+                handle_message(json.dumps(message), mock_event_producer)
+
+                self.assertEqual(mock_event_producer.get_message_key(), m.return_value[0]["id"])
```
Glutexo: If we have the _patch_ call here, we can even use directly the _host_id_ variable.
```suggestion
                self.assertEqual(mock_event_producer.get_message_key(), host_id)
```

But nothing prevents us from patch from a decorator and then the _return_value_ access would be good.

```python
    @patch(
        "app.queue.ingress.host_repository.add_host",
        return_value=({"id": uuid.uuid4(), "insights_id": None}, AddHostResults.created)
    )
    def test_handle_message_verify_message_key(self, add_host):
        ...
        with self.app.app_context():
            mock_event_producer = mockEventProducer()

            self.assertEqual(mock_event_producer.get_message_key(), add_host.return_value[0]["id"])

File: test_mq_service.py (hunk 16 above)
Comment (tahmidefaz) [reply test_mq_service.py]: @Glutexo I went for the `MagicMock()` here!

File: test_mq_service.py (hunk 22 above)
Comment (tahmidefaz) [reply test_mq_service.py]: @Glutexo Your decorator solution looks great, so, I decided to go for it here.

File: test_mq_service.py (hunk 17 above)
Comment (tahmidefaz) [reply test_mq_service.py]: @Glutexo I can try thinking of ways to combine them together, here

**Comment by Glutexo** [comment]
@tahmidefaz Is this ready for a new review? If so, please ping me or click the re-request button. Thanks!

**Comment by tahmidefaz** [comment]
@Glutexo I am trying to merge some of the test cases together as you mentioned earlier. I'll be ready for re-review soon!

**Comment by tahmidefaz** [comment]
This is ready for review again.

Also opened [RHCLOUD-4903](https://projects.engineering.redhat.com/browse/RHCLOUD-4903) to address the integration testing stuff.

File: test_mq_service.py [hunk 23]
```diff
@@ -34,12 +35,16 @@ class mockEventProducer:
...
-    def write_event(self, val):
+    def write_event(self, val, key=None):
         self.__data["write_event"] = val
+        self.__data["key"] = key
 
     def get_write_event(self):
         return self.__data["write_event"]
 
+    def get_message_key(self):
+        return self.__data["key"]
```
Glutexo: This whole class is simply terrible. I wouldn’t mind refactoring it, but I am not pushing it. Added it to [my fork](https://github.com/Glutexo/insights-host-inventory/tree/tahmidefaz).

```python
class MockEventProducer:
    def __init__(self):
        self.event = None
        self.key = None

    def write_event(self, event, key=None):
        self.event = event
        self.key = key
```

test_mq_service.py
```diff
@@ -200,21 +205,22 @@ def _base_add_host_test(self, host_data, expected_results, host_keys_to_check):
...
             "account": "0000001",
             "stale_timestamp": "2019-12-16T10:10:06.754201+00:00",
             "reporter": "test",
         }
 
-        message = {"operation": "add_host", "platform_metadata": metadata, "data": host_data}
+    def test_handle_message_verify_metadata_pass_through(self):
+        request_id = str(uuid.uuid4())
+
+        metadata = {"request_id": request_id, "archive_url": "https://some.url"}
```
Glutexo: Just making this more concise.

```suggestion
        metadata = {"request_id": str(uuid.uuid4()), "archive_url": "https://some.url"}
```

test_mq_service.py
```diff
@@ -200,21 +205,22 @@ def _base_add_host_test(self, host_data, expected_results, host_keys_to_check):
...
+
+        message = {"operation": "add_host", "platform_metadata": metadata, "data": self._host_data()}
 
         expected_results = {"platform_metadata": {**metadata}}
```
Glutexo: This variable is completely useless. We can just use the _metadata_ variable down there in the comparison.

```python
        event = json.loads(mock_event_producer.event)
        self.assertEqual(event["platform_metadata"], metadata)
```

test_mq_service.py
```diff
@@ -227,26 +233,20 @@ def test_handle_message_verify_metadata_pass_through(self):
...
-        }
+    @patch("app.queue.ingress.host_repository.add_host")
+    def test_handle_message_verify_message_key_and_metadata_not_required(self, add_host):
+        add_host.return_value = (self._host_data(), AddHostResults.created)
 
-        message = {"operation": "add_host", "data": host_data}
+        message = {"operation": "add_host", "data": self._host_data()}
 
-        expected_results = {"host": {**host_data}}
+        expected_results = {"host": {**add_host.return_value[0]}}
```
Glutexo: Let’s be consistent and use `add_host.return_value[0]` in both cases. Or even better, assign the host data to a varibale and use that.

```python
        host_data = self._host_data()
        add_host.return_value = (host_data, AddHostResults.created)

        expected_results = {"host": {**host_data}}
```

Glutexo: Dtto here, this variable makes no sense. The host data can be used directly.

test_mq_service.py
```diff
@@ -227,26 +233,20 @@ def test_handle_message_verify_metadata_pass_through(self):
...
 
         mock_event_producer = mockEventProducer()
         with self.app.app_context():
             handle_message(json.dumps(message), mock_event_producer)
+            self.assertEqual(mock_event_producer.get_message_key(), expected_results["host"]["id"])
```
Glutexo:
```suggestion
            self.assertEqual(mock_event_producer.get_message_key(), host_data["id"])
```

test_mq_service.py
```diff
@@ -227,26 +233,20 @@ def test_handle_message_verify_metadata_pass_through(self):
...
 
-        for key in host_data.keys():
+        for key in expected_results["host"].keys():
             self.assertEqual(
                 json.loads(mock_event_producer.get_write_event())["host"][key], expected_results["host"][key]
             )
```
Glutexo: No no no.

- You are unmarshalling the whole event JSON string, which is an expensive operation, for every host key. It is possible to do the _json.loads_ just once before the loop.
- The loop itself is unnecessary. We can compare the whole `event["host"]` and `host_data` dictionaries directly.

```suggestion
        event = json.loads(mock_event_producer.event)
        self.assertEqual(event["host"], host_data)
```

app/queue/egress.py
```diff
@@ -26,11 +26,13 @@ def __init__(self, config):
         self._kafka_producer = KafkaProducer(bootstrap_servers=config.bootstrap_servers)
         self._topic = config.host_egress_topic
 
-    def write_event(self, event):
-        logger.debug(f"Topic: {self._topic} => {event}")
+    def write_event(self, event, key=None):
```
jharting: Can we actually make the key required i.e. the code should fail if we attempt to send a message without specifying the key?

tasks/__init__.py
```diff
@@ -47,8 +47,8 @@ def _init_event_producer(config):
         return NullProducer()
 
 
-def emit_event(e):
-    producer.send(cfg.event_topic, value=e.encode("utf-8"))
+def emit_event(e, key=None):
```
jharting: same as above - let's make the key required

jharting:
@tahmidefaz rebase on current master plus squash please

test_mq_service.py
```diff
@@ -190,67 +190,51 @@ def tearDown(self):
     def _base_add_host_test(self, host_data, expected_results, host_keys_to_check):
         message = {"operation": "add_host", "data": host_data}
 
-        mock_event_producer = mockEventProducer()
+        mock_event_producer = MockEventProducer()
```
jharting: can we add an assertion here that verifies that message key matches the host id?

test_mq_service.py
```diff
@@ -31,15 +32,14 @@ def __init__(self, message):
...
-        self.__data = {}
+        self.event = None
+        self.key = None
 
-    def write_event(self, val):
-        self.__data["write_event"] = val
-
-    def get_write_event(self):
-        return self.__data["write_event"]
+    def write_event(self, event, key=None):
```
jharting: Let's make this strict and fail if key we get is None - that way we can discover any caller who does not include key when calling a producer.

jharting:
merged

PR #584, opened 2020-02-20
Fill in stale_timestamp and reporter

Glutexo:
Added migrations that

* [fill](https://github.com/Glutexo/insights-host-inventory/blob/08689b9e27528b239293f79659b876fb0fa7ad3c/migrations/versions/ff39a83abac3_fill_in_stale_timestamp_and_reporter.py) the missing stale timestamp a reporter.
* [alter](https://github.com/Glutexo/insights-host-inventory/blob/08689b9e27528b239293f79659b876fb0fa7ad3c/migrations/versions/f48f739902ed_alter_stale_timestamp_and_reporter_to_.py) the columns to not accept a NULL value.

The [downgrade](https://github.com/Glutexo/insights-host-inventory/blob/08689b9e27528b239293f79659b876fb0fa7ad3c/migrations/versions/ff39a83abac3_fill_in_stale_timestamp_and_reporter.py#L53) is not 100% symmetric: the values are cleared for all records marked as migrated. It shouldn’t have been ever possible though to add hosts with only stale timestamp or only reporter filled in.

### DIFF
--- a/migrations/helpers.py
+++ b/migrations/helpers.py
@@ -1,0 +1,20 @@
+from contextlib import contextmanager
+
+from alembic import op
+from sqlalchemy.orm.session import Session
+
+
+__all__ = ("session",)
+
+
+@contextmanager
+def session():
+    bind = op.get_bind()
+    session = Session(bind)
+    try:
+        yield session
+    except Exception:
+        session.rollback()
+        raise
+    else:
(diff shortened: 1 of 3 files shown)


Glutexo:
Merging. I am not sure what happens to the migrations though.

PR #588, opened 2020-02-27
RHCLOUD-2766 - Graceful shutdown

tahmidefaz:
Jira: [Here](https://projects.engineering.redhat.com/browse/RHCLOUD-2766)

Current approach: _I am extending the Kafka Consumer class to add a graceful shutdown method and some flags to check and halt shutdown when an in-flight operation is in progress_

There is a failing test that still needs to be addressed. But, please let me know if this approach can be improved or needs to be changed entirely. :slightly_smiling_face:

### DIFF
--- a/app/queue/ingress.py
+++ b/app/queue/ingress.py
@@ -1,4 +1,5 @@
 import json
+import signal
 
 from marshmallow import fields
 from marshmallow import Schema
@@ -21,6 +22,7 @@ from lib import host_repository
 logger = get_logger(__name__)
 
 EGRESS_HOST_FIELDS = DEFAULT_FIELDS + ("tags", "system_profile")
+CONSUMER_POLL_TIMEOUT_MS = 1000
 
 
 class OperationSchema(Schema):
@@ -134,17 +136,22 @@ def handle_message(message, event_producer):
         event_producer.write_event(event, output_host["id"])
 
 
-def event_loop(consumer, flask_app, event_producer, handler=handle_message):
+def event_loop(consumer, flask_app, event_producer, handler, shutdown_handler):
     with flask_app.app_context():
-        logger.debug("Waiting for message")
-        for msg in consumer:
-            logger.debug("Message received")
-            try:
-                handler(msg.value, event_producer)
-                metrics.ingress_message_handler_success.inc()
-            except Exception:
-                metrics.ingress_message_handler_failure.inc()
-                logger.exception("Unable to process message")
+
+        signal.signal(signal.SIGTERM, shutdown_handler.signal_handler)  # For Openshift
+        signal.signal(signal.SIGINT, shutdown_handler.signal_handler)  # For Ctrl+C
+        while not shutdown_handler.shut_down():
+            msgs = consumer.poll(timeout_ms=CONSUMER_POLL_TIMEOUT_MS)
+            for topic_partition, messages in msgs.items():
+                for message in messages:
+                    logger.debug("Message received")
+                    try:
+                        handler(message.value, event_producer)
+                        metrics.ingress_message_handler_success.inc()
+                    except Exception:
+                        metrics.ingress_message_handler_failure.inc()
+                        logger.exception("Unable to process message")
 
 
 def initialize_thread_local_storage(metadata):
--- a/inv_mq_service.py
+++ b/inv_mq_service.py
@@ -1,4 +1,5 @@
 import os
+from signal import Signals
 
 from kafka import KafkaConsumer
 from prometheus_client import start_http_server
@@ -9,8 +10,22 @@ from app.config import RuntimeEnvironment
 from app.logging import get_logger
 from app.queue.egress import create_event_producer
 from app.queue.ingress import event_loop
+from app.queue.ingress import handle_message
 
 logger = get_logger("mq_service")
+
+
+class ShutdownHandler:
+    def __init__(self):
+        self._shutdown = False
+
+    def signal_handler(self, signum, frame):
+        signame = Signals(signum).name
+        logger.info("Gracefully Shutting Down. Received: %s", signame)
+        self._shutdown = True
+
+    def shut_down(self):
+        return self._shutdown
 
 
 def main():
@@ -31,7 +46,13 @@ def main():
 
     event_producer = create_event_producer(config, "kafka")
 
-    event_loop(consumer, application, event_producer)
+    try:
+        event_loop(consumer, application, event_producer, handle_message, ShutdownHandler())
+    finally:
+        logger.info("Closing consumer")
+        consumer.close(autocommit=True)
+        logger.info("Closing producer")
+        event_producer.close()
 
 
 if __name__ == "__main__":
--- a/test_mq_service.py
+++ b/test_mq_service.py
@@ -56,13 +56,18 @@ class MQServiceTestCase(MQServiceBaseTestCase):
         Test to ensure that an exception in message handler method does not cause the
         event loop to stop processing messages
         """
-        # fake_consumer = [FakeKafkaMessage("blah"), FakeKafkaMessage("fred"), FakeKafkaMessage("ugh")]
-        fake_consumer = [Mock(), Mock(), Mock()]
+        fake_consumer = Mock()
+        fake_consumer.poll.return_value = {"poll1": [Mock(), Mock(), Mock()]}
+
         fake_event_producer = None
-
         handle_message_mock = Mock(side_effect=[None, KeyError("blah"), None])
-        event_loop(fake_consumer, self.app, fake_event_producer, handler=handle_message_mock)
-
+        event_loop(
+            fake_consumer,
+            self.app,
+            fake_event_producer,
+            handler=handle_message_mock,
+            shutdown_handler=Mock(**{"shut_down.side_effect": (False, True)}),
+        )
         self.assertEqual(handle_message_mock.call_count, 3)
 
     def test_handle_message_failure_invalid_json_message(self):
@@ -115,6 +120,22 @@ class MQServiceTestCase(MQServiceBaseTestCase):
                         "type": "created",
                     },
                 )
+
+    def test_shutdown_handler(self):
+        fake_consumer = Mock()
+        fake_consumer.poll.return_value = {"poll1": [Mock(), Mock()]}
+
+        fake_event_producer = None
+        handle_message_mock = Mock(side_effect=[None, None])
+        event_loop(
+            fake_consumer,
+            self.app,
+            fake_event_producer,
+            handler=handle_message_mock,
+            shutdown_handler=Mock(**{"shut_down.side_effect": (False, True)}),
+        )
+        fake_consumer.poll.assert_called_once()
+        self.assertEqual(handle_message_mock.call_count, 2)
 
     # Leaving this in as a reminder that we need to impliment this test eventually
     # when the problem that it is supposed to test is fixed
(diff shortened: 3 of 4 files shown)


app/queue/ingress.py
```diff
@@ -138,17 +139,35 @@ def handle_message(message, event_producer):
...
+                for message in messages:
+                    try:
+                        handle_message(message.value, event_producer)
+                        metrics.ingress_message_handler_success.inc()
+                    except Exception:
+                        metrics.ingress_message_handler_failure.inc()
+                        logger.exception("Unable to process message")
+            if shutdown_handler.shutdown:
+                logger.info("Shutting down gracefully.")
+                consumer.close()
```
jharting: this commits the offsets by default but it would be better to be explicit in this case
```
consumer.close(autocommit=True)
```
so that it is clear what is happening

app/queue/ingress.py
```diff
@@ -138,17 +139,35 @@ def handle_message(message, event_producer):
...
+        self.shutdown = False
+
+    def signal_handler(self, signal, frame):
+        logger.info("Processing any in-flight operations before shutting down...")
+        self.shutdown = True
+
+
 def event_loop(consumer, flask_app, event_producer, handler=handle_message):
     with flask_app.app_context():
         logger.debug("Waiting for message")
```
jharting: This statement is no longer relevant

app/queue/ingress.py
```diff
@@ -138,17 +139,35 @@ def handle_message(message, event_producer):
...
-            try:
-                handler(msg.value, event_producer)
-                metrics.ingress_message_handler_success.inc()
-            except Exception:
-                metrics.ingress_message_handler_failure.inc()
-                logger.exception("Unable to process message")
+        shutdown_handler = ShutdownHandler()
+        signal.signal(signal.SIGTERM, shutdown_handler.signal_handler)
+        signal.signal(signal.SIGINT, shutdown_handler.signal_handler)
+        while True:
```
jharting: can we do `while not shutdown_handler.shutdown` instead of `while True` + `break`?

app/queue/ingress.py
```diff
@@ -138,17 +139,35 @@ def handle_message(message, event_producer):
...
+class ShutdownHandler:
+    def __init__(self):
+        self.shutdown = False
+
+    def signal_handler(self, signal, frame):
+        logger.info("Processing any in-flight operations before shutting down...")
+        self.shutdown = True
+
+
 def event_loop(consumer, flask_app, event_producer, handler=handle_message):
```
jharting: Whatever part of the code is responsible for initializing KafkaConsumer should be also responsible for closing. I.e. if a initialized reference is passed to us then we should not be closing the instance. Instead, this part of code should end the loop and give control back to https://github.com/RedHatInsights/insights-host-inventory/blob/master/inv_mq_service.py#L35 where close() will be called

app/queue/ingress.py
```diff
@@ -138,17 +139,35 @@ def handle_message(message, event_producer):
         event_producer.write_event(event)
 
 
+class ShutdownHandler:
+    def __init__(self):
+        self.shutdown = False
+
+    def signal_handler(self, signal, frame):
+        logger.info("Processing any in-flight operations before shutting down...")
```
jharting: can we log what type of signal we received?

tahmidefaz:
Addressing the failing test now

inv_mq_service.py
```diff
@@ -33,6 +33,8 @@ def main():
 
     event_loop(consumer, application, event_producer)
 
+    consumer.close(autocommit=True)
```
jharting: We should make sure this gets called even if the event_loop somehow produces an uncaught error (i.e. in a finally block). Even better if we can make it work using `with` python construct

jharting: Also, can we also close the kafka producer?

tahmidefaz: @jharting I went for the finally block instead of the `with` construct as mentioned earlier since  `__exit__()` isn't defined [here ](https://github.com/dpkp/kafka-python/blob/34dc36d755abe6ec647efff74e284595497634ea/kafka/consumer/group.py#L26).

Glutexo:
I believe this shouldn’t be that hard to test. I’ll try to think something out and keep you updated.

tahmidefaz:
@jharting Thanks! I haven't thought of it yet. I'm mostly battling the failing test now.

app/queue/ingress.py
```diff
@@ -138,17 +139,34 @@ def handle_message(message, event_producer):
         event_producer.write_event(event)
 
 
-def event_loop(consumer, flask_app, event_producer, handler=handle_message):
+class ShutdownHandler:
+    def __init__(self):
+        self._shutdown = False
+
+    def signal_handler(self, signum, frame):
+        signame = "SIGINT" if signum == 2 else "SIGTERM"
```
Glutexo:
```suggestion
        from signal import Signals
        signame = Signals(signum).name
```

Glutexo:
Confirming that the SIGTERM and SIGINT signals are caught correctly and that the application stops. Did not verify the processing completion and Kafka buffer flushing yet.

app/queue/ingress.py
```diff
@@ -138,17 +139,34 @@ def handle_message(message, event_producer):
...
+        logger.info("Gracefully Shutting Down. Received: %s", signame)
+        self._shutdown = True
+
+    def shutdown(self):
+        return self._shutdown
```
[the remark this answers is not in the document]
tahmidefaz: Yes, I initially had it set as a boolean. But, it made quite challenging to break stop the polling loop for the [MQServiceTestCase.test_event_loop_exception_handling](https://github.com/RedHatInsights/insights-host-inventory/pull/588/files#diff-145bac01820de07cb0ff908174f7ecf2L52) test. Having the `shutdown()` method instead of a boolean allows me to stop the polling loop after a certain amount of iteration in [FakeShutdownHandler ](https://github.com/RedHatInsights/insights-host-inventory/pull/588/files#diff-145bac01820de07cb0ff908174f7ecf2R44)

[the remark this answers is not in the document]
tahmidefaz: @Glutexo Yes, I initially had it set as a boolean. But, it made quite challenging to break stop the polling loop for the [MQServiceTestCase.test_event_loop_exception_handling](https://github.com/RedHatInsights/insights-host-inventory/pull/588/files#diff-145bac01820de07cb0ff908174f7ecf2L52) test. Having the `shutdown()` method instead of a boolean allows me to stop the polling loop after a certain amount of iteration in [FakeShutdownHandler ](https://github.com/RedHatInsights/insights-host-inventory/pull/588/files#diff-145bac01820de07cb0ff908174f7ecf2R44)

Glutexo:
It should be possible to set the timeout for iteration by passing _consumer_timeout_ms_ to the _KafkaConsumer_ constructor. [\[docs\]](https://kafka-python.readthedocs.io/en/master/apidoc/KafkaConsumer.html#kafka.KafkaConsumer)

> consumer_timeout_ms (int) – number of milliseconds to block during message iteration before raising StopIteration (i.e., ending the iterator). Default block forever [float(‘inf’)].

test_mq_service.py
```diff
@@ -41,6 +41,22 @@ def get_write_event(self):
...
+
+    def signal_handler(self, signum, frame):
+        pass
+
+    def shutdown(self):
+        if self.times_called < self.max_call_num:
+            self.times_called += 1
+            return False
+        else:
+            return True
```
Glutexo: This is overly complicated considering it is used only in one place and with default values. `Mock(**{"shutdown.side_effect": (False, True)})` would do the job. If we had more or better tests for the shutdown handler, this might become handy, but with the current state, we don’t even need the class.
I‘d consider more descriptive name for the class. Something saying that this things signals shutdown after its first/n-th use.

test_mq_service.py
```diff
@@ -56,12 +72,18 @@ def test_event_loop_exception_handling(self):
         event loop to stop processing messages
         """
         # fake_consumer = [FakeKafkaMessage("blah"), FakeKafkaMessage("fred"), FakeKafkaMessage("ugh")]
```
Glutexo: This is no longer relevant.

Glutexo:
I saw this comment:

> Also note that we're fetching messages in batches. By default batches of 10 messages. When shutting down gracefully we need to commit the right offset so that the messages are neither lost neither processed multiple times.

The documentation says this about polling and offset: [(source)](https://kafka-python.readthedocs.io/en/master/apidoc/KafkaConsumer.html#kafka.KafkaConsumer.poll)

> On each poll, consumer will try to use the last consumed offset as the starting offset and fetch sequentially. The last consumed offset can be manually set through seek() or automatically set as the last committed offset for the subscribed list of partitions.

We implemented the signal handler in a way that it finishes the current batch and then shuts down. That means offset shouldn’t be an issue. The default graceful termination period is 30 seconds, which should be enough fo process 10 messages. [(source)](https://docs.openshift.com/container-platform/3.7/dev_guide/deployments/advanced_deployment_strategies.html#graceful-termination)

> After the graceful termination period expires, a process that has not exited will be sent the KILL signal, which immediately ends the process. The terminationGracePeriodSeconds attribute of a pod or pod template controls the graceful termination period (default 30 seconds) and may be customized per application as necessary.

It‘s however good to know that if we wanted to interrupt in the middle of a batch, we’d need to manually seek. Thanks for bringing that up!

app/queue/ingress.py
```diff
@@ -138,17 +139,21 @@ def handle_message(message, event_producer):
...
+        while not shutdown_handler.shutdown():
+            msgs = consumer.poll(timeout_ms=1000)
+            for topic_partition, messages in msgs.items():
+                for message in messages:
+                    try:
+                        handler(message.value, event_producer)
+                        metrics.ingress_message_handler_success.inc()
+                    except Exception:
+                        metrics.ingress_message_handler_failure.inc()
+                        logger.exception("Unable to process message")
```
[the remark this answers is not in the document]
tahmidefaz: @Glutexo This looks good, but I have a concern about doing it this way though. This essentially does not shut down until it reads all the messages in the queue and this might lead to an indefinite blocking period even with the `consumer_timeout_ms`.
For example, if we have 15 messages in the queue when we send out the shutdown signal, it will finish reading those 15 messages before shutting. However, for some reason, if we receive 20 more messages while processing those 15 messages, it won't shut down until it finishes reading those new messages as well.
For this to actually shut down after we send out the signal, it would need a period of no messages in the queue. I am not sure how busy the MQ service is in the prod & pre-prod environments. So, I am not sure whether to stick to `consumer.poll()` here or iterate over the consumer itself like before. Do let me know what you think about the aforementioned issue.

app/queue/ingress.py
```diff
@@ -138,17 +139,21 @@ def handle_message(message, event_producer):
...
-            try:
-                handler(msg.value, event_producer)
-                metrics.ingress_message_handler_success.inc()
-            except Exception:
-                metrics.ingress_message_handler_failure.inc()
-                logger.exception("Unable to process message")
+        signal.signal(signal.SIGTERM, shutdown_handler.signal_handler)
+        signal.signal(signal.SIGINT, shutdown_handler.signal_handler)
+        while not shutdown_handler.shutdown():
+            msgs = consumer.poll(timeout_ms=1000)
```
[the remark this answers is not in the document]
tahmidefaz: I went for a constant here, as it was a much simpler fix.

app/queue/ingress.py
tahmidefaz: Ah, I thought this was for the `logger.info` statement! :100:

PR #590, opened 2020-02-29
Test account conditions in host queries

Glutexo:
We have already encountered a data leak because of a missing account number condition. There are however still no tests present that would catch a regression. Adding these to prevent further problems.


jharting:
@Glutexo this is awesome! Note that we wanted to do this for a while already. Tracked under https://projects.engineering.redhat.com/browse/RHCLOUD-2922

Glutexo:
Tracked now by [RHCLOUD-8504](https://projects.engineering.redhat.com/browse/RHCLOUD-8504). I would really want this revived, but I don’t want to work on it outside of the Sprint. Closing now, reopen if revived.

PR #585, opened 2020-02-21
Rhcloud2463/searching in host tags

yashnalla:
refinements from https://github.com/RedHatInsights/insights-host-inventory/pull/554/files#

PR #591, opened 2020-03-02
Rhcloud2463v2

yashnalla:
searching in host tags refinements

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -354,7 +354,7 @@ def get_host_tags(host_id_list, page=1, per_page=100, order_by=None, order_how=N
     return _build_paginated_host_tags_response(query.total, page, per_page, tags)
 
 
-def _build_serialized_tags(host_list, search=None):
+def _build_serialized_tags(host_list, search):
     response_tags = {}
 
     for host in host_list:
@@ -365,22 +365,8 @@ def _build_serialized_tags(host_list, search=None):
         tag_dictionaries = []
         for tag in tags:
             tag_dictionaries.append(tag.data())
-
         response_tags[str(host.id)] = tag_dictionaries
-
     return response_tags
-
-
-# def _build_serialized_tags(host_list, search):
-#     search_filter = Tag.filter_tags(search)
-#     return {str(host.id): _build_serialized_tags_for_host(host, search_filter) for host in host_list}
-
-
-# def _build_serialized_tags_for_host(host, search_filter):
-#     tags = Tag.create_tags_from_nested(host.tags)
-#     if search_filter:
-#         tags = filter(search_filter, tags)
-#     return [tag.data() for tag in tags]
 
 
 def _build_paginated_host_tags_response(total, page, per_page, tags_list):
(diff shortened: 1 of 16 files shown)


yashnalla:
https://github.com/RedHatInsights/insights-host-inventory/pull/554

Glutexo:
Can we close #585 now?

yashnalla:
Yea I went ahead and closed it

api/host.py
```diff
@@ -352,7 +352,7 @@ def get_host_tags(host_id_list, page=1, per_page=100, order_by=None, order_how=N
     return _build_paginated_host_tags_response(query.total, page, per_page, tags)
 
 
-def _build_serialized_tags(host_list, search):
+def _build_serialized_tags(host_list, search=None):
```
Glutexo: Is the default necessary? The function is called from [one single place](https://github.com/RedHatInsights/insights-host-inventory/blob/f94c8f8cb78cfee061168bbf5482edf875d47b94/api/host.py#L350) and the argument is always passed in.

test_unit.py
```diff
@@ -495,10 +493,15 @@ def test_search_matches_tag_key(self):
         self._base_structured_to_filtered_test(structured_tags, expected_filtered_tags, "key1")
 
     def test_complex_filter(self):
```
Glutexo: I still don’t know, why is this test called “complex”. 😄 But whatever.

test_unit.py
```diff
@@ -495,10 +493,15 @@ def test_search_matches_tag_key(self):
...
-        structured_tags = [Tag("NS1", "key1", "val"), Tag(None), Tag("NS2", "key2"), Tag("NS3", "key3", "value3")]
-        expected_filtered_tags = [Tag("NS1", "key1", "val"), Tag("NS3", "key3", "value3")]
+        structured_tags = [
+            Tag("NS1erTys", "kindLey1", "v!2#eyal"),
+            Tag(None),
+            Tag("yeNS2", "ke56ty2"),
+            Tag("NYES12E3", "keingRty3", "value3uiley"),
+        ]
+        expected_filtered_tags = [Tag("NS1erTys", "kindLey1", "v!2#eyal"), Tag("NYES12E3", "keingRty3", "value3uiley")]
```
Glutexo: Okay, let’s consider this nice _v!2#eyal_ string complex. 😅

Glutexo:
Now I see that the commented out pieces of code are those that would make the [filter](https://github.com/RedHatInsights/insights-host-inventory/blob/f94c8f8cb78cfee061168bbf5482edf875d47b94/app/utils.py#L343) function possible to use with Python’s native filtering. Are you going to use those pieces or throw them out?

api/host.py
yashnalla: that was an artifact from earlier I will go ahead an remove it

Glutexo:
Overridden by #613. Still I’d like to see what was failing here for the sake of gaining experience and sharing knowledge.

PR #592, opened 2020-03-02
Replica identity full

fijshion:
This is required to prevent truncated messages from being sent through the xjoin pipeline. See https://debezium.io/blog/2019/10/08/handling-unchanged-postgres-toast-values/

### DIFF
--- a/migrations/versions/5544cd265053_set_replica_identity_to_full_on_hosts_.py
+++ b/migrations/versions/5544cd265053_set_replica_identity_to_full_on_hosts_.py
@@ -15,6 +15,13 @@ branch_labels = None
 depends_on = None
 
 
+# Can be verified by:
+#
+#   SELECT "relreplident" FROM "pg_class" WHERE "oid" = 'hosts'::regclass;
+#
+# 'd' means DEFAULT, 'f' means FULL.
+
+
 def upgrade():
     op.execute('ALTER TABLE "hosts" REPLICA IDENTITY FULL')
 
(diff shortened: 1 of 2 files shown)


migrations/versions/5544cd265053_set_replica_identity_to_full_on_hosts_.py
```diff
@@ -0,0 +1,24 @@
...
+# revision identifiers, used by Alembic.
+revision = '5544cd265053'
+down_revision = 'f48f739902ed'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    conn.execute("ALTER TABLE hosts REPLICA IDENTITY FULL");
```
Glutexo:
```suggestion
    op.execute("ALTER TABLE \"hosts\" REPLICA IDENTITY FULL")
```

migrations/versions/5544cd265053_set_replica_identity_to_full_on_hosts_.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def downgrade():
+    conn.execute("ALTER TABLE hosts REPLICA IDENTITY DEFAULT");
```
Glutexo:
```suggestion
    op.execute("ALTER TABLE \"hosts\" REPLICA IDENTITY DEFAULT")
```

Glutexo:
I verified that the statement really sets the replica identity flag to the table. ✅ I didn’t check however how this affects WAL, Debezium, ElasticSearch or Cross Join. I trust your explanation of the intended effect.

fijshion:
thanks for the review. I made the suggested changes. I'm OK waiting to merge this until the other migration is in prod on Thursday.

PR #593, opened 2020-03-02
Emit update event when host is patched (RHCLOUD-1102)

ryandillinfelton:


ryandillinfelton:
@Glutexo Could you take a look at this and see if I'm going at it the right way? The unit test failure has me a little worried about my approach

app/queue/egress.py
```diff
@@ -92,3 +100,4 @@ class HostEvent(Schema):
     host = fields.Nested(HostSchema())
     timestamp = fields.DateTime(format="iso8601")
     platform_metadata = fields.Dict()
+    inventory_metadata = fields.Dict()
```
jharting: Let's define the schema for this

app/queue/egress.py
```diff
@@ -48,11 +48,19 @@ def write_event(self, event, key=None):
...
     if event_type in ("created", "updated"):
         return (
             HostEvent(strict=True)
-            .dumps({"type": event_type, "host": host, "platform_metadata": metadata, "timestamp": datetime.utcnow()})
+            .dumps(
+                {
+                    "type": event_type,
+                    "host": host,
+                    "platform_metadata": platform_metadata,
+                    "inventory_metadata": inventory_metadata,
```
jharting: Since inventory is who's producing these events I would call it just `metadata`. The inventory prefix is implied.

Glutexo:
@ryandillinfelton Sorry for the delay, I was out. I’ll take a look!

app/queue/egress.py
Glutexo: The prefix is however nicely symmetric with the _platform_metadata_ key. 🤷🏻‍♂️

app/queue/egress.py
```diff
@@ -48,11 +48,19 @@ def write_event(self, event, key=None):
...
     if event_type in ("created", "updated"):
         return (
             HostEvent(strict=True)
-            .dumps({"type": event_type, "host": host, "platform_metadata": metadata, "timestamp": datetime.utcnow()})
+            .dumps(
+                {
+                    "type": event_type,
+                    "host": host,
+                    "platform_metadata": platform_metadata,
+                    "metadata": metadata,
```
Glutexo, line 60: The only metadata that is put in here is request ID. The existing delete event has it directly in the object root. Since the original Egress event didn’t have this field, we are free to put it there too, aren’t we? Or is there a contract saying that the request ID is expected under the _metadata_ subfield.

```suggestion
                    "request_id": request_id,
```

app/events.py
```diff
@@ -34,3 +36,10 @@ def delete(host):
         )
         .data
     )
+
+
+def emit_patch_event(host, metadata=None):
+    key = str(host.id)
+    metadata = {"request_id": threadctx.request_id}
+    event = build_event("updated", host, None, metadata)
```
Glutexo, line 44: Or as suggested later on, we can ditch the metadata argument entirely.
```suggestion}
    event = build_event("updated", host, request_id=threadctx.request_id)
```

app/queue/egress.py
```diff
@@ -87,8 +95,13 @@ class HostSchema(Schema):
     system_profile = fields.Nested(SystemProfileSchema)
 
 
+class metadataSchema(Schema):
+    request_id = fields.Str()
+
+
```
Glutexo, line 101: We may not even need this at all. See other comments.

app/queue/egress.py
```diff
@@ -87,8 +95,13 @@ class HostSchema(Schema):
...
 class HostEvent(Schema):
     type = fields.Str()
     host = fields.Nested(HostSchema())
     timestamp = fields.DateTime(format="iso8601")
     platform_metadata = fields.Dict()
+    metadata = fields.Nested(metadataSchema())
```
Glutexo, line 107: This schema would be universal for both the delete event and the create/update event. Remove _metadata_ if request ID in root is ok.

```suggestion
class HostEvent(Schema):
    type = fields.Str()
    host = fields.Nested(HostSchema())
    timestamp = fields.DateTime(format="iso8601")
    platform_metadata = fields.Dict()
    insights_id = fields.Str()
    request_id = fields.Str()
    metadata = fields.Nested(HostEventMetadataSchema())
```

Glutexo:
- I suggested some code movements, but they are far from the original refactoring.
- Also, there is a conflict in the tests. Please fix.
- Removed the _on hold_ label as there is finally some work being done on this.

jharting:
@Glutexo the `metadata` dictionary is on purpose. We want to create a place for metadata that is common across all the events that we produce. Later, more than just request_id will be added.

The delete event defines metadata fields as top-level fields.
Created/updated events use `platform_metadata` which however is not defined by us but by (some of the) reporters.

Therefore, we want to introduce `metadata` here and then retroactively add it into the existing events also. Also see https://projects.engineering.redhat.com/browse/RHCLOUD-5224

app/queue/egress.py
```diff
@@ -48,11 +48,19 @@ def write_event(self, event, key=None):
 
 
 @metrics.egress_event_serialization_time.time()
-def build_event(event_type, host, metadata):
+def build_event(event_type, host, platform_metadata, metadata=None):
```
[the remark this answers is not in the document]
ryandillinfelton: Added in the first suggestion. :+1: 

As for the second, I agree we should refactor a lot of the queue stuff, but I'll hold off on that so we can have one big refactoring task.

Rerding the third, Like Jozef said below, the metadata field should eventually become more than just request ID in the future

app/queue/egress.py
ryandillinfelton: Jozef addressed this in his comment below. Marking as resolved.

Glutexo:
Based on @jharting’s explanation I gave some _Uptated:_ lines to my earlier comments. Some things I’d like to note:

- Even if there is going to be the _metadata_ dictionary in the event, we can still [pass in](https://github.com/RedHatInsights/insights-host-inventory/blob/f919931b145c8be0b57534bd05b11c03b5e3dc6c/app/queue/egress.py#L51) to the function just the request ID. It communicates better what are the properties of the event. Passing in the whole dictionary (and not an object) looks more like any arbitrary data can be passed it. I acknowledge that platform metadata is still a dictionary and would be nice to have a schema for it, but we just relay what we got, so it’s still different. I don’t insist on that change though.
- I can live with stuff not being moved from [_app.queue.egress_](https://github.com/RedHatInsights/insights-host-inventory/blob/f919931b145c8be0b57534bd05b11c03b5e3dc6c/app/queue/egress.py) as it’s just modifying the existing functions. I’d however insist on moving the new stuff to where it belongs and renaming the new stuff in _egress_.
- The metric is still incorrect and we should resolve it. Although not necessarily in this pull request.

Glutexo:
Host stale timestamp in the new event is not converted to UTC. The time is correct, but it is in a local time zone. Steps to reproduce:

1. Have a host in the database with a stale timestamp in a UTC time zone.
2. Listen to the _platform.inventory.events_ topic.
3. Set a local timestamp to something non-UTC.
4. PATCH the host with a new information, say a new display name.
5. See that the _host/stale_timestamp_ field in the event in in the local time zone and not in UTC.

test_api.py
```diff
@@ -1885,7 +1885,7 @@ def test_patch_produces_update_event(self, emit_event):
         patch_doc = {"display_name": "patch_event_test"}
         host_to_patch = self.added_hosts[0].id
         self.patch(f"{HOST_URL}/{host_to_patch}", patch_doc, 200)
-        print(emit_event.call_args)
+        # print(emit_event.call_args)
```
Glutexo, line 1888: 🤦🏻‍♂️

app/queue/egress.py
```diff
@@ -48,11 +48,19 @@ def write_event(self, event, key=None):
...
         return (
             HostEvent(strict=True)
-            .dumps({"type": event_type, "host": host, "platform_metadata": metadata, "timestamp": datetime.utcnow()})
+            .dumps(
+                {
+                    "type": event_type,
+                    "host": host,
+                    "platform_metadata": platform_metadata,
+                    "metadata": metadata,
+                    "timestamp": datetime.utcnow(),
```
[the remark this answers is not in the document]
ryandillinfelton: Switched

jharting:
Wrt the metric name I think it would be best to have a single metric for serialization times but use a label to distinguish different event types. No need to do that in this PR.

utils/kafka_consumer.py
```diff
@@ -8,7 +8,7 @@
 # from app.models import SystemProfileSchema
 
 
-HOST_EGRESS_TOPIC = os.environ.get("KAFKA_HOST_EGRESS_TOPIC", "platform.inventory.host-egress")
+HOST_EGRESS_TOPIC = os.environ.get("KAFKA_HOST_EGRESS_TOPIC", "platform.inventory.events")
```
Glutexo, line 11: Do we want to do this here? Like, are we already ditching the whole egress topic?

test_api.py
```diff
@@ -1889,13 +1894,38 @@ def test_invalid_data(self):
...
+                    "display_name": "patch_event_test",
+                    "reporter": "test",
+                    "id": self.added_hosts[0].id,
+                    "account": "000501",
+                    "ansible_host": None,
+                    "tags": [{}],
+                },
+                "platform_metadata": {},
+                "metadata": {"request_id": "-1"},
+                "timestamp": self.now_timestamp.replace(tzinfo=timezone.utc).isoformat(),
```
Glutexo, line 1924: Just use `datetime.now(timezone.utc)` instead of `datetime.utcnow` in [setUp](https://github.com/RedHatInsights/insights-host-inventory/blob/a8c07151bd386b8d1c90bddafa13a42b55df3c71/app/queue/egress.py#L65) and you can save the replace.

```python
        self.now_timestamp = datetime.now(timezone.utc)
```

```suggestion
                "timestamp": self.now_timestamp.isoformat(),
```

test_api.py
```diff
@@ -2869,43 +2899,50 @@ def test_get_system_profile_ignores_culled(self):
...
+    def test_put_facts_ignores_culled(self, emit_patch_event):
         url = HOST_URL + "/" + self.culled_host["id"] + "/facts/ns1"
 
         self.put(url, {"ARCHITECTURE": "patched"}, 400)
 
-    def test_put_facts_works_on_non_culled(self):  # broken
+    @patch("api.host.emit_patch_event")
+    def test_put_facts_works_on_non_culled(self, emit_patch_event):  # broken
         url = HOST_URL + "/" + self.fresh_host["id"] + "/facts/ns1"
         print(url)
```
Glutexo, line 2935: Forgotten print statement.

```suggestion
```

test_api.py
```diff
@@ -2869,43 +2899,50 @@ def test_get_system_profile_ignores_culled(self):
...
         self.put(url, {"ARCHITECTURE": "patched"}, 200)
 
-    def test_delete_ignores_culled(self):
+    @patch("api.host.emit_patch_event")
+    def test_delete_ignores_culled(self, emit_patch_event):
         url = HOST_URL + "/" + self.culled_host["id"]
 
         self.delete(url, 404)
 
     @patch("lib.host_delete.emit_event", new_callable=MockEmitEvent)
```
Glutexo, line 2944: Let’s remove the _new_callable_, it’s not used at all.

```suggestion
    @patch("lib.host_delete.emit_event")
```

test_api.py
```diff
@@ -2869,43 +2899,50 @@ def test_get_system_profile_ignores_culled(self):
...
+    @patch("api.host.emit_patch_event")
+    def test_put_facts_works_on_non_culled(self, emit_patch_event):  # broken
         url = HOST_URL + "/" + self.fresh_host["id"] + "/facts/ns1"
         print(url)
         self.put(url, {"ARCHITECTURE": "patched"}, 200)
 
-    def test_delete_ignores_culled(self):
+    @patch("api.host.emit_patch_event")
+    def test_delete_ignores_culled(self, emit_patch_event):
         url = HOST_URL + "/" + self.culled_host["id"]
```
Glutexo, line 2941:
```suggestion
```

test_api.py
```diff
@@ -2869,43 +2899,50 @@ def test_get_system_profile_ignores_culled(self):
 
 
 class QueryStalenessGetHostsIgnoresCulledTestCase(QueryStalenessGetHostsBaseTestCase, DeleteHostsBaseTestCase):
-    def test_patch_ignores_culled(self):
+    @patch("api.host.emit_patch_event")
```
Glutexo, line 2902: To be consistent with the [_delete_ tests](https://github.com/RedHatInsights/insights-host-inventory/blob/a8c07151bd386b8d1c90bddafa13a42b55df3c71/test_api.py#L2939), the very producing function (_tasks.emit_event_) should be mocked. Mocking _api.host.emit_patch_event_ would be analogical to mocking _lib.host_delete._emit_event_.

```suggestion
    @patch("api.host.emit_event")
```

test_api.py
```diff
@@ -2869,43 +2899,50 @@ def test_get_system_profile_ignores_culled(self):
...
+    def test_patch_ignores_culled(self, emit_event):
         url = HOST_URL + "/" + self.culled_host["id"]
 
         self.patch(url, {"display_name": "patched"}, 404)
 
-    def test_patch_works_on_non_culled(self):
+    @patch("api.host.emit_patch_event")
```
Glutexo, line 2908:
```suggestion
    @patch("api.host.emit_event")
```

test_api.py
```diff
@@ -2869,43 +2899,50 @@ def test_get_system_profile_ignores_culled(self):
...
+    def test_patch_works_on_non_culled(self, emit_patch_event):
         url = HOST_URL + "/" + self.fresh_host["id"]
 
         self.patch(url, {"display_name": "patched"}, 200)
 
-    def test_patch_facts_ignores_culled(self):
+    @patch("api.host.emit_patch_event")
```
Glutexo, line 2914:
```suggestion
    @patch("api.host.emit_event")
```

test_api.py
```diff
@@ -2869,43 +2899,50 @@ def test_get_system_profile_ignores_culled(self):
...
+    def test_patch_facts_ignores_culled(self, emit_patch_event):
         url = HOST_URL + "/" + self.culled_host["id"] + "/facts/ns1"
 
         self.patch(url, {"ARCHITECTURE": "patched"}, 400)
 
-    def test_patch_facts_works_on_non_culled(self):  # broken
+    @patch("api.host.emit_patch_event")
```
Glutexo, line 2920:
```suggestion
    @patch("api.host.emit_event")
```

test_api.py
```diff
@@ -2869,43 +2899,50 @@ def test_get_system_profile_ignores_culled(self):
...
+    def test_patch_facts_works_on_non_culled(self, emit_patch_event):  # broken
         url = HOST_URL + "/" + self.fresh_host["id"] + "/facts/ns1"
 
         self.patch(url, {"ARCHITECTURE": "patched"}, 200)
 
-    def test_put_facts_ignores_culled(self):
+    @patch("api.host.emit_patch_event")
```
Glutexo, line 2926:
```suggestion
    @patch("api.host.emit_event")
```

test_api.py
```diff
@@ -2869,43 +2899,50 @@ def test_get_system_profile_ignores_culled(self):
...
+    def test_put_facts_ignores_culled(self, emit_patch_event):
         url = HOST_URL + "/" + self.culled_host["id"] + "/facts/ns1"
 
         self.put(url, {"ARCHITECTURE": "patched"}, 400)
 
-    def test_put_facts_works_on_non_culled(self):  # broken
+    @patch("api.host.emit_patch_event")
```
Glutexo, line 2932:
```suggestion
    @patch("api.host.emit_event")
```

test_api.py
```diff
@@ -2869,43 +2899,50 @@ def test_get_system_profile_ignores_culled(self):
 
 
 class QueryStalenessGetHostsIgnoresCulledTestCase(QueryStalenessGetHostsBaseTestCase, DeleteHostsBaseTestCase):
```
Glutexo, line 2901: We can split this class int, one with the _patch_ event tests and one with the _delete_ event tests. Then, the _patch_ decorator could be only once to the whole class instead of repeating at every test method. Moreover we don’t even need to inherit from _DeleteHostsBaseTestCase_, including the _delete_ tests.

```python
@patch("api.host.emit_event")
class QueryStalenessPatchIgnoresCulledTestCase(QueryStalenessGetHostsBaseTestCase):
    ...

@patch("lib.host_delete.emit_event")
class QueryStalenessDeleteIgnoresCulledTestCase(QueryStalenessGetHostsBaseTestCase):
    ...
```

utils/kafka_consumer.py
ryandillinfelton: Nope, sorry. Didn't mean to commit this change, it was for testing

test_api.py
```diff
@@ -1889,13 +1894,38 @@ def test_invalid_data(self):
...
+                    "stale_timestamp": self.added_hosts[0].stale_timestamp.replace("T", " "),
+                    "display_name": "patch_event_test",
+                    "reporter": "test",
+                    "id": self.added_hosts[0].id,
+                    "account": "000501",
+                    "ansible_host": None,
+                    "tags": [{}],
+                },
+                "platform_metadata": {},
+                "metadata": {"request_id": "-1"},
```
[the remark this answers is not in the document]
ryandillinfelton: :+1:

api/host.py
```diff
@@ -234,6 +236,15 @@ def get_host_system_profile_by_id(host_id_list, page=1, per_page=100, order_by=N
     return flask_json_response(json_output)
 
 
+def emit_patch_event(host):
+    # Convert the host timestamp to UTC before outputting the event
+    host.stale_timestamp = host.stale_timestamp.astimezone(pytz.utc)
```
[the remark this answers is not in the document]
ryandillinfelton: If I just remove it, won't we then still have the bug? Would it be better to add a note to address this when we make the change to converting the timestamp when we pull it out of the database?

ryandillinfelton:
@Glutexo I think I got all the changes you requested in. 

I am curious about the interim solution for fixing the patch event timezone issue. I've left the current fix in for now and added a note for when we switch to correcting the timezone when we pull from the database. Please let me know if you think of a better solution.

Glutexo:
Note: If RHCLOUD-5000 gets merged first, the new event should have headers too.

api/host.py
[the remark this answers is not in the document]
Glutexo: We won’t because I [suggested](https://github.com/RedHatInsights/insights-host-inventory/pull/593#discussion_r405449407) calling [_serialize_host_](https://github.com/RedHatInsights/insights-host-inventory/blob/6a23e6549e43d592f87de0e7cbf5642441248cf7/app/serialization.py#L83). By that, time zone is [converted](https://github.com/RedHatInsights/insights-host-inventory/blob/6a23e6549e43d592f87de0e7cbf5642441248cf7/app/serialization.py#L160)[for](https://github.com/RedHatInsights/insights-host-inventory/blob/6a23e6549e43d592f87de0e7cbf5642441248cf7/app/serialization.py#L108) the stale timestamp.

Glutexo:
@ryandillinfelton I hope the suggestions and comments answer your question about the time zone patching. If not, feel free to ping me.

test_api.py
```diff
@@ -1889,13 +1896,54 @@ def test_invalid_data(self):
...
 
-    def test_invalid_host_id(self):
+    def test_invalid_host_id(self, emit_event):
         patch_doc = {"display_name": "branch_id_test"}
         host_id_lists = ["notauuid", f"{self.added_hosts[0].id},notauuid"]
         for host_id_list in host_id_lists:
             with self.subTest(host_id_list=host_id_list):
                 self.patch(f"{HOST_URL}/{host_id_list}", patch_doc, 400)
 
+    def _base_patch_produces_update_event_test(self, emit_event, request_id):
```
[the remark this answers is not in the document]
ryandillinfelton: Gotcha. I'll keep it in mind going forward

Glutexo:
Oh no, a conflict! 😱 That’s because #618 got merged. Please resolve before re-requesting a review. Thanks!

utils/kafka_consumer.py
Glutexo: It’s still there.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/home/rfelton/.local/share/virtualenvs/insights-host-inventory-Tltzlw87/bin/python"
+}
```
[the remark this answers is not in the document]
ryandillinfelton: Thanks for the suggestion! Got it set up now

jharting:
Putting "on hold" label as the actual rollout of this change will need to be coordinated. Implementation and reviews should continue despite the label though

test_api.py
```diff
@@ -1929,13 +1939,64 @@ def test_invalid_data(self):
...
+                ).isoformat(),
+                "subscription_manager_id": host.subscription_manager_id,
+                "updated": host.updated,
+            },
+            "platform_metadata": None,
+            "metadata": {"request_id": expected_request_id},
+            "timestamp": self.now_timestamp.isoformat(),
+        }
+
+        event_message = json.loads(emit_event.call_args[0][0])
```
Glutexo, line 1990:
```suggestion
        emit_event.assert_called_once()
        event_message = json.loads(emit_event.call_args[0][0])
```

Glutexo:
Still, we need one more ✅. @jharting, @tahmidefaz, @shdunning, @yashnalla?

test_api.py
```diff
@@ -1929,13 +1940,64 @@ def test_invalid_data(self):
...
+                "ip_addresses": host.ip_addresses,
+                "mac_addresses": host.mac_addresses,
+                "reporter": host.reporter,
+                "rhel_machine_id": host.rhel_machine_id,
+                "satellite_id": host.satellite_id,
+                "stale_timestamp": host.stale_timestamp,
+                "stale_warning_timestamp": (
+                    dateutil.parser.parse(host.stale_timestamp) + timedelta(weeks=1)
+                ).isoformat(),
+                "subscription_manager_id": host.subscription_manager_id,
```
jharting, line 1981: `tags` and `system_profile` are missing in the message

Glutexo: They are, but that’s not an issue. The message shape should have been the same as for create/update and there these fields are absent too. It’s probably a remnant from the REST API, but it’s always been like that. If we want these fields, let’s add it to both events. But that’s out of scope of this pull request.

jharting: No, in created/updated events these fields are present https://github.com/RedHatInsights/insights-host-inventory/blob/master/test_mq_service.py#L322

jharting:

Glutexo:
sorry

Glutexo:
Just reminding, since this is **on_hold**.

> Note: If RHCLOUD-5000 (#633) gets merged first, the new event should have headers too.

There needs to be tests for that too.

Glutexo:
Closed by accident. The buttons are too close to each other. Sorry. Reopening.

test_api.py
Glutexo: This was not always the case and I didn’t notice that we have [_EGRESS_HOST_FIELDS_](https://github.com/RedHatInsights/insights-host-inventory/blob/6f34e553d1ca6e43650e90c2cfc3cf534dea435d/app/queue/ingress.py#L24) now. It was me who instructed to remove the _tags_, at that time it was still produced with incorrect values though. Thanks for clarification!

test_api.py
```diff
@@ -1880,13 +1891,71 @@ def test_invalid_data(self):
...
+                "updated": self.added_hosts[0].updated,
+            },
+            "platform_metadata": None,
+            "metadata": {"request_id": expected_request_id},
+            "timestamp": self.now_timestamp.isoformat(),
+        }
+
+        emit_event.assert_called_once()
+        event_message = json.loads(emit_event.call_args[0][0])
+        self.assertEqual(event_message, expected_event_message)
```
Glutexo, line 1949: Let’s test the _key_ too and also the headers once added.

```suggestion
        emitted_event = emit_event.call_args[0]
        event_message = json.loads(emitted_event[0])
        self.assertEqual(event_message, expected_event_message)
        self.assertEqual(emitted_event[1], self.added_hosts[0].id)
        self.assertEqual(emitted_event[2], "updated")
```

jharting:
@ryandillinfelton with https://github.com/RedHatInsights/insights-host-inventory/pull/640 merged this will need to be rebased. One functional change that will be needed is to add an event_type header for the event you're introducing in this PR similarly to how it's done here https://github.com/RedHatInsights/insights-host-inventory/blob/master/lib/host_delete.py#L48

app/events.py
```diff
@@ -10,6 +10,7 @@
 logger = logging.getLogger(__name__)
 
 DELETE_EVENT_NAME = "delete"
+UPDATE_EVENT_NAME = "updated"
```
Glutexo, line 13: Maybe we could reuse Maybe we can reuse [_lib.host_repository.AddHostResults.updated.name_](https://github.com/RedHatInsights/insights-host-inventory/blob/c7e7c06cde4d5ae6bc4b24ce305015d0e38a660e/lib/host_repository.py#L29), but not necessary.

test_api.py
```diff
@@ -1882,13 +1893,74 @@ def test_invalid_data(self):
...
+            "metadata": {"request_id": expected_request_id},
+            "timestamp": self.now_timestamp.isoformat(),
+        }
+
+        emit_event.assert_called_once()
+        emitted_event = emit_event.call_args[0]
+        event_message = json.loads(emitted_event[0])
+        self.assertEqual(event_message, expected_event_message)
+        self.assertEqual(emitted_event[1], self.added_hosts[0].id)
+        self.assertEqual(emitted_event[2]["event_type"], "updated")
```
Glutexo, line 1954:
```suggestion
        self.assertEqual(emitted_event[2], {"event_type": "updated"})
```

jharting:
Looks great. I am removing the "on hold" label and I am going to merge this as soon as the tests finish.

Glutexo:
🎉🎉

PR #564, opened 2020-01-29
edit regex to accept all 'mark' characters

yashnalla:

PR #596, opened 2020-03-04
regex expansion

yashnalla:
New branch for 3819 due to merge hell

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,13 @@ pytest test_json_validators.py
 Depending on the environment, it might be necessary to set the DB related environment
 variables (INVENTORY_DB_NAME, INVENTORY_DB_HOST, etc).
 
+## Generate migration script
+
+Run this command to generate a new revision in `migrations/versions`
+```
+python manage.py db revision -m "Description of revision"
+```
+
 ## Running the server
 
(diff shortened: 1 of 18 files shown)


yashnalla:
referencing this earlier PR: https://github.com/RedHatInsights/insights-host-inventory/pull/564

yashnalla:
So I've been working on this for a while but I have concluded that it might not be possible to actually expand the regex to include the mark characters at all. While testing the tags endpoint at https://github.com/RedHatInsights/insights-host-inventory/pull/596/files#diff-dc4308fba8c70f6d0b9ab17c57fb8ab2L2202 I always kept failing the test with a status response code of 400, this is because the Flask http get has a url regex that is as follow: \'^[\\\\\\\\w.\\\\\\\\-%~]+\\\\\\\\/[\\\\\\\\w.\\\\\\\\-%~]+=[\\\\\\\\w.\\\\\\\\-%~]+, this regex does not allow any of the newer mark characters that our regex has been expanded to accept. As far as I can tell this regex is an inbuilt feature of Flask itself and unless we fork Flask or utilize a different web infrastructure altogether I can't think of a way to fix this.

Glutexo:
Fixed by #613. Closing, going to examine.

PR #597, opened 2020-03-04
Add feature to filter out hosts not registered with insights (RHCLOUD-4859)

ryandillinfelton:


ryandillinfelton:
only is a placeholder name for the parameter. If anyone reading this has a suggestion of what to call it that would make sense when we expand the functionality I'm open.

jharting:
@ryandillinfelton looks good. We should also add this filter to `/tags` resource

karelhala:
API client has been released as well - https://www.npmjs.com/package/@redhat-cloud-services/host-inventory-client/v/1.0.55

PR #600, opened 2020-03-09
RHCLOUD-4217: Add reporter to find existing host log statements

shdunning:
To help with debugging, log more data about the host during ingress.

### DIFF
--- a/app/queue/ingress.py
+++ b/app/queue/ingress.py
@@ -103,7 +103,11 @@ def add_host(host_data):
                 add_results.name, host_data.get("reporter", "null")
             ).inc()  # created vs updated
             # log all the incoming host data except facts and system_profile b/c they can be quite large
-            logger.info("Host %s: %s", add_results.name, {i: host_data[i] for i in host_data if i not in ('facts', 'system_profile')})
+            logger.info(
+                "Host %s: %s",
+                add_results.name,
+                {i: host_data[i] for i in host_data if i not in ("facts", "system_profile")},
+            )
             payload_tracker_processing_ctx.inventory_id = output_host["id"]
             return (output_host, add_results)
         except InventoryException:
--- a/app/queue/ingress.py
+++ b/app/queue/ingress.py
@@ -102,7 +102,8 @@ def add_host(host_data):
             metrics.add_host_success.labels(
                 add_results.name, host_data.get("reporter", "null")
             ).inc()  # created vs updated
-            logger.info("Host added")  # This definitely needs to be more specific (added vs updated?)
+            # log all the incoming host data except facts and system_profile b/c they can be quite large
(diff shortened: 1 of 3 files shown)


app/queue/ingress.py
```diff
@@ -124,17 +125,8 @@ def handle_message(message, event_producer):
...
-
-        # FIXME: verify operation type
         (output_host, add_results) = add_host(validated_operation_msg["data"])
-
-        if add_results == host_repository.AddHostResults.created:
-            event_type = "created"
-        else:
-            event_type = "updated"
-
-        event = build_event(event_type, output_host, metadata)
```
shdunning: no need for this b/c `add_resutls` in Enum type

app/queue/ingress.py
```diff
@@ -102,7 +102,8 @@ def add_host(host_data):
             metrics.add_host_success.labels(
                 add_results.name, host_data.get("reporter", "null")
             ).inc()  # created vs updated
-            logger.info("Host added")  # This definitely needs to be more specific (added vs updated?)
+            # log all the incoming host data except facts and system_profile b/c they can be quite large
+            logger.info("Host %s: %s", add_results.name, {i: host_data[i] for i in host_data if i not in ('facts', 'system_profile')})
```
shdunning: i wasn't sure whether to use `host_data` or `output_host` here -- either will work fine. also not sure if we should include the `id`, which might be useful

lib/host_repository.py
```diff
@@ -20,7 +20,7 @@
     "update_existing_host",
 )
 
-# FIXME:  rename this
+# FIXME:  rename this, but more importantly why use Enum over a str constant?
 AddHostResults = Enum("AddHostResults", ["created", "updated"])
```
shdunning: i don't understand how an Enum benefits us with how `AddHostResults` is being used

test_mq_service.py
```diff
@@ -154,6 +154,7 @@ def test_handle_message_unicode_not_damaged(self, add_host, build_event):
         for message in messages:
             with self.subTest(message=message):
                 add_host.reset_mock()
+                add_host.return_value = ({}, AddHostResults.updated)
```
shdunning: fixes borked unittest

shdunning:
ok @jharting, i reworked it a bit to modify the logging statement in ingress. i left some comments in the PR to hopefully explain myself.

lib/host_repository.py
Glutexo: String constants can work just as well. _Enum_ provides additional semantics though. I think this is a correct usage of _Enum_. We can speculate on whether we need any constants/enum here at all, but it’s still beter to have an _Enum_ than a magic number/string.

shdunning: @Glutexo but Enums _are_ both magic numbers and _magic_ strings :-) 
AFAICT from reading code, the only main value here is having `add_host` return to the caller which operation was performed so it can be logged and/or conditionally branched on.

PR #603, opened 2020-03-11
Change chatty log stmt from INFO to DEBUG to ease prod log surfing

shdunning:
To help with Kibana storage limitations, change this one log statement to DEBUG so it doesn't end up in Kibana.

### DIFF
--- a/app/queue/ingress.py
+++ b/app/queue/ingress.py
@@ -93,9 +93,9 @@ def add_host(host_data):
     ) as payload_tracker_processing_ctx:
 
         try:
-            logger.info("Attempting to add host...")
             input_host = deserialize_host(host_data)
             staleness_timestamps = Timestamps.from_config(inventory_config())
+            logger.info("Attempting to add host", extra={"canonical_facts": input_host.canonical_facts})
             (output_host, add_results) = host_repository.add_host(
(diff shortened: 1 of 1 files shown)


jharting:
@shdunning at this point where inventory barely logs anything I am hesitant to remove one of the few log statements that works. Are you sure this solves a real problem?

shdunning:
@jharting i doubt it will resolve the issue we're trying to track down with platform-tools, i was just questioning the value of this particular log statement. you would know better than i if it's useful in practice so feel free to do the right thing with this PR (nuke it if need be).

app/queue/ingress.py
```diff
@@ -104,9 +104,9 @@ def add_host(host_data):
             ).inc()  # created vs updated
             # log all the incoming host data except facts and system_profile b/c they can be quite large
             logger.info(
-                "Host %s: %s",
+                "Host %s",
                 add_results.name,
-                {i: output_host[i] for i in output_host if i not in ("facts", "system_profile")},
+                extra={"host": {i: output_host[i] for i in output_host if i not in ("facts", "system_profile")}},
```
shdunning: whoa @jharting was this the fix to get unstructured data to show up in Kibana?

jharting: Yes I think this fixed it

PR #610, opened 2020-03-18
[WIP] New schema migration (RHCLOUD-4531)

ryandillinfelton:
[Link to Jira](https://projects.engineering.redhat.com/browse/RHCLOUD-4531)

Migrating all of the GraphQL queries to work with the [new schema in xjoin-search](https://github.com/RedHatInsights/xjoin-search/pull/48)

### DIFF
--- a/api/host_query_xjoin.py
+++ b/api/host_query_xjoin.py
@@ -52,13 +52,14 @@ ORDER_HOW_MAPPING = {"modified_on": "DESC", "display_name": "ASC"}
 
 
 def build_tag_query_dict_array(tags):
-    query_tag_array = []
+    query_tag_array = ()
     for string_tag in tags:
         query_tag_dict = {}
         tag_dict = Tag().from_string(string_tag).data()
         for key in tag_dict.keys():
             query_tag_dict[key] = {"eq": tag_dict[key]}
         query_tag_array += ({"tag": query_tag_dict},)
+    # TODO: REMOVE LOGGING
     logger.info("query_tag_array: %s", query_tag_array)
     return query_tag_array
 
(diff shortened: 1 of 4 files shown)


ryandillinfelton:
I had to change the tests to match the new schema as expected. But I also had trouble with some tests thinking the tag dictionaries in the filter would be in a list, and some seeming to expect it to be in a tuple. If someone could look into that and let me know which it really should be I'd appreciate it

api/host_query_xjoin.py
```diff
@@ -112,13 +124,15 @@ def _query_filters(fqdn, display_name, hostname_or_id, insights_id, tags, stalen
             hostname_or_id_filters += ({"id": str(id)},)
```
jharting: this `id` filter needs to be converted also

api/host_query_xjoin.py
```diff
@@ -51,6 +51,18 @@
...
 
+def build_tag_query_dict_array(tags):
+    query_tag_array = []
+    for string_tag in tags:
+        query_tag_dict = {}
+        tag_dict = Tag().from_string(string_tag).data()
+        for key in tag_dict.keys():
+            query_tag_dict[key] = {"eq": tag_dict[key]}
+        query_tag_array += ({"tag": query_tag_dict},)
+    logger.info("query_tag_array: %s", query_tag_array)
```
jharting: this should not be info level

api/host_query_xjoin.py
```diff
@@ -112,13 +124,15 @@ def _query_filters(fqdn, display_name, hostname_or_id, insights_id, tags, stalen
...
         query_filters = ({"OR": hostname_or_id_filters},)
     elif insights_id:
-        query_filters = ({"insights_id": insights_id},)
+        query_filters = ({"insights_id": {"eq": insights_id}},)
     else:
         query_filters = ()
 
     if tags:
-        query_filters += tuple({"tag": Tag().from_string(string_tag).data()} for string_tag in tags)
+        query_filters += (build_tag_query_dict_array(tags),)
```
jharting: This does not work. Seems that the problem is that an array is returned (instead of tuple being extended for each tag condition)

PR #613, opened 2020-03-19
Fix tags

Glutexo:
Fix tag handling. Intended features:

- [x] Little refactoring
- [x] More tests
- [x] Better regex, allow more chars
- [x] Allow empty namespace
- [x] Allow whitespace

Will probably eventurally replace #591. If so, sorry @yashnalla. 😞

There are more issues I didn’t address:

- The _Tag_ class:
  - Contains methods that are not used at all.
  - Can be just a named tuple. Like that it wouldn’t expose its _data_ as a mutable object etc.
  - Shouldn’t raise _ValidationException_.

And maybe more.

### DIFF
--- a/api/host_query_db.py
+++ b/api/host_query_db.py
@@ -127,7 +127,7 @@ def _find_hosts_by_tag(string_tags, query):
     tags = []
 
     for string_tag in string_tags:
-        tags.append(Tag().from_string(string_tag))
+        tags.append(Tag.from_string(string_tag))
 
     tags_to_find = Tag.create_nested_from_tags(tags)
 
--- a/app/utils.py
+++ b/app/utils.py
@@ -209,7 +209,7 @@ class Tag:
 
     @staticmethod
     def from_string(string_tag):
-        match = re.match(r"((?P<namespace>[^=/\s]+)/)?(?P<key>[^=/\s]+)(=(?P<value>[^=/\s]+))?$", string_tag)
+        match = re.match(r"((?P<namespace>[^=/]+)/)?(?P<key>[^=/]+)(=(?P<value>[^=/]+))?$", string_tag)
         encoded_tag_data = match.groupdict()
         decoded_tag_data = {}
         for k, v in encoded_tag_data.items():
--- a/app/utils.py
+++ b/app/utils.py
@@ -209,7 +209,7 @@ class Tag:
 
     @staticmethod
     def from_string(string_tag):
-        match = re.match(r"((?P<namespace>[\w.\-%~]+)\/)?(?P<key>[\w.\-%~]+)(=(?P<value>[\w.\-%~]+))?$", string_tag)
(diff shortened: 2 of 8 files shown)


Glutexo:
_RHCLOUD-2463_ is fixed. All special characters apart from delimiters _/_, _=_ and _,_ are now accepted unescaped in the _tags_ parameter. That applies both to the Host query and to the Tags query.

What remains is to fix the two adjacent bugs:

* Whitespace characters should be accepted too. Applies to Host save too, not only to search.
* Namespace should not be mandatory when searching by tags.

api/host_query_db.py
```diff
@@ -110,7 +110,7 @@ def _find_hosts_by_tag(string_tags, query):
     tags = []
 
     for string_tag in string_tags:
-        tags.append(Tag().from_string(string_tag))
+        tags.append(Tag.from_string(string_tag))
```
shdunning: Why not just have `__init__` take a string and do what `from_string` is doing? Curious on the value of a `@staticmethod` VS just using the constructor?

shdunning: Oh nevermind, I think I see. There are multiple ways a Tag() obj can be constructed... 🤔

shdunning: I guess the alternative would be to check the type in the constructor 🤷‍♂and call the `from_string` or `from_nested` if the ctor arg is `str` or `dict` respectively. i dunno...

Glutexo: I was thinking about that too. There is however a complementary [_from_nested_](https://github.com/Glutexo/insights-host-inventory/blob/9b576f49b8f488a2bd6e59a2177b808b46aac1f0/app/utils.py#L225) method, so I like the explicitness of the name.

I we switch this class to a _namedtuple_ later, which I’d love to, the we’d already have a constructor. With that a method that clearly accepts a single string representation of a tag makes sense. Strings like `"a/b=c"` are still valid values for any tag part.

Glutexo:
🎉  Lot of fixed stuff, even more tests. I agree with @yashnalla that this pull request became a bit expensive. I hope the fact that the fixes themselves are quite tiny and closely related would make it still reviewable.

Glutexo:
Oh oh oh, conflicts and failing tests. Will resolve.

app/utils.py
```diff
@@ -201,96 +201,64 @@ def value(self):
...
-        return self
-
-    def from_nested(self, nested_tag):
-        namespace, key, value = None, None, None
+    def __eq__(self, other):
+        return self.data() == other.data()
+
+    @staticmethod
+    def from_string(string_tag):
+        match = re.match(r"((?P<namespace>[^=/]+)/)?(?P<key>[^=/]+)(=(?P<value>[^=/]+))?$", string_tag)
```
jharting: I didn't know named capturing groups before but using them here makes this code very elegant!

Glutexo:
Resolved. Also fixed the tests.

Glutexo:
I see this merged. That’s awesome! There were quite some changes in tests I knew it will be tough to go through and verify. I am glad it went through well. Thanks!

PR #617, opened 2020-03-24
Add exit code for Host Reaper exception

Glutexo:
On failure, the Host Reaper now [exits](https://github.com/Glutexo/insights-host-inventory/blob/8652f0f568fa9ece9e179e557938355891c8f801/host_reaper.py#L81) with a status code 1. Just as if the exception wasn’t caught at all.

No tests at the Reaper script scaffold is not tested.

### DIFF
--- a/host_reaper.py
+++ b/host_reaper.py
@@ -1,3 +1,5 @@
+import sys
+from functools import partial
 from os import getenv
 
 from prometheus_client import CollectorRegistry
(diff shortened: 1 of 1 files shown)


Glutexo:
I was thinking about not catching the exception at all. But like this, we get the advantage of our [_get_logger_](https://github.com/Glutexo/insights-host-inventory/blob/8652f0f568fa9ece9e179e557938355891c8f801/app/logging.py#L117) function. If we sometime in the future decide to not rely on this logging setup, it will be possible to just ditch the whole [_except_](https://github.com/Glutexo/insights-host-inventory/blob/8652f0f568fa9ece9e179e557938355891c8f801/host_reaper.py#L78) block and keep only the [_finally_](https://github.com/Glutexo/insights-host-inventory/blob/8652f0f568fa9ece9e179e557938355891c8f801/host_reaper.py#L82) one.

jharting:
hey @Glutexo can you add a JIRA number into the commit message?

Glutexo:
Good point, @jharting. Used Python native exception handling mechanism to use our logger and exit naturally.

Glutexo:
Jenkins is stuck. 😞 Is there anything we can do about that? @rodrigonull

PR #618, opened 2020-03-24
RHCLOUD-4824 - Culled Host No Dedup

tahmidefaz:
Jira: https://projects.engineering.redhat.com/browse/RHCLOUD-4824

### DIFF
--- a/lib/host_repository.py
+++ b/lib/host_repository.py
@@ -15,12 +15,12 @@ from lib.db import session_guard
 
 __all__ = (
     "add_host",
-    "ALL_STALENESS_STATES",
     "canonical_facts_host_query",
     "create_new_host",
     "find_existing_host",
     "find_host_by_canonical_facts",
     "find_hosts_by_staleness",
+    "find_non_culled_hosts",
     "stale_timestamp_filter",
     "update_existing_host",
 )
@@ -83,7 +83,7 @@ def canonical_fact_host_query(account_number, canonical_fact, value):
     query = Host.query.filter(
         (Host.account == account_number) & (Host.canonical_facts[canonical_fact].astext == value)
     )
-    return find_hosts_by_staleness(ALL_STALENESS_STATES, query)
+    return find_non_culled_hosts(query)
 
 
 def canonical_facts_host_query(account_number, canonical_facts):
@@ -94,7 +94,7 @@ def canonical_facts_host_query(account_number, canonical_facts):
             | Host.canonical_facts.comparator.contained_by(canonical_facts)
         )
     )
-    return find_hosts_by_staleness(ALL_STALENESS_STATES, query)
+    return find_non_culled_hosts(query)
 
 
 def find_host_by_canonical_fact(account_number, canonical_fact, value):
@@ -133,6 +133,10 @@ def find_hosts_by_staleness(staleness, query):
         staleness_conditions += (Host.stale_timestamp == NULL,)
 
     return query.filter(or_(*staleness_conditions))
+
+
+def find_non_culled_hosts(query):
+    return find_hosts_by_staleness(ALL_STALENESS_STATES, query)
 
 
 @metrics.new_host_commit_processing_time.time()
--- a/lib/host_repository.py
+++ b/lib/host_repository.py
@@ -85,15 +85,14 @@ def exclude_culled_hosts(query):
     return query.filter(or_(*staleness_conditions))
 
 
-def canonical_fact_host_query(account_number, canonical_fact, value, include_culled=True):
-    query = Host.query.filter((Host.account == account_number) & (Host.canonical_facts[canonical_fact].astext == value))
-    if include_culled == False:
-        logger.debug("Checking if include culled works %s!", str(include_culled))
-        return exclude_culled_hosts(query)
-    return query
+def canonical_fact_host_query(account_number, canonical_fact, value):
+    query = Host.query.filter(
+        (Host.account == account_number) & (Host.canonical_facts[canonical_fact].astext == value)
+    )
+    return exclude_culled_hosts(query)
 
 
-def canonical_facts_host_query(account_number, canonical_facts, include_culled = True):
+def canonical_facts_host_query(account_number, canonical_facts):
     query = Host.query.filter(
         (Host.account == account_number)
         & (
@@ -101,10 +100,7 @@ def canonical_facts_host_query(account_number, canonical_facts, include_culled =
             | Host.canonical_facts.comparator.contained_by(canonical_facts)
         )
     )
-    if include_culled == False:
-        logger.debug("Checking if include culled works %s!", str(include_culled))
-        return exclude_culled_hosts(query)
-    return query
+    return exclude_culled_hosts(query)
 
 
 def find_host_by_canonical_fact(account_number, canonical_fact, value):
@@ -113,7 +109,7 @@ def find_host_by_canonical_fact(account_number, canonical_fact, value):
     """
     logger.debug("find_host_by_canonical_fact(%s, %s)", canonical_fact, value)
 
-    host = canonical_fact_host_query(account_number, canonical_fact, value, include_culled=False).first()
+    host = canonical_fact_host_query(account_number, canonical_fact, value).first()
 
     if host:
         logger.debug("Found existing host using canonical_fact match: %s", host)
@@ -127,7 +123,7 @@ def find_host_by_canonical_facts(account_number, canonical_facts):
     """
     logger.debug("find_host_by_canonical_facts(%s)", canonical_facts)
 
-    host = canonical_facts_host_query(account_number, canonical_facts, include_culled=False).first()
+    host = canonical_facts_host_query(account_number, canonical_facts).first()
 
     if host:
         logger.debug("Found existing host using canonical_fact match: %s", host)
--- a/lib/host_repository.py
+++ b/lib/host_repository.py
@@ -15,10 +15,12 @@ from lib.db import session_guard
 
 __all__ = (
     "add_host",
+    "ALL_STALENESS_STATES",
     "canonical_facts_host_query",
     "create_new_host",
     "find_existing_host",
     "find_host_by_canonical_facts",
+    "find_hosts_by_staleness",
     "stale_timestamp_filter",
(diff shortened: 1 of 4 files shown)


tahmidefaz:
I had to rewrite a version of [api.host_query_db.find_host_by_staleness](https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/host_query_db.py#L70) to avoid cyclic import exceptions.

lib/host_repository.py
```diff
@@ -71,18 +77,30 @@ def _find_host_by_elevated_ids(account_number, canonical_facts):
     return None
 
 
+def exclude_culled_hosts(query):
```
Glutexo: This function is almost an exact copy of [_app.host_query_db.find_hosts_by_staleness_](https://github.com/RedHatInsights/insights-host-inventory/blob/4ea57696717eb8a9e2fc3661f576cab917d5c888/api/host_query_db.py#L53). Can we move that _find_hosts_by_staleness_ here along with [_api.host.ALL_STALENESS_STATES_](https://github.com/RedHatInsights/insights-host-inventory/blob/4ea57696717eb8a9e2fc3661f576cab917d5c888/api/host.py#L44)?

lib/host_repository.py
```diff
@@ -29,6 +32,9 @@
 # the priority.
 ELEVATED_CANONICAL_FACT_FIELDS = ("insights_id", "subscription_manager_id")
 
+STALENESS_STATES = ("fresh", "stale", "stale_warning", "unknown")
```
Glutexo: We already have this in [_api.host_](https://github.com/RedHatInsights/insights-host-inventory/blob/4ea57696717eb8a9e2fc3661f576cab917d5c888/api/host.py#L44). Let’s keep this one and remove the other.

lib/host_repository.py
```diff
@@ -71,18 +77,30 @@ def _find_host_by_elevated_ids(account_number, canonical_facts):
     return None
 
 
+def exclude_culled_hosts(query):
+    config = inventory_config()
+    staleness_conditions = tuple(staleness_to_conditions(config, STALENESS_STATES, stale_timestamp_filter))
```
Glutexo: You don’t need the condition here. [_STALENESS_STATES_](https://github.com/RedHatInsights/insights-host-inventory/blob/4ea57696717eb8a9e2fc3661f576cab917d5c888/lib/host_repository.py#L35) is a constant and you know exactly what’s in there. It’s always _True_.

Glutexo:
Oh, I see, @tahmidefaz already stated my concerns in his comment. I’ll try to resolve.

Glutexo:
@tahmidefaz I tried to DRY the two functions and constants and I couldn’t reproduce the cyclic import issue. Please take a look at [my fork](https://github.com/Glutexo/insights-host-inventory/tree/dedup-no-culled-host). There is a [commit](https://github.com/Glutexo/insights-host-inventory/commit/0d4c6436db9b6571da230636bea04bb97538b246) that tries to achieve the merge.

test_api.py
```diff
@@ -809,6 +809,28 @@ def test_create_host_without_ansible_host_then_update(self):
...
+        response = self.post(HOST_URL, [host_data.data()], 207)
+
+        self._verify_host_status(response, 0, 201)
+
+        created_host = self._pluck_host_from_response(response, 0)
+
+        original_id = created_host["id"]
+
+        # Update the host
+        host_data.subscription_manager_id = generate_uuid()
```
Glutexo: It’s a bit strange to update the host with an elevated ID, but whatever, it works.

test_api.py
```diff
@@ -809,6 +809,28 @@ def test_create_host_without_ansible_host_then_update(self):
...
+
+        new_response = self.post(HOST_URL, [host_data.data()], 207)
```
Glutexo: Let’s verify the new status too.

```suggestion
        new_response = self.post(HOST_URL, [host_data.data()], 207)
        
        self._verify_host_status(new_response, 0, 201)
```

tahmidefaz:
@Glutexo Your suggestions make sense. I will take a look. Thanks!

tahmidefaz:
> @tahmidefaz I tried to DRY the two functions and constants and I couldn’t reproduce the cyclic import issue. Please take a look at [my fork](https://github.com/Glutexo/insights-host-inventory/tree/dedup-no-culled-host). There is a [commit](https://github.com/Glutexo/insights-host-inventory/commit/0d4c6436db9b6571da230636bea04bb97538b246) that tries to achieve the merge.

I ran into that cyclic import error when I tried to import _find_host_by_staleness_ function. I guess, moving the function entirely makes more sense than rewriting a version of it  :+1:

Glutexo:
> find_hosts_by_staleness is only from one single place called with something else than with the ALL_STALENESS_STATES constants. It may be worth to either make this a default or even better (more verbose and explicit) create another function, something like find_non_culled_hosts(query).

There is only [one place](https://github.com/RedHatInsights/insights-host-inventory/blob/e56d9396172c8a075dc2de6d97ea21eba0fdb225/api/host_query_db.py#L39) from which the [_find_hosts_by_staleness_](https://github.com/RedHatInsights/insights-host-inventory/blob/e56d9396172c8a075dc2de6d97ea21eba0fdb225/lib/host_repository.py#L128) function is called with a different value than that of the [_ALL_STALENESS_STATES_](https://github.com/RedHatInsights/insights-host-inventory/blob/e56d9396172c8a075dc2de6d97ea21eba0fdb225/lib/host_repository.py#L37) constant. I was thinking about introducing a new function that would DRY this. It would make those calls nicely expressive.

```python
def find_non_culled_hosts(query):
    return find_hosts_by_staleness(ALL_STALENESS_STATES, query)
```

Don’t forget to add its name into the [_\_\_all\_\__](https://github.com/RedHatInsights/insights-host-inventory/blob/e56d9396172c8a075dc2de6d97ea21eba0fdb225/lib/host_repository.py#L16) list on the top of the file.

Glutexo:
Also there is a conflict, please resolve and rebase. Thanks!

tahmidefaz:
@Glutexo done!

This is ready for a final review, now!

test_api.py
```diff
@@ -815,6 +815,30 @@ def test_create_host_without_ansible_host_then_update(self):
...
+        response = self.post(HOST_URL, [host_data.data()], 207)
+
+        self._verify_host_status(response, 0, 201)
+
+        created_host = self._pluck_host_from_response(response, 0)
+
+        original_id = created_host["id"]
+
+        # Update the host
+        host_data.subscription_manager_id = generate_uuid()
```
Glutexo, line 832:
```suggestion
```

test_api.py
```diff
@@ -815,6 +815,30 @@ def test_create_host_without_ansible_host_then_update(self):
 
                 self._validate_host(host_lookup_results["results"][0], host_data, expected_id=original_id)
 
+    def test_ignore_culled_host_on_update(self):
```
Glutexo, line 818:
```suggestion
    def test_ignore_culled_host_on_update_by_canonical_facts(self):
```

test_api.py
```diff
@@ -815,6 +815,30 @@ def test_create_host_without_ansible_host_then_update(self):
...
+        host_data = HostWrapper(test_data(facts=None))
```
Glutexo, line 819: You don’t need to use _HostWrapper_ at all. You just use the data dictionary. Also, you can pass in the _stale_timestamp_ as another argument along the _facts_ instead of doing that in a separate instruction.

```python
host_data = test_data(facts=None, stale_timestamp=(now() - timedelta(weeks=3)).isoformat())
```
And then use just `host_data` instead of `host_data.data()`.

test_api.py
```diff
@@ -815,6 +815,30 @@ def test_create_host_without_ansible_host_then_update(self):
...
+        host_data.subscription_manager_id = generate_uuid()
+
+        new_response = self.post(HOST_URL, [host_data.data()], 207)
+
+        self._verify_host_status(new_response, 0, 201)
+
+        updated_host = self._pluck_host_from_response(new_response, 0)
+
+        self.assertNotEqual(original_id, updated_host["id"])
+
```
Glutexo, line 841:
```suggestion

    def test_ignore_culled_host_on_update_by_elevated_id(self):
        host_to_create_data = test_data(
            insights_id=generate_uuid(), facts=None, stale_timestamp=(now() - timedelta(weeks=3)).isoformat()
        )  # culled host

        # Create the host
        response = self.post(HOST_URL, [host_to_create_data], 207)
        self._verify_host_status(response, 0, 201)

        created_host = self._pluck_host_from_response(response, 0)

        # Update the host
        host_to_update_data = {**host_to_create_data, "ip_addresses": ["10.10.0.2"]}
        new_response = self.post(HOST_URL, [host_to_update_data], 207)
        self._verify_host_status(new_response, 0, 201)

        updated_host = self._pluck_host_from_response(new_response, 0)
        self.assertNotEqual(created_host["id"], updated_host["id"])

```

Glutexo:
To merge this:

1. I’d like to insist on my additional test suggestion.
2. We need one more review.

tahmidefaz:
@Glutexo I think I was able to address all your suggestions. Let me know if you see anything that needs changing.

This is ready for review again.

lib/host_repository.py
```diff
@@ -29,6 +34,9 @@
 # the priority.
 ELEVATED_CANONICAL_FACT_FIELDS = ("insights_id", "subscription_manager_id")
 
+ALL_STALENESS_STATES = ("fresh", "stale", "stale_warning", "unknown")
```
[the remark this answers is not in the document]
shdunning: @Glutexo `ALL_STALENESS_STATES` is being used in a DB query tho - i dunno much about sqlalchemy, but will it even accept Enums and/or what is the benefit of using Enum over a tuple in this particular case?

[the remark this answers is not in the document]
Glutexo: I didn’t mean to replace _ALL_STALENESS_STATES_ by an Enum, but using Enum for the values. It is more semantically correct and would prevent possible bugs, because at any point in the code it will be known (and enforced) what the valid values are.

Glutexo:
Great work, @tahmidefaz! Thanks!

PR #620, opened 2020-03-26
Make mandatory fields NOT NULL

Glutexo:
[Altered](https://github.com/Glutexo/insights-host-inventory/blob/4acb9ad0bd8aa2c4fe096950a2f3ece98ab9c084/migrations/versions/c7c730559750_alter_mandatory_fields_to_not_null.py) _account_, _canonical_facts_, _created_on_ and _modified_on_ to _NOT NULL_. These fields are mandatory. Hosts with `NULL` or `'null'` in _canonical_facts_ [become](https://github.com/Glutexo/insights-host-inventory/blob/4acb9ad0bd8aa2c4fe096950a2f3ece98ab9c084/migrations/versions/33e0aca8516f_fill_in_canonical_facts.py) `{}`. This is still invalid, because a host is obliged to have at least one canonical fact. Moreover it’s still possible to insert a JSON `'null'` there into the field.

In case of `'null'`, the host is not retrievable – _500 Internal Server Error_ occurs, but is reapable – gets deleted just fine, event is emitted. Empty canonical facts dictionary `{}` results in a host that is retrievable just fine, but can’t be matched by the deduplication process.


migrations/versions/33e0aca8516f_fill_in_canonical_facts.py
```diff
@@ -0,0 +1,43 @@
...
+class Host(Base):
+    __tablename__ = "hosts"
+
+    id = Column(UUID, primary_key=True)
+    canonical_facts = Column(JSONB(none_as_null=True))
+
+
+def upgrade():
+    with session() as s:
+        s.query(Host).filter((Host.canonical_facts == NULL) | (Host.canonical_facts == JSONB.NULL)).update(
```
jharting: I don't think we currently have any hosts that would match this. Should we run this update anyway? If yes, should we also deal with hosts potentially having account/created_on/modified_on columns null?

migrations/versions/33e0aca8516f_fill_in_canonical_facts.py
```diff
@@ -0,0 +1,43 @@
...
+            {Host.canonical_facts: {}}
+        )
+
+
+def downgrade():
+    with session() as s:
+        s.query(Host).filter(Host.canonical_facts == {}).update({Host.canonical_facts: NULL})
```
jharting: This is not really an inverse action of `upgrade`. Even if upgrade modifies 0 columns (e.g. all records have canonical_facts={} already) this functions updates all of those records to NULL. Since we cannot do an inverse action I wonder if it's worth trying to do anything here

migrations/versions/c7c730559750_alter_mandatory_fields_to_not_null.py
```diff
@@ -0,0 +1,27 @@
...
+down_revision = "33e0aca8516f"
+branch_labels = None
+depends_on = None
+
+COLUMNS = ("created_on", "modified_on", "account", "canonical_facts")
+
+
+def upgrade():
+    for column in COLUMNS:
+        op.alter_column("hosts", column, nullable=False)
```
jharting: I wonder what the result is going to be should one of these fail. Are we going to be able to upgrade/downgrade afterwards? How about doing both UPDATES and DDL changes in a single migration wrapped within a transaction?

migrations/versions/33e0aca8516f_fill_in_canonical_facts.py
Glutexo: If there are no such hosts in any environments, I can remove this migration. It would be still more correct though to have one.

I can change this to _DELETE_. That wouldn’t emit an event, but such hosts are borked anyway and we don’t want them to contaminate our database. So maybe we can pretend they were never there?

The same can be done for hosts without the timestamps or account. For missing account, this is the only option, we can’t assume one. For timestamps, we can fill in current timestamp, but that would be false information – _NULL_ would be more correct.

Canonical facts is the only field we can safely fill in. Although hosts having no canonical facts should not exist, it’s at least more or less valid value.

migrations/versions/33e0aca8516f_fill_in_canonical_facts.py
Glutexo: I am aware of this. Since there should be no hosts that have `{}` anyway. If setting those imaginary hosts to NULL is not a good solution, I can remove this and call this migration either irreversible, or make this a no-op.

migrations/versions/c7c730559750_alter_mandatory_fields_to_not_null.py
Glutexo: Migrations are run within a transaction, so if this fails, all columns remain nullable. Everything is still reversible. It’s definitely possible to merge those two migrations, or possibly make more column-oriented ones instead of these two operation-oriented. That would result in logic duplication though.

Glutexo:
Replied to the comments. I think we should have a data migration and I have nothing against merging it with the DDL one. Since the data migration is more a “fix” than a “change”, I can make the downgrade no-op.

But what to do with invalid records? Canonical facts can be filled in, but what about the other fields? I‘d be in favor of filling in the canonical facts with `{}` and then if any record is still invalid, delete it. I’d still like to have a DELETE in the DDL change transaction, but we can do a graceful delete with events and stuff using the Reaper. The hard DELETE would then be no-op, only there for consistency.

So the only question that remains is – do we need to care about the delete event for such hosts?

Glutexo:
Pushed some new commits. The two migrations are merged into one, so they both happen in the same transaction. Canonical facts are fixed. Remaining hosts without timestamps or account are deleted without event emission. If we should reap them properly, please let me know and I’ll make a patch.

Glutexo:
Also rebased on current _master_.

Glutexo:
Added the matched rows number logging. Used our standard logging. It logs at least to standard output under Inventory namespace. It may eventually even log into CloudWatch if environment variables are passed, but I didn’t check that.

Glutexo:
Also rebased on current _master_.

PR #622, opened 2020-03-27
Create logger only once

Glutexo:
A little follow-up to #617. This slipped through my finger.

The logger is now [created right after](https://github.com/Glutexo/insights-host-inventory/blob/f6c5bd1853415f204aabf11c34bc220526b227f0/host_reaper.py#L93) logging configuration. Use this logger instance and pass it down to the functions instead of creating a new one.


jharting:
Looks like the tests need to be fixed on this one

Glutexo:
🤦🏻‍♂️ Fixed. Also fixed variable shadowing in the [_run_](https://github.com/Glutexo/insights-host-inventory/blob/584c3a75891c48830f535d812aea86a2f062abe2/host_reaper.py#L54) function.

PR #623, opened 2020-03-30
RHCLOUD-3524: Enforce stale timestamp timezone

tahmidefaz:
Jira: https://projects.engineering.redhat.com/browse/RHCLOUD-3524

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -337,6 +337,11 @@ class HostSchema(Schema):
         if len(mac_address_list) < 1:
             raise ValidationError("Array must contain at least one item")
 
+    @validates("stale_timestamp")
+    def validate_stale_timestamp(self, timestamp):
+        if timestamp.tzinfo is None:
+            raise ValidationError("Timestamp must contain timezone info")
+
 
 class PatchHostSchema(Schema):
     ansible_host = fields.Str(validate=validate.Length(min=0, max=255))
--- a/app/models.py
+++ b/app/models.py
@@ -318,8 +318,8 @@ class HostSchema(Schema):
     satellite_id = fields.Str(validate=verify_uuid_format)
     fqdn = fields.Str(validate=validate.Length(min=1, max=255))
     bios_uuid = fields.Str(validate=verify_uuid_format)
-    ip_addresses = fields.List(fields.Str(validate=validate.Length(min=1, max=255)))
-    mac_addresses = fields.List(fields.Str(validate=validate.Length(min=1, max=59)))
+    ip_addresses = fields.List(fields.Str(validate=validate.Length(min=1, max=255)), validate=validate.Length(min=1))
+    mac_addresses = fields.List(fields.Str(validate=validate.Length(min=1, max=59)), validate=validate.Length(min=1))
     external_id = fields.Str(validate=validate.Length(min=1, max=500))
     facts = fields.List(fields.Nested(FactsSchema))
     tags = fields.List(fields.Nested(TagsSchema))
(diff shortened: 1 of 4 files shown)


app/models.py
```diff
@@ -337,6 +337,11 @@ def validate_mac_addresses(self, mac_address_list):
         if len(mac_address_list) < 1:
             raise ValidationError("Array must contain at least one item")
 
+    @validates("stale_timestamp")
+    def validate_stale_timestamp(self, timestamp):
+        if timestamp.tzinfo is None:
+            raise ValidationError("Timestamp must contain timezone info")
```
shdunning, line 343: Is this function only used to validate `stale_timestamp`? If so, I would change the error message to indicate which timestamp requires TZ info (i.e., `"stale_timestamp must contain timezone info"`.

shdunning: or, change the name of the validator to be less specific: `validate_timestmap(self, timestamp)`

tahmidefaz: @shdunning Yes, it only uses `stale_timestamp`. Well, but they see this in the error message, though `{'stale_timestamp': ['Timestamp must contain timezone info']}`

test_db_model.py
```diff
@@ -150,6 +150,20 @@ def test_host_schema_invalid_tags(tags):
...
+        "account": "00102",
+        "stale_timestamp": "2020-03-31T10:10:06.754201",
+        "reporter": "test",
+    }
+    with pytest.raises(ValidationError) as excinfo:
+        _ = HostSchema(strict=True).load(host)
+
+    assert "Timestamp must contain timezone info" in str(excinfo.value)
+
+
```
Glutexo, line 166: Let’s add an [API test](https://github.com/RedHatInsights/insights-host-inventory/blob/ea2cf38f7baec082f1564a7dda7ef72d53248cc4/test_api.py) too. There are already [tests](https://github.com/RedHatInsights/insights-host-inventory/blob/ea2cf38f7baec082f1564a7dda7ef72d53248cc4/test_api.py#L1022) for (in)valid stale timestamp values. 
```python
def test_create_host_with_stale_timestamp_without_time_zone(self):
    self._add_host(400, stale_timestamp=datetime.now().isoformat())
```

Also a [MQ test](https://github.com/RedHatInsights/insights-host-inventory/blob/ea2cf38f7baec082f1564a7dda7ef72d53248cc4/test_mq_service.py) would come in handy. Again, there are already [some](https://github.com/RedHatInsights/insights-host-inventory/blob/ea2cf38f7baec082f1564a7dda7ef72d53248cc4/test_mq_service.py#L269) testing various valid and invalid input.

```python
    @patch("app.queue.egress.datetime", **{"utcnow.return_value": datetime.utcnow()})
    def test_add_host_with_invalid_stale_timestmap(self, datetime_mock):
        mock_event_producer = MockEventProducer()

        for stale_timestamp in ("invalid", datetime.now().isoformat()):
            with self.subTest(stale_timestamp=stale_timestamp):
                host_data = {
                    "display_name": "test_host",
                    "insights_id": str(uuid.uuid4()),
                    "account": "0000001",
                    "stale_timestamp": stale_timestamp,
                    "reporter": "test",
                }

                with self.app.app_context():
                    with self.assertRaises(ValidationException):
```

app/models.py
tahmidefaz: @Glutexo simplified the validation for `ip_addresses` and `mac_addresses`, however, decided not to go after simplifying `stale_timestamp` with the `lambda` function as it would loose the helpful missing time zone information.

tahmidefaz:
Done! This is ready for review again.

app/models.py
Glutexo: Awsum.

Glutexo:
Merde!

PR #625, opened 2020-04-01
Replacing intervals rhcloud3997

yashnalla:


Glutexo:
Merging with two approvals, thanks! I also appreciate that the closed/open intervals were picked correctly. The staleness starts exactly at the staleness point in time (closed) and not immediately after it (open). 👍🏻

PR #626, opened 2020-04-01
More Logging for Kibana

tahmidefaz:


tahmidefaz:
Anything else I need to add here?

app/queue/ingress.py
```diff
@@ -97,7 +97,17 @@ def add_host(host_data):
...
-            logger.info("Attempting to add host", extra={"canonical_facts": input_host.canonical_facts})
+            logger.info(
+                "Attempting to add host",
+                extra={
+                    "input_host": {
+                        "display_name": input_host.display_name,
+                        "canonical_facts": input_host.canonical_facts,
+                        "reporter": input_host.reporter,
+                        "stale_timestamp": input_host.stale_timestamp.isoformat(),
+                    }
```
shdunning, line 108: I _think_ this is supposed to be: `extra={"host": {i: input_host[i] for i in input_host if i not in ("facts", "system_profile")}},` right @jharting ?

tahmidefaz: I tried that first, but `input_host` is an object and does not support iteration by default. `output_host` is a dictionary, hence is iterable :arrow_up:

PR #628, opened 2020-04-06
update validation on satellite_id to support Sat 5

shdunning:
Fix to allow validation of satellite_ids that come from Sat 5 hosts.

See https://bugzilla.redhat.com/show_bug.cgi?id=1821321

### DIFF
--- a/app/validators.py
+++ b/app/validators.py
@@ -52,6 +52,6 @@ def verify_satellite_id(id_str):
     # satellite_id can either be a UUID or a 10 digit number depending on Sat version
     if verify_uuid_format(id_str):
         return True
-    elif id_str and re.match(r"\d{10}", id_str):
+    elif id_str and re.match(r"^\d{10}$", id_str):
         return True
     return False
--- a/app/validators.py
+++ b/app/validators.py
@@ -48,15 +48,10 @@ def check_empty_keys(data):
     return True
 
 
-"""
-satellite_id can either be a 10 digit number or a UUID depending on Satellite version
-"""
-
-
 def verify_satellite_id(id_str):
-    if not id_str:
-        return False
-
-    if re.match(r"\d{10}", id_str):
+    # satellite_id can either be a UUID or a 10 digit number depending on Sat version
+    if verify_uuid_format(id_str):
         return True
-    return verify_uuid_format(id_str)
+    elif id_str and re.match(r"\d{10}", id_str):
+        return True
+    return False
--- a/app/validators.py
+++ b/app/validators.py
@@ -1,3 +1,4 @@
+import re
 import uuid
 
 import validators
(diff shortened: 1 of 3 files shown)


app/validators.py
```diff
@@ -45,3 +46,12 @@ def check_empty_keys(data):
...
     return True
+
+
+def verify_satellite_id(id_str):
+    # satellite_id can either be a UUID or a 10 digit number depending on Sat version
+    if verify_uuid_format(id_str):
+        return True
+    elif id_str and re.match(r"^\d{10}$", id_str):
+        return True
+    return False
```
Glutexo, line 57:
```suggestion
def verify_satellite_id(satellite_id):
    return verify_uuid_format(satellite_id) or _verify_legacy_satellite_id(satellite_id)
```

def _verify_legacy_satellite_id(satellite_id):
    return satellite_id and re.fullmatch(r"\d{10}", satellite_id) is not None
```

- _fullmatch_ saves us from _^_ and _$_
- A separate method allows us to use simple boolean expression instead of if/else constructs yet retain readability and semantics.

File: test_validators.py [hunk 2]
```diff
@@ -18,6 +19,25 @@ def test_invalid_uuid(uuid):
...
 
+@pytest.mark.parametrize("satellite_id", ["4a8fb994-57fe-4dbb-ad2a-9e922560b6c1", "1000056432"])
+def test_valid_satellite_id(satellite_id):
+    assert verify_satellite_id(satellite_id) is True
+
+
+@pytest.mark.parametrize(
+    "satellite_id",
+    [
+        "123456789",
```
Glutexo, line 30:
```suggestion
        "123456789",
        "12345678901",
```

The first attempt to this pull request did a false positive match for too long IDs. A regression test would come in handy.

test_validators.py
```diff
@@ -18,6 +19,25 @@ def test_invalid_uuid(uuid):
     assert not verify_uuid_format(uuid)
 
 
+@pytest.mark.parametrize("satellite_id", ["4a8fb994-57fe-4dbb-ad2a-9e922560b6c1", "1000056432"])
+def test_valid_satellite_id(satellite_id):
+    assert verify_satellite_id(satellite_id) is True
```
Glutexo, line 24:
```suggestion
    assert verify_satellite_id(satellite_id)
```

Do we need to check that it’s exactly _True_? The validator is just fine with returning simply a value that acts **like** True.

test_validators.py
```diff
@@ -18,6 +19,25 @@ def test_invalid_uuid(uuid):
     assert not verify_uuid_format(uuid)
 
 
+@pytest.mark.parametrize("satellite_id", ["4a8fb994-57fe-4dbb-ad2a-9e922560b6c1", "1000056432"])
```
Glutexo, line 22:
```suggestion
@pytest.mark.parametrize(
    "satellite_id", ["4a8fb994-57fe-4dbb-ad2a-9e922560b6c1", "4a8fb99457fe4dbbad2a9e922560b6c1", "1000056432"]
)
```

A UUID without hyphens is ok too. Adding one doesn’t hurt.

PR #631, opened 2020-04-08
Remove percent encoding of the search string in /tags

jharting:
With RHCLOUD-5665 we're removing the need to urlencode the search string when making the request.

This is related to:
https://projects.engineering.redhat.com/browse/RHCLOUD-5656
https://projects.engineering.redhat.com/browse/RHCLOUD-5668

### DIFF
--- a/api/tag.py
+++ b/api/tag.py
@@ -1,5 +1,4 @@
 import re
-from urllib.parse import quote_plus as url_quote
 
 import flask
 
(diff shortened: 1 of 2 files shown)


Glutexo:
Do we want another approval? Maybe @shdunning?

api/tag.py
```diff
@@ -75,7 +74,7 @@ def get_tags(search=None, tags=None, order_by=None, order_how=None, page=None, p
     if search:
         variables["filter"] = {
             # Escaped to prevent ReDoS
```
Glutexo, line 76: This comment is inaccurate. Even though regex escaping prevents ReDoS, it also changes the search mechanics. It makes this injected string literal match, not itself being treated as a regex. Which is the only appropriate way for non-regex search fields.

jharting: @Glutexo good catch!. I'll update the comment. Thanks for the review.

PR #633, opened 2020-04-13
RHCLOUD-5000: Add msg headers to indicate msg type

tahmidefaz:
Jira: https://projects.engineering.redhat.com/browse/RHCLOUD-5000

### DIFF
--- a/app/events.py
+++ b/app/events.py
@@ -36,5 +36,8 @@ def delete(host):
     )
 
 
+delete_event_name = "delete"
+
+
 def message_headers(event_type):
-    return [("event_type", str.encode(event_type))]
+    return {"event_type": event_type}
--- a/app/events.py
+++ b/app/events.py
@@ -34,3 +34,7 @@ def delete(host):
         )
         .data
     )
+
+
+def message_headers(event_type):
+    return [("event_type", str.encode(event_type))]
--- a/app/events.py
+++ b/app/events.py
@@ -9,6 +9,8 @@ from app.serialization import serialize_canonical_facts
 
 logger = logging.getLogger(__name__)
 
+DELETE_EVENT_NAME = "delete"
+
 
 class HostEvent(Schema):
     id = fields.UUID()
@@ -34,9 +36,6 @@ def delete(host):
         )
         .data
     )
-
-
-delete_event_name = "delete"
 
 
 def message_headers(event_type):
--- a/app/queue/egress.py
+++ b/app/queue/egress.py
@@ -19,13 +19,13 @@ class KafkaEventProducer:
         self._kafka_producer = KafkaProducer(bootstrap_servers=config.bootstrap_servers)
         self._topic = config.host_egress_topic
 
-    def write_event(self, event, key):
-        logger.debug("Topic: %s, key: %s, event: %s", self._topic, key, event)
+    def write_event(self, event, key, headers):
+        logger.debug("Topic: %s, key: %s, event: %s, headers: %s", self._topic, key, event, headers)
 
         try:
             k = key.encode("utf-8") if key else None
             v = event.encode("utf-8")
-            self._kafka_producer.send(self._topic, key=k, value=v)
+            self._kafka_producer.send(self._topic, key=k, value=v, headers=headers)
             metrics.egress_message_handler_success.inc()
         except Exception:
             logger.exception("Failed to send event")
--- a/app/queue/egress.py
+++ b/app/queue/egress.py
@@ -25,7 +25,8 @@ class KafkaEventProducer:
         try:
             k = key.encode("utf-8") if key else None
             v = event.encode("utf-8")
-            self._kafka_producer.send(self._topic, key=k, value=v, headers=headers)
+            h = [(hk, hv.encode("utf-8")) for hk, hv in headers.items()]
+            self._kafka_producer.send(self._topic, key=k, value=v, headers=h)
             metrics.egress_message_handler_success.inc()
         except Exception:
             logger.exception("Failed to send event")
--- a/app/queue/ingress.py
+++ b/app/queue/ingress.py
@@ -7,6 +7,7 @@ from marshmallow import ValidationError
 
 from app import inventory_config
 from app.culling import Timestamps
+from app.events import message_headers
 from app.exceptions import InventoryException
 from app.logging import get_logger
 from app.logging import threadctx
@@ -131,10 +132,6 @@ def add_host(host_data):
             logger.exception("Error while adding host", extra={"host": host_data})
             metrics.add_host_failure.labels("Exception", host_data.get("reporter", "null")).inc()
             raise
-
-
-def message_headers(event_type):
-    return [("event_type", str.encode(event_type))]
 
 
 @metrics.ingress_message_handler_time.time()
--- a/app/queue/ingress.py
+++ b/app/queue/ingress.py
@@ -133,6 +133,10 @@ def add_host(host_data):
             raise
 
 
+def message_headers(event_type):
+    return [("event_type", str.encode(event_type))]
+
+
 @metrics.ingress_message_handler_time.time()
 def handle_message(message, event_producer):
     validated_operation_msg = parse_operation_message(message)
@@ -144,7 +148,7 @@ def handle_message(message, event_producer):
     with PayloadTrackerContext(payload_tracker, received_status_message="message received"):
         (output_host, add_results) = add_host(validated_operation_msg["data"])
         event = build_event(add_results.name, output_host, metadata)
-        event_producer.write_event(event, output_host["id"])
+        event_producer.write_event(event, output_host["id"], message_headers(add_results.name))
 
 
 def event_loop(consumer, flask_app, event_producer, handler, shutdown_handler):
--- a/lib/host_delete.py
+++ b/lib/host_delete.py
@@ -1,6 +1,7 @@
 from sqlalchemy.orm.base import instance_state
 
 from app.events import delete as delete_event
+from app.events import message_headers
 from app.models import Host
 from lib.metrics import delete_host_count
 from lib.metrics import delete_host_processing_time
@@ -43,5 +44,5 @@ def _deleted_by_this_query(host):
 def _emit_event(host):
     event = delete_event(host)
     key = str(host.id)
-    headers = [("event_type", b"deleted")]
+    headers = message_headers("deleted")
     emit_event(event, key, headers)
--- a/lib/host_delete.py
+++ b/lib/host_delete.py
@@ -1,6 +1,7 @@
 from sqlalchemy.orm.base import instance_state
 
 from app.events import delete as delete_event
+from app.events import delete_event_name
 from app.events import message_headers
 from app.models import Host
 from lib.metrics import delete_host_count
@@ -44,5 +45,5 @@ def _deleted_by_this_query(host):
 def _emit_event(host):
     event = delete_event(host)
     key = str(host.id)
-    headers = message_headers("deleted")
+    headers = message_headers(delete_event_name)
     emit_event(event, key, headers)
--- a/lib/host_delete.py
+++ b/lib/host_delete.py
@@ -43,4 +43,5 @@ def _deleted_by_this_query(host):
 def _emit_event(host):
     event = delete_event(host)
     key = str(host.id)
-    emit_event(event, key)
+    headers = [("event_type", b"deleted")]
+    emit_event(event, key, headers)
--- a/lib/host_delete.py
+++ b/lib/host_delete.py
@@ -1,7 +1,7 @@
 from sqlalchemy.orm.base import instance_state
 
 from app.events import delete as delete_event
-from app.events import delete_event_name
+from app.events import DELETE_EVENT_NAME
 from app.events import message_headers
 from app.models import Host
 from lib.metrics import delete_host_count
@@ -45,5 +45,5 @@ def _deleted_by_this_query(host):
 def _emit_event(host):
     event = delete_event(host)
     key = str(host.id)
-    headers = message_headers(delete_event_name)
+    headers = message_headers(DELETE_EVENT_NAME)
     emit_event(event, key, headers)
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -21,7 +21,6 @@ def _init_event_producer():
     logger.info("Starting event KafkaProducer()")
     return KafkaProducer(bootstrap_servers=cfg.bootstrap_servers)
 
-    
 
 def emit_event(event, key, headers):
     k = key.encode("utf-8") if key else None
@@ -30,6 +29,7 @@ def emit_event(event, key, headers):
     producer.send(cfg.event_topic, key=k, value=v, headers=h)
     logger.info("Event message produced: topic %s, key %s", cfg.event_topic, key)
     logger.debug("Event message body: %s", event)
+
 
 def flush():
     producer.flush()
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -27,7 +27,6 @@ def emit_event(event, key):
     logger.info("Event message produced: topic %s, key %s", cfg.event_topic, key)
     logger.debug("Event message body: %s", event)
 
-
 def flush():
     producer.flush()
     logger.info("Event messages flushed")
--- a/tasks/__init__.py
+++ b/tasks/__init__.py
@@ -21,9 +21,13 @@ def _init_event_producer():
     logger.info("Starting event KafkaProducer()")
     return KafkaProducer(bootstrap_servers=cfg.bootstrap_servers)
 
+    
 
-def emit_event(event, key):
-    producer.send(cfg.event_topic, key=key.encode("utf-8") if key else None, value=event.encode("utf-8"))
+def emit_event(event, key, headers):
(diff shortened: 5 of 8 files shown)


tasks/__init__.py
```diff
@@ -16,8 +16,14 @@
 
 
 class NullProducer:
-    def send(self, topic, value=None, key=None):
-        logger.debug("NullProducer - logging message:  topic (%s) - message: %s - message_id: %s", topic, value, key)
+    def send(self, topic, value=None, key=None, headers=None):
+        logger.debug(
+            "NullProducer - logging message:  topic (%s) - message: %s - message_id: %s - headers: %s",
```
shdunning, line 21: I wonder what `"NullProducer` is supposed to indicate in this logger message?

app/queue/egress.py
```diff
@@ -47,8 +47,8 @@ class NullEventProducer:
     def __init__(self):
         logger.info("Starting NullEventProducer()")
 
-    def write_event(self, event, key=None):
-        logger.debug("NullEventProducer - logging key: %s, event: %s", key, event)
+    def write_event(self, event, key=None, headers=None):
```
shdunning, line 50: is this method, with optional keyword args, even used anywhere?

test_mq_service.py
```diff
@@ -265,6 +267,25 @@ def test_handle_message_verify_message_key_and_metadata_not_required(self, add_h
...
+        message = {"operation": "add_host", "data": host_data}
+
+        mock_event_producer = MockEventProducer()
+        with self.app.app_context():
+            handle_message(json.dumps(message), mock_event_producer)
+            self.assertEqual(mock_event_producer.headers[0][1], b"created")  # event_type
+
+            add_host.reset_mock()
+            host_data.update(display_name="updated_test_host")
+            print(host_data)
```
jharting, line 284: print

app/queue/ingress.py
```diff
@@ -133,6 +133,10 @@ def add_host(host_data):
             raise
 
 
+def message_headers(event_type):
```
jharting, line 136: can we define this in a common place so that even the delete event code path can reuse this function instead of duplicating?

test_api.py
```diff
@@ -1187,6 +1187,8 @@ def _assert_event_is_valid(self, event, key, host, timestamp):
 
         self.assertEqual(key, host.id)
 
+        self.assertEqual(headers[0][1], b"deleted")
```
jharting, line 1190: beside the header value can we assert the right header key is used as well?

test_mq_service.py
```diff
@@ -265,6 +267,25 @@ def test_handle_message_verify_message_key_and_metadata_not_required(self, add_h
...
+    def test_handle_message_verify_message_headers(self, add_host):
+        host_data = self._host_data()
+        add_host.return_value = (host_data, AddHostResults.created)
+
+        message = {"operation": "add_host", "data": host_data}
+
+        mock_event_producer = MockEventProducer()
+        with self.app.app_context():
+            handle_message(json.dumps(message), mock_event_producer)
+            self.assertEqual(mock_event_producer.headers[0][1], b"created")  # event_type
```
jharting, line 280: The two parts of the test (create and update) seem to be independent (each sets the mock independently, calls handle message and then asserts). Can we turn this into separate test functions or even better a single test function that is parameterized?

test_mq_service.py
```diff
@@ -265,6 +267,25 @@ def test_handle_message_verify_message_key_and_metadata_not_required(self, add_h
...
+
+            add_host.reset_mock()
+            host_data.update(display_name="updated_test_host")
+            print(host_data)
+            add_host.return_value = (host_data, AddHostResults.updated)
+            handle_message(json.dumps(message), mock_event_producer)
+            self.assertEqual(mock_event_producer.headers[0][1], b"updated")  # event_type
```
jharting, line 287: besides the header value let's also test that the right header key is used

app/queue/egress.py
Glutexo: @shdunning _NullEventProducer_ should have the same interface as all other producers. Although I don’t think we even need the class at all.

Glutexo: @shdunning #640 for the rescue.

app/queue/ingress.py
Glutexo: @jharting What is the common place though. The message emission logic is now sprinkled around many modules. 😞 I’m afraid the best place is [_tasks_](https://github.com/RedHatInsights/insights-host-inventory/blob/c2f8afc8f517803a8dda9900d80d3345025a4716/tasks/__init__.py) and I‘m not very happy about that.

jharting: @Glutexo I was thinking `app/events.py` perhaps?

test_mq_service.py
tahmidefaz: :man_facepalming:

app/queue/ingress.py
tahmidefaz: @Glutexo @jharting I just relocated that function to `app/events.py` and was able to use it in all the events. Please let me know if there a better place to put it.

test_mq_service.py
tahmidefaz: @jharting I just added a function named `_check_headers`, please let me know if this is not what you meant.

lib/host_delete.py
```diff
@@ -43,4 +44,5 @@ def _deleted_by_this_query(host):
 def _emit_event(host):
     event = delete_event(host)
     key = str(host.id)
-    emit_event(event, key)
+    headers = message_headers("deleted")
```
jharting, line 47: Even though it is inconsistent with the other events the "type" field of this event is actually `delete`, not `deleted`, see https://github.com/RedHatInsights/insights-host-inventory/blob/92bac2a3d6f3a21fe60b0a1b5eef266b1eef2e55/app/events.py#L32

Might be worth defining this as a constant and referencing from both places.

app/events.py
```diff
@@ -34,3 +34,7 @@ def delete(host):
         )
         .data
     )
+
+
+def message_headers(event_type):
+    return [("event_type", str.encode(event_type))]
```
jharting, line 40: Looks good. A possible nitpick would be to define `event_type` as a constant rather than a magic string.

test_api.py
```diff
@@ -1187,6 +1187,9 @@ def _assert_event_is_valid(self, event, key, host, timestamp):
 
         self.assertEqual(key, host.id)
 
+        self.assertEqual(headers[0][0], "event_type")
```
jharting, line 1190: :+1:

test_mq_service.py
jharting: @tahmidefaz looks good. I think it would be better to keep the handle_message call out of `_check_headers` (i.e. do call `handle_message` in the test function directly) making `_check_headers` a complex assertion similar to e.g. https://github.com/RedHatInsights/insights-host-inventory/blob/master/test_api.py#L1218

That way it's possible to reuse this function in other tests in the future.

jharting:

test_mq_service.py
```diff
@@ -265,6 +267,25 @@ def test_handle_message_verify_message_key_and_metadata_not_required(self, add_h
...
 
+    @patch("app.queue.ingress.host_repository.add_host")
+    def test_handle_message_verify_message_headers(self, add_host):
+        host_data = self._host_data()
+        add_host.return_value = (host_data, AddHostResults.created)
+
+        message = {"operation": "add_host", "data": host_data}
+
+        mock_event_producer = MockEventProducer()
+        with self.app.app_context():
```
Glutexo, line 283: You don’t need to wrap everything by the _app_context_. Wrapping just _handle_message_ is enough.

test_mq_service.py
Glutexo: Parametrized.

```suggestion
    @patch("app.queue.ingress.host_repository.add_host")
    def test_handle_message_verify_message_headers(self, add_host):
        host_data = self._host_data()
        message = {"operation": "add_host", "data": host_data}
        for add_host_result in AddHostResults:
            with self.subTest(add_host_result=add_host_result):
                add_host.reset_mock()
                add_host.return_value = (host_data, add_host_result)

                mock_event_producer = MockEventProducer()
                with self.app.app_context():
                    handle_message(json.dumps(message), mock_event_producer)

                expected_headers = [("event_type", add_host_result.name)]
                self.assertEqual(mock_event_producer.headers, expected_headers)
```

lib/host_delete.py
```diff
@@ -43,4 +43,5 @@ def _deleted_by_this_query(host):
 def _emit_event(host):
     event = delete_event(host)
     key = str(host.id)
-    emit_event(event, key)
+    headers = [("event_type", b"deleted")]
```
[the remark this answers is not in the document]
tahmidefaz: This was already taken care of earlier. Resolving.

test_mq_service.py
tahmidefaz: @Glutexo showed me a better way to parameterize that, removing the function.

app/events.py
tahmidefaz: Changed that function to return a dictionary now. I don't think we would need to do this anymore, but let me know if you think otherwise.

tahmidefaz:
@jharting @Glutexo I think I was able to address both of your suggestions. Please let me know if I need to make any more changes. Thanks!

test_api.py
```diff
@@ -1298,8 +1308,7 @@ def test_culled_host_is_removed(self, emit_event):
 
         self.assertEqual(len(emit_event.events), 1)
 
-        event, key = emit_event.events[0]
-        self._assert_event_is_valid(json.loads(event), key, added_host, self.now_timestamp)
+        self._assert_event_is_valid(emit_event.events[0], added_host, self.now_timestamp)
```
Glutexo, line 1311:
```suggestion
        emit_event.assert_called_once()
        event = emitted_event(emit_event.mock_calls[0])
        self._assert_event_is_valid(event, added_host, self.now_timestamp)
```

test_api.py
```diff
@@ -1912,33 +1921,32 @@ def _delete(self, url_query="", header=None):
             with patch("app.events.datetime", **{"utcnow.return_value": self.timestamp}):
                 url = f"{self.delete_url}{url_query}"
                 self.delete(url, 200, header, return_response_as_json=False)
-                event, key = m.events[0]
-                return json.loads(event), key
+                return m.events[0]
```
Glutexo, line 1924: Let’s simplify this. 

- Move the patch to the class as a decorator `@patch("lib.host_delete.emit_event")`.
- Add the mock as an argument to the test methods. `def test_something(self, emit_event):`
- Dont’t return anything from here. The mock is available in the test methods.

```suggestion
```


```python
@patch("lib.host_delete.emit_event")
class DeleteHostsEventTestCase(PreCreatedHostsBaseTestCase, DeleteHostsBaseTestCase):
    ...

    def _delete(self, url_query="", header=None):
        with patch("app.events.datetime", **{"utcnow.return_value": self.timestamp}):
            url = f"{self.delete_url}{url_query}"
            self.delete(url, 200, header, return_response_as_json=False)
```

test_api.py
```diff
@@ -1912,33 +1921,32 @@ def _delete(self, url_query="", header=None):
...
 
     def test_create_then_delete(self):
         self._check_hosts_are_present((self.host_to_delete.id,))
-        event, key = self._delete()
-        self._assert_event_is_valid(event, key, self.host_to_delete, self.timestamp)
+        event = self._delete()
```
Glutexo, line 1928:
```suggestion
        self._delete()

        emit_event.assert_called_once()
        event = emitted_event(emit_event.mock_calls[0])
```

test_api.py
```diff
@@ -1912,33 +1921,32 @@ def _delete(self, url_query="", header=None):
...
+        self._assert_event_is_valid(event, self.host_to_delete, self.timestamp)
         self._check_hosts_are_deleted((self.host_to_delete.id,))
 
     def test_create_then_delete_with_branch_id(self):
         self._check_hosts_are_present((self.host_to_delete.id,))
-        event, key = self._delete(url_query="?branch_id=1234")
-        self._assert_event_is_valid(event, key, self.host_to_delete, self.timestamp)
+        event = self._delete(url_query="?branch_id=1234")
```
Glutexo, line 1934:
```suggestion
        self._delete(url_query="?branch_id=1234")

        emit_event.assert_called_once()
        event = emitted_event(emit_event.mock_calls[0])
```

test_api.py
```diff
@@ -1912,33 +1921,32 @@ def _delete(self, url_query="", header=None):
...
+        self._assert_event_is_valid(event, self.host_to_delete, self.timestamp)
         self._check_hosts_are_deleted((self.host_to_delete.id,))
 
     def test_create_then_delete_with_request_id(self):
         request_id = generate_uuid()
         header = {"x-rh-insights-request-id": request_id}
-        event, key = self._delete(header=header)
-        self._assert_event_is_valid(event, key, self.host_to_delete, self.timestamp)
-        self.assertEqual(request_id, event["request_id"])
+        event = self._delete(header=header)
```
Glutexo, line 1941:
```suggestion
        self._delete(header=header)

        emit_event.assert_called_once()
        event = emitted_event(emit_event.mock_calls[0])
```

test_api.py
```diff
@@ -1912,33 +1921,32 @@ def _delete(self, url_query="", header=None):
...
+        self._assert_event_is_valid(event, self.host_to_delete, self.timestamp)
+        self.assertEqual(request_id, event.value["request_id"])
 
     def test_create_then_delete_without_request_id(self):
         self._check_hosts_are_present((self.host_to_delete.id,))
-        event, key = self._delete(header=None)
-        self._assert_event_is_valid(event, key, self.host_to_delete, self.timestamp)
-        self.assertEqual("-1", event["request_id"])
+        event = self._delete(header=None)
```
Glutexo, line 1947:
```suggestion
        self._delete()

        emit_event.assert_called_once()
        event = emitted_event(emit_event.mock_calls[0])
```

I know this is not very DRY, but these are tests. They need to be straightforward. The closer are the mocks, patches and asserts to the actual test classes/methods, the better.

test_mq_service.py
```diff
@@ -37,10 +37,12 @@ class MockEventProducer:
     def __init__(self):
         self.event = None
         self.key = None
+        self.headers = None
 
-    def write_event(self, event, key):
+    def write_event(self, event, key, headers):
         self.event = event
         self.key = key
+        self.headers = [(hk, hv.encode("utf-8")) for hk, hv in headers.items()]
```
Glutexo, line 45: Well, actually we don’t need to replicate the internal logic of [_KafkaEventProducer_](https://github.com/RedHatInsights/insights-host-inventory/blob/71241da25ea6e952b42ac7a20e3f28d36e632c7b/app/queue/egress.py#L16). The transformation is done only because _KafkaProducer_ needs it, but here we just compare the values.

```suggestion
        self.headers = headers
```

I think we don’t even necessarily need the class, but here it’s used on many places, so let’s keep it for now. But if you want to do the same as in the API tests, it’d be awesome, although maybe out of the scope of this pull request.

Also, [_FakeKafkaMessage_](https://github.com/RedHatInsights/insights-host-inventory/blob/71241da25ea6e952b42ac7a20e3f28d36e632c7b/test_mq_service.py#L29) just above is not used at all. Also unrelated, but… remove?

test_mq_service.py
```diff
@@ -266,6 +268,24 @@ def test_handle_message_verify_message_key_and_metadata_not_required(self, add_h
...
+            with self.subTest(add_host_result=add_host_result):
+                add_host.reset_mock()
+                add_host.return_value = (host_data, add_host_result)
+
+                mock_event_producer = MockEventProducer()
+                with self.app.app_context():
+                    handle_message(json.dumps(message), mock_event_producer)
+
+                expected_headers = [("event_type", add_host_result.name.encode())]
+                self.assertEqual(mock_event_producer.headers, expected_headers)
```
Glutexo, line 287:
```suggestion
                self.assertEqual(mock_event_producer.headers, {"event_type": add_host_result.name})
```

test_api.py
```diff
@@ -163,8 +166,12 @@ class MockEmitEvent:
     def __init__(self):
         self.events = []
 
-    def __call__(self, e, key=None):
-        self.events.append((e, key))
+    def __call__(self, e, key=None, headers=None):
+        value = json.loads(e)
+        if headers:
+            headers = [(hk, hv.encode("utf-8")) for hk, hv in headers.items()]
```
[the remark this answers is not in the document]
tahmidefaz: Done!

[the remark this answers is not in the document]
Glutexo: In [_test_unknown_host_is_not_removed_](https://github.com/RedHatInsights/insights-host-inventory/blob/83e33bfe0eac750bde8320a1a3c7aa6b49b11887/test_api.py#L1323), _emit_event.events_ comparison must be replaced with _assert_not_called_. Not doing that doesn’t make the test fail, but it doesn’t test correctly that the event is not emitted. _emit_event_ now being a plain _MagicMock_ creates another _MagicMock_ instance on any attribute access: _emit_event.events_ is thus just _MagicMock_ too. For some reason, calling _len_ on such returns _0_, making the test pass.

Glutexo:
Added a follow-up pull request to fix the broken test. #642

PR #636, opened 2020-04-16
hotfix to not update display_name if update comes from Satellite

ryandillinfelton:
Attempt to fix the issue with Satellite display name updates

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -110,7 +110,7 @@ class Host(db.Model):
         self.update_canonical_facts(input_host.canonical_facts)
 
         # TODO: Address this. Hotfix satellite 6.7 issue wher display name is set to fqdn when it
-        if hasattr(input_host, "reporter") or input_host.reporter == "puptoo":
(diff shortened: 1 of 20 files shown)


app/models.py
```diff
@@ -106,10 +106,12 @@ def __init__(
     def save(self):
         db.session.add(self)
 
-    def update(self, input_host, update_system_profile=False):
+    def update(self, input_host, update_system_profile=False, from_REST_API=False):
         self.update_canonical_facts(input_host.canonical_facts)
 
-        self.update_display_name(input_host.display_name)
+        # TODO: Address this. Hotfix satellite 6.7 issue wher display name is set to fqdn when it
+        if from_REST_API or (input_host.reporter and input_host.reporter == "puptoo"):
```
jharting, line 113: @ryandillinfelton @shdunning why is REST_API allowed to change this and MQ not?

jharting:
Can we also have an api/mq test for this?

PR #637, opened 2020-04-17
cleanup Sat 6.7 hotfix

shdunning:
Summary of changes:

- change guard to not update display name for specific reporter == "yupana" (instead of only allowing display name updates for puptoo, which might have been too restrictive)
- remove the `from_REST_API` bool in favor of the change above
- add test an MQ test

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -106,11 +106,12 @@ class Host(db.Model):
     def save(self):
         db.session.add(self)
 
-    def update(self, input_host, update_system_profile=False, from_REST_API=False):
+    def update(self, input_host, update_system_profile=False):
         self.update_canonical_facts(input_host.canonical_facts)
 
         # TODO: Remove this eventually when Sat 6.7 stops sending fqdns as display_names (See RHCLOUD-5954)
-        if from_REST_API or (input_host.reporter and input_host.reporter == "puptoo"):
+        # NOTE: For this particular issue, the reporter is set to "yupana"
+        if input_host.reporter != "yupana":
             self.update_display_name(input_host.display_name)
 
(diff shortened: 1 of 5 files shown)


shdunning:
> Can we:
> 
> * add JIRA reference to the code comments that reference "satellite bug"
Yep - good suggestion.

> * add and actual MQ test
We tried and learned that the MQ tests don't actually test anything other than payloads we are sending are actually making it to the mocked Kafka producer, which in my opinion isn't really testing anything other than Python's mock functionality is working correctly.

> * remove the `from_REST_API` bit - no reason to treat REST ingest differently
We had, just in case an Update to the hostname was being performed via a different REST API other than `PATCH` (just to be safe until we kill the `POST add_hostlists` completely).

app/models.py
```diff
@@ -106,11 +106,12 @@ def __init__(
...
 
-    def update(self, input_host, update_system_profile=False, from_REST_API=False):
+    def update(self, input_host, update_system_profile=False):
         self.update_canonical_facts(input_host.canonical_facts)
 
-        # TODO: Address this. Hotfix satellite 6.7 issue wher display name is set to fqdn when it
-        if from_REST_API or (input_host.reporter and input_host.reporter == "puptoo"):
+        # TODO: Remove this eventually when Sat 6.7 stops sending fqdns as display_names (See RHCLOUD-5954)
+        # NOTE: For this particular issue, display_name changes from "puptoo" and "yupana" are ignored
+        if input_host.reporter != "yupana" and input_host.reporter != "rhsm-conduit":
```
shdunning, line 114: @jharting do we feel pretty good about this list of reporters?

PR #639, opened 2020-04-20
log stmt for every time HBI produces a message (RHCLOUD-4285)

yashnalla:


yashnalla:
https://projects.engineering.redhat.com/browse/RHCLOUD-4285

app/queue/egress.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, config):
 
     def write_event(self, event, key):
         logger.debug("Topic: %s, key: %s, event: %s", self._topic, key, event)
+        logger.info("Topic: %s, key: %s, event: %s", self._topic, key, event)
```
jharting, line 31: Does it make sense to log the same info twice on two different log levels?

app/queue/ingress.py
```diff
@@ -145,6 +145,7 @@ def handle_message(message, event_producer):
         (output_host, add_results) = add_host(validated_operation_msg["data"])
         event = build_event(add_results.name, output_host, metadata)
         event_producer.write_event(event, output_host["id"])
+        logger.info("Writing event of type=" + add_results.name + " at host=" + str(output_host["id"]))
```
jharting, line 148: It would be better to keep the key, event type structured using the extra field. See https://github.com/RedHatInsights/insights-host-inventory/blob/84b91b00628bd7d2be031484983b48835d3beb73/app/queue/ingress.py#L62 for an example

app/queue/ingress.py
```diff
@@ -145,6 +145,9 @@ def handle_message(message, event_producer):
         (output_host, add_results) = add_host(validated_operation_msg["data"])
         event = build_event(add_results.name, output_host, metadata)
         event_producer.write_event(event, output_host["id"])
+        logger.info(
```
jharting, line 148: if the EventProducer does the logging there's no need to log here also

jharting:

app/queue/egress.py
```diff
@@ -27,8 +27,7 @@ def __init__(self, config):
         self._topic = config.host_egress_topic
 
     def write_event(self, event, key):
-        logger.debug("Topic: %s, key: %s, event: %s", self._topic, key, event)
-
+        logger.info("Topic: " + self._topic, extra={"input_host": {"key": key, "event": event}})
```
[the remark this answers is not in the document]
Glutexo: Added headers to the logged info. The event type is in there.

tasks/__init__.py
```diff
@@ -49,6 +49,7 @@ def _init_event_producer(config):
 
 def emit_event(e, key):
     producer.send(cfg.event_topic, key=key.encode("utf-8") if key else None, value=e.encode("utf-8"))
+    logger.info("Producer sending event", extra={"input_host": {"host_id": key, "event_type": cfg.event_topic}})
```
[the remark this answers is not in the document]
Glutexo: Done. Extracted a [_app.logging.log_produced_message_](https://github.com/RedHatInsights/insights-host-inventory/blob/17ec61424ad613fbf3185fbad20e73ab894d282a/app/logging.py#L125) function.

Glutexo:
Rebased to resolve conflicts. I tried to address all the mentioned issues and all those that I found. Also extracted the logging to a common [function](https://github.com/RedHatInsights/insights-host-inventory/blob/17ec61424ad613fbf3185fbad20e73ab894d282a/app/logging.py#L125).

I went the way of logging the metadata (topic, key, headers) on the INFO level and the actual message body on the DEBUG level. As [suggested](https://projects.engineering.redhat.com/browse/RHCLOUD-4285?focusedCommentId=2262485&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2262485) by @shdunning, I tried to leverage the logger _extra_ parameter  to pass in the data in a structured format. I am not sure what good practices are, so please tell me if there is a better way to do it.

Messages are logged _after_ they’re produced, so they copy what really went on the message queue. Or should they be logged beforehand instead, to be logged even if producing fails?

app/queue/egress.py
```diff
@@ -21,13 +22,12 @@ def __init__(self, config):
...
 
     def write_event(self, event, key, headers):
-        logger.debug("Topic: %s, key: %s, event: %s, headers: %s", self._topic, key, event, headers)
-
         try:
             k = key.encode("utf-8") if key else None
             v = event.encode("utf-8")
             h = [(hk, hv.encode("utf-8")) for hk, hv in headers.items()]
             self._kafka_producer.send(self._topic, key=k, value=v, headers=h)
+            log_produced_message(self._topic, event, key, headers)
```
jharting, line 30: Since `send()` is asynchronous it would be better to register a callback here and only declare victory in the logs once we receive the ack.

```suggestion
            self._kafka_producer.send(self._topic, key=k, value=v, headers=h).add_callback(log_event_produced).add_errback(log_event_error)
```

app/logging.py
```diff
@@ -120,3 +120,13 @@ def get_logger(name):
     logger.addFilter(ContextualFilter())
     logger.setLevel(log_level)
     return logger
+
+
+def log_produced_message(logger, topic, value, key, headers):
```
jharting, line 125: I don't think that instrumentation logic like this belongs to this module which already has the responsibility of logger configuration and initialization. Since we don't have a module appointed with logging/metric instrumentation it might be a good time to create one. There are many places in the codebase that could benefit from having low-level logging/metric code replaced with calls to probles. See https://martinfowler.com/articles/domain-oriented-observability.html for details.

app/queue/egress.py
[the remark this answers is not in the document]
Glutexo: Oh, really. My bad. 😞

Glutexo: 🤔

app/logging.py
```diff
@@ -120,3 +120,13 @@ def get_logger(name):
...
+    extra_message = {"key": key, "headers": headers}
+
+    info_extra = {"topic": topic, "message": extra_message}
```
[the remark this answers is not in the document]
Glutexo: Oh. Why this didn’t happen to me? Reproduced.

app/logging.py
Glutexo: Hey, thanks for the great article! I moved the observer event function to a new [module](https://github.com/RedHatInsights/insights-host-inventory/blob/49b38b355344cbbcb2089074649e2a64cdb18211/app/instrumentation.py). I didn’t change the actual logic of simply passing in the logger instance, since that’s just how we currently log stuff. Keeping logger usage and configuration consolidation to a specifically targeted logging overhaul.

Glutexo:
- Introduced the [module](https://github.com/RedHatInsights/insights-host-inventory/blob/3a931e181290a07addab6db2edc8e033b708ab3a/app/instrumentation.py), although still crude and specific for now.
- Used callbacks to log real details about message production.

It was quite challenging to simulate realistic message production malfunction. To reproduce:

1. Launch the app as usual.
2. Stop Kafka or use other means to block connection to its port.
3. Try to produce a message.
4. Wait 30 seconds for timeout. If that doesn’t work, try to stop the application and then wait.

In case of the MQ service that uses Kafka for input, this is trickier. Message production has to be hindered after consumption, but before production. _time.sleep_ can help here.

PR #640, opened 2020-04-20
Remove unused MQ related code

Glutexo:
Removed a few pieces of message queue related code that are no longer in use. These include:

- Mock producers (_NullEventProducer_) only logging messages instead of actually producing them. Added _INFO/DEBUG_ logging to the real producer instead.
- System Profile consumer task and its metrics. This service has been entirely replaced by Ingress.
- Configuration options controlling the tasks and event production. No need to have _another_ way to disable this one specific part of the application.

I didn’t do any further refactoring, although there is still a lot of work to do. When its time comes, it’s at least going to be easier.

No JIRA for this. Induced by #633.


jharting:
@Glutexo should we remove this one also? https://github.com/RedHatInsights/insights-host-inventory/blob/master/tasks/__init__.py#L18

tasks/__init__.py
```diff
@@ -33,26 +25,28 @@ def init_tasks(config, flask_app=None):
...
-    else:
-        logger.info("Starting NullProducer()")
-        return NullProducer()
+    logger.info("Starting KafkaProducer()")
+    return KafkaProducer(bootstrap_servers=config.bootstrap_servers)
 
 
 def emit_event(e, key):
-    producer.send(cfg.event_topic, key=key.encode("utf-8") if key else None, value=e.encode("utf-8"))
+    logger.info("Produced message to topic %s, key %s.", cfg.event_topic, key)
```
jharting, line 40: This is misleading. We should log that an event is being put into a topic only when it's actually being put into a topic

Glutexo: You’re right! 😮

Glutexo:
Amended the log messages. Tested its workings. Thanks for the ✅ ’s!

Glutexo:
After discussing with @jharting removed more pieces of the message queue related code. See the updated description. Re-requesting reviews.

Glutexo:
Oh, I forgot to remove some tests. Gonna fix. :D

Glutexo:
Fixed.

PR #643, opened 2020-04-23
remove metadata field from events

ryandillinfelton:
There is concern that adding the new metadata field at the same time as the patch event will be too much for consumers. This PR simply removes the metadata fields from the messages and updates some tests to expect that. I did not however go through and remove all of the supporting logic so that when we eventually add the field back it should be simple.


shdunning:
@ryandillinfelton can you plz clarify this statement:

> ...new metadata field at the same time as the patch event will be too much for consumers

PR #645, opened 2020-04-27
Incoming tag sets should be replaced on a namespace (RHCLOUD-5156)

Glutexo:
Allows passing tags in host ingress in the nested format. That allows to delete a namespace by including it in the object, not listing any actual tags for it. The original structured format is still supported. REST API remains unchanged, still supporting only the structured format.

Resolves [_RHCLOUD-5156_](https://projects.engineering.redhat.com/browse/RHCLOUD-5156).

## Steps to reproduce:

1. Create a host with some tags. `{"namespace": {"key": ["value"]}}`
2. Update the host with a set of tags that contains the namespace, but no values. `{"namespace": {"key": []}}`
3. Get the host and see it has no tags under that namespace.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -25,6 +25,10 @@ logger = get_logger(__name__)
 
 db = SQLAlchemy()
 
+TAG_NAMESPACE_VALIDATION = validate.Length(max=255)
+TAG_KEY_VALIDATION = validate.Length(min=1, max=255)
+TAG_VALUE_VALIDATION = validate.Length(max=255)
+
 
 def _set_display_name_on_save(context):
     """
@@ -98,12 +102,13 @@ class Host(db.Model):
         self._update_ansible_host(ansible_host)
         self.account = account
         self.facts = facts
-        self.tags = tags
+        self.tags = tags or {}
         self.system_profile_facts = system_profile_facts or {}
         self.stale_timestamp = stale_timestamp
         self.reporter = reporter
 
     def save(self):
+        self._cleanup_tags()
         db.session.add(self)
 
     def update(self, input_host, update_system_profile=False):
@@ -118,7 +123,7 @@ class Host(db.Model):
 
         self.update_facts(input_host.facts)
 
-        self.update_tags(input_host.tags)
+        self._update_tags(input_host.tags)
 
         if update_system_profile:
             self._update_system_profile(input_host.system_profile_facts)
@@ -189,18 +194,19 @@ class Host(db.Model):
         self.facts[namespace] = facts_dict
         orm.attributes.flag_modified(self, "facts")
 
-    def update_tags(self, tags_dict):
-        if tags_dict:
-            if not self.tags:
-                self.tags = tags_dict
-                return
+    def _update_tags(self, tags_dict):
+        if not self.tags:  # fixme: Host tags should never be None, in DB neither NULL nor 'null'
+            self.tags = {}
 
-            for input_namespace, input_tags in tags_dict.items():
-                self.replace_tags_in_namespace(input_namespace, input_tags)
-
-    def replace_tags_in_namespace(self, namespace, tags_dict):
-        self.tags[namespace] = tags_dict
+        for input_namespace, input_tags in tags_dict.items():
+            if input_tags:
+                self.tags[input_namespace] = input_tags
+            elif input_namespace in self.tags:
+                del self.tags[input_namespace]
         orm.attributes.flag_modified(self, "tags")
+
+    def _cleanup_tags(self):
+        self._update_tags(self.tags)
 
     def merge_facts_in_namespace(self, namespace, facts_dict):
         if not facts_dict:
@@ -306,10 +312,10 @@ class FactsSchema(Schema):
     facts = fields.Dict(validate=check_empty_keys)
 
 
-class TagsSchema(Schema):
-    namespace = fields.Str(required=False, allow_none=True, validate=[validate.Length(min=1, max=255)])
-    key = fields.Str(required=True, allow_none=False, validate=[validate.Length(min=1, max=255)])
-    value = fields.Str(required=False, allow_none=True, validate=[validate.Length(min=1, max=255)])
+class StructuredTagsSchema(Schema):
+    namespace = fields.Str(required=False, allow_none=True, validate=TAG_NAMESPACE_VALIDATION)
+    key = fields.Str(required=True, allow_none=False, validate=TAG_KEY_VALIDATION)
+    value = fields.Str(required=False, allow_none=True, validate=TAG_VALUE_VALIDATION)
 
 
 class HostSchema(Schema):
@@ -326,7 +332,8 @@ class HostSchema(Schema):
     mac_addresses = fields.List(fields.Str(validate=validate.Length(min=1, max=59)), validate=validate.Length(min=1))
     external_id = fields.Str(validate=validate.Length(min=1, max=500))
     facts = fields.List(fields.Nested(FactsSchema))
-    tags = fields.List(fields.Nested(TagsSchema))
+    # tags = [fields.Dict(), fields.List(fields.Nested(StructuredTagsSchema))]
+    tags = fields.Raw()
     system_profile = fields.Nested(SystemProfileSchema)
     stale_timestamp = fields.DateTime(required=True, timezone=True)
     reporter = fields.Str(required=True, validate=validate.Length(min=1, max=255))
@@ -335,6 +342,39 @@ class HostSchema(Schema):
     def has_timezone_info(self, timestamp):
         if timestamp.tzinfo is None:
             raise ValidationError("Timestamp must contain timezone info")
+
+    @validates("tags")
+    def validate_tags(self, tags):
+        if isinstance(tags, list):
+            return self._validate_tags_list(tags)
+        elif isinstance(tags, dict):
+            return self._validate_tags_dict(tags)
+        else:
+            raise ValidationError("Tags must be either an object or an array.")
+
+    @staticmethod
+    def _validate_tags_list(tags):
+        StructuredTagsSchema(many=True, strict=True).validate(tags)
+        return True
+
+    @staticmethod
+    def _validate_tags_dict(tags):
+        for namespace, ns_tags in tags.items():
+            TAG_NAMESPACE_VALIDATION(namespace)
+            if ns_tags is not None and not isinstance(ns_tags, dict):
+                raise ValidationError("Tags in a namespace must be an object or null.")
+
+            for key, values in ns_tags.items():
+                TAG_KEY_VALIDATION(key)
+                if values is not None and not isinstance(values, list):
+                    raise ValidationError("Tag values must be an array or null.")
+
+                for value in values:
+                    if value is not None and not isinstance(value, str):
+                        raise ValidationError("Tag value must be a string or null.")
+                    TAG_VALUE_VALIDATION(value)
+
+        return True
 
 
 class PatchHostSchema(Schema):
--- a/app/models.py
+++ b/app/models.py
@@ -317,7 +317,7 @@ class FactsSchema(Schema):
     facts = fields.Dict(validate=check_empty_keys)
 
 
-class StructuredTagsSchema(Schema):
+class TagsSchema(Schema):
     namespace = fields.Str(required=False, allow_none=True, validate=TAG_NAMESPACE_VALIDATION)
     key = fields.Str(required=True, allow_none=False, validate=TAG_KEY_VALIDATION)
     value = fields.Str(required=False, allow_none=True, validate=TAG_VALUE_VALIDATION)
@@ -360,7 +360,7 @@ class HostSchema(Schema):
 
     @staticmethod
     def _validate_tags_list(tags):
-        StructuredTagsSchema(many=True, strict=True).validate(tags)
+        TagsSchema(many=True, strict=True).validate(tags)
         return True
 
     @staticmethod
--- a/app/models.py
+++ b/app/models.py
@@ -198,11 +198,16 @@ class Host(db.Model):
         if not self.tags:  # fixme: Host tags should never be None, in DB neither NULL nor 'null'
             self.tags = {}
 
+        namespaces_to_delete = []
         for input_namespace, input_tags in tags_dict.items():
             if input_tags:
                 self.tags[input_namespace] = input_tags
             elif input_namespace in self.tags:
-                del self.tags[input_namespace]
+                namespaces_to_delete.append(input_namespace)
+
+        for namespace in namespaces_to_delete:
+            del self.tags[namespace]
+
         orm.attributes.flag_modified(self, "tags")
 
     def _cleanup_tags(self):
@@ -332,8 +337,7 @@ class HostSchema(Schema):
     mac_addresses = fields.List(fields.Str(validate=validate.Length(min=1, max=59)), validate=validate.Length(min=1))
     external_id = fields.Str(validate=validate.Length(min=1, max=500))
     facts = fields.List(fields.Nested(FactsSchema))
-    # tags = [fields.Dict(), fields.List(fields.Nested(StructuredTagsSchema))]
-    tags = fields.Raw()
+    tags = fields.Raw(allow_none=True)
     system_profile = fields.Nested(SystemProfileSchema)
     stale_timestamp = fields.DateTime(required=True, timezone=True)
     reporter = fields.Str(required=True, validate=validate.Length(min=1, max=255))
@@ -349,8 +353,10 @@ class HostSchema(Schema):
             return self._validate_tags_list(tags)
         elif isinstance(tags, dict):
             return self._validate_tags_dict(tags)
+        elif tags is None:
+            return True
         else:
-            raise ValidationError("Tags must be either an object or an array.")
+            raise ValidationError("Tags must be either an object, an array or null.")
 
     @staticmethod
     def _validate_tags_list(tags):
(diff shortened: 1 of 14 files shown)


Glutexo:
One failing test in every file. A winning sequence.

app/models.py
```diff
@@ -336,6 +347,47 @@ def has_timezone_info(self, timestamp):
...
+
+            for key, values in ns_tags.items():
+                TAG_KEY_VALIDATION(key)
+                if values is None:
+                    continue
+                if not isinstance(values, list):
+                    raise ValidationError("Tag values must be an array or null.")
+
+                for value in values:
+                    if value is None:
```
jharting, line 383: I think we should validate that "values" is a list of **non-null** values. See my comment at https://projects.engineering.redhat.com/browse/RHCLOUD-5805

app/models.py
```diff
@@ -98,12 +102,13 @@ def __init__(
...
         self.account = account
         self.facts = facts
-        self.tags = tags
+        self.tags = tags or {}
         self.system_profile_facts = system_profile_facts or {}
         self.stale_timestamp = stale_timestamp
         self.reporter = reporter
 
     def save(self):
+        self._cleanup_tags()
```
Glutexo, line 111: This may look fishy as this effectively introduces state. 😱 That’s however unfortunately an inseparable feature of:

- Using a DB model class for business logic; during a save, this already happens.
- Using the same object for the input data (not exactly a host representation, more a create host request) and for the created/updated host and its output representation.

During the [update](https://github.com/Glutexo/insights-host-inventory/blob/32ad5800d604400a70e3efb19ddae74893c630f5/lib/host_repository.py#L156), this is better, because we use the input host only as a set of new host data / update commands. Here, on [create](https://github.com/Glutexo/insights-host-inventory/blob/32ad5800d604400a70e3efb19ddae74893c630f5/lib/host_repository.py#L145) this is worse, because after the save, the input host object becomes a created host object. As already stated, it’s just how things are.

app/models.py
```diff
@@ -189,19 +194,25 @@ def replace_facts_in_namespace(self, namespace, facts_dict):
...
+        namespaces_to_delete = []
+        for input_namespace, input_tags in tags_dict.items():
+            if input_tags:
+                self.tags[input_namespace] = input_tags
+            elif input_namespace in self.tags:
+                namespaces_to_delete.append(input_namespace)
 
-            for input_namespace, input_tags in tags_dict.items():
-                self.replace_tags_in_namespace(input_namespace, input_tags)
+        for namespace in namespaces_to_delete:
```
Glutexo, line 208: Not doing this the original iteration, because it would break if _self.tags_ was passed as _ tags_dict_. Not really happening, just a pre-caution.

app/models.py
```diff
@@ -189,19 +194,25 @@ def replace_facts_in_namespace(self, namespace, facts_dict):
...
+            del self.tags[namespace]
 
-    def replace_tags_in_namespace(self, namespace, tags_dict):
-        self.tags[namespace] = tags_dict
         orm.attributes.flag_modified(self, "tags")
 
+    def _cleanup_tags(self):
+        self._update_tags(self.tags)
+
```
Glutexo, line 215: For the sake of semantics instead of explanatory comments.

app/models.py
```diff
@@ -326,7 +337,7 @@ class HostSchema(Schema):
     mac_addresses = fields.List(fields.Str(validate=validate.Length(min=1, max=59)), validate=validate.Length(min=1))
     external_id = fields.Str(validate=validate.Length(min=1, max=500))
     facts = fields.List(fields.Nested(FactsSchema))
-    tags = fields.List(fields.Nested(TagsSchema))
+    tags = fields.Raw(allow_none=True)
```
Glutexo, line 340: Being benevolent allowed _null_. It is equivalent to an empty array or empty dictionary. It is always [saved](https://github.com/Glutexo/insights-host-inventory/blob/52fa371dbe0de11cab73eceae73e98eb52aaae15/app/models.py#L105) as an empty object.

The REST API still [requires](https://github.com/Glutexo/insights-host-inventory/blob/52fa371dbe0de11cab73eceae73e98eb52aaae15/swagger/api.spec.yaml#L683) an empty array or a missing _tags_ key. I didn’t change the specification to save work on a soon to be deprecated subsystem.

app/models.py
```diff
@@ -336,6 +347,47 @@ def has_timezone_info(self, timestamp):
...
+        elif isinstance(tags, dict):
+            return self._validate_tags_dict(tags)
+        elif tags is None:
+            return True
+        else:
+            raise ValidationError("Tags must be either an object, an array or null.")
+
+    @staticmethod
+    def _validate_tags_list(tags):
+        StructuredTagsSchema(many=True, strict=True).validate(tags)
```
Glutexo: Leveraging the side-effect of _strict=True_. _ValidationError_ is raised if _tags_ doesn’t conform.

app/models.py
```diff
@@ -336,6 +347,47 @@ def has_timezone_info(self, timestamp):
...
+        return True
+
+    @staticmethod
+    def _validate_tags_dict(tags):
+        for namespace, ns_tags in tags.items():
+            TAG_NAMESPACE_VALIDATION(namespace)
+            if ns_tags is None:
+                continue
```
Glutexo, line 371: Being benevolent, allowing _null_ instead of the actual sub-object. Why not. The internal representation remains the same – an empty object.

app/models.py
```diff
@@ -336,6 +347,47 @@ def has_timezone_info(self, timestamp):
...
+            if not isinstance(ns_tags, dict):
+                raise ValidationError("Tags in a namespace must be an object or null.")
+
+            for key, values in ns_tags.items():
+                TAG_KEY_VALIDATION(key)
+                if values is None:
+                    continue
```
Glutexo, line 378: The same for the list of values. Again, internally always stored as an empty array.

app/models.py
```diff
@@ -336,6 +347,47 @@ def has_timezone_info(self, timestamp):
...
+                if not isinstance(values, list):
+                    raise ValidationError("Tag values must be an array or null.")
+
+                for value in values:
+                    if value is None:
+                        continue
```
Glutexo, line 384: Allowing none just as empty strings. They’re not going to be stored.

app/serialization.py
```diff
@@ -172,8 +172,67 @@ def _serialize_uuid(u):
...
+        return []
+    else:
+        raise ValueError("Tags must be dict, list or None.")
+
+
+def _deserialize_tags_list(tags):
+    deserialized = {}
+
+    for tag_data in tags:
+        namespace = Tag.deserialize_namespace(tag_data.get("namespace"))
```
Glutexo, line 189: Keeping the logic of treating _"null"_ same as an empty value. I still think that’s not correct (reported as [_RHCLOUD-5803_](https://projects.engineering.redhat.com/browse/RHCLOUD-5803)), but keeping that for consistency.

app/serialization.py
```diff
@@ -172,8 +172,67 @@ def _serialize_uuid(u):
...
+            deserialized_key.append(value)
+
+    return deserialized
+
+
+def _deserialize_tags_dict(tags):
+    deserialized_tags = {}
+
+    for namespace, tags_ns in tags.items():
+        deserialized_namespace = Tag.deserialize_namespace(namespace)
```
Glutexo, line 212: Dtto.

app/utils.py
```diff
@@ -352,28 +349,8 @@ def create_tags_from_nested(nested_tags):
...
-                    tags.append(Tag(Tag._if_null(namespace), key))
+                    tags.append(Tag(Tag.deserialize_namespace(namespace), key))
                 else:
                     for value in nested_tags[namespace][key]:
-                        tags.append(Tag(Tag._if_null(namespace), key, value))
+                        tags.append(Tag(Tag.deserialize_namespace(namespace), key, value))
         return tags
-
-    @staticmethod
-    def create_structered_tags_from_tag_data_list(tag_data_list):
```
Glutexo, line 362: Sneakily stripping flesh from the _Tag_ class, moving it to [_app.serialization_](https://github.com/Glutexo/insights-host-inventory/blob/52fa371dbe0de11cab73eceae73e98eb52aaae15/app/serialization.py#L185).

swagger/api.spec.yaml
```diff
@@ -322,13 +322,13 @@ paths:
...
-                schema:
-                  $ref: '#/components/schemas/ActiveTags'
+          description: Tags
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ActiveTags'
         '400':
-            description: Invalid request.
+          description: Invalid request.
```
Glutexo: Automatic reformatting. 🤷🏻‍♂️

test_api.py
```diff
@@ -239,7 +239,7 @@ def _make_http_call(self, http_method, path, data, status, return_response_as_js
         return self._response_check(response, status, return_response_as_json)
 
     def _response_check(self, response, status, return_response_as_json):
-        self.assertEqual(response.status_code, status)
+        self.assertEqual(status, response.status_code)
```
Glutexo, line 242: The first argument should be “expected” and the second “actual”. At least PyCharm’s test suite output thinks so. Swapped those here in these broadly used methods, so if a test fails, the log output of the IDE is closer to the truth.

test_api.py
```diff
@@ -281,7 +281,7 @@ def _build_host_id_list_for_url(self, host_list):
         return ",".join(host_id_list)
 
     def _verify_host_status(self, response, host_index, expected_status):
-        self.assertEqual(response["data"][host_index]["status"], expected_status)
+        self.assertEqual(expected_status, response["data"][host_index]["status"])
```
Glutexo, line 284: Dtto.

test_api.py
```diff
@@ -879,15 +879,79 @@ def test_create_host_with_invalid_ansible_host(self):
...
+        )
+
+        for facts in samples:
+            with self.subTest(facts=facts):
+                host_data = HostWrapper(test_data(facts=facts))
+                response = self.post(HOST_URL, [host_data.data()], 207)
+                self._verify_host_status(response, 0, 400)
+
+
+class CreateHostsWithTagsTestCase(DBAPITestCase):
```
Glutexo, line 941: Although nothing changed for the REST API, I found the tests not exhaustive. I needed to be sure that nothing broke, so I added some more verifying the actual behavior.

test_api.py
```diff
@@ -879,15 +879,79 @@ def test_create_host_with_invalid_ansible_host(self):
...
+    def test_create_host_with_null_tag_key(self):
```
Glutexo, line 942: [_StructuredTag_](https://github.com/Glutexo/insights-host-inventory/blob/52fa371dbe0de11cab73eceae73e98eb52aaae15/swagger/api.spec.yaml#L508) doesn’t list _key_ as _required_. That means that it’s ok to omit it, but it’s not to include it, but with _null_ as its value. So this is the only case caught by the request body validation.

test_api.py
```diff
@@ -879,15 +879,79 @@ def test_create_host_with_invalid_ansible_host(self):
...
+        tag = ({"namespace": "ns", "key": None, "value": "val"},)
+        host_data = HostWrapper(test_data(tags=[tag]))
+        self.post(HOST_URL, [host_data.data()], 400)
+
     def test_create_host_with_invalid_tags(self):
```
Glutexo, line 947: The length constraints are missing in the API [specification](https://github.com/Glutexo/insights-host-inventory/blob/52fa371dbe0de11cab73eceae73e98eb52aaae15/swagger/api.spec.yaml#L508) as well. So these violations are caught by Marshmallow, but not by Connexion.

test_db_model.py
```diff
@@ -194,22 +194,6 @@ def test_host_schema_timezone_enforced():
     assert "Timestamp must contain timezone info" in str(excinfo.value)
 
 
-def test_tag_deserialization():
```
Glutexo, line 197: The _create_structered_tags_from_tag_data_list_ is gone, replaced by _ [_app.serialization.:deserialize_tags_list_](https://github.com/Glutexo/insights-host-inventory/blob/52fa371dbe0de11cab73eceae73e98eb52aaae15/app/serialization.py#L185). Now exhaustively tested [elsewhere](https://github.com/Glutexo/insights-host-inventory/blob/52fa371dbe0de11cab73eceae73e98eb52aaae15/test_unit.py#L1464).

test_mq_service.py
```diff
@@ -353,8 +356,13 @@ def test_add_host_with_tags(self, datetime_mock):
                 {"namespace": "NS1", "key": "key3", "value": "val3"},
```
Glutexo, line 356: Again, trying to make the tests as exhaustive as possible to catch any possible regression.

test_api.py
```diff
@@ -879,15 +879,83 @@ def test_create_host_with_invalid_ansible_host(self):
...
+        for facts in samples:
+            with self.subTest(facts=facts):
+                host_data = HostWrapper(test_data(facts=facts))
+                response = self.post(HOST_URL, [host_data.data()], 207)
+                self._verify_host_status(response, 0, 400)
+
+
+class CreateHostsWithTagsTestCase(DBAPITestCase):
+    def test_create_host_with_null_tags(self):
+        host_data = HostWrapper(test_data(tags=None))
```
Glutexo, line 943: Tags is not listed as _required_, but also not as _nullable_. That means it can be omitted, but when present it can’t be _null_. That’s not correct, but it’s already like that, so added a test to prevent regressions.

app/models.py
Glutexo: It’s been like this ever since. Both OpenAPI [specification](https://github.com/RedHatInsights/insights-host-inventory/blob/23eb955783381e544e5bef50230be01b82ccc29e/swagger/api.spec.yaml#L518) and _TagsSchema_ explicitly [allowed](https://github.com/RedHatInsights/insights-host-inventory/blob/23eb955783381e544e5bef50230be01b82ccc29e/app/models.py#L312) _null_ value. These are [then](https://github.com/RedHatInsights/insights-host-inventory/blob/23eb955783381e544e5bef50230be01b82ccc29e/app/utils.py#L301) [ignored](https://github.com/RedHatInsights/insights-host-inventory/blob/23eb955783381e544e5bef50230be01b82ccc29e/app/utils.py#L310) and the key is saved without the empty value.

It had to be like this. Otherwise it wouldn’t be possible to add a value-less tag using the structured format: `{"namespace": "namespace", "key": "key", "value": null}`

test_api.py
rodrigonull: According to the tagging documentation, the "key" should be mandatory.
https://docs.google.com/document/d/1Gfkc3QRWezyHXr8jnrHNYSUsBsetG8bZyx9uCmcRxEg

app/models.py
jharting: It's not clear to my why this would break. Can you elaborate and perhaps add it as a code comment also so that the second loop does not get removed on next refactoring?

app/serialization.py
```diff
@@ -172,8 +172,67 @@ def _serialize_uuid(u):
...
+        if namespace not in deserialized:
+            deserialized[namespace] = {}
+        deserialized_ns = deserialized[namespace]
+
+        if not tag_data.get("key"):
+            raise ValueError("Key cannot be empty.")
+
+        if tag_data["key"] not in deserialized_ns:
+            deserialized_ns[tag_data["key"]] = []
+        deserialized_key = deserialized_ns[tag_data["key"]]
```
jharting, line 199: Looks like this is more like `deserialized_values` rather than `deserialized_key`

utils/payloads.py
```diff
@@ -513,12 +513,10 @@ def build_host_chunk():
...
         "fqdn": fqdn,
         "display_name": fqdn,
-        "tags": [
-            {"namespace": "SPECIAL", "key": "key", "value": "val"},
-            {"namespace": "NS3", "key": "key3", "value": "val3"},
-            {"namespace": "NS1", "key": "key3", "value": "val3"},
-            {"namespace": "Sat", "key": "prod", "value": None},
-        ],
+        "tags": None,
+        # "tags": {
```
jharting, line 517: intention or accident?

Glutexo:
I’ll address the issues mentioned in comments.

app/serialization.py
Glutexo: I tried to make the code more readable by shortening the long two-level depp dictionary item references to a single variable. As I see now, it only caused confusion.

Instead, I rather replaced `tag_data["key"]` with just `key`. Like that, even the deepest reference is `deserialized[namespace][key]`, which is not that awful, yet it’s clear it’s a value inside a structure.

test_api.py
Glutexo: @rodrigonull Yeah. It’s a bug of our HTTP API specification. Because we are going to remove tags update through HTTP, I didn’t want to change (fix) it. So I just updated the tests so they test the current behavior, making them fail if anything changes unexpectedly.

utils/payloads.py
Glutexo: Accident. 😞

Glutexo: Accident. 😞

Glutexo: The _None_ was an accident. The commented out dictionary version of tags is intentional.

Glutexo:
I addressed the review comments. Refactored the tags update, so it’s now less fishy and more semantic.

Also I found out that there are no tests for correctly handling legacy hosts with `NULL` or `'null'` in _tags_. Added those.

Glutexo:
Found a bug. It’s not possible to delete a _null_ namespace. Working on a test and a fix.

Glutexo:
Fixed, tested. Added a few more tests verifying that all no namespace representations (missing key, _null_, empty string and string _"null"_, are treated the same on update. 🎉

Glutexo:
🥳

PR #647, opened 2020-04-28
Disable updating/creating tags through rest [RHCLOUD-5593]

ryandillinfelton:
This update disables updating and creating tags through the REST interface. Requests with tags will succeed, but will not change the stored tag values. 
This PR also changes a lot in tests. We now have to create hosts with tags through ingress so that has had some impact. Some other tests have also been changed to remove data that is no present in the local list of created hosts, but is not returned by the endpoints that are being tested/

[Link to Jira](https://projects.engineering.redhat.com/browse/RHCLOUD-5593)

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -29,6 +29,7 @@ from app.models import PatchHostSchema
 from app.payload_tracker import get_payload_tracker
 from app.payload_tracker import PayloadTrackerContext
 from app.payload_tracker import PayloadTrackerProcessingContext
+from app.probes import countTagsIgnored
 from app.queue.egress import build_event_topic_event
 from app.queue.ingress import EGRESS_HOST_FIELDS
 from app.serialization import deserialize_host
@@ -40,6 +41,7 @@ from lib.host_repository import add_host
 from lib.host_repository import AddHostResults
 from lib.host_repository import find_non_culled_hosts
 from tasks import emit_event
+
 
 FactOperations = Enum("FactOperations", ("merge", "replace"))
 TAG_OPERATIONS = ("apply", "remove")
@@ -112,7 +114,10 @@ def _add_host(input_host):
 
     # RHCLOUD-5593
     # Remove tags from host object here to keep from creating/updating tags from REST
-    input_host.tags = {}
+    if input_host.tags:
+        input_host.tags = {}
+        countTagsIgnored()
+        logger.info("tags were ignored")
 
     return add_host(input_host, staleness_timestamps(), update_system_profile=False)
 
--- a/api/host.py
+++ b/api/host.py
@@ -18,6 +18,8 @@ from app import db
 from app import inventory_config
 from app.auth import current_identity
 from app.config import BulkQuerySource
+from app.events import message_headers
+from app.events import UPDATE_EVENT_NAME
 from app.exceptions import InventoryException
 from app.exceptions import ValidationException
 from app.logging import get_logger
@@ -27,14 +29,17 @@ from app.models import PatchHostSchema
 from app.payload_tracker import get_payload_tracker
 from app.payload_tracker import PayloadTrackerContext
 from app.payload_tracker import PayloadTrackerProcessingContext
+from app.queue.egress import build_event_topic_event
+from app.queue.ingress import EGRESS_HOST_FIELDS
 from app.serialization import deserialize_host
+from app.serialization import serialize_host
 from app.serialization import serialize_host_system_profile
 from app.utils import Tag
 from lib.host_delete import delete_hosts
 from lib.host_repository import add_host
 from lib.host_repository import AddHostResults
 from lib.host_repository import find_non_culled_hosts
-
+from tasks import emit_event
 
 FactOperations = Enum("FactOperations", ("merge", "replace"))
 TAG_OPERATIONS = ("apply", "remove")
@@ -51,7 +56,7 @@ def add_host_list(host_list):
     response_host_list = []
     number_of_errors = 0
 
-    payload_tracker = get_payload_tracker(account=current_identity.account_number, payload_id=threadctx.request_id)
+    payload_tracker = get_payload_tracker(account=current_identity.account_number, request_id=threadctx.request_id)
 
     with PayloadTrackerContext(payload_tracker, received_status_message="add host operation"):
 
@@ -169,7 +174,7 @@ def get_host_list(
 @api_operation
 @metrics.api_request_time.time()
 def delete_by_id(host_id_list):
-    payload_tracker = get_payload_tracker(account=current_identity.account_number, payload_id=threadctx.request_id)
+    payload_tracker = get_payload_tracker(account=current_identity.account_number, request_id=threadctx.request_id)
 
     with PayloadTrackerContext(payload_tracker, received_status_message="delete operation"):
         query = _get_host_list_by_id_list(current_identity.account_number, host_id_list)
@@ -234,6 +239,13 @@ def get_host_system_profile_by_id(host_id_list, page=1, per_page=100, order_by=N
     return flask_json_response(json_output)
 
 
+def _emit_patch_event(host):
+    key = host["id"]
+    event = build_event_topic_event("updated", host, request_id=threadctx.request_id)
+    headers = message_headers(UPDATE_EVENT_NAME)
+    emit_event(event, key, headers)
+
+
 @api_operation
 @metrics.api_request_time.time()
 def patch_by_id(host_id_list, host_data):
@@ -253,6 +265,7 @@ def patch_by_id(host_id_list, host_data):
 
     for host in hosts_to_update:
         host.patch(validated_patch_host_data)
+        _emit_patch_event(serialize_host(host, staleness_timestamps(), EGRESS_HOST_FIELDS))
 
     db.session.commit()
 
--- a/api/host.py
+++ b/api/host.py
@@ -67,11 +67,9 @@ def add_host_list(host_list):
                 with PayloadTrackerProcessingContext(
                     payload_tracker, processing_status_message="adding/updating host"
                 ) as payload_tracker_processing_ctx:
-                    # RHCLOUD-5593
-                    # Remove tags from host object here to keep from creating/updating tags from REST
                     if host.get("tags"):
                         ignored_tags_from_http_request_count.inc()
-                        logger.info("tags from an HTTP request were ignored")
+                        logger.info("Tags from an HTTP request were ignored")
 
                     input_host = deserialize_host_http(host)
                     (output_host, add_result) = _add_host(input_host)
--- a/api/host.py
+++ b/api/host.py
@@ -104,6 +104,9 @@ def _add_host(input_host):
             detail="The account number associated with the user does not match the account number associated with the "
             "host",
         )
+    # #Remove tags from host object here
+    # Will this cause updates to remove all of the tags from an existing host with tags???????!
+    input_host.tags = {}
 
     return add_host(input_host, staleness_timestamps(), update_system_profile=False)
 
--- a/api/host.py
+++ b/api/host.py
@@ -14,7 +14,7 @@ from api.host_query import staleness_timestamps
 from api.host_query_db import get_host_list as get_host_list_db
 from api.host_query_db import params_to_order_by
 from api.host_query_xjoin import get_host_list as get_host_list_xjoin
-from api.metrics import ignored_tags_from_http_request_count
+from api.metrics import tags_ignored_from_http_count
 from app import db
 from app import inventory_config
 from app.auth import current_identity
@@ -68,7 +68,7 @@ def add_host_list(host_list):
                     payload_tracker, processing_status_message="adding/updating host"
                 ) as payload_tracker_processing_ctx:
                     if host.get("tags"):
-                        ignored_tags_from_http_request_count.inc()
+                        tags_ignored_from_http_count.inc()
                         logger.info("Tags from an HTTP request were ignored")
 
                     input_host = deserialize_host_http(host)
--- a/api/host.py
+++ b/api/host.py
@@ -14,6 +14,7 @@ from api.host_query import staleness_timestamps
 from api.host_query_db import get_host_list as get_host_list_db
 from api.host_query_db import params_to_order_by
 from api.host_query_xjoin import get_host_list as get_host_list_xjoin
+from api.metrics import ignored_tags_from_http_request_count
 from app import db
 from app import inventory_config
 from app.auth import current_identity
@@ -29,10 +30,9 @@ from app.models import PatchHostSchema
 from app.payload_tracker import get_payload_tracker
 from app.payload_tracker import PayloadTrackerContext
 from app.payload_tracker import PayloadTrackerProcessingContext
-from app.probes import countTagsIgnored
 from app.queue.egress import build_event_topic_event
 from app.queue.ingress import EGRESS_HOST_FIELDS
-from app.serialization import deserialize_host_REST
+from app.serialization import deserialize_host_http
 from app.serialization import serialize_host
 from app.serialization import serialize_host_system_profile
 from app.utils import Tag
@@ -67,7 +67,13 @@ def add_host_list(host_list):
                 with PayloadTrackerProcessingContext(
                     payload_tracker, processing_status_message="adding/updating host"
                 ) as payload_tracker_processing_ctx:
-                    input_host = deserialize_host_REST(host)
+                    # RHCLOUD-5593
+                    # Remove tags from host object here to keep from creating/updating tags from REST
+                    if host.get("tags"):
+                        ignored_tags_from_http_request_count.inc()
+                        logger.info("tags from an HTTP request were ignored")
+
+                    input_host = deserialize_host_http(host)
                     (output_host, add_result) = _add_host(input_host)
                     status_code = _convert_host_results_to_http_status(add_result)
                     response_host_list.append({"status": status_code, "host": output_host})
@@ -111,13 +117,6 @@ def _add_host(input_host):
             detail="The account number associated with the user does not match the account number associated with the "
             "host",
         )
-
-    # RHCLOUD-5593
-    # Remove tags from host object here to keep from creating/updating tags from REST
-    if input_host.tags:
-        input_host.tags = {}
-        countTagsIgnored()
-        logger.info("tags were ignored")
 
     return add_host(input_host, staleness_timestamps(), update_system_profile=False)
 
--- a/api/host.py
+++ b/api/host.py
@@ -109,8 +109,9 @@ def _add_host(input_host):
             detail="The account number associated with the user does not match the account number associated with the "
             "host",
         )
-    # #Remove tags from host object here
-    # Will this cause updates to remove all of the tags from an existing host with tags???????!
+
+    # RHCLOUD-5593
+    # Remove tags from host object here to keep from creating/updating tags from REST
     input_host.tags = {}
 
     return add_host(input_host, staleness_timestamps(), update_system_profile=False)
--- a/api/host.py
+++ b/api/host.py
@@ -351,7 +351,7 @@ def _count_tags(host_list):
 
     for host in host_list:
         host_tag_count = 0
-        if host.tags is not None:
+        if host.tags is not None:  # fixme: Host tags should never be None, in DB neither NULL nor 'null'
             for namespace in host.tags:
                 for tag in host.tags[namespace]:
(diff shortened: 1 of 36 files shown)


jharting:
@ryandillinfelton @Glutexo heads up: watch out for conflicts with https://github.com/RedHatInsights/insights-host-inventory/pull/645

api/host.py
```diff
@@ -110,6 +110,10 @@ def _add_host(input_host):
             "host",
         )
 
+    # RHCLOUD-5593
+    # Remove tags from host object here to keep from creating/updating tags from REST
+    input_host.tags = {}
+
```
shdunning, line 116: 🤔 should we return an error here if tags are provided? if someone uses this API before we have a chance to remove it they would be surprised to learn that tags are being silently ignored.

jharting: @shdunning @ryandillinfelton I think at this point we should make the change as little breaking as possible i.e. keep the formats and just change the behavior. To deal with the silent bit we can add log/metric/alert for when the tags are actually ignored.

Glutexo:
In the current state, the tags are not saved, but they are still validated and thus can cause 400 _Bad Request_. For them to be fully ignored, they need to be removed from the OpenAPI [specification](https://github.com/RedHatInsights/insights-host-inventory/blob/3e6a594342592d97defc5ed785aa75ac3b253a73/swagger/api.spec.yaml#L683). The field would however still be validate by the [schema](https://github.com/RedHatInsights/insights-host-inventory/blob/3e6a594342592d97defc5ed785aa75ac3b253a73/app/models.py#L329) that is shared with the Message Queue API.

I believe splitting the schema would be the correct solution because and  HTTP API now really has different request format. That would however require removing the validation (load) from the [deserialization](https://github.com/RedHatInsights/insights-host-inventory/blob/3e6a594342592d97defc5ed785aa75ac3b253a73/app/serialization.py#L45) function or parametrizing it. Again, I believe that would be the correct solution, because the function is called only from two places, one for each API. Later, we can think of a way to leverage Marshmallow with custom function to completely handle the (de)serialization, but for now, let’s separate those.

jharting:
@Glutexo since this entire operation is going to be deprecated soon I suggest to keep the change simple, keep the current schema and just not write the incoming tags to the database, as @ryandillinfelton implemented

Glutexo:
#645 has been merged. There are now conflicts because of the _tags_ schema field format change.

app/serialization.py
```diff
@@ -40,9 +41,9 @@
 )
 
 
-def deserialize_host(raw_data):
+def deserialize_host(raw_data, schema):
     try:
-        validated_data = HostSchema(strict=True).load(raw_data).data
+        validated_data = schema.load(raw_data).data
```
Glutexo, line 46: It’s this function’s responsibility to validate the schema in a way that it raises a _ValidationException_. Thus a Schema class should be passed in instead of its instance.

```suggestion
        validated_data = schema(strict=True).load(raw_data).data
```

swagger/api.spec.yaml
```diff
@@ -680,11 +680,6 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/FactSet'
-        tags:
```
Glutexo, line 683: It’s also necessary to remove _tags_ from _CreateHostOut_. https://github.com/RedHatInsights/insights-host-inventory/blob/92cfa0a337fd264a1ea347b15d9a3bcea1987870/swagger/api.spec.yaml#L807 If that’s not done, it breaks [exception handling](https://github.com/RedHatInsights/insights-host-inventory/blob/92cfa0a337fd264a1ea347b15d9a3bcea1987870/api/host.py#L75). If an exception is raised during the host addition, the input host with all its bad data is put into the response. The response is validated by Connexion and if _tags_ contained anything else than valid [structured tag format](https://github.com/RedHatInsights/insights-host-inventory/blob/92cfa0a337fd264a1ea347b15d9a3bcea1987870/swagger/api.spec.yaml#L508), it would explode with _500 Internal Server Error_.

swagger/api.spec.yaml
```diff
@@ -680,11 +680,6 @@ components:
...
-          description: The tags on a host
-          type: array
-          items:
-            $ref: '#/components/schemas/StructuredTag'
```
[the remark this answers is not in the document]
jharting: @ryandillinfelton  @Glutexo hmm OK I was under impression that we're disallowing additional properties i.e. after this change if someone keeps sending tags they would get 400 due to an unexpected `tags` field. However, that does not seem to be the case here i.e. if they keep sending `tags` the sheer fact that the field is present won't automatically cause the request to be rejected. In that case I am fine either way. Sorry for confusion.

test_api.py
```diff
@@ -880,126 +882,56 @@ def test_create_host_with_invalid_ansible_host(self):
...
-        original_id = created_host["id"]
+        # attempt to update
+        self.post(HOST_URL, [host_data.data()], 207)
 
-        host_lookup_results = self.get(f"{HOST_URL}/{original_id}", 200)
+        tags_after_update = self.get(f"{HOST_URL}/{host_data.id}/tags")
 
-        self._validate_host(host_lookup_results["results"][0], host_data, expected_id=original_id)
+        # check the tags haven't updated
+        self.assertNotEqual(host_data.tags, tags_after_update)
```
Glutexo, line 924: This doesn’t test the right thing. _tags_after_update_ contains a full get tags response that should look like this:

```python
{'total': 1, 'count': 1, 'page': 1, 'per_page': 50, 'results': {'eeaa3525-fddc-49cc-8145-c1620c7976c4': [{'namespace': 'ns', 'key': 'some_key', 'value': 'val'}]}}
```

We need to test that the tags were not updated. So we need to compare `tags_after_update["results"][host_id]["tags"]` to the original tags.

```suggestion
        self.assertEqual(host_data.tags, tags_response["results"][host_id])
```

test_api.py
```diff
@@ -880,126 +882,56 @@ def test_create_host_with_invalid_ansible_host(self):
...
 
-        created_host = self._pluck_host_from_response(response, 0)
+        with self.app.app_context():
+            mock_event_producer = Mock()
+            handle_message(json.dumps(message), mock_event_producer)
+            json.loads(mock_event_producer.write_event.call_args[0][0])
 
-        original_id = created_host["id"]
+        # attempt to update
+        self.post(HOST_URL, [host_data.data()], 207)
```
Glutexo, line 919: This does not try to change the tags, so we can’t distinguish the correct behavior (nothing changes) with the original one (tags are updated). You have to post a different tags and then check that even after the update the hosts still have the original set of tags. To achieve that it’d be also good to verify that the response is _200_ (updated) and not _201_ (created).

```suggestion
        update_host_data = HostWrapper(
            test_data(tags=[{"namespace": "other_ns", "key": "other_key", "value": "other_val"}], fqdn="fqdn")
        )
        update_response = self.post(HOST_URL, [update_host_data.data()], 207)
        self._verify_host_status(update_response, 0, 200)
```

It’s possible to get the host ID from here too.

```python
        updated_host = self._pluck_host_from_response(update_response, 0)
        host_id = updated_host["id"]
```

test_api.py
```diff
@@ -880,126 +882,56 @@ def test_create_host_with_invalid_ansible_host(self):
...
+                "facts": host_data.facts,
+                "fqdn": host_data.fqdn,
+            },
+        }
 
-        created_host = self._pluck_host_from_response(response, 0)
+        with self.app.app_context():
+            mock_event_producer = Mock()
+            handle_message(json.dumps(message), mock_event_producer)
+            json.loads(mock_event_producer.write_event.call_args[0][0])
```
Glutexo, line 916: The message is parsed, but nothing is done with it. Here we need to read the created host ID. It’s not _host_data.id_, that field is ignored, a new one is assigned on database insert.

```suggestion
            event = json.loads(mock_event_producer.write_event.call_args[0][0])
            host_id = event["host"]["id"]
```

test_api.py
```diff
@@ -880,126 +882,56 @@ def test_create_host_with_invalid_ansible_host(self):
...
 
-        original_id = created_host["id"]
+        # attempt to update
+        self.post(HOST_URL, [host_data.data()], 207)
 
-        host_lookup_results = self.get(f"{HOST_URL}/{original_id}", 200)
+        tags_after_update = self.get(f"{HOST_URL}/{host_data.id}/tags")
```
Glutexo, line 921:
```suggestion
        tags_response = self.get(f"{HOST_URL}/{host_id}/tags")
```

test_api.py
```diff
@@ -880,126 +882,56 @@ def test_create_host_with_invalid_ansible_host(self):
...
+        host_id = create_response["data"][0]["host"]["id"]
+        response = self.get(f"{HOST_URL}/{host_id}/tags")
 
-        host_data = HostWrapper(test_data(system_profile=system_profile))
+        self.maxDiff = None
+        self.assertEqual(response["results"][host_id], [])
 
-        response = self.post(HOST_URL, [host_data.data()], 207)
+    def test_update_host_with_tags_doesnt_change_tags(self):
+        host_data = HostWrapper(
```
Glutexo, line 895:
```suggestion
        create_host_data = HostWrapper(
```

test_api.py
```diff
@@ -880,126 +882,56 @@ def test_create_host_with_invalid_ansible_host(self):
...
+            "operation": "add_host",
+            "data": {
+                "id": host_data.id,
+                "account": host_data.account,
+                "display_name": host_data.display_name,
+                "tags": host_data.tags,
+                "reporter": host_data.reporter,
+                "stale_timestamp": host_data.stale_timestamp,
+                "facts": host_data.facts,
+                "fqdn": host_data.fqdn,
```
Glutexo, line 909:
```suggestion
                "account": create_host_data.account,
                "display_name": create_host_data.display_name,
                "tags": create_host_data.tags,
                "reporter": create_host_data.reporter,
                "stale_timestamp": create_host_data.stale_timestamp,
                "facts": create_host_data.facts,
                "fqdn": create_host_data.fqdn,
```

test_api.py
```diff
@@ -880,126 +882,56 @@ def test_create_host_with_invalid_ansible_host(self):
...
+            },
```
Glutexo, line 910:
```suggestion
            "data": create_host_data.data(),
```

test_api.py
```diff
@@ -880,126 +882,56 @@ def test_create_host_with_invalid_ansible_host(self):
...
 
-        host_lookup_results = self.get(f"{HOST_URL}/{original_id}", 200)
+        tags_after_update = self.get(f"{HOST_URL}/{host_data.id}/tags")
 
-        self._validate_host(host_lookup_results["results"][0], host_data, expected_id=original_id)
+        # check the tags haven't updated
+        self.assertNotEqual(host_data.tags, tags_after_update)
 
-    def test_create_host_with_too_long_MAC_address(self):
+    def test_create_host_with_20_byte_MAC_address(self):
```
Glutexo, line 926:
```suggestion
    def test_create_host_with_20_byte_mac_address(self):
```

test_api.py
```diff
@@ -880,126 +882,56 @@ def test_create_host_with_invalid_ansible_host(self):
...
+        create_response = self.post(HOST_URL, [host_data.data()], 207)
+        host_id = create_response["data"][0]["host"]["id"]
+        response = self.get(f"{HOST_URL}/{host_id}/tags")
 
-        host_data = HostWrapper(test_data(system_profile=system_profile))
+        self.maxDiff = None
+        self.assertEqual(response["results"][host_id], [])
 
-        response = self.post(HOST_URL, [host_data.data()], 207)
+    def test_update_host_with_tags_doesnt_change_tags(self):
```
Glutexo, line 894: ### Important!

To verify this test, change _deserialize_host_http_ to _deserialize_host_mq_ in [_api.host.add_host_list_](https://github.com/RedHatInsights/insights-host-inventory/blob/9bf13d684743fd676f0093c113627f8748004f33/api/host.py#L76). In that state the tags get updated and the test fails. If not, the test is not testing correctly.

ryandillinfelton: Check it out, and the new version fails when I which it to mq deserialization. Thanks for all the help

test_db_model.py
```diff
@@ -175,7 +176,7 @@ def test_host_schema_valid_tags(tags):
 def test_host_schema_invalid_tags(tags):
     host = {"fqdn": "fred.flintstone.com", "display_name": "display_name", "account": "00102", "tags": tags}
```
Glutexo, line 177:
```suggestion
    host = {
        "fqdn": "fred.flintstone.com",
        "display_name": "display_name",
        "account": "00102",
        "tags": tags,
        "stale_timestamp": "2019-12-16T10:10:06.754201+00:00",
        "reporter": "test",
    }
```

test_db_model.py
```diff
@@ -189,7 +190,7 @@ def test_host_schema_timezone_enforced():
         "reporter": "test",
     }
     with pytest.raises(ValidationError) as excinfo:
-        _ = HostSchema(strict=True).load(host)
+        _ = HttpHostSchema(strict=True).load(host)
```
Glutexo, line 193: Here, we can use _MqHostSchema_ (as the primary one) too.

```suggestion
        _ = MqHostSchema(strict=True).load(host)
```

test_db_model.py
```diff
@@ -175,7 +176,7 @@ def test_host_schema_valid_tags(tags):
 def test_host_schema_invalid_tags(tags):
```
Glutexo, line 176: It may be good to add a new test that verifies that _HttpHostSchema_ ignores the tags regardless on their validity.

```python
@pytest.mark.parametrize("tags", [[{"key": "good_tag"}], [{"value": "bad_tag"}]])
def test_host_schema_ignored_tags(tags):
    host = {
        "fqdn": "fred.flintstone.com",
        "display_name": "display_name",
        "account": "00102",
        "tags": tags,
        "stale_timestamp": "2019-12-16T10:10:06.754201+00:00",
        "reporter": "test",
    }

    validated_host = HttpHostSchema(strict=True).load(host)
    assert "tags" not in validated_host.data
```

test_unit.py
```diff
@@ -745,7 +747,7 @@ def test_with_all_fields(
         }
         host_schema.return_value.load.return_value.data = host_input
 
-        result = deserialize_host(host_input)
+        result = deserialize_host(input, host_schema)
```
Glutexo, line 750: Here again, the variable name is _host_input_. _input_ is a Python built-in. It works only because we are mocking the schema and it doesn’t 💥  on invalid type.

In the other tests, we are passing in just `{}` to make it evident that the input value is ignored. Doing that it’d be at least consistent.

```suggestion
        result = deserialize_host({}, host_schema)
```

test_unit.py
```diff
@@ -683,17 +685,17 @@ def test_with_invalid_input(self):
...
+                expected_errors = HttpHostSchema().load(input).errors
                 self.assertEqual(str(expected_errors), str(context.exception))
 
 
 @patch("app.serialization.Host")
 @patch("app.serialization._deserialize_tags")
 @patch("app.serialization._deserialize_facts")
 @patch("app.serialization._deserialize_canonical_facts")
-@patch("app.serialization.HostSchema")
+@patch("app.serialization.HttpHostSchema")
```
Glutexo, line 698: We don’t need to patch the schema, when it’s now passed in as an argument.

```suggestion
```

If you remove it, just don’t forget to remove it from the test method definitions:

```python
    def test_with_all_fields(self, deserialize_canonical_facts, deserialize_facts, deserialize_tags, host):
```

test_unit.py
```diff
@@ -745,7 +747,7 @@ def test_with_all_fields(
         }
         host_schema.return_value.load.return_value.data = host_input
```
Glutexo, line 748:
```suggestion
        host_schema = Mock(**{"return_value.load.return_value.data": host_input})
```

test_unit.py
```diff
@@ -783,7 +785,7 @@ def test_without_facts(self, host_schema, deserialize_canonical_facts, deseriali
...
```
Glutexo, line 786:
```suggestion
        host_schema = Mock(**{"return_value.load.return_value.data": host_input})
```

test_unit.py
```diff
@@ -821,7 +823,7 @@ def test_without_tags(self, host_schema, deserialize_canonical_facts, deserializ
...
```
Glutexo, line 824:
```suggestion
        host_schema = Mock(**{"return_value.load.return_value.data": host_input})
```

test_unit.py
```diff
@@ -864,7 +866,7 @@ def test_without_display_name(
...
```
Glutexo, line 867:
```suggestion
        host_schema = Mock(**{"return_value.load.return_value.data": host_input})
```

test_unit.py
```diff
@@ -902,7 +904,7 @@ def test_without_system_profile(
         }
         host_schema.return_value.load.return_value.data = input
```
Glutexo, line 905: Here, the variable name really is _input_. It works, but it is a bad practice. (And it’s probably my fault it’s like that. 😞) I’d like to ask you for a favor of renaming those to _host_input_ throughout the whole file. Thanks!

```suggestion
        host_schema = Mock(**{"return_value.load.return_value.data": host_input})
```

test_unit.py
```diff
@@ -951,7 +953,7 @@ def test_without_culling_fields(
                 host_schema.return_value.load.return_value.data = all_data
```
Glutexo, line 953:
```suggestion
                host_schema = Mock(**{"return_value.load.return_value.data": all_data)
```

test_unit.py
```diff
@@ -963,7 +965,7 @@ def test_host_validation(
     ):
         host_input = {"ansible_host": "some ansible host", "account": "some acct"}
 
-        deserialize_host(host_input)
+        deserialize_host(host_input, host_schema)
```
Glutexo, line 968:
```suggestion
        from unittest.mock import MagicMock
        host_schema = MagicMock()
        deserialize_host(host_input, host_schema)
```

_MagicMock_ instead of just _Mock_, because it supports magic methods like _\_\_getitem\_\__. Item access is used inside of the deserialization.

test_unit.py
```diff
@@ -976,7 +978,7 @@ def test_invalid_host_error(
         host_schema.return_value.load.side_effect = caught_exception
```
Glutexo, line 978:
```suggestion
        host_schema = Mock(**{"return_value.load.side_effect": caught_exception})
```

test_unit.py
```diff
@@ -641,61 +641,173 @@ def test_with_only_required_fields(self):
...
         for inp in inputs:
             with self.subTest(input=inp):
                 with self.assertRaises(ValidationException) as context:
-                    deserialize_host(input, HttpHostSchema)
+                    deserialize_host(inp, MqHostSchema)
+
+                    expected_errors = MqHostSchema().load(inp).errors
+                    self.assertEqual(str(expected_errors), str(context.exception))
+
+    # test that both of the host schamas will pass all of these fields
```
Glutexo, line 720:
```suggestion
    # Test that both of the host schemas will pass all of these fields
```

test_unit.py
```diff
@@ -960,22 +1063,21 @@ def test_without_culling_fields(
                 deserialize_tags.assert_called_once_with(common_data["tags"])
                 host.assert_not_called()
 
-    def test_host_validation(
-        self, host_schema, deserialize_canonical_facts, deserialize_facts, deserialize_tags, host
-    ):
+    def test_host_validation(self, deserialize_canonical_facts, deserialize_facts, deserialize_tags, host):
         host_input = {"ansible_host": "some ansible host", "account": "some acct"}
 
+        from unittest.mock import MagicMock
```
Glutexo, line 1069: Imports should be all gathered at the top of the file.

test_api.py
```diff
@@ -2166,14 +1987,21 @@ def _assert_host_ids_in_response(self, response, expected_hosts):
...
     def _base_query_test(self, host_id_list, expected_host_list):
         url = f"{HOST_URL}/{host_id_list}"
         response = self.get(url)
 
         self.assertEqual(response["count"], len(expected_host_list))
         self.assertEqual(len(response["results"]), len(expected_host_list))
 
-        host_data = [host.data() for host in expected_host_list]
+        host_data = [host for host in self._expected_host_list(expected_host_list)]
         for host in host_data:
```
Glutexo, line 2005: Both iterations and comparisons can be replaces by a single smart assertion.

```python
        self.assertCountEqual(host_data, response["results"])
```

Glutexo:
Pushed a couple of tiny fixes. Now it’s ✅. One more? @shdunning @tahmidefaz @jharting 

LGTM

PR #649, opened 2020-04-30
Add env var for creating log group, default to False

bsquizz:
In AppSRE configurations, apps do not have permissions to create the log group in cloudwatch. It is already created for them. This PR adds an env var (default: False) that allows the `create_log_group` option to be set to True or False on the `CloudWatchLogHandler`

### DIFF
--- a/app/logging.py
+++ b/app/logging.py
@@ -43,6 +43,7 @@ def _configure_watchtower_logging_handler():
     aws_region_name = os.getenv("AWS_REGION_NAME", None)
     log_group = os.getenv("AWS_LOG_GROUP", "platform")
     stream_name = os.getenv("AWS_LOG_STREAM", _get_hostname())  # default to hostname
+    create_log_group = str(os.getenv("AWS_CREATE_LOG_GROUP")) in ["True", "true"]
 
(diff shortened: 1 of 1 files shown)


app/logging.py
```diff
@@ -54,7 +55,10 @@ def _configure_watchtower_logging_handler():
 
         root = logging.getLogger()
         handler = watchtower.CloudWatchLogHandler(
-            boto3_session=boto3_session, log_group=log_group, stream_name=stream_name
+            boto3_session=boto3_session,
+            log_group=log_group,
+            stream_name=stream_name,
+            create_log_group=create_log_group,
```
shdunning, line 61: According to the docs, omitting `create_log_group` will default the value to `True` which is the opposite of what your change does (makes default value `False`):
 https://github.com/kislyuk/watchtower/blob/master/watchtower/__init__.py#L75-L76

Do we know if there are any unintended side effects from this change?

bsquizz: Yeah, this is the change we need, we no longer can have it default to "True" because apps don't have permission to create the log group in App-SRE

Since the 'platform'/'platform-dev' log group already exists in our current environments, we don't need the flag set to true in our current environments either.

PR #650, opened 2020-04-30
RHCLOUD-4283 - Enabling stdout logs

tahmidefaz:


tahmidefaz:
Is this solution any good? Tried hard not to use `global` variables. If you can think of a more elegant solution, let me know.
You can check out the `platform-dev` log group's `test_logs_inv` stream to check out the test logs I produced (towards the end).

app/logging.py
```diff
@@ -52,16 +54,23 @@ def _configure_watchtower_logging_handler():
...
+def _cloudwatch_handler():
+    global session_boto3
+    group = os.getenv("AWS_LOG_GROUP", "platform")
+    stream = os.getenv("AWS_LOG_STREAM", _get_hostname())
+
+    if session_boto3 is not None:
+        handler = watchtower.CloudWatchLogHandler(boto3_session=session_boto3, log_group=group, stream_name=stream)
+        handler.setFormatter(logstash_formatter.LogstashFormatterV1())
+        return handler
+
```
shdunning, line 72: does python have an implicit `return None` here? i forget

app/logging.py
```diff
@@ -118,5 +127,9 @@ def get_logger(name):
     log_level = os.getenv("INVENTORY_LOG_LEVEL", "INFO").upper()
     logger = logging.getLogger(LOGGER_PREFIX + name)
     logger.addFilter(ContextualFilter())
+
+    if _cloudwatch_handler() is not None:
```
shdunning, line 131: if you make `if session_boto3 is not None:` then 2 things happen:
1. you don't need to initialize a `CloudWatchLogHandler` more than once
2. you can simplify your `_cloudwatch_handler()` function into a getter:

```
def _get_cloudwatch_handler():
    global session_boto3
    group = os.getenv("AWS_LOG_GROUP", "platform")
    stream = os.getenv("AWS_LOG_STREAM", _get_hostname())
    handler = watchtower.CloudWatchLogHandler(boto3_session=session_boto3, log_group=group, stream_name=stream)
    handler.setFormatter(logstash_formatter.LogstashFormatterV1())
    return handler
```

app/logging.py
tahmidefaz: Yes, if a function does not return anything, the function call is basically a `None`

app/logging.py
tahmidefaz: done! :+1:

tahmidefaz:
@jharting just set the env var to use `logconfig.ini` as default. :+1:

app/logging.py
```diff
@@ -53,19 +54,24 @@ def _configure_watchtower_logging_handler():
...
-        )
-        handler.setFormatter(logstash_formatter.LogstashFormatterV1())
-        root.addHandler(handler)
+        global session_boto3
+        session_boto3 = boto3_session
     else:
         print("Unable to configure watchtower logging.  Please verify watchtower logging configuration!")
 
 
+def _get_cloudwatch_handler():
```
Glutexo, line 63: I wonder, do we need to create a new handler every time? Can we reuse the whole handler instead of reusing just the session?

app/logging.py
```diff
@@ -13,10 +13,12 @@
 
 threadctx = local()
 
+session_boto3 = None
+
```
Glutexo, line 17:
```suggestion
cloudwatch_handler = None
```

app/logging.py
```diff
@@ -43,7 +45,6 @@ def _configure_watchtower_logging_handler():
     aws_region_name = os.getenv("AWS_REGION_NAME", None)
     log_group = os.getenv("AWS_LOG_GROUP", "platform")
```
Glutexo, line 46:
```suggestion
```

This variable is not in the condition few lines below.

app/logging.py
```diff
@@ -43,7 +45,6 @@ def _configure_watchtower_logging_handler():
...
     stream_name = os.getenv("AWS_LOG_STREAM", _get_hostname())  # default to hostname
-    create_log_group = str(os.getenv("AWS_CREATE_LOG_GROUP")).lower() == "true"
 
     if all([aws_access_key_id, aws_secret_access_key, aws_region_name, stream_name]):
         print(f"Configuring watchtower logging (log_group={log_group}, stream_name={stream_name})")
```
Glutexo, line 50:
```python
        aws_log_group = os.getenv("AWS_LOG_GROUP", "platform")
        create_log_group = str(os.getenv("AWS_CREATE_LOG_GROUP")).lower() == "true"

        global cloudwatch_handler
        cloudwatch_handler = _get_cloudwatch_handler(
            aws_access_key_id, aws_secret_access_key, aws_region_name, aws_log_group, aws_stream_name, create_log_group
        )
```

Let’s create a whole handler instead of just the session. The whole handler can be reused.

app/logging.py
```diff
@@ -53,19 +54,24 @@ def _configure_watchtower_logging_handler():
...
 
 
+def _get_cloudwatch_handler():
+    global session_boto3
+    group = os.getenv("AWS_LOG_GROUP", "platform")
+    stream = os.getenv("AWS_LOG_STREAM", _get_hostname())
+    create_log_group = str(os.getenv("AWS_CREATE_LOG_GROUP")).lower() == "true"
+    handler = watchtower.CloudWatchLogHandler(
+        boto3_session=session_boto3, log_group=group, stream_name=stream, create_log_group=create_log_group
+    )
```
Glutexo, line 70: Let’s keep environment variable retrieval in one place, the closest to the module entry point as possible.

```suggestion
def _get_cloudwatch_handler(
    aws_access_key_id, aws_secret_access_key, aws_region_name, aws_log_group, aws_stream_name, create_log_group
):
    print(f"Configuring watchtower logging (log_group={aws_log_group}, stream_name={aws_stream_name})")
    boto3_session = Session(
        aws_access_key_id=aws_access_key_id,
        aws_secret_access_key=aws_secret_access_key,
        region_name=aws_region_name,
    )

    handler = watchtower.CloudWatchLogHandler(
        boto3_session=boto3_session,
        log_group=aws_log_group,
        stream_name=aws_stream_name,
        create_log_group=create_log_group
    )
```

app/logging.py
```diff
@@ -122,5 +128,10 @@ def get_logger(name):
     log_level = os.getenv("INVENTORY_LOG_LEVEL", "INFO").upper()
     logger = logging.getLogger(LOGGER_PREFIX + name)
     logger.addFilter(ContextualFilter())
+
+    global session_boto3
+    if session_boto3 is not None:
+        logger.addHandler(_get_cloudwatch_handler())
```
Glutexo, line 134:
```suggestion
    global cloudwatch_handler
    if cloudwatch_handler:
        logger.addHandler(cloudwatch_handler)
```

Glutexo:
Okay. Hopefully found a way how to do it.

app/logging.py
Glutexo, line 17:
```suggestion
```

app/logging.py
```diff
@@ -43,7 +45,6 @@ def _configure_watchtower_logging_handler():
     aws_region_name = os.getenv("AWS_REGION_NAME", None)
     log_group = os.getenv("AWS_LOG_GROUP", "platform")
     stream_name = os.getenv("AWS_LOG_STREAM", _get_hostname())  # default to hostname
```
Glutexo, line 47:
```suggestion
    aws_stream_name = os.getenv("AWS_LOG_STREAM", _get_hostname())  # default to hostname
```

app/logging.py
```diff
@@ -43,7 +45,6 @@ def _configure_watchtower_logging_handler():
...
-    create_log_group = str(os.getenv("AWS_CREATE_LOG_GROUP")).lower() == "true"
 
     if all([aws_access_key_id, aws_secret_access_key, aws_region_name, stream_name]):
```
Glutexo, line 49:
```suggestion
    if all([aws_access_key_id, aws_secret_access_key, aws_region_name, aws_log_group, aws_stream_name]):
```

app/logging.py
Glutexo: Function definition
```python
def _configure_watchtower_logging_handler(logger):
```

The same for __configure_contextual_logging_filter_.
```python
def _configure_contextual_logging_filter(logger):
```

```suggestion
    aws_log_group = os.getenv("AWS_LOG_GROUP", "platform")
```

app/logging.py
```diff
@@ -53,19 +54,24 @@ def _configure_watchtower_logging_handler():
...
-        handler = watchtower.CloudWatchLogHandler(
-            boto3_session=boto3_session,
-            log_group=log_group,
-            stream_name=stream_name,
-            create_log_group=create_log_group,
-        )
-        handler.setFormatter(logstash_formatter.LogstashFormatterV1())
-        root.addHandler(handler)
+        global session_boto3
+        session_boto3 = boto3_session
```
Glutexo, line 58:
```suggestion
handler = _get_cloudwatch_handler(aws_access_key_id, aws_secret_access_key, aws_region_name, aws_log_group, aws_stream_name, create_log_group)
logger = logging.addHandler(handler)
```

app/logging.py
Glutexo, line 63:
```suggestion
def _get_cloudwatch_handler(aws_access_key_id, aws_secret_access_key, aws_region_name, aws_log_group, aws_stream_name, create_log_group):
```

app/logging.py
```diff
@@ -53,19 +54,24 @@ def _configure_watchtower_logging_handler():
...
     else:
         print("Unable to configure watchtower logging.  Please verify watchtower logging configuration!")
 
 
+def _get_cloudwatch_handler():
+    global session_boto3
```
Glutexo, line 64:
```suggestion
```

app/logging.py
```diff
@@ -53,19 +54,24 @@ def _configure_watchtower_logging_handler():
...
+    group = os.getenv("AWS_LOG_GROUP", "platform")
+    stream = os.getenv("AWS_LOG_STREAM", _get_hostname())
+    create_log_group = str(os.getenv("AWS_CREATE_LOG_GROUP")).lower() == "true"
```
Glutexo, line 67:
```suggestion
```

app/logging.py
```diff
@@ -13,10 +13,12 @@
...
+session_boto3 = None
+
 
 def configure_logging(config_name):
     env_var_name = "INVENTORY_LOGGING_CONFIG_FILE"
-    log_config_file = os.getenv(env_var_name)
+    log_config_file = os.getenv(env_var_name, "logconfig.ini")
     if log_config_file is not None:
         # The logging module throws an odd error (KeyError) if the
         # config file is not found.  Hopefully, this makes it more clear.
```
Glutexo, line 24: On top of the file.
```python
LOGGER_NAME = "inventory"
```

Later in this function.
```python
    logger = getLogger(LOGGER_NAME)
    log_level = os.getenv("INVENTORY_LOG_LEVEL", "INFO").upper()
    logger.setLevel(log_level)

    if config_name != "testing":
        _configure_watchtower_logging_handler(logger)
        _configure_contextual_logging_filter(logger)
```

app/logging.py
```diff
@@ -122,5 +128,10 @@ def get_logger(name):
     log_level = os.getenv("INVENTORY_LOG_LEVEL", "INFO").upper()
     logger = logging.getLogger(LOGGER_PREFIX + name)
     logger.addFilter(ContextualFilter())
+
+    global session_boto3
+    if session_boto3 is not None:
+        logger.addHandler(_get_cloudwatch_handler())
+
     logger.setLevel(log_level)
     return logger
```
Glutexo, line 137: return logging.getLogger(f"{LOGGER_NAME}.{name}")

Glutexo:
There is now a conflict, caused by my #657 being merged in. It shouldn’t affect your code though.

tahmidefaz:
@Glutexo I think I took your suggestions into account, but sadly it does not send any logs to cloudwatch. I will keep digging, but let me know if you see something I did that diverges from the changes you suggested.

tahmidefaz:
Also, regarding changing _logconfig.ini_ did you meant taking out app, api, tasks, mq_service, lib, utils out? and replace it with only one `logger_inventory` with `qualname=inventory`? @Glutexo

tahmidefaz:
Ok, fixed it. :sweat: 
We def need that global `cloudwatch_handler`, didn't find a way to make it work without it. This solution diverges slightly from @Glutexo 's proposal. Lmk, if there are any concerns.

app/logging.py
tahmidefaz: @Glutexo this stops us from sending cloudwatch logs, no idea why. We need to have the `cloudwatch_handler` as a global variable and add it here rather than in `configure_logging()`, or at least that's the only way I got it to work again :man_shrugging:

app/logging.py
tahmidefaz: see my comment below

app/logging.py
```diff
@@ -9,14 +9,17 @@
...
 threadctx = local()
 
+cloudwatch_handler = None
 
-def configure_logging(runtime_environment):
+
+def configure_logging(config_name):
     env_var_name = "INVENTORY_LOGGING_CONFIG_FILE"
-    log_config_file = os.getenv(env_var_name)
+    log_config_file = os.getenv(env_var_name, "logconfig.ini")
```
shdunning, line 22: where does `logconfig.ini` get created?

tahmidefaz: It's provided with the codebase: https://github.com/RedHatInsights/insights-host-inventory/blob/master/logconfig.ini

app/logging.py
Glutexo: Found out why. It’s because many Inventory modules are configured in [_logconfig.ini_](https://github.com/RedHatInsights/insights-host-inventory/blob/188767e2f1724c49a30ca99d6a3581f3b1137e92/logconfig.ini) not to propagate. They then just don’t pass their messages to the _inventory_ logger that has the CloudWatch handler attached.

Even though explicitly adding the handler to every logger is a working solution. We can just simply remove the faulty configuration and everything will start working automatically.

app/logging.py
```diff
@@ -32,45 +35,55 @@ def configure_logging(runtime_environment):
...
         logging.config.fileConfig(fname=log_config_file)
 
-    if runtime_environment.logging_enabled:
-        _configure_watchtower_logging_handler()
-        _configure_contextual_logging_filter()
+    logger = logging.getLogger(LOGGER_NAME)
+    log_level = os.getenv("INVENTORY_LOG_LEVEL", "INFO").upper()
+    logger.setLevel(log_level)
+
+    if config_name != "testing":
```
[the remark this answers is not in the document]
tahmidefaz: @Glutexo I got confused by one of your earlier suggestions here :sweat_smile:

tahmidefaz:
@Glutexo it's working now that propagation issue in the `logconfig.ini`was the main culprit :tada: :tada: 

Also, this is ready for a final review now

Glutexo:
Looking into this.

Glutexo:
I found the culprit, why contextual information is not attached to the logs anymore. The reason is that the contextual filter is attached to the _inventory_ logger and filters are not propagated to children loggers.

It can be however attached to a handler instead and because handers are propagated, so are filters attached to them. But there is still one more problem:

- The Logstash handler that logs to the standard output and Kibana is instantiated in _logconfig.ini_.
- The legacy _fileConfig_ function that loads logging configuration does not support filters.

I see some possible solutions. They need more exploration though.

- Get the Logstash handler instance created by the file config somehow. Then, the filter can be attached to it. Is it even possible?
- Do not create the Logstash handler in the file config, but in the code, like the CloudWatch one. But is it possible to define loggers without explicit handlers?
- Switch to more modern logging configuration format that supports filters. _dictConfig_ may work here.
- Instantiate the filter in the code and then attach the instance to every logger created. This is just as ugly as it was with the CloudWatch handler.

Glutexo:
@jharting What do we need to override in the logging configuration in production? Is it only the log level, or are there other use cases? Asking what we would lose if we ended up entirely ditching the file configuration.

Also one more note. Moving from _fileConfig_ to other means of configuration would mean we couldn’t use Gunicorn’s _--log-config_ argument any more. Which I wanted to instead of loading it manually. But it’s possible it supports other formats and I just don’t know it yet.

Glutexo:
From today’s discussion with @jharting. When attaching a filter to handlers, make sure that the values are in there in the thread context. Like, the handlers may use some threading for asynchronousness. I think this won’t be a problem, because filters are applied before the log record even gets to the handler, but I’m not 100% here and it should be verified.

Glutexo:
Logger has _handlers_ property, but it’s not documented. Thus I’d assume it’s not safe to use. It looks like there is no way to get an existing logger instance. Also there is _logger.manager.loggerDict_, which, again, is undocumented. Sigh. 😤 So it looks like that the first point – get the instance – is no go.

Glutexo:
It is not possible to define a logger without a handler using _fileConfig_. So, it looks like we’re doing _dictConfig_. Also, the “problem” with disabling Gunicorn’s own logger applies to _dictConfig_ too. See https://github.com/benoitc/gunicorn/issues/1051. I don’t know yet, whether it’s possible to instruct Gunicorn to use a dictionary (YAML/JSON) configuration.

jharting:
@Glutexo the use-case I am aware of is to separately configure the log level of a handler (e.g. different log level for cloudwatch than for std out). Otherwise I don't think any "dynamic" overrides are needed.

We'll also need a way to silence e.g. a noisy lib (i.e. for certain loggers set the threshold to WARN even though we're using INFO globally) but that can be statically defined either in a file or some conf dictionary.

Glutexo:
@jharting I see. A configuration file allows to silence a library or to lower the log level immediately. Having that in a code would require to do a full deploy to change a setting. It looks like the former is more favorable.

Glutexo:
It is possible to instantiate a filter using _dictConfig_ and attach it to a handler. I’m pretty positive that it’s even possible to instantiate the CloudWatch handler in there. Because it doesn’t use the application config object, but directly accesses the environment variables, there is no need for any hacks with global variables.

Glutexo:
Feel free to take a look at my fork. https://github.com/Glutexo/insights-host-inventory/tree/stdout-logging I’m going to wrap this up so @tahmidefaz can use it.

tahmidefaz:
Thank you! I am currently focusing on one of the other issues that are on my plate. I plan on getting back to this sometime today.

tahmidefaz:
I can confirm stdout log works with the addition of the `logconfig.yaml`. Researching what can be done for the cloudwatch handler. It seems to be always missing a positional argument in formatters and filters, but obviously the problem lies somewhere else.

Glutexo:
Investigating the error. I already had this working.

```
ValueError: Unable to configure handler 'cloudwatch': setFormatter() missing 1 required positional argument: 'fmt'
```

Glutexo:
Okay, this was stupid. In case of disabled CloudWatch, I was returning a _NullHandler_ class instead of an instance. 🤦🏻‍♂️ https://github.com/RedHatInsights/insights-host-inventory/blob/3f29d5f14f9b1ad7a1fa6dc165186e9d14a4cf8f/app/logging.py#L56

tahmidefaz:
Oops! Didn't realize that was causing the issue :disappointed:

tahmidefaz:
@Glutexo In `configure_logging()`, we can just check whether the `cloudwatch` credentials were provided, and if not we can simply remove that info from the dictionary before loading it via `dictConfig`. What do you think?

tahmidefaz:
^ This simply removes the cloudwatch handler info from the dictionary if creds aren't present. Let me know if you can think of a better way to do this.

Glutexo:
@tahmidefaz What’s wrong with the _NullHandler_? 😦 Now when we now we need to return an instance instead of a class.

tahmidefaz:
@Glutexo sorry, I misunderstood. I will revert :man_facepalming:

app/logging.py
```diff
@@ -6,74 +6,58 @@
...
-            create_log_group=create_log_group,
-        )
-        handler.setFormatter(logstash_formatter.LogstashFormatterV1())
-        root.addHandler(handler)
-    else:
-        print("Unable to configure watchtower logging.  Please verify watchtower logging configuration!")
+    log_config_file = os.getenv("INVENTORY_LOGGING_CONFIG_FILE", DEFAULT_LOGGING_CONFIG_FILE)
+
+    with open(log_config_file) as log_config_file:
+        logconfig_dict = safe_load(log_config_file)
```
Glutexo, line 23: This is actually the only line that needs to be under the _open_ context manager. Once _logconfig_dict_ is loaded into the memory, the file can be safely closed.

app/logging.py
```diff
@@ -6,74 +6,58 @@
...
+        if all((AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION_NAME)):
```
Glutexo: Let’s move this to the CloudWatch handler constructor and return _NullHandler_ if credentials are not provided.

app/logging.py
```diff
@@ -6,74 +6,58 @@
...
+            logging.config.dictConfig(logconfig_dict)
+        else:
+            print("Unable to configure watchtower logging.  Please verify watchtower logging configuration!")
+            logconfig_dict["loggers"]["inventory"]["handlers"].remove("cloudwatch")
```
Glutexo: This mangling with the configuration feels no better than the original solution. Back then, some configuration was there in the config file while other parts were in the code. This does the same, just in the opposite direction. Let’s solve this by the _NullHandler_ instead.

app/logging.py
```diff
@@ -6,74 +6,58 @@
...
+            logconfig_dict["handlers"].pop("cloudwatch", None)
+            logging.config.dictConfig(logconfig_dict)
```
Glutexo:
```suggestion
    logging.config.dictConfig(logconfig_dict)
```

app/logging.py
```diff
@@ -1,79 +1,63 @@
...
-                f"Error reading the logging configuration file.  Verify the {env_var_name} environment variable is "
-                "set correctly. Aborting..."
-            )
-            raise
-
-        logging.config.fileConfig(fname=log_config_file)
+    log_config_file = os.getenv("INVENTORY_LOGGING_CONFIG_FILE", DEFAULT_LOGGING_CONFIG_FILE)
+    with open(log_config_file) as log_config_file:
+        logconfig_dict = safe_load(log_config_file)
+        logging.config.dictConfig(logconfig_dict)
```
Glutexo, line 24:
```suggestion
    logging.config.dictConfig(logconfig_dict)
```

app/logging.py
```diff
@@ -1,79 +1,63 @@
...
 OPENSHIFT_ENVIRONMENT_NAME_FILE = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
 DEFAULT_AWS_LOGGING_NAMESPACE = "inventory-dev"
-LOGGER_PREFIX = "inventory."
+DEFAULT_LOGGING_CONFIG_FILE = "logconfig.yaml"
+LOGGER_NAME = "inventory"
 
 threadctx = local()
 
 
 def configure_logging(runtime_environment):
```
Glutexo, line 20: The _runtime_envirnoment_ argument is no longer use. Let’s remove it from here and from all the calls.

```suggestion
def configure_logging():
```

app/logging.py
```diff
@@ -1,79 +1,63 @@
...
-    create_log_group = str(os.getenv("AWS_CREATE_LOG_GROUP")).lower() == "true"
 
-    if all([aws_access_key_id, aws_secret_access_key, aws_region_name, stream_name]):
-        print(f"Configuring watchtower logging (log_group={log_group}, stream_name={stream_name})")
+    logger = get_logger(__name__)
+    if all((aws_access_key_id, aws_secret_access_key, aws_region_name)):
+        aws_log_group = os.getenv("AWS_LOG_GROUP", "platform")
+        aws_log_stream = os.getenv("AWS_LOG_STREAM", _get_hostname())
+        create_log_group = str(os.getenv("AWS_CREATE_LOG_GROUP")).lower() == "true"
+        logger.info(f"Configuring watchtower logging (log_group={aws_log_group}, stream_name={aws_log_stream})")
```
Glutexo, line 41: This unfortunately doesn’t work properly. When this is encountered, logging is not configured yet. Prints are not ideal, but work just as fine. Let’s keep those for now and fix it once we figure out how to do the logging configuration earlier.

Moreover if this worked, the message would be logged twice. Once by this logging statement and the second time by the _print_ below.

app/logging.py
```diff
@@ -1,79 +1,63 @@
...
     aws_access_key_id = os.getenv("AWS_ACCESS_KEY_ID", None)
     aws_secret_access_key = os.getenv("AWS_SECRET_ACCESS_KEY", None)
     aws_region_name = os.getenv("AWS_REGION_NAME", None)
-    log_group = os.getenv("AWS_LOG_GROUP", "platform")
-    stream_name = os.getenv("AWS_LOG_STREAM", _get_hostname())  # default to hostname
-    create_log_group = str(os.getenv("AWS_CREATE_LOG_GROUP")).lower() == "true"
 
-    if all([aws_access_key_id, aws_secret_access_key, aws_region_name, stream_name]):
-        print(f"Configuring watchtower logging (log_group={log_group}, stream_name={stream_name})")
+    logger = get_logger(__name__)
```
Glutexo, line 36: This won’t be needed if we remove the _info_ statement below.

app/logging.py
```diff
@@ -1,79 +1,63 @@
...
-            stream_name=stream_name,
+            log_group=aws_log_group,
+            stream_name=aws_log_stream,
             create_log_group=create_log_group,
         )
-        handler.setFormatter(logstash_formatter.LogstashFormatterV1())
-        root.addHandler(handler)
     else:
         print("Unable to configure watchtower logging.  Please verify watchtower logging configuration!")
+        return NullHandler()
```
Glutexo, line 56: 🎉

logconfig.ini
```diff
@@ -1,5 +1,6 @@
 [loggers]
 keys=root, gunicorn.error, gunicorn.access, sqlalchemy.engine, xjoin, app, api, tasks, mq_service, lib, utils
```
Glutexo: This line should not be here.

```suggestion
```

But this whole file is no longer used, so it should be removed.

logconfig.ini
```diff
@@ -1,5 +1,6 @@
...
+keys=root, gunicorn.error, gunicorn.access, sqlalchemy.engine, inventory, xjoin
```
Glutexo: This should replace the previous line, not add another one. But this whole file should be removed as it’s no longer used.

logconfig.yaml
```diff
@@ -0,0 +1,54 @@
...
+    handlers:
+    - logstash
+    propagate: true
+  sqlalchemy.engine:
+    level: WARNING
+    handlers:
+    - logstash
+    propagate: true
+  inventory:
+    level: DEBUG
```
Glutexo, line 24: This should be _INFO_.

- To match the original log level from _logconfig.ini_.
- To be immediately usable in production.

run_gunicorn.py
```diff
@@ -30,7 +30,7 @@ def run_server():
     variables.
     """
     bind = f"0.0.0.0:{LISTEN_PORT}"
-    run(("gunicorn", "--log-config=logconfig.ini", "--log-level=debug", f"--bind={bind}", "run"))
+    run(("gunicorn", "--log-config=logconfig.yaml", "--log-level=debug", f"--bind={bind}", "run"))
```
Glutexo, line 33: Does this actually work? Not only I am afraid that it doesn’t but we also found out that it’s this loading the configuration by the _--log-config_ argument that causes problems. Let’s remove it instead.

```suggestion
    run(("gunicorn", "--log-level=debug", f"--bind={bind}", "run"))
```

Glutexo:
It looks like you already fix most of my comments. We’re getting there!

run_gunicorn.py
Glutexo: Confirmed. It tries to use the _configparser_.

```
configparser.MissingSectionHeaderError: File contains no section headers.
file: 'logconfig.yaml', line: 1
'---\n'
```

Glutexo: [This](https://github.com/benoitc/gunicorn/issues/1087#issuecomment-322502603) may be a solution. We already use _gunicorn.conf.py_ in production. I think we can move the logging configuration there! Not in this PR, but it’d definitely worth exploring.

tahmidefaz:
@Glutexo sorry for the hiccup earlier, today is a day of major faceplams :man_facepalming: 
I did a revert to where things were in your branch and added the `NullHandler()`. I also removed the `logconfig.ini` file. I am going to address the other concerns you have.
I will also need to open a subsequent PR in e2e-deploy to incorporate the new `logconfig.yaml`.

Glutexo:
@tahmidefaz Do you know what needs to be done there and how? Can you post a link here once you have it?

tahmidefaz:
@Glutexo yes, I will mention this PR in e2e-deploy once I get around to it

Makefile
```diff
@@ -15,10 +15,10 @@ run_inv_web_service:
 	#
 	# KAFKA_TOPIC="platform.system-profile" KAFKA_GROUP="inventory" KAFKA_BOOTSTRAP_SERVERS="localhost:29092"
 	#
-	INVENTORY_LOGGING_CONFIG_FILE=logconfig.ini INVENTORY_LOG_LEVEL=DEBUG gunicorn -b :8080 run
+	INVENTORY_LOGGING_CONFIG_FILE=logconfig.yaml INVENTORY_LOG_LEVEL=DEBUG gunicorn -b :8080 run
```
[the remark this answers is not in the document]
tahmidefaz: I don't think we need it since we are now setting a default value for that env variable.

app/__init__.py
```diff
@@ -40,6 +38,7 @@ def create_app(runtime_environment):
     # This feels like a hack but it is needed.  The logging configuration
     # needs to be setup before the flask app is initialized.
     configure_logging(runtime_environment)
+    logger = get_logger(__name__)  # we need to call the logger after configuring it, not before
```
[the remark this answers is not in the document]
tahmidefaz: I believe the reason I moved this line around was that due to the way logging was setup earlier, some of the logs weren't being produced. However, after the logging refactoring, that seems to be fixed. I will move it back to where it was earlier. :+1:

tahmidefaz:
^ @Glutexo I think in terms of the changes in e2e-deploy, I think that should be good. Please take a look.

Glutexo:
Tests passed. @jharting, are you okay with merging?

PR #651, opened 2020-05-01
Event refactoring [RHCLOUD-5644]

ryandillinfelton:
[Jira](https://projects.engineering.redhat.com/browse/RHCLOUD-5644)

This PR aims to refactor the code surrounding event handling and emission. Read the Jira for more granular details.


ryandillinfelton:
@Glutexo @jharting 
I think this is currently in a pretty good spot but I'd appreciate a high-level review to make sure I'm on the right path. 

I'm interested in trying to completely remove the tasks subsystem since it feels pretty redundant in a lot of ways, but I worry that would be overstepping the goals of this re-factor. Let me know what you think.

Thanks

Glutexo:
I am a fan of removing that overpompously named Task subsystem. _kafka-python_ uses its own threads to perform non-blocking operations. Everything we need is to keep an instance alive and correctly synchronously flush on exit including errors.

app/queue/queue.py
```diff
@@ -145,7 +145,7 @@ def handle_message(message, event_producer):
     with PayloadTrackerContext(payload_tracker, received_status_message="message received"):
         (output_host, add_results) = add_host(validated_operation_msg["data"])
         event = build_egress_topic_event(add_results.name, output_host, platform_metadata)
-        event_producer.write_event(event, output_host["id"], message_headers(add_results.name))
+        event_producer.write_event_egress_topic(event, output_host["id"], message_headers(add_results.name))
```
Glutexo, line 148:
```suggestion
        from app.queue.event_producer import Topic
        event_producer.write_event(Topic.egress, event, output_host["id"], message_headers(add_results.name))
```

app/queue/event_producer.py
```diff
@@ -0,0 +1,40 @@
...
+
+logger = get_logger(__name__)
+
+
+class KafkaEventProducer:
+    def __init__(self, config):
+        logger.info("Starting KafkaEventProducer()")
+        self._kafka_producer = KafkaProducer(bootstrap_servers=config.bootstrap_servers)
+        self._egress_topic = config.host_egress_topic
+        self._event_topic = config.event_topic
```
Glutexo, line 14:
```suggestion
        from enum import Enum
        self.topics = {Topics.egress: config.host_egress_topic, Topics.event: config.event_topic}
```

app/queue/events.py
```diff
@@ -93,12 +50,67 @@ class HostSchema(Schema):
...
+                    "timestamp": datetime.now(timezone.utc),
+                }
+            )
+            .data
+        )
+    else:
+        raise ValueError(f"Invalid event type ({event_type})")
+
+
+@metrics.egress_event_serialization_time.time()
```
Glutexo, line 110: Using the metric is not much this module’s responsibility. Let’s use it as a context manager in the caller and get rid of these two functions.

app/queue/events.py
```diff
@@ -93,12 +50,67 @@ class HostSchema(Schema):
...
+def build_egress_topic_event(event_type, host, platform_metadata=None):
+    return _build_event(event_type, host, platform_metadata=platform_metadata)
+
+
+def build_event_topic_event(event_type, host, request_id=None):
+    return _build_event(event_type, host, request_id=request_id)
```
Glutexo, line 116:
```suggestion
```

app/queue/queue.py
```diff
@@ -145,7 +145,7 @@ def handle_message(message, event_producer):
     with PayloadTrackerContext(payload_tracker, received_status_message="message received"):
         (output_host, add_results) = add_host(validated_operation_msg["data"])
         event = build_egress_topic_event(add_results.name, output_host, platform_metadata)
```
Glutexo, line 147:
```suggestion
        with metrics.egress_event_serialization_time.time():
            event = build_event(add_results.name, output_host, platform_metadata=platform_metadata)
```

And [analogically](https://github.com/RedHatInsights/insights-host-inventory/blob/aca1a0e48d0858746369dacdacffa186f7b9f572/api/host.py#L241):

```python
    event = build_event("updated", host, request_id=threadctx.request_id)
```

app/queue/events.py
```diff
@@ -93,12 +50,67 @@ class HostSchema(Schema):
...
+    )
+
+
+def message_headers(event_type):
+    return {"event_type": event_type}
+
+
+# Event Constructors
+def _build_event(event_type, host, *, platform_metadata=None, request_id=None):
+    if event_type in ("created", "updated"):
```
Glutexo, line 93: Let’s use an _Enum_ for the event type. Then we won’t need the validation.

```python
from enum import Enum
EventTypes = Enum("created", "updated", "delete")
```

app/queue/events.py
```diff
@@ -93,12 +50,67 @@ class HostSchema(Schema):
...
+                "request_id": threadctx.request_id,
+                "type": "delete",
+            }
+        )
+        .data
+    )
+
+
+def message_headers(event_type):
+    return {"event_type": event_type}
```
Glutexo, line 88: With the _enum_:

```suggestion
    return {"event_type": event_type.name}
```

app/queue/events.py
```diff
@@ -93,12 +50,67 @@ class HostSchema(Schema):
...
+
+
+# Event Constructors
+def _build_event(event_type, host, *, platform_metadata=None, request_id=None):
```
Glutexo, line 92:
```suggestion
def build_event(event_type, host, *, platform_metadata=None, request_id=None):
```

api/host.py
```diff
@@ -244,9 +245,9 @@ def get_host_system_profile_by_id(host_id_list, page=1, per_page=100, order_by=N
 
 def _emit_patch_event(host):
     key = host["id"]
-    event = build_event_topic_event("updated", host, request_id=threadctx.request_id)
-    headers = message_headers(UPDATE_EVENT_NAME)
-    emit_event(event, key, headers)
+    headers = message_headers(EventTypes.updated)
+    event = build_event(EventTypes.updated, host, request_id=threadctx.request_id)
```
shdunning, line 249: `EventTypes.updated` VS `UPDATE_EVENT_NAME` and `"updated"` -- nice!

app/queue/events.py
```diff
@@ -0,0 +1,121 @@
...
+                **serialize_canonical_facts(host.canonical_facts),
+                "account": host.account,
+                "request_id": request_id,
+            }
+        )
+        .data
+    )
+
+
+# Event Constructors
```
shdunning, line 107: Not a constructor b/c the obj is not being returned, right? A data representation of an obj (dict) is being returned (if i'm reading the code correctly). I would remove or update this comment.

app/queue/queue.py
```diff
@@ -7,15 +7,15 @@
```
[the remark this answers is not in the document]
shdunning: +1 to what @Glutexo said.

app/queue/events.py
ryandillinfelton: removed

app/queue/queue.py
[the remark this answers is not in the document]
ryandillinfelton: Calling it __init__.py causes a lot of trouble since it makes it so that any time you import something from a file in /queue you have to import everything in queue/__init.py__. This especially causes issues because queue/event_producer is used in app/init in create_app, but queue/__init__.py imports something from app/__init__.py that isn't defined until that function finishes running.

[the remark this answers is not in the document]
Glutexo: Imports generally shouldn’t have side effects. But I get your point.

api/host.py
Glutexo: Enums for the rescue!

app/queue/events.py
```diff
@@ -0,0 +1,120 @@
...
+        return dumpHostDeleteEvent(event_type, host, request_id)
+    else:
+        raise ValueError(f"Invalid event type ({event_type})")
+
+
+def add_host_results_to_event_type(results):
+    if results == AddHostResults.created:
+        return EventTypes.created
+    if results == AddHostResults.updated:
+        return EventTypes.updated
```
Glutexo, line 120: Instead of a function like this I’d rather use a map.

```python
ADD_HOST_RESULT_TO_EVENT_TYPE = {
    AddHostResults.created: EventTypes.created,
    AddHostResults.updated: EventTypes.updated,
}
event_type = ADD_HOST_RESULT_TO_EVENT_TYPE[add_host_result]
```

One advantage of that is that it raises _KeyError_ if invalid type is passed it. (But that should never happen!) This function of yours doesn’t raise anything in that case.

Glutexo, line 120: But we can cheat here a bit and rely on the fact that the add host result names and the event type names are the same. 😈

```python
event_type = EventType[AddHostResults.name]
```

```suggestion
```

app/queue/events.py
```diff
@@ -0,0 +1,120 @@
...
+            {
+                "timestamp": datetime.now(timezone.utc),
+                "type": event_type.name,
+                "host": host,
+                "platform_metadata": platform_metadata,
+                "metadata": {"request_id": request_id},
+            }
+        )
+        .data
+    )
```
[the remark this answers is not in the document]
Glutexo: As I am looking onto this with fresh eyes, it would have been nicer to have the Schemas in the map dictionary. I had a bad feeling about that, but somehow it didn’t come up to my mind. But whatever, like this it works too.

```python
{
    EventTypes.created: (HostCreateUpdateEvent, host_create_update_event),
    ...
}
```

app/queue/event_producer.py
```diff
@@ -0,0 +1,34 @@
+from enum import Enum
+
+from kafka import KafkaProducer
+
+from app.logging import get_logger
+from app.queue import metrics
+
+logger = get_logger(__name__)
+
+Topics = Enum("Topics", ("egress", "event"))
```
Glutexo, line 10: Let’s make this singular.

```suggestion
Topic = Enum("Topic", ("egress", "event"))
```

I even considered moving this to _app.config_, but here it’s probably fine as this is the module that deals with topics.

app/queue/event_producer.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class EventProducer:
+    def __init__(self, config):
+        logger.info("Starting EventProducer()")
+        self._kafka_producer = KafkaProducer(bootstrap_servers=config.bootstrap_servers)
+        self.topics = {"egress": config.host_egress_topic, "event": config.event_topic}
```
Glutexo, line 17: **This does not work!**

The keys must be _Topics_, not strings. Otherwise the event production does not work. _write_event_ is called with _Topics.*_, that doesn’t get matched to the dictionary keys that are strings.

```suggestion
        self.topics = {Topic.egress: config.host_egress_topic, Topic.event: config.event_topic}
```

app/queue/event_producer.py
ryandillinfelton: I changed it anyway, it was just a few spots. It seemed worth it to just fix it now even if it wasn't a huge deal

app/queue/queue.py
```diff
@@ -145,8 +147,13 @@ def handle_message(message, event_producer):
 
     with PayloadTrackerContext(payload_tracker, received_status_message="message received"):
         (output_host, add_results) = add_host(validated_operation_msg["data"])
-        event = build_egress_topic_event(add_results.name, output_host, platform_metadata)
-        event_producer.write_event(event, output_host["id"], message_headers(add_results.name))
+        event = build_event(
+            add_host_results_to_event_type(add_results),
+            output_host,
+            platform_metadata=platform_metadata,
+            request_id=threadctx.request_id,
```
Glutexo, line 154: Is there a reason why is the request ID [stored](https://github.com/RedHatInsights/insights-host-inventory/blob/40ed45773e3fde8f84cee0a4cdac66d171246cb6/app/queue/queue.py#L144) in the local thread context? The [payload tracker](https://github.com/RedHatInsights/insights-host-inventory/blob/40ed45773e3fde8f84cee0a4cdac66d171246cb6/app/payload_tracker/__init__.py) doesn’t use it and I don’t see any other _threadctx_ access. Am I missing something?

app/queue/queue.py
```diff
@@ -145,8 +147,13 @@ def handle_message(message, event_producer):
...
+        )
```
Glutexo, line 155:
```suggestion
        request_id = platform_metadata.get("request_id", "-1")
        event_type = add_host_results_to_event_type(add_results)
        event = build_event(event_type, output_host, platform_metadata=platform_metadata, request_id=request_id)
```

app/queue/queue.py
```diff
@@ -145,8 +147,13 @@ def handle_message(message, event_producer):
...
+        event_producer.write_event(event, output_host["id"], message_headers(add_results), Topics.egress)
```
Glutexo, line 156:
```suggestion
        event_producer.write_event(event, output_host["id"], message_headers(add_results), Topic.egress)
```

app/queue/queue.py
ryandillinfelton: No idea why it was set up like that initially, it's just been that way

Glutexo:
Pushed in some trivial fixes:

- Converted enum names to singular.
- Decoupled [Host Delete](https://github.com/RedHatInsights/insights-host-inventory/blob/a233040ec3cfd5a104f4bac63792d389133b69ff/lib/host_delete.py) from Flask.
- Removed unnecessary Thread Context usages.
- Consolidated Request ID in the [Host Reaper](https://github.com/RedHatInsights/insights-host-inventory/blob/a233040ec3cfd5a104f4bac63792d389133b69ff/host_reaper.py).

It’s possible that I broke the tests. I‘m going to take a look at those.

app/queue/event_producer.py
```diff
@@ -0,0 +1,34 @@
...
+            k = key.encode("utf-8") if key else None
+            v = event.encode("utf-8")
+            h = [(hk, hv.encode("utf-8")) for hk, hv in headers.items()]
+            self._kafka_producer.send(self.topics[topic], key=k, value=v, headers=h)
+            metrics.egress_message_handler_success.inc()
+        except Exception:
+            logger.exception("Failed to send event")
+            metrics.egress_message_handler_failure.inc()
+
+    def close(self):
```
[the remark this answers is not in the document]
Glutexo: Added the call to the Host Reaper as it was missing there.

Glutexo:
I made a mistake and broke Request ID logging through filters. Going to fix.

app/queue/event_producer.py
```diff
@@ -0,0 +1,34 @@
...
+
+from app.logging import get_logger
+from app.queue import metrics
+
+logger = get_logger(__name__)
+
+Topics = Enum("Topics", ("egress", "events"))
+
+
+class EventProducer:
```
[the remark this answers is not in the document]
Glutexo: Asked @ryandillinfelton to write tests for this.

app/queue/events.py
```diff
@@ -0,0 +1,115 @@
...
+            "metadata": {"request_id": request_id},
+        },
+    )
+
+
+def host_delete_event(event_type, host, request_id):
+    return (
+        HostDeleteEvent,
+        {
+            "timestamp": datetime.utcnow(),
```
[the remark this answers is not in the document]
Glutexo: Fixed.

app/queue/queue.py
Glutexo: I figured that out. It’s because of the logging. https://github.com/RedHatInsights/insights-host-inventory/blob/b31e258712a1cd6a0968e27af9548a9070a50259/app/logging.py#L87

I would be awesome to change this and stop relying on thread local storage at all. But let’s not add more stuff into this pull request. There is the JIRA ticket, so let’s do it as a part of it when it comes to a Sprint.

Glutexo:
Fixed the Request ID local thread context problem I introduced. There should be tests for that.

ryandillinfelton:
There are some more tests that need to be changed from the current_app mocking approach. I'll push up the fixes as soon as I figure them out.

test_api.py
```diff
@@ -1871,32 +1895,37 @@ def __iter__(self, *args, **kwargs):
...
+    def test_delete_when_all_hosts_are_deleted(self, current_app_mock):
         host_id_list = [self.added_hosts[0].id, self.added_hosts[1].id]
         url = HOST_URL + "/" + ",".join(host_id_list)
         with patch("api.host.delete_hosts", self.DeleteHostsMock.create_mock(host_id_list)):
-            # Two hosts queried, but both deleted by a different process. No event emitted yet
-            # returning 200 OK.
-            self.delete(url, 200, return_response_as_json=False)
-            emit_event.assert_not_called()
+            with self.app.app_context():
+                with patch("lib.host_delete.current_app.event_producer.write_event") as write_event:
```
Glutexo: It’s not necessary to patch _current_app.event_producer_. If a mock is attached to _self.app_, it is used by _delete_hosts_ and it’s possible to assert _self.app.event_producer.write_event.assert_not_called()_.

test_api.py
```diff
@@ -1871,32 +1895,37 @@ def __iter__(self, *args, **kwargs):
...
+                    # Two hosts queried, but both deleted by a different process. No event emitted yet
+                    # returning 200 OK.
+                    self.delete(url, 200, return_response_as_json=False)
+                    write_event.assert_not_called()
```
```suggestion
                # Two hosts queried, but both deleted by a different process. No event emitted yet
                # returning 200 OK.
                self.delete(url, 200, return_response_as_json=False)
                self.app.event_producer.write_event.assert_not_called()
```

test_api.py
```diff
@@ -1871,32 +1895,37 @@ def __iter__(self, *args, **kwargs):
             self._delete_hosts()
             return iterator
```
Glutexo, line 1899:
```suggestion

def setUp():
    self.app.event_producer = mock.Mock()
```

test_api.py
```diff
@@ -2773,42 +2802,44 @@ def test_get_system_profile_ignores_culled(self):
...
 class QueryStalenessDeleteIgnoresCulledTestCase(QueryStalenessGetHostsBaseTestCase):
-    def test_delete_ignores_culled(self, emit_event):
+    def test_delete_ignores_culled(self):
         url = HOST_URL + "/" + self.culled_host["id"]
         self.delete(url, 404)
 
-    def test_delete_works_on_non_culled(self, emit_event):
-        url = HOST_URL + "/" + self.fresh_host["id"]
-        self.delete(url, 200, return_response_as_json=False)
+    @patch("lib.host_delete.current_app")
```
Glutexo, line 2846: Neither here it’s necessary to mock _current_app_. A better approach would be to attach a mock to _self.app.event_producer_.

Glutexo:
@ryandillinfelton Perfect! I gave you some inspiration in the inline comments. As I wrote in my previous comment, I found a way how to make it work, so take a look if you’re stuck. I’ll take a look tomorrow morning at the results.

app/queue/event_producer.py
```diff
@@ -0,0 +1,34 @@
...
+Topic = Enum("Topic", ("egress", "events"))
+
+
+class EventProducer:
+    def __init__(self, config):
+        logger.info("Starting EventProducer()")
+        self._kafka_producer = KafkaProducer(bootstrap_servers=config.bootstrap_servers)
+        self.topics = {Topic.egress: config.host_egress_topic, Topic.events: config.event_topic}
+
+    def write_event(self, event, key, headers, topic):
```
Glutexo, line 19: All our messages are JSON encoded. This class/method is an abstraction. Can we pass a serialized, but not stringified (dumps’ed) object as _event_ argument? It would make testing a lot easier and would be more semantic.

app/queue/event_producer.py
```diff
@@ -0,0 +1,34 @@
...
+        logger.debug("Topic: %s, key: %s, event: %s, headers: %s", topic, key, event, headers)
+
+        try:
+            k = key.encode("utf-8") if key else None
+            v = event.encode("utf-8")
```
Glutexo, line 24:
```suggestion
            v = json.dumps(event).encode("utf-8")
```

app/queue/event_producer.py
Glutexo: To explain this better:

Our _EventProducer_ is a higher-level abstraction over _KafkaProducer_. Although it already deals with keys, headers, messages etc. it still does some transformations like string to bytes encoding. I believe that instead of passing in a JSON-encoded string, we can pass in a JSON-serializable object instead and leave the serialization for the _EventProducer_. Doing so would save us from deserialization in the tests. Semantically this would be correct too as both encodings (string and JSON) would be gathered to a single place.

I would even think that even the schema validation etc. can be more tightly coupled here. But no need to dive any deeper with the refactoring.

test_api.py
```diff
@@ -1795,61 +1794,65 @@ def setUp(self):
...
+                return cls(hosts_ids_to_delete, select_query, event_producer)
 
             return _constructor
 
-        def __init__(self, host_ids_to_delete, original_query):
+        def __init__(self, host_ids_to_delete, original_query, event_producer):
             self.host_ids_to_delete = host_ids_to_delete
-            self.original_query = delete_hosts(original_query)
+            with self.app.app_context():
+                self.original_query = delete_hosts(original_query, "-1", self.app.event_producer)
```
Glutexo, line 1855: You created a different [_event_producer_](https://github.com/RedHatInsights/insights-host-inventory/blob/e1c4e8373c802569bfc767358d7cfa74d397c2b5/test_api.py#L1842) in _setUp_, but here you are passing in _self.app.event_producer_. But even if you used _self.event_producer_ here, it still wouldn’t work. Here, _self_ refers to a _DeleteHostsMock_ class instance, but the event producer is attached to _DeleteHostsRaceConditionTestCase_.

Just use the _event_producer_ you receive as an argument. It’s passed in by the caller that uses _current_app.event_producer_. That is the same object as _self.app.event_producer_ here in the tests.

```suggestion
                self.original_query = delete_hosts(original_query, "-1", event_producer)
```

test_api.py
```diff
@@ -1795,61 +1794,65 @@ def setUp(self):
...
+            event = json.loads(self.app.event_producer.event)
+            self.assertEqual("-1", event["request_id"])
 
 
-@patch("lib.host_delete.emit_event")
 class DeleteHostsRaceConditionTestCase(PreCreatedHostsBaseTestCase):
+    def setUp(self):
+        super().setUp()
+        self.event_producer = mock.Mock()
+
```
Glutexo, line 1843: This is not necessary. _self.event_producer_ is not used anywhere. Moreover, _self.app.event_producer_ is automatically used when _delete_hosts_ is called, because there in the calling place, it is referenced as _current_app.event_producer_.

```suggestion
```

test_api.py
```diff
@@ -1795,61 +1794,65 @@ def setUp(self):
...
+            def _constructor(select_query, request_id, event_producer):
+                return cls(hosts_ids_to_delete, select_query, event_producer)
 
             return _constructor
 
-        def __init__(self, host_ids_to_delete, original_query):
+        def __init__(self, host_ids_to_delete, original_query, event_producer):
             self.host_ids_to_delete = host_ids_to_delete
-            self.original_query = delete_hosts(original_query)
+            with self.app.app_context():
```
Glutexo, line 1854: This causes the race condition. It’s because _self_ here refers to _ DeleteHostsMock_ and not _DeleteHostsRaceConditionTestCase_. _DeleteHostsMock_ has a _\_\_getattr\_\__ magic method that gets called whenever a non-existent property is accessed. In this case, _self.app_, because _DeleteHostsMock_ does not have the _app_. But what _\_\_getattr\_\__ does is that is forwards all calls to _self.original_query_ and this is not set yet. As a result, _\_\_getattr\_\__ is called again and again.

Moreover, you don’t need to wrap this to the application context at all. _delete_hosts_ is a generator and calling it doesn’t do any executive stuff by itself. It starts deleting only after it’s iterated over.

test_api.py
```diff
@@ -1871,32 +1874,35 @@ def __iter__(self, *args, **kwargs):
...
         host_id = self.added_hosts[0].id
         url = HOST_URL + "/" + host_id
         with patch("api.host.delete_hosts", self.DeleteHostsMock.create_mock([host_id])):
             # One host queried, but deleted by a different process. No event emitted yet returning
             # 200 OK.
-            self.delete(url, 200, return_response_as_json=False)
-            emit_event.assert_not_called()
+            with self.app.app_context():
+                self.delete(url, 200, return_response_as_json=False)
+                self.event_producer.write_event.assert_not_called()
```
Glutexo, line 1885: It’s _self.app.event_producer_ that is used by _delete_hosts_. This assertion would thus pass, even when it shouldn’t. Since _self.app.event_producer_ is _MockEventProducer_, we can change the assertion like this:

```suggestion
                self.assertIsNone(self.app.event_producer.event)
```

test_api.py
```diff
@@ -1871,32 +1874,35 @@ def __iter__(self, *args, **kwargs):
...
         with patch("api.host.delete_hosts", self.DeleteHostsMock.create_mock(host_id_list)):
-            # Two hosts queried, but both deleted by a different process. No event emitted yet
-            # returning 200 OK.
-            self.delete(url, 200, return_response_as_json=False)
-            emit_event.assert_not_called()
+            with self.app.app_context():
+                # Two hosts queried, but both deleted by a different process. No event emitted yet
+                # returning 200 OK.
+                self.delete(url, 200, return_response_as_json=False)
+                self.app.event_producer.write_event.assert_not_called()
```
Glutexo, line 1895: Same here.

```suggestion
                self.assertIsNone(self.app.event_producer.event)
```

test_api.py
```diff
@@ -1871,32 +1874,35 @@ def __iter__(self, *args, **kwargs):
...
         with patch("api.host.delete_hosts", self.DeleteHostsMock.create_mock(host_id_list[0:1])):
-            # Two hosts queried, one of them deleted by a different process. Only one event emitted,
-            # returning 200 OK.
-            self.delete(url, 200, return_response_as_json=False)
-            emit_event.assert_called_once()
+            with self.app.app_context():
+                # Two hosts queried, one of them deleted by a different process. Only one event emitted,
+                # returning 200 OK.
+                self.delete(url, 200, return_response_as_json=False)
+                self.app.write_event.assert_called_once()
```
Glutexo, line 1905: Here, we can even test that the event is emitted for the correct host.

```suggestion
                self.assertEqual(host_id_list[1], self.app.event_producer.key)
```

utils/validate_hosts_for_delete.py
```diff
@@ -1,18 +1,19 @@
...
 
 logger = get_logger("utils")
 
 
 def test_validations(host):
     schema = HostEvent()
-    event = events.delete(host)
+    event = build_event(EventType.delete, host)
     deserialized = schema.loads(event)
     return deserialized.errors
```
Glutexo, line 18: This is all handled by [_build_event_](https://github.com/RedHatInsights/insights-host-inventory/blob/294e1760096cf97139c64449bf23d0bc4704064d/app/queue/events.py#L107) now. Moreover it uses strict validation that raises a ValidationError on failure.

```suggestion
    try:
        build_event(EventType.delete, host)
    except ValidationError as error:
        return error.messages
    else:
        return {}
``

**Comment by Glutexo** [comment]
@shdunning ’s review is still _changes requested_. Please approve. But only if it really works and you are OK with the code. No pro forma approvals, no no.

Does @jharting want to take a fresh look?

**Comment by Glutexo** [comment]
I asked @ryandillinfelton to extract new regression tests for bugs we managed to introduce to a follow up pull request. I don’t want to delay this any longer because of new stuff. These things were:

- EventProducer unit tests.
- Logging filter tests.

**Comment by Glutexo** [comment]
Tests passed. 
scared-cat

File: host_reaper.py [hunk 35]
```diff
@@ -76,15 +74,17 @@ def main(logger):
...
         with session_guard(session):
-            run(config, logger, session)
+            run(config, logger, session, event_producer)
     finally:
-        flush()
-
         job = _prometheus_job(config.kubernetes_namespace)
         push_to_gateway(config.prometheus_pushgateway, job, registry)
 
+        event_producer.close()
```
jharting, line 86: the producer should be closed even if pushing to gateway fails for some reason

app/queue/event_producer.py
```diff
@@ -0,0 +1,34 @@
...
+            h = [(hk, hv.encode("utf-8")) for hk, hv in headers.items()]
+            self._kafka_producer.send(self.topics[topic], key=k, value=v, headers=h)
+            metrics.egress_message_handler_success.inc()
```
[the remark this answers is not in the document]
Glutexo: The callbacks are now there, but this _inc()_ remained here. 😢

Glutexo:
No!

Glutexo:
Every event emission would cause _NameError_, causing a failure message to be logged. But the event is still emitted. I’ll fix it.

Glutexo:
```log
[2020-06-23 17:33:07,511] [12819] [4566195648] [inventory.app.queue.event_producer] [ERROR] Failed to send event
  File "/Users/stomsa/Vývoj/Insights/Repozitáře/insights-host-inventory/app/queue/event_producer.py", line 30, in write_event
    send_future.add_errback(message_not_produced, logger, self._topic, event, key, headers)
AttributeError: 'EventProducer' object has no attribute '_topic'
[2020-06-23 17:33:07,571] [12819] [123145509335040] [inventory.app.queue.event_producer] [INFO] Message PRODUCED offset=1326 timestamp=1592926387511 topic=platform.inventory.host-egress, key=2771c070-fd2e-401a-bdbb-ec8aaecab1c2, headers={'event_type': 'created'}
```

jharting:
@Glutexo hmm I did run inventory locally during the last review round. No clue how I missed that :confused:

Glutexo:
I see it‘s fixed now by #668. Awesome! 🎉

PR #654, opened 2020-05-07
Set More Message Headers (RHCLOUD-5964)

tahmidefaz:
[RHCLOUD-5964](https://projects.engineering.redhat.com/browse/RHCLOUD-5964) asks to add the following headers:

- [x] request_id
- [x] producer
- [x] whether or not a host is associated with insights
- [ ] whether or not a host is triggered by ingress event (we have different producers for ingress and rest, so, do we need this one?)

Also, I couldn't think of any other info that might be a good candidate for message headers. If you have any suggestions, let me know.

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -242,8 +242,7 @@ def _emit_patch_event(host):
     request_id = threadctx.request_id
     event = build_event_topic_event("updated", host, request_id=request_id)
     headers = message_headers(
-        event_type=UPDATE_EVENT_NAME,
-        registered_with_insights="true" if "insights_id" in host else "false",
+        event_type=UPDATE_EVENT_NAME, registered_with_insights="true" if "insights_id" in host else "false"
     )
     emit_event(event, key, headers)
 
--- a/api/host.py
+++ b/api/host.py
@@ -68,8 +68,9 @@ def add_host_list(host_list):
                     input_host = deserialize_host(host)
                     (output_host, add_result) = _add_host(input_host)
                     status_code = _convert_host_results_to_http_status(add_result)
-                    response_host_list.append({"status": status_code, "host": output_host})
-                    payload_tracker_processing_ctx.inventory_id = output_host["id"]
+                    serialized_host = serialize_host(output_host, staleness_timestamps())
+                    response_host_list.append({"status": status_code, "host": serialized_host})
+                    payload_tracker_processing_ctx.inventory_id = output_host.id
             except ValidationException as e:
                 number_of_errors += 1
                 logger.exception("Input validation error while adding host", extra={"host": host})
@@ -110,7 +111,7 @@ def _add_host(input_host):
             "host",
         )
 
-    return add_host(input_host, staleness_timestamps(), update_system_profile=False)
+    return add_host(input_host, update_system_profile=False)
 
 
 def get_bulk_query_source():
--- a/api/host.py
+++ b/api/host.py
@@ -18,7 +18,6 @@ from app import db
 from app import inventory_config
 from app.auth import current_identity
 from app.config import BulkQuerySource
-from app.events import HOSTNAME
 from app.events import message_headers
 from app.events import UPDATE_EVENT_NAME
 from app.exceptions import InventoryException
@@ -243,8 +242,6 @@ def _emit_patch_event(host):
     event = build_event_topic_event("updated", host, request_id=request_id)
     headers = message_headers(
         event_type=UPDATE_EVENT_NAME,
-        request_id=request_id,
-        producer=HOSTNAME,
         registered_with_insights="true" if "insights_id" in host else "false",
     )
     emit_event(event, key, headers)
--- a/api/host.py
+++ b/api/host.py
@@ -18,6 +18,7 @@ from app import db
 from app import inventory_config
 from app.auth import current_identity
 from app.config import BulkQuerySource
+from app.events import HOSTNAME
 from app.events import message_headers
 from app.events import UPDATE_EVENT_NAME
 from app.exceptions import InventoryException
@@ -238,8 +239,14 @@ def get_host_system_profile_by_id(host_id_list, page=1, per_page=100, order_by=N
 
 def _emit_patch_event(host):
     key = host["id"]
-    event = build_event_topic_event("updated", host, request_id=threadctx.request_id)
-    headers = message_headers(UPDATE_EVENT_NAME)
+    request_id = threadctx.request_id
+    event = build_event_topic_event("updated", host, request_id=request_id)
+    headers = message_headers(
+        event_type=UPDATE_EVENT_NAME,
+        request_id=request_id,
+        producer=HOSTNAME,
+        registered_with_insights="true" if "insights_id" in host else "false",
+    )
     emit_event(event, key, headers)
 
 
--- a/app/events.py
+++ b/app/events.py
@@ -39,5 +39,10 @@ def delete(host):
     )
 
 
-def message_headers(event_type):
-    return {"event_type": event_type}
+def message_headers(event_type, request_id, producer, registered_with_insights):
+    return {
+        "event_type": event_type,
+        "request_id": request_id,
+        "producer": producer,
+        "registered_with_isights": registered_with_insights,
+    }
--- a/app/events.py
+++ b/app/events.py
@@ -13,7 +13,8 @@ logger = logging.getLogger(__name__)
 DELETE_EVENT_NAME = "delete"
 UPDATE_EVENT_NAME = "updated"
 
-HOSTNAME = os.uname().nodename
+def hostname():
+    return os.uname().nodename
 
 
 class HostEvent(Schema):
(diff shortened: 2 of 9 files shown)


app/events.py
```diff
@@ -12,6 +13,8 @@
 DELETE_EVENT_NAME = "delete"
 UPDATE_EVENT_NAME = "updated"
 
+HOSTNAME = os.uname()[1]
```
jharting, line 16: \<nitpick\/\> `os.uname().nodename` can be used instead of accessing the field using numeric index to make the code a little more explicit

app/events.py
```diff
@@ -39,5 +42,10 @@ def delete(host):
...
 
 
-def message_headers(event_type):
-    return {"event_type": event_type}
+def message_headers(event_type, request_id, producer, registered_with_insights):
+    return {
+        "event_type": event_type,
+        "request_id": request_id,
+        "producer": producer,
+        "registered_with_insights": registered_with_insights,
```
jharting, line 50: Context: This entry is added to provide a way for apps to tell whether the host included in the message is associated with insights or not (similar to the registered_with query param).

I am on the fence myself whether to do this using a single-purpose parameter like this one or instead do a more generic thing such as including the `insights_id` in header - apps you check for its presence to know if the host is associated with Insights or not. Curious what @Glutexo thinks

tahmidefaz: @jharting Either is fine by me, though I think using `insights_id` might reduce confusion with the query param.

@Glutexo any thoughts on this?

Glutexo: I didn’t like the _registered_with_ query parameter from the very beginning. It overlaps to some extent with _reporter_ and essentially it’s just a set of boolean rules on the existing data.

Adding a header with _insights_id_ sounds more useful to me. It being an elevated ID for querying hints that consumers may be interested in filtering messages by it. I am however a bit afraid of adding host information to the headers.

- It violates the paradigm that the host ID is ultimate ID used for dealing with Inventory. Then, why Insights ID and not Subscription Manager ID? Both are elevated IDs used for host search. And what about the other IDs, and other canonical facts used for search (fqdn) and deduplication (all of them).
- It breaks the envelope/payload distinction. Like that we can put all IDs and eventually all canonical facts in there.

So, it depends on what is the intended usage. Like what information does the consumer to need to determine whether it cares about its contents? Is it the ID or the application? We may introduce a concept of “registerer” (is that different from “reporter”?) and list those in a field _registered_with_. It would use the same logic as the query parameter.

Those two approaches are not mutually exclusive. Even though one piece of information (system is registered with insights) can be inferred from another (system has an Insights ID), it may still make sense to include both as the determination logic lies in the Inventory.

app/events.py
```diff
@@ -12,6 +13,8 @@
 DELETE_EVENT_NAME = "delete"
 UPDATE_EVENT_NAME = "updated"
 
+HOSTNAME = os.uname().nodename
```
Glutexo, line 16: I wonder whether this should be a constant. Although unames tend not change much during runtime, it’s still a function call. And as such is is executed on import. I know we do that a lot with our own [_get_logger_](https://github.com/RedHatInsights/insights-host-inventory/blob/44f85329d88983292bbfe924acec326855529063/app/logging.py#L121) function a lot, but still.

api/host.py
```diff
@@ -238,8 +239,14 @@ def get_host_system_profile_by_id(host_id_list, page=1, per_page=100, order_by=N
...
     key = host["id"]
-    event = build_event_topic_event("updated", host, request_id=threadctx.request_id)
-    headers = message_headers(UPDATE_EVENT_NAME)
+    request_id = threadctx.request_id
+    event = build_event_topic_event("updated", host, request_id=request_id)
+    headers = message_headers(
+        event_type=UPDATE_EVENT_NAME,
+        request_id=request_id,
+        producer=HOSTNAME,
+        registered_with_insights="true" if "insights_id" in host else "false",
```
Glutexo, line 248: `json.dumps("insights_id" in host)` 😃 Just kidding.

api/host.py
```diff
@@ -238,8 +239,14 @@ def get_host_system_profile_by_id(host_id_list, page=1, per_page=100, order_by=N
...
 def _emit_patch_event(host):
     key = host["id"]
-    event = build_event_topic_event("updated", host, request_id=threadctx.request_id)
-    headers = message_headers(UPDATE_EVENT_NAME)
+    request_id = threadctx.request_id
+    event = build_event_topic_event("updated", host, request_id=request_id)
+    headers = message_headers(
+        event_type=UPDATE_EVENT_NAME,
+        request_id=request_id,
+        producer=HOSTNAME,
```
Glutexo, line 247: These values are always the same for all event messages. They don’t have to be passed in as arguments.

```suggestion
```

app/events.py
```diff
@@ -39,5 +42,10 @@ def delete(host):
     )
 
 
-def message_headers(event_type):
-    return {"event_type": event_type}
+def message_headers(event_type, request_id, producer, registered_with_insights):
+    return {
+        "event_type": event_type,
+        "request_id": request_id,
+        "producer": producer,
```
Glutexo, line 49:
```suggestion
        "request_id": threadctx.request_id,
        "producer": HOSTNAME,
```

api/host.py
Glutexo: This doesn’t work, because serialized host has all the canonical fact keys present. The missing fields are filled with _None_.

```suggestion
        registered_with_insights="true" if host["insights_id"] else "false",
```

I’d like to DRY this logic with the  other appearances.But where to put it? The problem is that this operates on a serialized object. The information can be easily pulled out of it, but it’s not the correct approach.

I took a look and it wouldn’t be that big of a change to move the serialization closer to the actual outlet, where it belongs. In this context it means here to the __emit_patch_event_ method.

```python
...
def patch_by_id(host_id_list, host_data):
    ...
    for host in hosts_to_update:
        ...
        _emit_patch_event(host)
    ...
...
def _emit_patch_event(host):
    event = build_event_topic_event("updated", serialized_host, threadctx.request_id)
    headers = message_headers(UPDATE_EVENT_NAME, registered_with_insights(host))
    emit_event(event, host.id, headers)
```

That would allow the logic function to operate on a Host object:

```python
def registered_with_insights(host):
    return "insights_id" in host.canonical_facts
```

or it can be even a Host model method, which is probably the best approach:

```python
class Host(db.Model):
    ...
    def registered_with_insights(self):
        "insights_id" in self.canonical_facts
```

rodrigonull:
Regarding the event types, shouldn't we standardize that? because some are in the past like "created", "updated" and the "delete" is in the present tense.

jharting:
> Regarding the event types, shouldn't we standardize that? because some are in the past like "created", "updated" and the "delete" is in the present tense.

The discrepancy has exited already in message bodies and unfortunately we cannot fix it without breaking the existing consumers :-/

app/queue/ingress.py
```diff
@@ -146,7 +147,16 @@ def handle_message(message, event_producer):
...
         event = build_egress_topic_event(add_results.name, output_host, platform_metadata)
-        event_producer.write_event(event, output_host["id"], message_headers(add_results.name))
+        event_producer.write_event(
+            event,
+            output_host["id"],
+            message_headers(
+                event_type=add_results.name,
+                request_id=threadctx.request_id,
+                producer=HOSTNAME,
+                registered_with_insights="true" if "insights_id" in output_host else "false",
```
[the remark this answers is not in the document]
tahmidefaz: Changes suggested in app.queue.ingress.add_host, specifically,
`"host": {i: getattr(output_host, i) for i in output_host if i not in ("facts", "system_profile")}`
doesn't work. Throwing errors even when I serialize the output_host. Investigating... :thinking:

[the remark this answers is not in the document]
tahmidefaz: @Glutexo This throws me a `sqlalchemy.orm.exc.DetachedInstanceError: Instance Host is not bound to a Session` error. Need help figuring out what is up.

tahmidefaz:
@Glutexo I left a comment for you above. Basically, when I try to move the serialization around, it gets angry with me. It would be great if you can take a look/clarify a little more here. :frowning_face:

Glutexo:
From what I remember, the problem there was that we were applying some logic on serialized data. That is not a good approach as the serialization should happen at the very output. It wasn’t however that simple to move it to another place, because of a way how database session is handled there. Unfortunately it breaks subsequent database model attribute access.

I’ll take take a look at the new changes, but I see that the tests are failing. Hopefully I’ll find out why.

jharting:
@tahmidefaz with https://github.com/RedHatInsights/insights-host-inventory/pull/651 nearing completion I think you can start rebasing this branch on top of the refactoring

tahmidefaz:
@Glutexo All of the test fails are for the MQ service and I think I figured out why. It's because we are patching the `host_repository.add_host` and using a dictionary as a return value, whereas this PR makes that function return a `host` object other than a dictionary. I will try a fix. :crossed_fingers:

Glutexo:
I’ll take a fresh look after this is rebased.

tahmidefaz:
Closing this one in favor of #676

PR #655, opened 2020-05-11
Tests refactor

rodrigonull:
This PR is part of the card: https://projects.engineering.redhat.com/browse/RHCLOUD-6424

The main goal here is to try to organize the test code making it easier to maintain.

- [x] Moved all tests files to /tests
- [x] Splitted large files into small ones (test_api.py -> test_api_auth, test_api_delete, test_api_get, test_api_patch, test_api_post, test_api_put, test_api_utils)
- [x] Created a new fixture to handle the test database creation/drop to avoid renaming tables during the test execution (note: this only applies to the tests that are using the flask_app_fixture)
- [x] Added pytest markers to be able to run specific tests (I hope you can give me suggestions on which markers would be useful)
- [x] Converted test_api_auth.py to pytest

I had to add a new lib "sqlalchemy-utils" to be able to create/drop the test database.

Any comments or suggestions are welcome :)


Glutexo:
Removed “WIP” from the title as this is no longer WIP.

app/config.py
```diff
@@ -120,3 +120,10 @@ def log_configuration(self, config_name):
             elif self._environment == RuntimeEnvironment.job:
                 self.logger.info("Metrics Pushgateway: %s", self.prometheus_pushgateway)
                 self.logger.info("Kubernetes Namespace: %s", self.kubernetes_namespace)
+
+
+class TestConfig(Config):
+    def __init__(self, environment):
+        super().__init__(environment)
+        self._db_name = os.getenv("INVENTORY_TEST_DB_NAME", "insights-test")
+        self.db_uri = self._build_db_uri(self._db_ssl_mode)
```
Glutexo, line 129: Can we leverage the _config_name_ string instead instead? It’s not the best solution around here, but it’s already there.

```python
class Config:
    def __init__(self, environment, config_name):
        ...
        default_db_name = "insights-test" if configname == "testing" else "insights"
        self._db_name = os.getenv("INVENTORY_DB_NAME", default_db_name)
```

```suggestion
```

This would require adding the config name to other places where _Config_ is instantiated. In all of them, _config_name_ is available. In the tests, it should be _testing_ and in the explicit config tests, it should be _development_.

tests/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+
+    yield config.db_uri
+
+    drop_database(config.db_uri)
+
+
+@pytest.fixture(scope="function")
+def flask_app(database):
+    app = create_app(config_name="testing")
+    app.config["SQLALCHEMY_DATABASE_URI"] = database
```
Glutexo, line 26:
```suggestion
```

tests/conftest.py
```diff
@@ -0,0 +1,43 @@
+import pytest
+from sqlalchemy_utils import create_database
+from sqlalchemy_utils import database_exists
+from sqlalchemy_utils import drop_database
```
Glutexo, line 4: Do we actually need the utils? We already have [alembic](https://github.com/rodrigonull/insights-host-inventory/blob/4c99a5d71276c0d38ec2d198854a03824201cf66/Pipfile.lock#L19). It can be used for this too, not just for migrations. It would just mean explicitly putting it into _Pipfile_, but it would be no-op as it’s already there because of dependencies.

I was originally thinking about doing some cleanup of the dependencies. I thought we already have too many of them and I have suspicion that they are often used only for one some simple task. I’d rather not add more if we already have ways to to this.

rodrigonull: @Glutexo I tried to change the code to create / drop databases using alembic, but I didn't find a way to do it. The only other way I found besides using this library was to use sqlalchemy directly, but the code turned out to be very ugly. So, if you know how to do it using alembic, could you show me how?

Glutexo: Oh. It looks I was wrongly assuming that Alembic is able to handle this. But it looks like it migrates only inside a single database. Let’s keep SQLAlchemy Utils then.

tests/conftest.py
```diff
@@ -0,0 +1,42 @@
...
+from app import db
+from app.config import Config
+from app.config import RuntimeEnvironment
+
+
+@pytest.fixture(scope="session")
+def database():
+    config = Config(RuntimeEnvironment.server, "testing")
+    if not database_exists(config.db_uri):
+        create_database(config.db_uri)
```
Glutexo, line 16: Is this the right thing to do? If the database exists, it means it hasn’t been properly dropped. In such case, it may contain unexpected data. Don’t we want to rather drop it and re-create instead of skipping the create?

rodrigonull: I agree with you. I think the best option is to drop and them re-create the database.

jharting:
@Glutexo waiting for your :+1:

Glutexo:
FinalLooking into this now.

tests/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+from app import db
+from app.config import Config
+from app.config import RuntimeEnvironment
+
+
+@pytest.fixture(scope="session")
+def database():
+    config = Config(RuntimeEnvironment.TEST)
+    if database_exists(config.db_uri):
+        drop_database(config.db_uri)
```
Glutexo, line 16: Awesome! 👍🏻

tests/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+        yield app
+
+        ctx.pop()
+        db.session.remove()
+        db.drop_all()
+
+
+@pytest.fixture(scope="function")
+def flask_client(flask_app):
+    flask_app.testing = True
```
Glutexo, line 39: We are not testing any exceptions, so this doesn’t get used at all. (If it's not here, the tests still pass.) It‘s however good to have it here. 👍🏻

tests/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@pytest.fixture(scope="function")
+def flask_app(database):
+    app = create_app(RuntimeEnvironment.TEST)
+
+    # binds the app to the current context
+    with app.app_context() as ctx:
+        db.create_all()
+        ctx.push()
```
Glutexo: Is this really necessary? If I understand the documentation correctly, pushing/popping the application (and request) context is exactly what the _app_context_ context manager does.

Neither the [testing guide](https://flask.palletsprojects.com/en/1.1.x/testing) nor the API [documentation](https://flask.palletsprojects.com/en/1.1.x/testing/) shows this usage. There is [one example](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Flask.teardown_appcontext) showing _push_/_pop_, but that one doesn’t use the context manager.

I suggest removing this until something breaks. Currently, if this manual context mangling is not there, our tests still pass.

PR #658, opened 2020-05-19
RHCLOUD-6508 - Add support to search by os_release

dLobatog:
This adds the ability to search systems by os_release, and introduces
searching by any system_profile_fact to the codebase for any future search.

Draft for now because there are a few topics I need to have resolved before getting this in:

 * First and foremost *tests*
 * I've seen hosts have *all sorts* of os_release fields. From compliance (or any app), I guess all I want is to search by `os_release=RHEL7`. But the os_release field contains things like "Red Hat EL 7.0.1", or "Red Hat Enterprise Linux Server release 7.5 (Maipo)".
 * Maybe it's worth having a special param "os" which matches on `(Red*Hat|RHEL).*$version` or something.

### DIFF
--- a/api/host_query_db.py
+++ b/api/host_query_db.py
@@ -30,7 +30,7 @@ def get_host_list(
     order_how,
     staleness,
     registered_with,
-    os_release,
+    os_major,
 ):
     if fqdn:
         query = _find_hosts_by_canonical_fact("fqdn", fqdn)
@@ -40,8 +40,8 @@ def get_host_list(
         query = _find_hosts_by_hostname_or_id(hostname_or_id)
     elif insights_id:
         query = _find_hosts_by_canonical_fact("insights_id", insights_id)
-    elif os_release:
-        query = _find_hosts_by_system_profile_fact("os_release", os_release)
+    elif os_major:
+        query = _find_hosts_by_system_profile_fact("os_release", os_major, '((\d)\.*\d*)')
     else:
         query = _find_all_hosts()
 
@@ -108,8 +108,8 @@ def _find_hosts_by_canonical_fact(canonical_fact, value):
     return canonical_fact_host_query(current_identity.account_number, canonical_fact, value)
 
 
-def _find_hosts_by_system_profile_fact(system_profile_fact, value):
-    return system_profile_fact_host_query(current_identity.account_number, system_profile_fact, value)
+def _find_hosts_by_system_profile_fact(system_profile_fact, value, regex=None):
+    return system_profile_fact_host_query(current_identity.account_number, system_profile_fact, value, regex)
 
 
 def _find_hosts_by_tag(string_tags, query):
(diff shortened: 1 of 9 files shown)


lib/host_repository.py
```diff
@@ -97,6 +99,22 @@ def canonical_facts_host_query(account_number, canonical_facts):
...
+        logger.info("SYSTEMPROFILEFACTS 2 (%s)", Host.system_profile_facts[system_profile_fact].astext)
+        query = Host.query.filter(
+            (Host.account == account_number) &
+            (re.match(regex, Host.system_profile_facts[system_profile_fact].astext).group(2) == value)
+        )
+    else:
+        query = Host.query.filter(
+            (Host.account == account_number) & (Host.system_profile_facts[system_profile_fact].astext == value)
+        )
+    return find_non_culled_hosts(query)
```
Glutexo, line 115: In this function the _Host.query.filter_ with the account condition repeats twice. The function essentially does two different things. Maybe we can have two functions instead. But one of them would be currently not used at all. So just keep the one?

Another option would be to just DRY the common part.

```suggestion
def system_profile_fact_host_query(account_number, system_profile_fact, value,
                                   regex):
    query = Host.query.filter(
        (Host.account == account_number) &
        (re.match(regex, Host.system_profile_facts[system_profile_fact].astext).group(2) == value)
        )
    return find_non_culled_hosts(query)
```

api/host_query_db.py
```diff
@@ -38,6 +40,8 @@ def get_host_list(
         query = _find_hosts_by_hostname_or_id(hostname_or_id)
     elif insights_id:
         query = _find_hosts_by_canonical_fact("insights_id", insights_id)
+    elif os_major:
+        query = _find_hosts_by_system_profile_fact("os_release", os_major, '((\d)\.*\d*)')
```
Glutexo, line 44: If we changed the parameter type to integer, we should stringify it for the regex matching.

```suggestion
        query = _find_hosts_by_system_profile_fact("os_release", str(os_major), '((\d)\.*\d*)')
```

lib/host_repository.py
```diff
@@ -97,6 +99,22 @@ def canonical_facts_host_query(account_number, canonical_facts):
     return find_non_culled_hosts(query)
 
 
+def system_profile_fact_host_query(account_number, system_profile_fact, value,
+                                   regex=None):
+    if regex:
+        logger.info("FINDING OSMAJOR (%s, %s)", regex, value)
+        logger.info("SYSTEMPROFILEFACTS 2 (%s)", Host.system_profile_facts[system_profile_fact].astext)
```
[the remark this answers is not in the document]
dLobatog: Agreed, this is just a draft PR

swagger/api.spec.yaml
```diff
@@ -39,6 +39,12 @@ paths:
             format: uuid
           description: Search for a host by insights_id
           required: false
+        - in: query
+          name: os_major
```
jharting, line 43: We're being asked to implement the same format for fact filtering that we're using for tags. It's unclear yet what exactly it means but it may be something like ?facts=os_release=10.0.0.1

^^ is an FYI only. I don't this the plan is actionable just yet.

PR #659, opened 2020-05-20
RHCLOUD-6643 - Query extra Host parameters via sparse fieldset

dLobatog:
Inspiration from https://jsonapi.org/format/#fetching-sparse-fieldsets

This allows to query attributes via `hosts?fields[system_profile]=os_release,arch`. It includes support for `hosts?fields[system_profile]=*` which includes all of the fields. The method is generic enough to support any sort of possible extra field in the future.

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -48,7 +48,7 @@ TAG_OPERATIONS = ("apply", "remove")
 GET_HOST_LIST_FUNCTIONS = {BulkQuerySource.db: get_host_list_db, BulkQuerySource.xjoin: get_host_list_xjoin}
 XJOIN_HEADER = "x-rh-cloud-bulk-query-source"  # will be xjoin or db
 REFERAL_HEADER = "referer"
-SUPPORTED_FIELDS = {"system_profile", "host"}
+SUPPORTED_FIELDSETS = {"system_profile", "host"}
 
 logger = get_logger(__name__)
 
@@ -172,14 +172,16 @@ def get_host_list(
     except ValueError as e:
         flask.abort(400, str(e))
 
-    unsupported_fields = fields.keys() - SUPPORTED_FIELDS
-    if unsupported_fields:
-        flask.abort(400, f"Unsupported fields {unsupported_fields}. Supported fields include {SUPPORTED_FIELDS}.")
+    unsupported_fieldsets = fields.keys() - SUPPORTED_FIELDSETS
+    if unsupported_fieldsets:
+        flask.abort(
+            400, f"Unsupported fields {unsupported_fieldsets}. Supported fields include {SUPPORTED_FIELDSETS}."
+        )
 
     try:
         json_data = build_paginated_host_list_response(total, page, per_page, host_list, fields)
     except KeyError as error:
-        flask.abort(400, f"Unsupported field {error}. Supported fields include {SUPPORTED_FIELDS}.")
+        flask.abort(400, f"Unsupported field: {error}. Supported fields include {SUPPORTED_FIELDSETS}.")
     return flask_json_response(json_data)
 
 
--- a/api/host.py
+++ b/api/host.py
@@ -48,6 +48,7 @@ TAG_OPERATIONS = ("apply", "remove")
 GET_HOST_LIST_FUNCTIONS = {BulkQuerySource.db: get_host_list_db, BulkQuerySource.xjoin: get_host_list_xjoin}
 XJOIN_HEADER = "x-rh-cloud-bulk-query-source"  # will be xjoin or db
 REFERAL_HEADER = "referer"
+SUPPORTED_FIELDS = ("system_profile", "host", "canonical_facts")
 
 logger = get_logger(__name__)
 
@@ -170,6 +171,10 @@ def get_host_list(
         )
     except ValueError as e:
         flask.abort(400, str(e))
+
+    for field in fields:
+        if field not in SUPPORTED_FIELDS:
+            flask.abort(400, "Supported fields include %s", SUPPORTED_FIELDS)
 
     json_data = build_paginated_host_list_response(total, page, per_page, host_list, fields)
     return flask_json_response(json_data)
--- a/api/host.py
+++ b/api/host.py
@@ -48,7 +48,7 @@ TAG_OPERATIONS = ("apply", "remove")
 GET_HOST_LIST_FUNCTIONS = {BulkQuerySource.db: get_host_list_db, BulkQuerySource.xjoin: get_host_list_xjoin}
 XJOIN_HEADER = "x-rh-cloud-bulk-query-source"  # will be xjoin or db
 REFERAL_HEADER = "referer"
-SUPPORTED_FIELDS = ("system_profile", "host")
+SUPPORTED_FIELDS = {"system_profile", "host"}
 
 logger = get_logger(__name__)
 
@@ -172,9 +172,9 @@ def get_host_list(
     except ValueError as e:
         flask.abort(400, str(e))
 
-    for field in fields:
-        if field not in SUPPORTED_FIELDS:
-            flask.abort(400, "Supported fields include %s", SUPPORTED_FIELDS)
+    unsupported_fields = fields.keys() - SUPPORTED_FIELDS
+    if unsupported_fields:
+        flask.abort(400, f"Supported fields include {SUPPORTED_FIELDS}")
 
     json_data = build_paginated_host_list_response(total, page, per_page, host_list, fields)
     return flask_json_response(json_data)
--- a/api/host.py
+++ b/api/host.py
@@ -175,13 +175,14 @@ def get_host_list(
     unsupported_fieldsets = fields.keys() - SUPPORTED_FIELDSETS
     if unsupported_fieldsets:
         flask.abort(
-            400, f"Unsupported fields {unsupported_fieldsets}. Supported fields include {SUPPORTED_FIELDSETS}."
+            400,
+            f"Unsupported fields {unsupported_fieldsets}. Supported fields include {', '.join(SUPPORTED_FIELDSETS)}.",
         )
 
     try:
         json_data = build_paginated_host_list_response(total, page, per_page, host_list, fields)
     except KeyError as error:
-        flask.abort(400, f"Unsupported field: {error}. Supported fields include {SUPPORTED_FIELDSETS}.")
+        flask.abort(400, f"Unsupported field: {error}. Supported fields include {', '.join(SUPPORTED_FIELDSETS)}.")
     return flask_json_response(json_data)
 
 
--- a/api/host.py
+++ b/api/host.py
@@ -145,6 +145,7 @@ def get_host_list(
     order_how=None,
     staleness=None,
     registered_with=None,
+    fields=None,
 ):
     total = 0
     host_list = ()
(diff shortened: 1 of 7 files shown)


app/models.py
```diff
@@ -66,6 +66,7 @@ class Host(db.Model):
     tags = db.Column(JSONB)
     canonical_facts = db.Column(JSONB)
     system_profile_facts = db.Column(JSONB)
+    system_profile = orm.synonym("system_profile_facts")
```
dLobatog, line 69: Needed so that `hosts?fields[system_profile]=os_release,arch` could work, otherwise we'd have to leak the abstraction to the API and force users to query `hosts?fields[system_profile_facts]=os_release,arch`

dLobatog:
Huh, the pre-commit fail seems unrelated:
```

app/auth/identity.py:58:28: F541 f-string is missing placeholders

app/instrumentation.py:9:20: F541 f-string is missing placeholders

app/instrumentation.py:23:20: F541 f-string is missing placeholders

utils/validate_hosts_for_delete.py:25:17: F541 f-string is missing placeholders

utils/validate_hosts_for_delete.py:26:17: F541 f-string is missing placeholders
```

swagger/api.spec.yaml
```diff
@@ -814,6 +824,8 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/StructuredTag'
+        system_profile:
+          $ref: '#/components/schemas/SystemProfileIn'
```
Glutexo, line 828: This belongs to [_HostOut_](https://github.com/dLobatog/insights-host-inventory/blob/a1f992e6932f38688071b6c9024f35e7062c356b/swagger/api.spec.yaml#L850). That is the schema for GET operations. _CreateHostOut_ is for POST.

dLobatog: Thanks, addressed - it's now part of HostOut

app/serialization.py
```diff
@@ -120,6 +120,18 @@ def serialize_host(host, staleness_timestamps, fields=DEFAULT_FIELDS):
...
+            extra_field_attributes = extra_fields[extra_field].replace(' ', '').split(',')
+            serialized_host[extra_field] = {}
+            for extra_field_attribute in extra_field_attributes:
+                host_extra_field_object = getattr(host, extra_field, None)
+                if host_extra_field_object and extra_field_attribute in host_extra_field_object:
+                    serialized_host[extra_field][extra_field_attribute] = host_extra_field_object[extra_field_attribute]
+                elif host_extra_field_object and extra_field_attribute == '*':
+                    serialized_host[extra_field] = host_extra_field_object
+                else:
+                    serialized_host[extra_field][extra_field_attribute] = 'not found'
```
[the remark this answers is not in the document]
dLobatog: I've decided to omit it, it seems clear enough IMO. I can document this in the swagger if needed

api/host.py
```diff
@@ -164,7 +165,8 @@ def get_host_list(
     except ValueError as e:
         flask.abort(400, str(e))
 
-    json_data = build_paginated_host_list_response(total, page, per_page, host_list)
+    json_data = build_paginated_host_list_response(total, page, per_page,
+                                                   host_list, fields)
```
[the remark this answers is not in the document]
dLobatog: Style fix done

Glutexo:
“MUST NOT”. 😮 Okay, that sounds convincing.

app/serialization.py
```diff
@@ -89,7 +91,50 @@ def deserialize_host_xjoin(data):
...
+    serialized_host = {}
+
+    if "host" in sparse_fieldset:
+        host_attributes = sparse_fieldset["host"].replace(" ", "").split(",")
+        canonical_facts_attributes = list(set(host_attributes) & set(_CANONICAL_FACTS_FIELDS))
+        host_attributes = list(set(host_attributes) - set(canonical_facts_attributes))
+        serialized_host = {
+            **serialize_canonical_facts(host.canonical_facts, canonical_fields=canonical_facts_attributes)
+        }
+        for host_attribute in host_attributes:
```
Glutexo, line 113:
```suggestion
        for host_attribute in other_host_attributes:
```

app/serialization.py
```diff
@@ -89,7 +91,50 @@ def deserialize_host_xjoin(data):
...
+def _sparse_fieldset_serialization(host, staleness_timestamps, sparse_fieldset):
+    serialized_host = {}
+
+    if "host" in sparse_fieldset:
+        host_attributes = sparse_fieldset["host"].replace(" ", "").split(",")
+        canonical_facts_attributes = list(set(host_attributes) & set(_CANONICAL_FACTS_FIELDS))
+        host_attributes = list(set(host_attributes) - set(canonical_facts_attributes))
+        serialized_host = {
+            **serialize_canonical_facts(host.canonical_facts, canonical_fields=canonical_facts_attributes)
+        }
```
Glutexo, line 112: What is the point of this dictionary comprehension here? If I understand it correctly, it has no effect.

```suggestion
        serialized_host = serialize_canonical_facts(host.canonical_facts, canonical_facts_attributes)
```

app/serialization.py
```diff
@@ -89,7 +91,50 @@ def deserialize_host_xjoin(data):
...
+        for host_attribute in host_attributes:
+            _serialize_host_field(host, host_attribute, staleness_timestamps, serialized_host)
```
Glutexo, line 114: Something like this should work:

```suggestion
        serialized_host = {
            **serialize_canonical_facts(host.canonical_facts, canonical_facts_attributes),
            **_serialize_host_fields(host, other_host_attributes, staleness_timestamps)
        }
```

api/host_query.py
```diff
@@ -13,9 +13,13 @@
 Order = namedtuple("Order", ("by", "how"))
 
 
-def build_paginated_host_list_response(total, page, per_page, host_list):
+def build_paginated_host_list_response(total, page, per_page, host_list, fields=None):
     timestamps = staleness_timestamps()
-    json_host_list = [serialize_host(host, timestamps) for host in host_list]
+    if fields:
```
dLobatog, line 18: Maybe there's a neater way of doing this - wasn't sure (note the methods are not the same and neither are the args)

app/serialization.py
```diff
@@ -90,49 +90,84 @@ def deserialize_host_xjoin(data):
...
+        if field == "created":
+            field = "created_on"
+        elif field == "updated":
+            field = "modified_on"
+
+        value = getattr(host, field)
+        return serialize(value)
+
+    def staleness_timestamp(field):
+        if not host.stale_timestamp:
```
dLobatog, line 144: If there's no stale_timestamp, we will get an error when we call `compute` later on in the method:

```
File "/home/dlobatog/workspace/insights/insights-host-inventory/app/serialization.py", line 146, in staleness_timestamp
    timestamp = compute(host.stale_timestamp)
  File "/home/dlobatog/workspace/insights/insights-host-inventory/app/culling.py", line 31, in stale_timestamp
    return self._add_days(stale_timestamp, 0)
  File "/home/dlobatog/workspace/insights/insights-host-inventory/app/culling.py", line 28, in _add_days
    return timestamp + timedelta(days=days)
```

app/serialization.py
```diff
@@ -90,49 +90,84 @@ def deserialize_host_xjoin(data):
...
+
+    return serialized_host
+
+
+def _serialize_host_fields(host, fields, staleness_timestamps):
+    def raw(field):
+        return getattr(host, field)
+
+    def function(field, serialize):
+        if field == "created":
```
dLobatog, line 135: I'd prefer using `orm.synonym` but I think this is your call

app/serialization.py
```diff
@@ -80,7 +84,51 @@ def deserialize_host_xjoin(data):
...
+        canonical_facts_attributes = sparse_fieldset["canonical_facts"].replace(" ", "").split(",")
+        serialized_host = {
+            **serialize_canonical_facts(host.canonical_facts, canonical_fields=canonical_facts_attributes)
+        }
+
+    for field in BASIC_FIELDS:
+        _serialize_host_field(host, field, staleness_timestamps, serialized_host)
+
+    if "system_profile" in sparse_fieldset:
+        system_profile_attributes = sparse_fieldset["system_profile"].replace(" ", "").split(",")
```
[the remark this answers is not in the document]
dLobatog: We could but I think the price to pay is quite little for something that makes the call much more user friendly - typos, or users just trying it out with spaces will not have an issue right away instead of a 400

app/serialization.py
```diff
@@ -90,49 +90,84 @@ def deserialize_host_xjoin(data):
...
     return serialized_host
 
 
+def serialize_host_sparse(host, staleness_timestamps, sparse_fieldset):
+    serialized_host = {}
+
+    if "host" in sparse_fieldset:
+        requested_host_attributes = set(sparse_fieldset["host"].replace(" ", "").split(","))
+        canonical_facts_attributes = requested_host_attributes & _CANONICAL_FACTS_FIELDS
+        other_host_attributes = requested_host_attributes - canonical_facts_attributes
```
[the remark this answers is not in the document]
dLobatog: Thanks, good catch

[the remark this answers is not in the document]
dLobatog: Test added now

test_unit.py
```diff
@@ -1291,7 +1292,7 @@ def test_with_all_fields(self, serialize_canonical_facts, serialize_facts, seria
         }
         self.assertEqual(expected, actual)
 
-        serialize_canonical_facts.assert_called_once_with(host_init_data["canonical_facts"])
+        serialize_canonical_facts.assert_called_once_with(host_init_data["canonical_facts"], _CANONICAL_FACTS_FIELDS)
```
dLobatog, line 1295: Had to add this since we changed the `serialize_host` method

app/serialization.py
Glutexo: I hear you and it’s true that it would save us the conditions here. I am however trying not to couple the logic with the database model more than it already is. I believe that it’s not DB model’s responsibility to know about serialized field names.

app/serialization.py
```diff
@@ -89,50 +89,81 @@ def deserialize_host_xjoin(data):
...
+    def raw(field):
+        return getattr(host, field)
+
+    def function(field, serialize):
+        if field == "created":
+            field = "created_on"
+        elif field == "updated":
+            field = "modified_on"
+
+        value = getattr(host, field)
```
Glutexo, line 136: If we want to get rid of the ifs and of the argument name override, we can use mapping like this:

```suggestion
        field_map = {"created": "created_on", "updated": "modified_on"}
        attr = field_map.get(field, field)
        value = getattr(host, field)
```

app/serialization.py
Glutexo: Nice catch! 👍🏻

swagger/api.spec.yaml
```diff
@@ -480,8 +481,24 @@ components:
...
         enum:
           - insights
-
-
+    fieldsParam:
+      in: query
+      name: fields
+      style: deepObject
+      allowReserved: true
+      description: "Include extra fields for each host. If the requested field does not exist, it will return null.
```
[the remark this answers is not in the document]
dLobatog: I've now changed it - if the field is valid, it will return null in both host or system_profile. If it's invalid, it'll be a 400

shdunning:
Closing this as it's not needed at this time.

We should pick this back up if Graphql Gateway investigation doesn't prove fruitful.

PR #661, opened 2020-06-04
Add a metric to track POST requests on the REST API [RHCLOUD-5004]

ryandillinfelton:
To prepare to deprecate host creation through the REST API we want to have some metrics to know how many people are still using it. This PR adds a new counter `rest_post_request_counter` to Prometheus.

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -56,7 +56,7 @@ logger = get_logger(__name__)
 @api_operation
 @metrics.api_request_time.time()
 def add_host_list(host_list):
-    rest_post_request_count.inc()
+    reporter = None
 
     response_host_list = []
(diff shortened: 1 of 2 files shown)


api/host.py
```diff
@@ -55,6 +56,8 @@
 @api_operation
 @metrics.api_request_time.time()
 def add_host_list(host_list):
+    rest_post_request_count.inc()
```
jharting, line 59: adding a label that identifies the reporter will make future investigation of "who's still using this deprecated API?" easier

ryandillinfelton: Ahh man, apparently I never pushed that change up. It's up now.

PR #663, opened 2020-06-08
Report errors to payload tracker [RHCLOUD-5005]

ryandillinfelton:
Reports errors to payload tracker.

Note: Not sure which cases should use `processing_error` vs `payload_error`, took my best guess, but let me know if any should be changed

Dependent on the [event refactor PR](https://github.com/RedHatInsights/insights-host-inventory/pull/651).


app/queue/queue.py
```diff
@@ -127,10 +129,12 @@ def add_host(host_data):
             return (output_host, add_results)
         except InventoryException:
             logger.exception("Error adding host ", extra={"host": host_data})
+            payload_tracker.payload_error(f"Error adding host: {host_data}")
```
jharting, line 132: Rather than the host data it would be more useful to report the exception message

ryandillinfelton:
"rebased" and ready for review now

api/host.py
```diff
@@ -85,14 +85,19 @@ def add_host_list(host_list):
             except ValidationException as e:
                 number_of_errors += 1
                 logger.exception("Input validation error while adding host", extra={"host": host})
+                payload_tracker.payload_error(
+                    f"Input validation error: {e} \nEncountered validation error adding host: {host}"
```
jharting, line 89: Putting the entire host body into the reported error may be overwhelming. I think it is enough to send the error message. The request id is sent automatically which makes it easy to fetch the host body and other things from Kibana if needed for investigation

api/host.py
```diff
@@ -85,14 +85,19 @@ def add_host_list(host_list):
...
+                )
                 response_host_list.append({**e.to_json(), "title": "Bad Request", "host": host})
             except InventoryException as e:
                 number_of_errors += 1
                 logger.exception("Error adding host", extra={"host": host})
+                payload_tracker.payload_error(f"Inventory exception: {e} \nEncountered error adding host: {host}")
```
jharting, line 95: I realized this actually is not correct. Turns out inventory reports errors already - it's just not setting the error status message. That's because these entire blocks are wrapped within a `with PayloadTrackerContext` block and the payload tracker context manager reports the error automatically should an exception escape the block. It however needs to be enhanced to report the error message. See https://github.com/RedHatInsights/insights-host-inventory/blob/ae86e927b0b03ebe2947dfb4f4c769374faef9dd/app/payload_tracker/__init__.py#L241 for details

jharting:
@ryandillinfelton this came to conflict with the test refactor PR

app/queue/queue.py
```diff
@@ -127,12 +127,14 @@ def add_host(host_data):
             )
             payload_tracker_processing_ctx.inventory_id = output_host["id"]
             return (output_host, add_result)
-        except InventoryException:
+        except InventoryException as e:
             logger.exception("Error adding host ", extra={"host": host_data})
+            payload_tracker.payload_error(f"Inventory exception: {e}\nEncountered error adding host: {host_data}")
```
jharting, line 132: These explicit error calls are redundant as they are made by PayloadTrackerProcessingContext. I thought you already removed these? :confused:

PR #664, opened 2020-06-09
Add metadata to delete events (RHCLOUD-5224)

tahmidefaz:
Adding metadata to delete events. From what I could tell, `created` and `updated` events already has it.

PR #667, opened 2020-06-23
Add metadata to delete events (RHCLOUD-5224)

tahmidefaz:
`created` and `updated` events already has it.

### DIFF
--- a/test_api.py
+++ b/test_api.py
@@ -1841,12 +1841,15 @@ class DeleteHostsEventTestCase(PreCreatedHostsBaseTestCase, DeleteHostsBaseTestC
     def test_create_then_delete_check_metadata(self):
         with self.app.app_context():
             self._check_hosts_are_present((self.host_to_delete.id,))
-            self._delete(header=None)
+
+            request_id = generate_uuid()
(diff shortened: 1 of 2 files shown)


tahmidefaz:
Closing #664 , let's use this one instead.

test_api.py
```diff
@@ -1835,6 +1838,16 @@ def test_create_then_delete_without_request_id(self):
...
 
+    def test_create_then_delete_check_metadata(self):
+        with self.app.app_context():
+            self._check_hosts_are_present((self.host_to_delete.id,))
+            self._delete(header=None)
+
+            self._assert_event_is_valid(self.app.event_producer, self.host_to_delete, self.timestamp)
+
+            event = json.loads(self.app.event_producer.event)
+            self.assertEqual(self._expected_metadata("-1"), event["metadata"])
```
jharting, line 1849: hey @tahmidefaz can you add/extend a test to verify that when a request id is passed to the delete request this id is propagated all the way through to the message metadata?

tahmidefaz:
@jharting I think this should be good now, thanks! :+1:

PR #669, opened 2020-06-24
Add deprecation notice to host create/update in API spec

ryandillinfelton:
As part of the deprecation of the host update/create functionality on the REST API we want to make sure users know about the deprecation. This PR marks the POST operation as deprecated in the swagger docs. Below is a preview of the change.

API_POST_deprecation

### DIFF
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -65,8 +65,9 @@ paths:
         hosts. A host is updated if there is already one with the same
         canonicals facts and belonging to the same account.
         <br /><br />
-        NOTICE: This operation is deprecated. We suggest you create and update hosts
-        using the [message queue interface](https://platform-docs.cloud.paas.psi.redhat.com/backend/inventory.html#mq-host-insertion) instead.
+        NOTICE: This operation is deprecated. The explicit creation of hosts is
+        no longer supported. Hosts are created automatically based on uploads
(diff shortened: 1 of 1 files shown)


swagger/api.spec.yaml
```diff
@@ -55,6 +55,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/HostQueryOutput'
     post:
+      deprecated: true
```
jharting, line 58: nice!

swagger/api.spec.yaml
```diff
@@ -63,6 +64,9 @@ paths:
         Create a new host and add it to the host list or update an existing
         hosts. A host is updated if there is already one with the same
         canonicals facts and belonging to the same account.
+        <br /><br />
+        NOTICE: This operation is deprecated. We suggest you create and update hosts
```
jharting, line 68: As this is a public-facing API documentation a customer using this resource (e.g. in their scripts) won't be able to access the internal platform docs neither to use MQ insertion if they wanted. Let's instead say that explicit creation of hosts is no longer supported and that there is no substitute - hosts are created automatically based on uploads processed by the Payload Ingress Servyce (docs/api/ingress#operations-default-post_upload)

ryandillinfelton:
Redone with a relative link. Out of curiosity, is `api` or `api-docs` preferred? They seem to behave the same

PR #671, opened 2020-06-24
Add unit tests for EventProducer [RHCLOUD-7246]

ryandillinfelton:
This PR adds a happy path test for EventProducer to insure that we are giving KafkaProducer the correct information. This should help us avoid unintentionally breaking event production.

Edit: In addition to the happy path test there is now a test that ensures that an exception is raised if the event producer tries to write to an invalid topic (one not defined in our `Topic` enum).


test_unit.py
```diff
@@ -1623,5 +1629,55 @@ def test_deserialize_nested_no_key_error(self):
...
+
+    # Check that the event is always sent to the right topic
+    # Check that the key makes it through correct and encoded
+    # check that the headers make it through and are encoded
+    # check thet the event makes it through and is encoded
+    def test_happy_path(self):
+        event = build_event(EventType.created, self.basic_host)
+        key = self.basic_host["id"]
+
+        for event_type in EventType:
```
ryandillinfelton, line 1654: @tahmidefaz pointed out to me that I am not actually using the event type in the test. Trying to fix this brought to my attention that the build_event function expects the host to be a different type depending on the type of event being created. For create/update it is expected to be serialized when passed to the function, while delete expects the non-serialized host type. The executive code currently facilitates this, but I don't like the inconsistency so I will be trying to change it in this or another PR depending on how complex the change needs to be. I'll add another comment when I decide what to do about it in the current PR.

ryandillinfelton: For the current PR I decided to just break the test into two instead of changing how the EventProducer actually works.

test_unit.py
```diff
@@ -1623,5 +1629,61 @@ def test_deserialize_nested_no_key_error(self):
...
+        cls.topic_name = {Topic.events: cls.config.event_topic, Topic.egress: cls.config.host_egress_topic}
+        cls.event_types = {EventType.created: "created", EventType.updated: "updated", EventType.delete: "delete"}
+        threadctx.request_id = "-1"
+        cls.basic_host = {
+            "id": str(uuid4()),
+            "stale_timestamp": datetime.now(timezone.utc).isoformat(),
+            "reporter": "test_reporter",
+            "account": "test",
+            "fqdn": "fqdn",
+        }
```
tahmidefaz, line 1647: Is there a specific reason to use a `class method` here over a generic `instance method` or even plain `class variables`?

utils/kafka_consumer.py
```diff
@@ -25,7 +26,7 @@ def msg_handler(parsed):
 
 
 def main():
-    consumer = KafkaConsumer(HOST_EGRESS_TOPIC, group_id=HOST_INGRESS_GROUP, bootstrap_servers=BOOTSTRAP_SERVERS)
+    consumer = KafkaConsumer(HOST_PAYLOAD_TRACKER_TOPIC, bootstrap_servers=BOOTSTRAP_SERVERS)
```
tahmidefaz, line 30: revert? :slightly_smiling_face:

ryandillinfelton: Yeah, it sneaked in :-1:

test_unit.py
ryandillinfelton: It's a unittest feature https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass
It's required to be a class method

tahmidefaz:
@ryandillinfelton double check if this needed to be done differently after the recent test refactoring. Otherwise, this is :heavy_check_mark: for me.

test_unit.py
Glutexo: That doesn’t really answer the question though. Why this needs to be in _setUpClass_ instead of just _setUp_?

Glutexo:
Can you please link to a JIRA ticket in the description? If there is none, let’s create it.

Glutexo:
Merging.

PR #665, opened 2020-06-10
Send create/update events to platform.inventory.events topic (RHCLOUD-5223)

tahmidefaz:
This PR will enable sending create/update events to the `platform.inventory.events` topic by default. I added the feature to toggle this functionality using the `KAFKA_SECONDARY_TOPIC_ENABLED` env var. 

This PR is intended to be merged after #664

PR #672, opened 2020-06-25
Send create/update events to platform.inventory.events topic (RHCLOUD-5223)

tahmidefaz:
Closing #665 and opening this one instead.

This functionality can be toggled using the `KAFKA_SECONDARY_TOPIC_ENABLED` environment variable.
This feature is intended to be in place for around 2 months, after which the `platform.inventory.egress` topic will be removed.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -6,7 +6,6 @@ from app.environment import RuntimeEnvironment
 from app.logging import get_logger
 
 BulkQuerySource = Enum("BulkQuerySource", ("db", "xjoin"))
-KAFKA_SECONDARY_TOPIC_ENABLED = os.environ.get("KAFKA_SECONDARY_TOPIC_ENABLED", "false")
 
 
 class Config:
@@ -41,6 +40,7 @@ class Config:
         self.system_profile_topic = os.environ.get("KAFKA_TOPIC", "platform.system-profile")
         self.bootstrap_servers = os.environ.get("KAFKA_BOOTSTRAP_SERVERS", "localhost:29092")
         self.event_topic = os.environ.get("KAFKA_EVENT_TOPIC", "platform.inventory.events")
+        self.secondary_topic_enabled = True if os.environ.get("KAFKA_SECONDARY_TOPIC_ENABLED") == "true" else False
 
         self.prometheus_pushgateway = os.environ.get("PROMETHEUS_PUSHGATEWAY", "localhost:9091")
         self.kubernetes_namespace = os.environ.get("NAMESPACE")
@@ -122,6 +122,7 @@ class Config:
                 self.logger.info("Kafka Host Ingress Topic: %s" % self.host_ingress_topic)
                 self.logger.info("Kafka Host Ingress Group: %s" % self.host_ingress_consumer_group)
                 self.logger.info("Kafka Host Egress Topic: %s" % self.host_egress_topic)
+                self.logger.info("Kafka Secondary Topic Enabled: %s" % self.secondary_topic_enabled)
 
             if self._runtime_environment.event_producer_enabled:
                 self.logger.info("Kafka Event Topic: %s" % self.event_topic)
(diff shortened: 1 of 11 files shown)


app/queue/queue.py
```diff
@@ -153,6 +154,10 @@ def handle_message(message, event_producer):
         event = build_event(event_type, output_host, platform_metadata=platform_metadata)
         event_producer.write_event(event, output_host["id"], message_headers(add_results), Topic.egress)
 
+        # for transition to platform.inventory.events
+        if os.environ.get("KAFKA_SECONDARY_TOPIC_ENABLED") == "true":
```
jharting, line 158: Instead of reading the env variable here it would be better to add that to app/config.py - that way all the configuration is in one place

test_mq_service.py
```diff
@@ -125,6 +126,31 @@ def test_shutdown_handler(self):
...
+                "stale_timestamp": "2019-12-16T10:10:06.754201+00:00",
+                "reporter": "test",
+            },
+        }
+
+        # setting envionment variable to enable the secondary topic
+        with patch.dict("os.environ", {"KAFKA_SECONDARY_TOPIC_ENABLED": "true"}):
+            with self.app.app_context():
+                with unittest.mock.patch("app.queue.queue.add_host") as m:
+                    m.return_value = ({"id": uuid.uuid4(), "insights_id": None}, AddHostResult.created)
```
jharting, line 144: Let's parameterize this for both created and updated events to be tested

tahmidefaz:
Sadly, directly patching the env var doesn't work anymore because of the way the tests are set up. Instead, I patched the variable that's being supplied from the config.

@jharting does this look good now?

jharting:
@tahmidefaz what I meant was to align this configuration option with how other stuff is configured, i.e.

- make this a field of the Config class
- do the conversion to boolean when reading the env variable (in Config class)
- log the value on startup

tahmidefaz:
@jharting Initially, I tried to not add the field to the Config class because I didn't find the `configs` to be very accessible from within `handle_message()`. Here, I just tweaked our `EventProducer` a little bit to pass the field's value.

app/config.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, runtime_environment):
         self.system_profile_topic = os.environ.get("KAFKA_TOPIC", "platform.system-profile")
         self.bootstrap_servers = os.environ.get("KAFKA_BOOTSTRAP_SERVERS", "localhost:29092")
         self.event_topic = os.environ.get("KAFKA_EVENT_TOPIC", "platform.inventory.events")
+        self.secondary_topic_enabled = True if os.environ.get("KAFKA_SECONDARY_TOPIC_ENABLED") == "true" else False
```
jharting, line 43: `os.environ.get("KAFKA_SECONDARY_TOPIC_ENABLED") == "true"` is already a boolean expression. No need to wrap it in a conditional statement that returns boolean.

Also, it is a good practice for boolean env variables to be case insensitive i.e. no matter if set to "TRUE" or "true" it should still work as expected. See e.g. `PAYLOAD_TRACKER_ENABLED` for how it's done.


```suggestion
        self.secondary_topic_enabled = os.environ.get("KAFKA_SECONDARY_TOPIC_ENABLED", "false").lower() == "true"
```

app/queue/event_producer.py
```diff
@@ -18,6 +18,9 @@ def __init__(self, config):
         self._kafka_producer = KafkaProducer(bootstrap_servers=config.bootstrap_servers)
         self.topics = {Topic.egress: config.host_egress_topic, Topic.events: config.event_topic}
 
+        # for transition to platform.inventory.events
+        self.secondary_topic_enabled = config.secondary_topic_enabled
```
jharting, line 22: The logic for sending the event to two topics is all implemented in `queue.py`. The event_producer does not need to know anything about it - it is or is not asked to write_event also to the secondary topic by `queue.py`. That's a nice separation of concerns.

Storing `secondary_topic_enabled` though breaks that separation of concerns. The event_producer does not work with that value neither should be responsible for providing it. Instead, queue.py should obtain that configuration value directly from the config.

tahmidefaz:
^ Just opened that PR on `e2e-deploy` to enable this in CI.

PR #674, opened 2020-07-07
Disable REST POST [RHCLOUD-7544]

ryandillinfelton:
This PR adds an environment variable that can toggle the REST POST functionality on and of, and defaults to on. Although we aren't removing the operation from production until the 25th of August we want to be able to disable the functionality in CI/QA earlier, this will allow us to do so.


api/host.py
```diff
@@ -57,56 +57,60 @@
...
-                        "detail": "Could not complete operation",
-                        "host": host,
-                    }
-                )
-
-        rest_post_request_count.labels(reporter=reporter).inc()
-
-        response = {"total": len(response_host_list), "errors": number_of_errors, "data": response_host_list}
-        return flask_json_response(response, status=207)
+    print(inventory_config().rest_post_enabled)
```
jharting, line 60: print

api/host.py
```diff
@@ -57,56 +57,60 @@
...
+    if inventory_config().rest_post_enabled:
```
jharting, line 61: In situations like this it is better to exit early on negative condition and then leave the main body of the function do the possitive code path.
```
if not inventory_config().rest_post_enabled:
    return flask_json_response(...)

... rest of the logic remains untouched
```

The benefit is that this is more readable, the main body of the function does not need extra level of nesting and even the diff is tiny compared to this. See http://lecterror.com/articles/view/code-formatting-and-readability for more info on this.

api/host.py
```diff
@@ -57,56 +57,60 @@
...
+                            "host": host,
+                        }
+                    )
+
+            rest_post_request_count.labels(reporter=reporter).inc()
+
+            response = {"total": len(response_host_list), "errors": number_of_errors, "data": response_host_list}
+            return flask_json_response(response, status=207)
+    else:
+        return flask_json_response({}, status=410)
```
jharting, line 113: 410 mean the entire resource (`/hosts` in this case) is gone entirely (i.e. no matter which method you use the resource just is not there any longer) That's not the case here - we're removing an operation i.e. the ability to use a certain method (POST) on the given resource but the resource is not going away. 405 Method Not Allowed is therefore more appropriate.

In addition, the message body should not be empty. What do you get if you try to use e.g. PATCH on /hosts? The result should be consistent.

app/config.py
```diff
@@ -34,6 +34,8 @@ def __init__(self, runtime_environment):
 
         self.api_urls = [self.api_url_path_prefix, self.legacy_api_url_path_prefix]
 
+        self.rest_post_enabled = os.environ.get("REST_POST_ENABLED", "true").lower() == "true"
```
jharting, line 37: :+1:

api/host.py
ryandillinfelton: removed

api/host.py
ryandillinfelton: :+1:

PR #675, opened 2020-07-09
Migration logging to cloudwatch - RHCLOUD-4284

tahmidefaz:


tahmidefaz:
We only need `INFO` level logging for `sqlalchemy.engine` when running migrations, hence the requirement for the `configure_migrations_logging()`. 

**Question:** I moved over the `console` handler from the place where it was set up earlier. Do we really need it? I think we should be fine with `logstash` here, unless I am missing something.

logconfig.yaml
```diff
@@ -44,9 +51,15 @@ handlers:
     formatter: logstash
     filters:
     - contextual
+  console:
```
jharting, line 54: What is this handler for?

logconfig.yaml
```diff
@@ -44,9 +51,15 @@ handlers:
...
+    class: logging.StreamHandler
+    formatter: generic
+    stream: ext://sys.stderr
 formatters:
   human_readable:
     format: "[%(asctime)s] [%(process)d] [%(thread)d] [%(name)s] [%(levelname)s] %(message)s"
+  generic:
+    format: "%(levelname)-5.5s [%(name)s] %(message)s"
```
jharting, line 62: Do we two formats?

Makefile
```diff
@@ -7,7 +7,7 @@ test:
 	pytest --cov=.
 
 upgrade_db:
-	python manage.py db upgrade
+	MIGRATIONS_LOG_LEVEL=INFO python manage.py db upgrade
```
jharting, line 10: That's the default value, right?

logconfig.yaml
tahmidefaz: It was being used by the migrations earlier, and I moved it here from the old  `alembic.ini`. I don't think we need it, it'd be better to use `logstash` imo. What do you think?

logconfig.yaml
tahmidefaz: Yes, the migrations was using this format earlier, we can switch this to the `human_readable` format as well

logconfig.yaml
jharting: Agreed, let's have it produce logs in the same format than the rest

Makefile
tahmidefaz: Yes, it doesn't make a difference atm. I put it there to increase visibility.

logconfig.yaml
tahmidefaz: done! :+1:

logconfig.yaml
tahmidefaz: removed :+1:

jharting:
@tahmidefaz I changed this to work differently. Take a look at https://github.com/RedHatInsights/insights-host-inventory/pull/675/commits/f577b63673cd4fb602798ad4c99eca7bf4319e48 when you're back

PR #676, opened 2020-07-13
Set More Message Headers (RHCLOUD-5964)

tahmidefaz:
Closing #654 and opening this one instead.


tahmidefaz:
I still need to write tests for this and fix the test fails (approx. 30).

tahmidefaz:
Ok, I identified that some host information is getting lost after moving the serialization around this way. Currently looking into what's up. 
@Glutexo this maybe what you warned me about in the meeting today :point_up:

tahmidefaz:
I changed the way how this works, since moving the serialization around was getting complicated and was causing host data loss. I still need write tests and fix tests for this.

I also thought having an `insights_id` header rather than `registered_with_insights` would be more useful for consumers and it also looks much simpler.

jharting:
> I also thought having an insights_id header rather than registered_with_insights would be more useful for consumers and it also looks much simpler.

@tahmidefaz that's fine with me

tahmidefaz:
This should be ready for review. I am not sure why `nacho-bot` is angry.

lib/host_delete.py
```diff
@@ -16,14 +16,17 @@ def delete_hosts(select_query, event_producer):
     while select_query.count():
         for host in select_query.limit(CHUNK_SIZE):
             host_id = host.id
+            insights_id = host.canonical_facts["insights_id"] if "insights_id" in host.canonical_facts else ""
```
jharting, line 19: Why empty string and not None?

app/serialization.py
```diff
@@ -101,6 +101,8 @@ def serialize_host(host, staleness_timestamps, fields=DEFAULT_FIELDS):
 
     serialized_host = {**serialize_canonical_facts(host.canonical_facts)}
 
+    insights_id = host.canonical_facts["insights_id"] if "insights_id" in host.canonical_facts else ""
```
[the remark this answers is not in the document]
tahmidefaz: I could've misunderstood @Glutexo 's request here, but I think Štěpán suggested that we don't touch the serialized_host data at all. @Glutexo , is this what your suggestion was?

lib/host_delete.py
tahmidefaz: I tried that first, but all Kafka header values need to be byte strings. Sadly, we cannot perform `encode()` on `NoneTypes`

app/serialization.py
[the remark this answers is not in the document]
jharting: That's true. Why put this code into `serialize()`, though?

lib/host_delete.py
jharting: At this point this should either be a uuid string or None. What you're describing is kafka-specific serialization requirement. Any kafka-specific serialization of headers should go into `event_producer.write_event`. That way the rest of the codebase remains neutral.

app/serialization.py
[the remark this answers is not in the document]
Glutexo: Wherever you can call _serialize_host_ you have the Host object and its _insights_id_. You need the extracted _insights_id_ only when dealing with event emission. My talks about tuples were meant only to avoid the DB session problem. I agree with @jharting that it doesn’t belong to the serialization function.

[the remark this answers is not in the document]
Glutexo: It belongs to _lib.host_repository.add_host_. Something like this. Untested, you get the idea.

```python
def add_host(input_host, staleness_offset, update_system_profile=True, fields=DEFAULT_FIELDS):
    with session_guard(db.session):
        existing_host = find_existing_host(input_host.account, input_host.canonical_facts)
        if existing_host:
            insights_id = existing_host.insights_id
            output_host = update_existing_host(existing_host, input_host, staleness_offset, update_system_profile, fields)
        else:
            insights_id = input_host.insights_id
            output_host = create_new_host(input_host, staleness_offset, fields)
    return output_host, insights_id
```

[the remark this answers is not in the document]
tahmidefaz: Ah, I see! Ok, I will change this. Thanks, all!

tahmidefaz:
Ok, I moved the logic around for fetching `insights_id` and moved the conversion of NoneType `insights_id` to `event_producer.write_event`. :+1: 

This is ready for review again!

app/queue/queue.py
```diff
@@ -113,7 +113,7 @@ def add_host(host_data):
                     }
                 },
             )
-            (output_host, add_result) = host_repository.add_host(
+            (output_host, insights_id, add_result) = host_repository.add_host(
```
Glutexo: We can retrieve Host ID here too so we don’t need to extract it from the serialized payload.

```suggestion
            output_host, host_id, insights_id, add_result = host_repository.add_host(
```

app/queue/queue.py
```diff
@@ -126,7 +126,7 @@ def add_host(host_data):
                 extra={"host": {i: output_host[i] for i in output_host if i not in ("facts", "system_profile")}},
             )
             payload_tracker_processing_ctx.inventory_id = output_host["id"]
-            return (output_host, add_result)
+            return (output_host, insights_id, add_result)
```
Glutexo: If we retrieve the Host ID above, we need to pass it out here too.

```suggestion
            return output_host, host_id, insights_id, add_result
```

app/queue/queue.py
```diff
@@ -150,14 +150,16 @@ def handle_message(message, event_producer):
     with PayloadTrackerContext(
         payload_tracker, received_status_message="message received", current_operation="handle_message"
     ):
-        (output_host, add_results) = add_host(validated_operation_msg["data"])
+        (output_host, insights_id, add_results) = add_host(validated_operation_msg["data"])
```
Glutexo: Here, the added Host ID is returned.

```suggestion
        output_host, host_id, insights_id, add_results = add_host(validated_operation_msg["data"])
```

app/queue/queue.py
```diff
@@ -150,14 +150,16 @@ def handle_message(message, event_producer):
...
         event_type = add_host_results_to_event_type(add_results)
         event = build_event(event_type, output_host, platform_metadata=platform_metadata)
-        event_producer.write_event(event, output_host["id"], message_headers(add_results), Topic.egress)
+        event_producer.write_event(event, output_host["id"], message_headers(add_results, insights_id), Topic.egress)
 
         # for transition to platform.inventory.events
         if inventory_config().secondary_topic_enabled:
-            event_producer.write_event(event, output_host["id"], message_headers(add_results), Topic.events)
+            event_producer.write_event(
+                event, output_host["id"], message_headers(add_results, insights_id), Topic.events
```
Glutexo:
```suggestion
                event, str(host_id), message_headers(add_results, insights_id), Topic.events
```

api/host.py
```diff
@@ -90,7 +90,7 @@ def add_host_list(body):
                         logger.info("Tags from an HTTP request were ignored")
 
                     input_host = deserialize_host_http(host)
-                    (output_host, add_result) = _add_host(input_host)
+                    (output_host, _, add_result) = _add_host(input_host)
```
Glutexo: Here, we can retrieve the Host ID. In contrast to the Insights ID that is not useful here.

```suggestion
                    output_host, host_id, _, add_result = _add_host(input_host)
```

api/host.py
```diff
@@ -90,7 +90,7 @@ def add_host_list(body):
...
                     status_code = _convert_host_results_to_http_status(add_result)
                     response_host_list.append({"status": status_code, "host": output_host})
                     payload_tracker_processing_ctx.inventory_id = output_host["id"]
```
Glutexo: This too extracts information from the serialized payload. If we retrieve the Host ID above, we can use it here.

```suggestion
                    payload_tracker_processing_ctx.inventory_id = host_id
```

lib/host_delete.py
```diff
@@ -23,7 +23,10 @@ def delete_hosts(select_query, event_producer):
             if host_deleted:
                 delete_host_count.inc()
                 event = build_event(EventType.delete, host)
-                event_producer.write_event(event, str(host.id), message_headers(EventType.delete), Topic.events)
+                insights_id = host.canonical_facts["insights_id"] if "insights_id" in host.canonical_facts else None
```
Glutexo: _dict.get_ does exactly the same thing.

```suggestion
                insights_id = host.canonical_facts.get("insights_id")
```

Glutexo:
Yet @tahmidefaz can’t give a ✅, because this is his pull request. If @jharting or @shdunning checks that I didn’t break anything, that would be awesome.

tahmidefaz:
Tested this locally, looks like it's working as expected. :+1:  :heavy_check_mark:

PR #678, opened 2020-07-14
Final touches to unit test refactor step 1

Glutexo:
Two little follow-up changes to #655.

- Consolidated abstract test names, so they all end with _BaseTestCase_.
- Extracted the Management test case from the Utils file.

The database locking should not be a problem. When an exception is raised, Pytest still tears down the fixtures gracefully. The problems happens only if the process is killed unconditionally before it even manages to do the teardown. Nothing we should care about.

Resolves [RHCLOUD-7870](https://projects.engineering.redhat.com/browse/RHCLOUD-7870), a subtask of [RHCLOUD-7076](https://projects.engineering.redhat.com/browse/RHCLOUD-7076).


Glutexo:
Thanks @rodrigonull for the approval! One more? @jharting

PR #677, opened 2020-07-13
Add sap facts sap_system and sap_local_instances

CodeHeeler:

PR #679, opened 2020-07-14
Add sap field  [RHCLOUD-7696]

ryandillinfelton:
Adds `sap_system` fact to `system_profile`. As well as tests to show it appears in MQ event and when querying system profile through the REST API.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ test:
 	pytest --cov=.
 
 upgrade_db:
-	python manage.py db upgrade
+	SQLALCHEMY_ENGINE_LOG_LEVEL=INFO python manage.py db upgrade
 
 run_inv_web_service:
 	# Set the "KAFKA_TOPIC", "KAFKA_GROUP", "KAFKA_BOOTSTRAP_SERVERS" environment variables
--- a/Pipfile
+++ b/Pipfile
@@ -15,10 +15,9 @@ flask-sqlalchemy = "~=2.4.0"
 psycopg2-binary = "~=2.8.3"
 flask-migrate = "~=2.5.2"
 itsdangerous = "~=1.1.0"
-request = "==2019.4.13"  # Not semver
 gunicorn = "~=19.9.0"
 connexion = {version = "~=2.6.0",extras = ["swagger-ui"]}
-pyyaml = "~=5.1.2"
+pyyaml = "~=5.3.1"
 prometheus-client = "~=0.7.1"
 logstash-formatter = "~=0.5.17"
 validators = "~=0.13.0"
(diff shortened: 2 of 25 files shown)


ryandillinfelton:
Pipenv install is failing because of a dependency that was removed from pypi
There's a PR out to fix it: https://github.com/RedHatInsights/insights-host-inventory/pull/681

CodeHeeler:
Someone pointed out in our PR that you are doing parallel work.
https://github.com/RedHatInsights/insights-host-inventory/pull/680

Would you be willing and able to add the SAP object here instead of just the plain bool as that is what we all need to be using as per relevant email discussions?  We really need the object code but like your tests which would need to be adapted to it.

If not, please let us know and maybe there's a way we could split the work across the two PRs in a way that would merge nicely.

jharting:
@CodeHeeler can you elaborate why it needs to be an object? The precedent so far in system profile is to use flat keys/values (except for arrays). Even multiple sap-related keys can be represented without using object e.g.
```
sap_system: true,
sap_local_instances: ...
```

CodeHeeler:
If I understand correctly, in conversation with Jerome Marc, @jhjaggars requested that we created a group or collection of the SAP facts to work with which is why we closed the original PR (https://github.com/RedHatInsights/insights-host-inventory/pull/677) in favor of this current one.

beav:
> @CodeHeeler can you elaborate why it needs to be an object? The precedent so far in system profile is to use flat keys/values (except for arrays). Even multiple sap-related keys can be represented without using object e.g.
> 
> ```
> sap_system: true,
> sap_local_instances: ...
> ```

We got feedback from @jhjaggars that object is better (I heard the feedback 2nd hand, so possibly it was to reduce the amount of top-level stuff in the profile but not 100% sure). However, we can keep them as multiple top-level keys and then collapse them down in the comparison report easily if needed. We are happy to do whichever works for yall :cowboy_hat_face:

jhjaggars:
@jharting asked me a great question just a bit ago: "What is it like to query a nested object?".  Would it be easier to query a flattened keyspace?  I tend towards an object to keep the global namespace smaller, but it might not be worth it if querying is difficult.

beav:
> @jharting asked me a great question just a bit ago: "What is it like to query a nested object?". Would it be easier to query a flattened keyspace? I tend towards an object to keep the global namespace smaller, but it might not be worth it if querying is difficult.

In our app, we currently access it via a python object that's created from the JSON, so it's just a matter of `profile["sap"]["local_instances"]` or `profile["sap_local_instances"]`. In the future we're going to try transforming with jsonata, but I'm pretty sure it would be the same or similar in how we reference it.

If anyone is querying via postgres JSON operators it might matter more.

jharting:
@beav @CodeHeeler we'll go ahead with sap_system to unblock Cyndi work. If more fields should be added please add those in a follow-up PR to https://github.com/redhatinsights/inventory-schemas and here.

PR #681, opened 2020-07-15
removed unused dependency

ryandillinfelton:
This package was removed from pypi and it looks like we never use it in the first place. This updated the pipfiles to remove it


ryandillinfelton:
Seems like any way I can remove the package with pipenv commands will result in all the dependencies updating.
I've tried `--keep-outdated` and got no error, but it still updated everything else. Let me know if that's a big problem, I might be able to edit the lockfile manually, but I'd rather avoid that if possible.

jharting:
@ryandillinfelton yes it's a common problem with pipenv. I am not aware of a workaround besides locking down the dependency versions. Let's see what happens.

PR #682, opened 2020-07-16
RHCLOUD-7520 Extract System Profile schema to a separate file

Glutexo:
Extracted System Profile schema from the OpenAPI specification file. [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03) is used to link the external schema to the main document.

[prance](https://pypi.org/project/prance/) library used to dereference and compile the schema into a single specification object.

For the schema to be valid for OpenAPI, the _/$def/SystemProfile/$id_ property had to be removed. I’ll remove it from _inventory-schemas_ too. Also the _x-body-name_ Connexion proprietary parameter had to be removed.

A part of [RHCLOUD-7520](https://projects.engineering.redhat.com/browse/RHCLOUD-7520).


Glutexo:
Hmmm… Facts tests are broken. Let’s see why.

Glutexo:
Okay. The culprit is that _prance_ doesn’t recognize a proprietary _x-body-value_ Connexion attribute. Sweet.

Glutexo:
Fixed.

app/__init__.py
```diff
@@ -35,9 +40,16 @@ def inventory_config():
 
 
 def create_app(runtime_environment):
+    # This feels like a hack but it is needed.  The logging configuration
+    # needs to be setup before the flask app is initialized.
+    configure_logging()
+
+    app_config = Config(runtime_environment)
```
jharting, line 47: is this supposed to be done twice?

app/__init__.py
```diff
@@ -35,9 +40,16 @@ def inventory_config():
...
+    app_config.log_configuration()
+
     connexion_options = {"swagger_ui": True}
 
-    # This feels like a hack but it is needed.  The logging configuration
+    # This feels like a hack but it is needed.  The loggingconfiguration
```
jharting, line 52: loggingconfiguration?

Glutexo: Hehe.

app/__init__.py
Glutexo: Oh noez, rebase conflict artifact.

jharting: @Glutexo This conversation is marked resolved but nothing changed. Did you forget to git push?

Glutexo: Resolved as in “All understood” not as in “Fixed and pushed”. 😃

Glutexo:
Thanks for spotting out those dumb bugs, @jharting. Hopefully fixed now!

logconfig.yaml
```diff
@@ -22,7 +22,7 @@ loggers:
     - logstash
     propagate: true
   sqlalchemy.engine:
-    level: WARNING
+    level: INFO
```
jharting, line 25: :warning: :arrow_right: :information_source: :question:

Glutexo: Good catch, thanks!

Glutexo:
Fixed that log configuration change committed by mistake. I wonder whether there is a way to ignore changes to files already committed to repository. Something like _.gitignore_, but for existing files.

Glutexo:
🎉  Yay. Thanks!

PR #683, opened 2020-07-17
Internal openapi references resolution

rodrigonull:
With the system profile schema separation we adopted a new lib to parse the swagger specification.

This new lib messed up the internal references on the generated openapi spec: https://qa.cloud.redhat.com/api/inventory/v1/openapi.json

To avoid this internal reference problem, let's make the lib explicitly only resolve references from files.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -6,6 +6,7 @@ from flask import current_app
 from flask import jsonify
 from flask import request
 from prance import ResolvingParser
+from prance.util.resolver import RESOLVE_FILES
 from prometheus_flask_exporter import PrometheusMetrics
 
 from api.mgmt import monitoring_blueprint
@@ -52,7 +53,7 @@ def create_app(runtime_environment):
     connexion_app = connexion.App("inventory", specification_dir="./swagger/", options=connexion_options)
 
     # Read the swagger.yml file to configure the endpoints
-    parser = ResolvingParser(SPECIFICATION_FILE)
+    parser = ResolvingParser(SPECIFICATION_FILE, resolve_types=RESOLVE_FILES)
     parser.parse()
 
     for api_url in app_config.api_urls:
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -683,7 +683,7 @@ components:
           items:
             $ref: '#/components/schemas/FactSet'
         system_profile:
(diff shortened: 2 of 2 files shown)


Glutexo:
Thanks @jharting for the additional commit. I see that it’s not only visually DRY, but the composed file then has the System Profile included only once. 👍🏻

rodrigonull:
> I see the difference the argument makes and it makes sense to me. Connexion handles local references correctly, but not the external ones. We use _prance_ to resolve only that issue.
> 
> But still I have a question, @rodrigonull. What actual problems does this excessive resolving cause? I see some increase in memory usage. May I ask you for steps to reproduce anything that doesn’t work with the old settings?

@Glutexo The auto-generated openapi client provide us objects that we can use in our automated tests, like this one for instance: https://gitlab.cee.redhat.com/insights-qe/iqe-host-inventory-plugin/-/blob/master/iqe_host_inventory_api/models/host_out.py.

The changes made the openapi client generator stop creating those objects and all we got from the requests to HBI after generating the client were dicts instead of the python objects.

before the changes we could do something like this: assert response.results[0].display_name == "test" and after the changes we need to do something like this: assert response.results[0]["display_name"] == "test"

Glutexo:
@rodrigonull The representation should be equivalent though. Did you investigate what exactly leads the OpenAPI Generator to treat the objects differently?

Glutexo:
@rodrigonull Moreover, I see that the modules are not generated at runtime. If so, the composition used to feed the generator should be independent on the _prance_ way used in the app initialization. Am I missing something here?

rodrigonull:
> @rodrigonull The representation should be equivalent though. Did you investigate what exactly leads the OpenAPI Generator to treat the objects differently?

@Glutexo as I mentioned before the openapi generator uses the: https://qa.cloud.redhat.com/api/inventory/v1/openapi.json to generate the client and as you can see here: https://www.diffchecker.com/JwODMrQz, there is a big difference between the openapi.json files. If you don't set he option: "resolve_types=RESOLVE_FILES" it generates a openapi.json file with 5k+ lines and if you generate with the mentioned option, it generates a openapi.json file like its was before, with around 1.9k lines.

I think the problem was caused because this new lib, when it is being executed with the default options, does not make the appropriate references to the schema components and ends up duplicating the properties of those components instead of referring to them.

PR #684, opened 2020-07-20
Monitor event producer errors [RHCLOUD-7400]

ryandillinfelton:


app/queue/metrics.py
```diff
@@ -28,12 +28,12 @@
 )
 version = Info("inventory_mq_service_version", "Build version for the inventory message queue service")
 version.info({"version": get_build_version()})
-egress_message_handler_success = Counter(
-    "inventory_egress_message_handler_successes", "Total amount of messages successfully written to the egress queue"
+event_producer_success = Counter(
+    "inventory_event_producer_successes", "Total amount of messages successfully written", ["event_type", "topic"]
```
jharting, line 32: Did you mean to use these labels?

app/__init__.py
```diff
@@ -30,6 +34,13 @@
 SPECIFICATION_FILE = join(SPECIFICATION_DIR, "api.spec.yaml")
 
 
+def initialize_metrics():
+    for event_type in EventType:
+        for topic in Topic:
+            event_producer_failure.labels(event_type=event_type, topic=topic)
```
jharting, line 40: This should initialize the topic with actual topic name, not ```topic="Topic.events"```.

app/__init__.py
```diff
@@ -30,6 +34,14 @@
 SPECIFICATION_FILE = join(SPECIFICATION_DIR, "api.spec.yaml")
 
 
+def initialize_metrics(config):
+    topic_names = {Topic.egress: config.host_egress_topic, Topic.events: config.event_topic}
+    for event_type in EventType:
+        for topic in Topic:
+            event_producer_failure.labels(event_type=event_type, topic=topic_names[topic])
+            event_producer_success.labels(event_type=event_type, topic=topic_names[topic])
```
jharting, line 42: The intialized event type label still does not match the value that's used at runtime:
```
event_type="EventType.created"
```
vs
```
event_type="created"
```

Glutexo:
Like, really?

tests/test_api_hosts_update.py
```diff
@@ -188,6 +188,53 @@ def test_patch_produces_update_event_no_insights_id(self):
         with self.app.app_context():
             self._base_patch_produces_update_event_test(created_host, {}, ANY)
 
+    def test_event_producer_instrumentation(self):
+        from unittest.mock import Mock
+        from unittest.mock import ANY
```
Glutexo, line 193: Imports should be at the top of the file. How could this even go through our linter?

app/__init__.py
```diff
@@ -30,6 +34,14 @@
 SPECIFICATION_FILE = join(SPECIFICATION_DIR, "api.spec.yaml")
 
 
+def initialize_metrics(config):
+    topic_names = {Topic.egress: config.host_egress_topic, Topic.events: config.event_topic}
+    for event_type in EventType:
+        for topic in Topic:
```
Glutexo, line 40: [_itertools.product_](https://docs.python.org/ja/3.6/library/itertools.html#itertools.product) may help here.

tests/test_api_hosts_update.py
jharting: @Glutexo interestingly it looks like flake8 does not actually detect

tests/test_api_hosts_update.py
```diff
@@ -188,6 +188,53 @@ def test_patch_produces_update_event_no_insights_id(self):
...
+        with patch("app.queue.event_producer.KafkaProducer", **{"return_value.send.return_value": mock_future}):
+            self.app.event_producer = EventProducer(self.app.config["INVENTORY_CONFIG"])
+
+        patch_doc = {"display_name": "patch_event_test"}
+        host_to_patch = self.added_hosts[0].id
+
+        with self.app.app_context():
+            with patch("app.queue.events.datetime", **{"now.return_value": self.now_timestamp}):
+                with patch("app.queue.event_producer.message_produced") as message_produced:
+                    with patch("app.queue.event_producer.message_not_produced") as message_not_produced:
```
Glutexo, line 231: The three _app.queue.*_ patch statements should be possible to contract to a single _patch.multiple_ statement. It would prevent this four level deep nesting.

Also, [_with_](https://docs.python.org/3.6/reference/compound_stmts.html#the-with-statement) statement support multiple items. `with x: with y:` can be replaced with `with x, y:` with only one level nesting.

tests/test_api_hosts_update.py
Glutexo: There is [E402](https://www.flake8rules.com/rules/E402.html), but it fires only if the import is top-level. It may make sense (mainly performance driven) to import inside a function. This is however not the case. Mocks are generic things used in tests and we always import them at the thop.

PR #685, opened 2020-07-22
Database name set by environment variable

rodrigonull:
https://projects.engineering.redhat.com/browse/RHCLOUD-8079

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
 from sqlalchemy_utils import create_database
 from sqlalchemy_utils import database_exists
 from sqlalchemy_utils import drop_database
@@ -7,6 +8,14 @@ from app import create_app
 from app import db
 from app.config import Config
 from app.config import RuntimeEnvironment
+
+
+@pytest.fixture(scope="session", autouse=True)
+def database_name():
+    patch = MonkeyPatch()
+    patch.setenv("INVENTORY_DB_NAME", "insights_test")
+    yield patch
+    patch.undo()
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -1,4 +1,5 @@
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
```
Glutexo, line 2: This uses stuff from _pytest_ private space. See the underscore. I am aware that the public _monkeypatch_ fixture is not usable with the session scope, yet this doesn’t seem right either.

tests/conftest.py
```diff
@@ -9,6 +10,14 @@
 from app.config import RuntimeEnvironment
 
 
+@pytest.fixture(scope="session", autouse=True)
+def database_name():
+    patch = MonkeyPatch()
+    patch.setenv("INVENTORY_DB_NAME", "insights_test")
+    yield patch
+    patch.undo()
```
Glutexo, line 18: We can use our [_set_environment_](https://github.com/rodrigonull/insights-host-inventory/blob/304d26d3ab045c82f839b9506dd7635ce7d6c83a/tests/test_utils.py#L7) utility function. It’s not the best one out there, but we use it already in some tests. It’d be at least consistent.

```suggestion
    from tests.test_utils import set_environment
    with set_environment({"INVENTORY_DB_NAME": "insights_test"}):
       yield
```

tests/conftest.py
```diff
@@ -9,6 +10,14 @@
 from app.config import RuntimeEnvironment
 
 
+@pytest.fixture(scope="session", autouse=True)
```
Glutexo, line 13: Is there a reason for the autouse? I see a possible increase of safety. Yet we don’t need to set the database name if we don’t use the database.

```suggestion
@pytest.fixture(scope="session")
```

rodrigonull:
Thanks for the suggestions @Glutexo 
I forgot about that we already had a set_environment function.

The PR was updated with your suggestions.

rodrigonull:
Closing this one since this change was merged with: https://github.com/RedHatInsights/insights-host-inventory/pull/656

PR #687, opened 2020-07-24
minor tweak to Makefile

jharting:
- run mq service with platform.inventory.events locally
- add alias for host reaper


Glutexo:
LGTM

I personally don’t care about Makefile. I wouldn’t be even mad if it just disappeared. Tested your changes nevertheless and they work fine. 👍🏻

PR #688, opened 2020-07-27
Graceful Shutdown API - RHCLOUD-5357

tahmidefaz:
I found [atexit](https://docs.python.org/3/library/atexit.html) that waits for requests to finish and also complies with gunicorn's own graceful shutdown handler.

I had a couple of questions regarding this PR:

1. I am not sure how to test this. Any ideas? Can we even test this, given the way our tests are configured atm?
2. Is `engine.dispose()` the correct way to close the db?


tahmidefaz:
Now... tests are failing. Need to investigate :man_facepalming:

app/__init__.py
```diff
@@ -110,6 +110,14 @@ def set_request_id():
             "The message queue based event notifications have been disabled."
         )
 
+    @atexit.register
```
jharting, line 113: since you're using atexit it would be better to add the shutdown logic hook registration near where the given object is initialized, e.g.

```python
    db.init_app(flask_app)

    def close_db():
        logger.info("Closing Database")
        db.get_engine(flask_app).dispose()

    atexit.register(close_db)
```

and 

```python
    if runtime_environment.event_producer_enabled:
        flask_app.event_producer = EventProducer(app_config)

        def close_event_producer():
            logger.info("Closing EventProducer()")
            flask_app.event_producer.close()

        atexit.register(close_event_producer)
```

jharting: Even better, you could create a higher-order function for creating shutdown handlers which, given the close function and a component name (for logging) creates a function to be registered with atexit. As a result you can specify the shutdown logic with a one-liner just next to where a given component is initialized.

```python
    if runtime_environment.event_producer_enabled:
        flask_app.event_producer = EventProducer(app_config)
        atexit.register(shutdown_hook(flask_app.event_producer.close, "event_producer"))

    else:
        logger.warning(
            "WARNING: The event producer has been disabled.  "
            "The message queue based event notifications have been disabled."
        )
```

jharting: the benefit of collocating create/destroy logic is that once we move e.g. event_producer creation out of application the shutdown logic can go with it easily

tahmidefaz:
Weird how it had a :x: before and now a :heavy_check_mark: after the latest test refactoring merge :thinking:

gunicorn.conf.py
```diff
@@ -1,5 +1,9 @@
+import os
+import tempfile
+
 from prometheus_client import multiprocess
 
 
 def child_exit(server, worker):
-    multiprocess.mark_process_dead(worker.pid)
+    multiproc_path = os.environ.get("prometheus_mutliproc_dir", tempfile.gettempdir() + "/prometheus-multiproc-dir/")
```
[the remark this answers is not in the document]
tahmidefaz: Yeah, this should work as-is on Openshift, since we set the env variable there. Removing

app/__init__.py
```diff
@@ -110,6 +110,14 @@ def set_request_id():
...
+    def pre_shutdown_cleaner():
+        if runtime_environment.event_producer_enabled:
+            logger.info("Closing EventProducer()")
+            flask_app.event_producer.close()
+            logger.info("Closing Database")
+            db.get_engine(flask_app).dispose()
```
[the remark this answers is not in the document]
tahmidefaz: changed this to work with `shutdown_hook()`

app/__init__.py
tahmidefaz: Went with the `shutdown_hook()` idea. I also put the `db dispose` in a if-statement since we don't have logging enabled in testing.

app/__init__.py
```diff
@@ -96,6 +101,9 @@ def create_app(runtime_environment):
 
     db.init_app(flask_app)
 
+    if runtime_environment.logging_enabled:
```
jharting, line 104: how is db shutdown related to logging?

app/__init__.py
```diff
@@ -48,13 +49,17 @@ def render_exception(exception):
     return response
 
 
+def shutdown_hook(close_function, name):
+    logger.info(f"Closing {name}")
```
jharting, line 53: Logger has its own syntax for interpolation. No need to use fstrings

app/__init__.py
tahmidefaz: hmm, I first I thought [CreateAppConfigTestCase.test_config_is_assigned](https://github.com/RedHatInsights/insights-host-inventory/blob/master/tests/test_unit.py#L286) was failing because of logging, but looking back again I realized that I might have to update the test.

tahmidefaz:
Ok, that test needed a `db_uri`, should be good now :+1:

app/__init__.py
```diff
@@ -104,6 +111,7 @@ def set_request_id():
 
     if runtime_environment.event_producer_enabled:
         flask_app.event_producer = EventProducer(app_config)
+        atexit.register(shutdown_hook, flask_app.event_producer.close, "EventProducer()")
```
shdunning, line 114: a bit of a nit-pick, but why `EventProducer()` instead of `EventProducer`? I'm trying to imagine myself reading through the logs in prod.

tahmidefaz: EventProducer would be more user friendly. Changed :+1:

tests/test_unit.py
```diff
@@ -283,7 +283,13 @@ def test_config_development_settings(self):
             self.assertEqual(conf.db_pool_timeout, 3)
 
 
-@patch("app.Config", **{"return_value.mgmt_url_path_prefix": "/"})
+@patch(
+    "app.Config",
+    **{
+        "return_value.mgmt_url_path_prefix": "/",
+        "return_value.db_uri": "postgresql://insights:insights@localhost/insights",
```
Glutexo, line 290: Even if this is not used for any destructive operation, I still feel better if we don’t use the “production” database name.

```suggestion
        "return_value.db_uri": "postgresql://insights:insights@localhost/insights_test",
```

But! I don’t think it is the right thing to do. This tests should only verify that the configuration has been loaded correctly. Now, the _atexit_ handler causes SQLAlchemy to do its teardown, potentially touching the database. I’d rather prevent somehow the teardown from being triggered.

Patching _db.get_engine_ should do the trick:

```python
@patch("app.db.get_engine")
```

I was thinking about adding some method into [_RuntimeEnvironment_](https://github.com/RedHatInsights/insights-host-inventory/blob/8f5c54164bbaff92a305114776f3ac3c761c9b10/app/environment.py). But we want to prevent it only in this specific tests. API tests use the database and thus should even do a proper teardown.

tahmidefaz:
Looks good to me :+1:

Glutexo:
One more? 😃 @shdunning , @ryandillinfelton

Glutexo:
Or maybe @rodrigonull can take a look at the PR and the new patch?

Glutexo:
As for your questions, @tahmidefaz:

- I took a look at _engine.dispose_ and it looks like it does the correct thing! 👍🏻
- For the testing… In this case it may be possible to emulate the signal being sent and see that the application does what it should. It may not be even necessary to try to spawn a new process manually, it looks like tests are run in separate processes anyways. But I am not sure. It would need some investigation and that would be over the scope of this PR.

tahmidefaz:
@Glutexo Yes, let's do that in a different PR. :+1:

Glutexo:
@shdunning, @ryandillinfelton, @rodrigonull May I ask somebody of you to tak a look at the new changes in this pull request?

PR #690, opened 2020-07-29
Graceful Shutdown Handler for Reaper - RHCLOUD-5358

tahmidefaz:


tahmidefaz:
This is the least invasive way I can think of, atm. Let me know if this can be improved.

tests/test_culling.py
```diff
@@ -304,7 +304,13 @@ def test_culled_host_is_removed(
     assert db_get_host(created_host.id)
 
     threadctx.request_id = UNKNOWN_REQUEST_ID_VALUE
-    host_reaper_run(inventory_config, mock.Mock(), db.session, event_producer_mock)
+    host_reaper_run(
+        inventory_config,
+        mock.Mock(),
+        db.session,
+        event_producer_mock,
+        shutdown_handler=mock.Mock(**{"shut_down.side_effect": (False, False)}),
```
Glutexo, line 312: We can even go with _return_value_ here. We are not testing the shutdown handler, so no interruptions at all.

```suggestion
        shutdown_handler=mock.Mock(**{"shut_down.return_value": False}),
```

tests/test_culling.py
```diff
@@ -333,14 +339,46 @@ def test_non_culled_host_is_not_removed(
     assert created_host_ids == [host.id for host in retrieved_hosts]
 
     threadctx.request_id = UNKNOWN_REQUEST_ID_VALUE
-    host_reaper_run(inventory_config, mock.Mock(), db.session, event_producer_mock)
+    host_reaper_run(
+        inventory_config,
+        mock.Mock(),
+        db.session,
+        event_producer_mock,
+        shutdown_handler=mock.Mock(**{"shut_down.side_effect": (False)}),
```
Glutexo, line 347: Same here.

```suggestion
        shutdown_handler=mock.Mock(**{"shut_down.return_value": False}),
```

tests/test_culling.py
```diff
@@ -333,14 +339,46 @@ def test_non_culled_host_is_not_removed(
...
+):
+    staleness_timestamps = get_staleness_timestamps()
+    created_hosts = []
+
+    for _ in range(3):
+        host = minimal_db_host(stale_timestamp=staleness_timestamps["culled"].isoformat(), reporter="some reporter")
+        created_hosts.append(db_create_host(host))
+
+    created_host_ids = [host.id for host in created_hosts]
+    assert db_get_hosts(created_host_ids)
```
Glutexo, line 368: For symmetry and to be sure that we are doing the right thing, let’s assert the exact count here as well.

```suggestion
    retrieved_hosts = db_get_hosts(created_host_ids)
    assert retrieved_hosts.count() == 3
```

tests/test_culling.py
```diff
@@ -333,14 +339,46 @@ def test_non_culled_host_is_not_removed(
...
     assert created_host_ids == [host.id for host in retrieved_hosts]
     assert event_producer_mock.event is None
 
 
+@pytest.mark.host_reaper
+def test_reaper_shutdown_handler(
+    event_producer_mock, event_datetime_mock, db_create_host, db_get_hosts, inventory_config
+):
+    staleness_timestamps = get_staleness_timestamps()
+    created_hosts = []
```
Glutexo, line 361: We can collect only IDs here and then there would be no need to extract them from this list.

tests/test_culling.py
```diff
@@ -333,14 +339,46 @@ def test_non_culled_host_is_not_removed(
...
+
+    created_host_ids = [host.id for host in created_hosts]
+    assert db_get_hosts(created_host_ids)
+    threadctx.request_id = UNKNOWN_REQUEST_ID_VALUE
+    host_reaper_run(
+        inventory_config,
+        mock.Mock(),
+        db.session,
+        event_producer_mock,
+        shutdown_handler=mock.Mock(**{"shut_down.side_effect": (False, False, True)}),
```
Glutexo, line 375: If the interrupt check is moved to the end of the loop, this needs to be changed.

```suggestion
        shutdown_handler=mock.Mock(**{"shut_down.side_effect": (False, True)}),
```

lib/host_delete.py
```diff
@@ -12,9 +12,12 @@
 CHUNK_SIZE = 1000
 
 
-def delete_hosts(select_query, event_producer):
+def delete_hosts(select_query, event_producer, reaper_shutdown_handler=None):
```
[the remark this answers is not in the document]
tahmidefaz: The gunicorn graceful shutdown handler already does this for us. It waits for the current request to finish and then gracefully shuts it down. So, the graceful shutdown PR for the API should be enough.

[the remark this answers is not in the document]
Glutexo: As already mentioned in the chit-chat: Then I’d like to see some state class in the flask app that could say that it is shutting down. This could be than used in the host delete. Something like

```python
# app.__init__
def create_app(...):
    ...
    flask_app.shutdown_handler = ShutdownHandler()
    ...
```

tahmidefaz:
I didn't had a chance to look at these today. I'll take a look first thing tomorrow.

host_reaper.py
```diff
@@ -86,14 +88,21 @@ def main(logger):
...
         with session_guard(session):
-            run(config, logger, session, event_producer)
+            run(config, logger, session, event_producer, shutdown_handler)
     finally:
         try:
             job = _prometheus_job(config.kubernetes_namespace)
             push_to_gateway(config.prometheus_pushgateway, job, registry)
         finally:
+            logger.info("Closing Database")
+            session.get_bind().dispose()
```
jharting, line 104: if an error is raised closing db connections that should not prevent EventProducer from being closed

Glutexo: Good point!

Glutexo: If we use _atexit_ here to register the handlers, would it work?

Glutexo: I experimented a bit with this and it should work. Pushed a [patch](https://github.com/Glutexo/insights-host-inventory/commit/d9638302a4a0a44dae71e0faad3df56b30a24260) to [my fork](https://github.com/Glutexo/insights-host-inventory/) as a proof of concept with the registration being DRIED into _lib.handlers_.

@tahmidefaz Please take a look and consider.

tahmidefaz:
I might need to update some tests. Taking a look. :eyes:

tahmidefaz:
Fixed some tests and verified that all the shutdown handlers are working as expected. This is ready for review. :+1:

tests/test_culling.py
```diff
@@ -353,7 +396,13 @@ def test_unknown_host_is_not_removed(
     assert retrieved_host.reporter is None
 
     threadctx.request_id = UNKNOWN_REQUEST_ID_VALUE
-    host_reaper_run(inventory_config, mock.Mock(), db.session, event_producer_mock)
+    host_reaper_run(
+        inventory_config,
+        mock.Mock(),
+        db.session,
+        event_producer_mock,
+        shutdown_handler=mock.Mock(**{"shut_down.side_effect": (False)}),
```
Glutexo: Here too a _return_value_ would be more appropriate.

```suggestion
        shutdown_handler=mock.Mock(**{"shut_down.return_value": False}),
```

PR #693, opened 2020-08-07
Add a script that replaces IMAGE_TAG in deploy.yml

Glutexo:
Created a simple script that replaces the _IMAGE_TAG_ value of the deployed services in App SRE’s [_deploy.yml_](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/host-inventory/deploy.yml). It uses standard input and output. Use [_sponge_](http://joeyh.name/code/moreutils/) to edit the file in place.

The YAML snipped in the test has the same format as the real _deploy.yml_, but I tried to stripped any potentially sensitive information from it. It only exposes some bits of the structure and service names.

The YAML file is not parsed by a YAML library, but as a string by regular expressions. This is generally a bad thing™️ and here it is done to preserve formatting. Not even _ruamel.yaml.RoundTripLoader/Dumper_ does not preserve all formatting. Not doing so would produce bad patches.

I would like to ask @rodrigonull to take a look as the test. Is this the right way to use fixtures?

How to use:

1. Clone the [_app-interface_](https://gitlab.cee.redhat.com/service/app-interface) repository.
2. Use the _deploy.py_ script to swap the _IMAGE_TAG_.
   ```
   $ pipenv run python utils/deploy.py abcd123 < ../app-interface/services/insights/host-inventory/deploy.yml | sponge ../app-interface/services/insights/host-inventory/deploy.yml
   ```
3. See that _IMAGE_TAG_ has been replaced for the Reaper, MQ and HTTP, but not for Delete.


Glutexo:
Successfully used the script to deploy _c6c2452_. https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/7674 For the next time I’d add support for input and output file path, since the standard input/output is quite cumbersome. I don’t want to burn more cycles on it now though.

May anybody please take a look? @tahmidefaz @rodrigonull @shdunning

Glutexo:
@tahmidefaz Added the command to README. Is this ok for you?

tahmidefaz:
The README looks good. I will take a look at the script. Thanks! :+1:

Glutexo:
As @tahmidefaz is not a code owner, I‘ll need one more ✅ . @rodrigonull, @ryandillinfelton, @shdunning, @jharting…

tests/test_utils.py
```diff
@@ -0,0 +1,143 @@
...
+
+    return _fixture
+
+
+@fixture(scope="function")
+def run_deploy(temporary_file):
+    def _fixture(promo_code, inp_):
+        args = Args(promo_code)
+
+        with temporary_file() as inp, temporary_file() as outp:
```
rodrigonull, line 103: Can we simplify this? In this case we won't need the temporary_file fixture.

```suggestion
        with TemporaryFile("r+", encoding="utf-8") as inp, TemporaryFile("r+", encoding="utf-8") as outp:
```

tests/test_utils.py
```diff
@@ -0,0 +1,143 @@
...
+            deploy(args, inp, outp)
+
+            outp.seek(0)
+            return outp.read()
+
+    return _fixture
+
+
+@fixture(scope="session")
+def head():
```
rodrigonull, line 116: I also think that we don't need a fixture for this. I could be a simple helper function.
```suggestion
def head(doc):
    lines = doc.split("\n")
    first_lines = lines[0:7]
    return "\n".join(first_lines)
```

tests/test_utils.py
```diff
@@ -0,0 +1,143 @@
...
+    @contextmanager
+    def _fixture():
+        with TemporaryFile("r+", encoding="utf-8") as file:
+            yield file
+
+    return _fixture
+
+
+@fixture(scope="function")
+def run_deploy(temporary_file):
```
rodrigonull, line 99: This also doesn't needs to be a fixture. It can be a helper function.
```suggestion
def run_deploy():
```

tests/test_utils.py
```diff
@@ -0,0 +1,143 @@
...
+def head():
+    def _fixture(doc):
+        lines = doc.split("\n")
+        first_lines = lines[0:7]
+        return "\n".join(first_lines)
+
+    return _fixture
+
+
+def test_deploy_image_tag_is_replaced(run_deploy):
```
rodrigonull, line 125:
```suggestion
def test_deploy_image_tag_is_replaced():
```

tests/test_utils.py
```diff
@@ -0,0 +1,143 @@
...
+        RESOURCE_TEMPLATES_INDEXES["insights-inventory"],
+    )
+    for rt_i, t_i in product(resource_templates, TARGETS_INDEXES.values()):
+        expected["resourceTemplates"][rt_i]["targets"][t_i]["parameters"]["IMAGE_TAG"] = promo_code
+
+    result = run_deploy(promo_code, DEPLOY_YML)
+    assert safe_load(result) == expected
+
+
+def test_deploy_formatting_is_not_broken(run_deploy, head):
```
rodrigonull, line 141:
```suggestion
def test_deploy_formatting_is_not_broken():
```

Glutexo:
Thanks @rodrigonull for your review! It makes sense to use plain helper function instead of fixtures if those are used only in this one specific test case. It saved some complexity, imports and lines of code.

Also included two tiny fixes related to naming.

@tahmidefaz Now I need only your ✅ .

Glutexo:
Used the script again to deploy _cd9eeba_.

README.md
```diff
@@ -330,6 +330,13 @@ This applies to the following components:
 The `IMAGE_TAG` parameter should be updated for both namespaces for each of the aforementioned components.
 Note that `insights-host-delete`, which uses a different image, should not be updated.
 
+There is a script _utils/deploy.py_ that can be used to automatically update the `IMAGE_TAG`. Possible usage using
+_sponge_ from [_moreutils_](http://joeyh.name/code/moreutils/):
+
+```bash
+$ pipenv run python utils/deploy.py < path/to/deploy.yml | sponge path/to/deploy.yml
+```
```
tahmidefaz, line 338: Can we mention the `PROMO_CODE` in that pipenv command?

Glutexo: Oh, sure! I forgot to include it here. 😠

Glutexo:
This has already two ✅’s, yet it can’t be (non-forcefully) merged. Probably because it needs one from a code owner? @jharting @shdunning

utils/deploy.py
```diff
@@ -0,0 +1,52 @@
...
+
+def _parse_args():
+    parser = ArgumentParser(
+        description="""Replaces IMAGE_TAG in the App Interface Host Inventory deploy.yml with
+the provided PROMO_CODE. Reads from stdin, writes to stdout.
+Possible usage: pipenv run python utils/deploy.py < path/to/deploy.yml | sponge path/to/deploy.yml
+sponge is part of moreutils""",
+        formatter_class=RawTextHelpFormatter,
+    )
+    parser.add_argument("promo_code", type=str, help="PROMO_CODE to deploy.")
```
jharting, line 22: It would be nice to define a validation function that either tells you if the PROMO_CODE that was provided does not have exactly 7 chars or even truncates it for you if longer

Glutexo: Nice suggestion! I’ll add in in a separate follow-up PR.

PR #695, opened 2020-08-13
Remove tags from add_host log message

kylape:
All of the fields in the "extra" dict passed in to the logger get
created as fields in Elasticsearch, recursively.  This means e.g. both
input_host.account and input_host.display_name are created as fields.

Each index in ES is configured by default to only allow 1024 fields due
to performances reasons.

Unfortunately this means that the whole tag hierarchy gets added as
fields as well.  A couple of the field names look like this:

input_host.tags.insights-client.ynJeCH
input_host.tags.insights-client.yvEXlrsP

Since it seems like every host creates a new field, we run out of fields
in our index every day.

### DIFF
--- a/app/queue/queue.py
+++ b/app/queue/queue.py
@@ -108,7 +108,6 @@ def add_host(host_data):
                         "canonical_facts": input_host.canonical_facts,
                         "reporter": input_host.reporter,
                         "stale_timestamp": input_host.stale_timestamp.isoformat(),
-                        "tags": input_host.tags,
                     }
                 },
             )


jharting:
@kylape logging the incoming tags has proven useful in the past for debugging. Instead of removing the field from logging entirely I think it would be better to alter the Kibana settings to store it but not index it. That can be done e.g. using dynamic templates. That way:

- the problem with limited number of fields would be resolved, and
- we will still be able to see incoming tags for a given host in Kibana (we won't be able to search by that field but that's fine)

kylape:
Sure, sounds reasonable to me. Since I’m on PTO this week, can you please
reach out to Asa or Feng to help make those changes in ES?

jharting:
@kruai @BumbleFeng does the counter-proposal sound reasonable to you?

kruai:
I'm fine with this. I'll work with @BumbleFeng to remove the "tags" field from the index.

BumbleFeng:
```
PUT _template/template_1
{
  "template": "cwl-*",
  "mappings": {
    "dynamic_templates": [
      {
        "input_host_tags_as_text": {
          "match_mapping_type": "string",
          "path_match": "input_host.tags.*",
          "mapping": {
            "type": "text"
          }
        }
      }
    ]
  }
}
```

This template should config input_host.tag.*  mapping to [text-only](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-templates.html#text-only-mappings-strings) and reduce half of the field count. (By default Elasticsearch will map string fields as a text field with a sub keyword field.) Wdyt?

jharting:
@BumbleFeng the total number of unique tags can grow very large. Therefore, cutting it in half will likely not be sufficient. I think it would be better to not index this field at all
```
"mapping": {
    "index": "no"
}
```

Also note that there are multiple places where we log tags. https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/queue/queue.py#L122 Changing the path_match to `.*host\.tags.*` will make it capture these other log statements also.

BumbleFeng:
replaced by https://github.com/RedHatInsights/kibana/pull/5

PR #698, opened 2020-08-17
Rbac middleware for HBI

tahmidefaz:


api/__init__.py
```diff
@@ -30,7 +33,12 @@ def new_func(*args, **kwargs):
 
         api_request_count.inc()
 
+        if current_app.config["INVENTORY_CONFIG"].rbac_enforced:
```
jharting, line 36: it would be better for this to be in a standalone decorator, rather than added to api_operation

lib/middlewares.py
```diff
@@ -0,0 +1,54 @@
...
+
+from flask import current_app
+from flask import request
+
+from app.logging import get_logger
+
+
+logger = get_logger(__name__)
+
+read_actions = {
```
jharting, line 12: instead listing the required permissions here with standalone decorator (see my comment above) you can define the required permission right next to the definition of an operation, e.g.

```
@api_operation
@rbac("hosts:read")
def get_host_list(
```

lib/middlewares.py
```diff
@@ -0,0 +1,54 @@
...
+        print("RBAC_DATA RESPONSE", resp_data)
+        return resp_data["data"]
+
+    # with open("utils/rbac-mock-data/inv-read-write.json", "r") as rbac_response:
+    #     resp_data = json.load(rbac_response)
+    #     return resp_data["data"]
+
+
+def check_rbac_permissions(action):
+    rbac_data = get_rbac_permissions()
```
jharting, line 39: Don't forget to check whether the x-rh-identity belongs to a user or to a system (cert auth). For cert auth we don't enforce RBAC.

tests/fixtures/api_fixtures.py
```diff
@@ -11,42 +11,60 @@ def flask_client(flask_app):
...
+        return do_request(flask_client.post, HOST_URL, data, query_parameters, extra_headers, auth_type, identity_type)
 
     return _api_create_or_update_host
 
 
 @pytest.fixture(scope="function")
 def api_patch(flask_client):
-    def _api_patch(url, host_data, query_parameters=None, extra_headers=None):
-        return do_request(flask_client.patch, url, host_data, query_parameters, extra_headers)
+    def _api_patch(url, host_data, query_parameters=None, extra_headers=None, identity_type="System"):
```
jharting, line 25: Why do these fixtures default the identity_type to System instead of User?

tests/test_api_facts.py
```diff
@@ -122,3 +125,35 @@ def test_replace_facts_on_multiple_culled_hosts(db_create_multiple_hosts, db_get
     # Try to replace the facts on a host that has been marked as culled
     response_status, response_data = api_put(facts_url, DB_NEW_FACTS)
     assert_response_status(response_status, expected_status=400)
+
+
+def test_patch_host_with_RBAC_allowed(subtests, mocker, api_put, db_create_host, event_producer_mock, enable_rbac):
```
jharting, line 130: What is the `event_producer_mock` parameter for?

tests/test_unit.py
```diff
@@ -99,7 +99,7 @@ class AuthIdentityConstructorTestCase(TestCase):
 
     @staticmethod
     def _identity():
-        return Identity(account_number="some acct")
+        return Identity(account_number="some acct", identity_type="System")
```
jharting, line 102: Same as above - why default to System?

jharting (changes requested):
Good progress since the last time. There are still problems that need fixing, though.

In addition:

- we'll need test(s) for error handling
- we'll need test(s) to verify that rbac is bypassed for System identity type

tests/test_unit.py
ryandillinfelton: We thought it would be necessary to have the old tests bypass rbac by default. On second examination rbac isn't actually enforced by default so I'll change these over to user by default

tests/fixtures/api_fixtures.py
ryandillinfelton: Answered below

tests/test_api_facts.py
ryandillinfelton: Unnecessary here, I must not have noticed I carried it over from another version of the test that needed it.

swagger/api.spec.yaml
```diff
@@ -11,7 +11,10 @@ paths:
       tags:
         - hosts
       summary: Read the entire list of hosts
-      description: Read the entire list of all hosts available to the account.
+      description: >-
+        Read the entire list of all hosts available to the account.
+        <br /><br />
+        Required permisions: inventory:hosts:read
```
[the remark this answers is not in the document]
ryandillinfelton: :facepalm:

lib/middlewares.py
```diff
@@ -0,0 +1,85 @@
...
+    resp_data = loads(rbac_response.content.decode("utf-8"))
+    logger.debug("Fetched RBAC Data", extra=resp_data)
+
+    return resp_data["data"]
+
+
+def rbac(requested_permission):
+    def other_func(func):
+        @wraps(func)
+        def modified_func(*args, **kwargs):
```
rodrigonull, line 61:
```suggestion
        def modified_func(*args, **kwargs):
            if not inventory_config().rbac_enforced:
                return func(*args, **kwargs)

            if current_identity.identity_type != "User":
                return func(*args, **kwargs)

            rbac_data = get_rbac_permissions()

            for rbac_permission in rbac_data:
                if (
                    rbac_permission["permission"] == "inventory:*:*"
                    or rbac_permission["permission"] == "inventory:hosts:*"
                    or rbac_permission["permission"] == requested_permission
                ):
                    return func(*args, **kwargs)

            rbac_failure(logger, "unauth")
            abort(status.HTTP_403_FORBIDDEN)

        return modified_func
```

rodrigonull: Your code executes the request before it validates the RBAC user's permission which is wrong because even if the user doesn't has the permissions to for example patch a host, the host would be patched anyway despite the user getting a 403 response.

shdunning: excellent catch!

jharting (approved):
Some more tests improvements should be done. Tracked in https://projects.engineering.redhat.com/browse/RHCLOUD-8775

PR #699, opened 2020-08-20
release process - add stage environment

jharting:


README.md
```diff
@@ -286,74 +286,63 @@ They trigger the QE testing jobs \[[1](https://jenkins-jenkins.5a9f.insights-dev
...
-The `IMAGE_TAG` parameter should be updated for both namespaces for each of the aforementioned components.
+The `IMAGE_TAG` parameter should be updated for the stage namespace for each of the aforementioned components.
 Note that `insights-host-delete`, which uses a different image, should not be updated.
 
 There is a script _utils/deploy.py_ that can be used to automatically update the `IMAGE_TAG`. Possible usage using
 _sponge_ from [_moreutils_](http://joeyh.name/code/moreutils/):
 
 ```bash
-$ pipenv run python utils/deploy.py PROMO_CODE < path/to/deploy.yml | sponge path/to/deploy.yml
+$ pipenv run python utils/deploy.py -s PROMO_CODE < ../app-interface/data/services/insights/host-inventory/deploy.yml | sponge ../app-interface/data/services/insights/host-inventory/deploy.yml
```
Glutexo, line 308: This assumes a directory structure on your machine. The _app-interface_ repository can be anywhere on your drive, or even not cloned at all. It is perfectly possible to create a _deploy.yml_ update MR just using GitLab UI. Even though the path is likely to be like this, it makes the command very long and pulls attention away from the command structure.

Glutexo (approved):
LGTM

With a little comment to the explicit path.

README.md
jharting: It indeed does. I did it for convenience

- if you adhere to this directory structure you can copy the command and only replace PROMO_CODE
- if not you will need to tweak the prefix of the path a bit

In either case you won't need to specify the long path (twice) each time you want to do a release.

PR #700, opened 2020-08-21
Use externalized schema for system profile validation

Glutexo:
Message Queue API Ingress uses externalized System Profile schema for validation. The schema has been updated to the [current version](https://github.com/RedHatInsights/inventory-schemas/blob/b4436190454610e93e2d73815d2510e15e817196/schemas/system_profile/v1.yaml) with all Marshmallow constrains included.

There are however some changes, so the schema can be easily used:

* Schema version downgraded to Draft 4 so it is compatible with Connexion that uses the file too.
* The recursive _NestedObject_ definition has been removed. It is not compatible with JSON Schema Draft 4. Empty key validation is done manually.
* Object types are explicitly defined, so Connexion’s type coercion used by our custom validation works properly. There is already a PR to fix this.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1,18 +1,25 @@
 import uuid
+from copy import deepcopy
 from datetime import datetime
 from datetime import timezone
+from os.path import join
 
+from connexion.decorators.validation import coerce_type
 from flask_sqlalchemy import SQLAlchemy
+from jsonschema import validate as jsonschema_validate
+from jsonschema import ValidationError as JsonSchemaValidationError
 from marshmallow import fields
+from marshmallow import pre_load
 from marshmallow import Schema
-from marshmallow import validate
+from marshmallow import validate as marshmallow_validate
 from marshmallow import validates
-from marshmallow import ValidationError
+from marshmallow import ValidationError as MarshmallowValidationError
 from sqlalchemy import Index
 from sqlalchemy import orm
 from sqlalchemy import text
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.dialects.postgresql import UUID
+from yaml import safe_load
 
 from app.exceptions import InventoryException
 from app.logging import get_logger
@@ -25,9 +32,12 @@ logger = get_logger(__name__)
 
 db = SQLAlchemy()
 
-TAG_NAMESPACE_VALIDATION = validate.Length(max=255)
-TAG_KEY_VALIDATION = validate.Length(min=1, max=255)
-TAG_VALUE_VALIDATION = validate.Length(max=255)
+TAG_NAMESPACE_VALIDATION = marshmallow_validate.Length(max=255)
+TAG_KEY_VALIDATION = marshmallow_validate.Length(min=1, max=255)
+TAG_VALUE_VALIDATION = marshmallow_validate.Length(max=255)
+
+SPECIFICATION_DIR = "./swagger/"
+SYSTEM_PROFILE_SPECIFICATION_FILE = "system_profile.spec.yaml"
 
 
 def _set_display_name_on_save(context):
@@ -239,40 +249,40 @@ class Host(db.Model):
 
 
 class DiskDeviceSchema(Schema):
-    device = fields.Str(validate=validate.Length(max=2048))
-    label = fields.Str(validate=validate.Length(max=1024))
+    device = fields.Str(validate=marshmallow_validate.Length(max=2048))
+    label = fields.Str(validate=marshmallow_validate.Length(max=1024))
     options = fields.Dict(validate=check_empty_keys)
-    mount_point = fields.Str(validate=validate.Length(max=2048))
-    type = fields.Str(validate=validate.Length(max=256))
+    mount_point = fields.Str(validate=marshmallow_validate.Length(max=2048))
+    type = fields.Str(validate=marshmallow_validate.Length(max=256))
 
 
 class YumRepoSchema(Schema):
-    id = fields.Str(validate=validate.Length(max=256))
-    name = fields.Str(validate=validate.Length(max=1024))
+    id = fields.Str(validate=marshmallow_validate.Length(max=256))
+    name = fields.Str(validate=marshmallow_validate.Length(max=1024))
     gpgcheck = fields.Bool()
     enabled = fields.Bool()
-    base_url = fields.Str(validate=validate.Length(max=2048))
+    base_url = fields.Str(validate=marshmallow_validate.Length(max=2048))
 
 
 class DnfModuleSchema(Schema):
-    name = fields.Str(validate=validate.Length(max=128))
-    stream = fields.Str(validate=validate.Length(max=128))
+    name = fields.Str(validate=marshmallow_validate.Length(max=128))
+    stream = fields.Str(validate=marshmallow_validate.Length(max=128))
 
 
 class InstalledProductSchema(Schema):
-    name = fields.Str(validate=validate.Length(max=512))
-    id = fields.Str(validate=validate.Length(max=64))
-    status = fields.Str(validate=validate.Length(max=256))
+    name = fields.Str(validate=marshmallow_validate.Length(max=512))
+    id = fields.Str(validate=marshmallow_validate.Length(max=64))
+    status = fields.Str(validate=marshmallow_validate.Length(max=256))
 
 
 class NetworkInterfaceSchema(Schema):
     ipv4_addresses = fields.List(fields.Str())
     ipv6_addresses = fields.List(fields.Str())
-    state = fields.Str(validate=validate.Length(max=25))
+    state = fields.Str(validate=marshmallow_validate.Length(max=25))
     mtu = fields.Int()
-    mac_address = fields.Str(validate=validate.Length(max=59))
-    name = fields.Str(validate=validate.Length(min=1, max=50))
-    type = fields.Str(validate=validate.Length(max=18))
+    mac_address = fields.Str(validate=marshmallow_validate.Length(max=59))
+    name = fields.Str(validate=marshmallow_validate.Length(min=1, max=50))
+    type = fields.Str(validate=marshmallow_validate.Length(max=18))
 
 
 class SystemProfileSchema(Schema):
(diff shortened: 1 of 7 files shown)


app/models.py
```diff
@@ -372,29 +392,52 @@ def _validate_tags_dict(tags):
...
+            raise MarshmallowValidationError(f"System profile does not conform to schema.\n{error}") from error
+
+        for dd_i, disk_device in enumerate(system_profile.get("disk_devices", [])):
+            if not check_empty_keys(disk_device.get("options")):
+                raise MarshmallowValidationError(f"Empty key in /system_profile/disk_devices/{dd_i}/options.")
+
 
 class HttpHostSchema(BaseHostSchema):
-    pass
+    system_profile = fields.Nested(SystemProfileSchema)
```
jharting, line 438: Should this be added here?

app/models.py
```diff
@@ -372,29 +392,52 @@ def _validate_tags_dict(tags):
...
+            coerce_type(
+                self.system_profile_spec["$defs"]["SystemProfile"], processed_data["system_profile"], "property"
+            )
+            return processed_data
+        else:
+            return raw_data
+
+    @validates("system_profile")
+    def system_profile_is_valid(self, system_profile):
+        schema = {**self.system_profile_spec, "$ref": "#/$defs/SystemProfile"}
```
jharting, line 426: This combines two static values. Should it be done once i.e. in the constructor rather than repeatedly on every validation?

jharting (changes requested):
This implementation does not deal with keys sent by reporters that are not explicitly defined in the schema.

The current (marshmallow) implementation tolerates such keys (does not consider them to be schema violations) but strips them away. As a result, such fields are never stored in the database.

The proposed implementation does not validate such keys but does not remove them either. As a result, such keys are stored in the database which is not desired.

app/models.py
Glutexo: Great point! Not only this is suboptimal, but  also the schema is loaded and parsed repeatedly every time the schema is used. That is even more performace heavy. I’ll fix both.

app/models.py
Glutexo: It should. I removed the field from the parent _BaseHostSchema_. The definition is now different for _MqHostSchema_ and _HttpHostSchema_.

Glutexo:
It is not actually easy to test that the Marshmallow schema is not used. It is instantiated on class declaration, not on its instantiation. Because of that it is barely impossible to patch it in any way. I added at least a test verifying that the _MqHostSchema_ is used for deserialization and that the schema class does not directly use a nested _SystemProfileSchema_ instance.

Glutexo:
One more test to verify that the external schema is actually used. It mocks the specification file and adds a validation that is not in Marshmallow. Then validation is checked for failure.

app/models.py
Glutexo: I wanted to prevent a side effect of loading and parsing a file on import, so I introduced a lazy-loaded variable. On its first use, the schema is loaded and cached, further on it is re-used. It requires some shenanigans in the test tampering with the schema, but saves a good deal of performance.

Glutexo:
Although Marshmallow did remove undefined fields on unmarshall, this is not correct behavior for JSON Schema. To reject unexpected object properties there is an _additionalProperties_ directive that can contain a _false_ schema. If not provided, it is equal to a _true_ schema. Adding this to our schema would be the right thing to do and pairs nicely with disabling the type coercion.

Now, to preserve the current behavior, we need to violate the JSON Schema semantics a bit. I‘d go with this:
- If _properties_ are explicitly defined, use it as a whitelist and silently (without failing) remove the undefined ones. This is what Marshmallow does.
- If there is no _properties_ definition, keep all keys. This makes us not remove everything from arbitrary (disk device-like) objects.
- Ignore (keep) anything that is not explicitly declared as _object_.
- Do not pay any attention to _additionalProperties_. It is not used in the current version of the schema anyway.

Glutexo:
I implemented the key filtering the way described above. Please review the approach and its implementation. Also please review the new tests verifying that the externalized schema is actually used instead of Marshmallow for System Profile.

jharting:
> I implemented the key filtering the way described above. Please review the approach and its implementation. Also please review the new tests verifying that the externalized schema is actually used instead of Marshmallow for System Profile.

Did you push all the commits? I cannot see the key filtering neither tests for it.

Glutexo:
😞🤦🏻‍♂️ Fixed.

app/models.py
```diff
@@ -45,6 +55,25 @@ def _time_now():
...
 
+def _filter_keys(schema, payload):
+    schema_type = schema.get("type")
+    if schema_type == "object" and "properties" in schema:
+        return {
+            key: _filter_keys(schema["properties"][key], value)
+            for key, value in payload.items()
+            if key in schema["properties"]
+        }
+    elif schema_type == "array" and "items" in schema:
```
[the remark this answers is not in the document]
Glutexo: I am not sure here. In both cases, this would make the code quite unnecessarily verbose. The inline strings here are string literal key names defined by JSON Schema.

- ```
   class SchemaTypes(Enum):
       ARRAY = "array"
       OBJECT = "object"
   
   if schme
   ```

app/models.py
```diff
@@ -372,29 +415,54 @@ def _validate_tags_dict(tags):
...
 
+    @pre_load
+    def coerce_system_profile(self, raw_data):
+        if "system_profile" not in raw_data:
+            return raw_data
+
+        schema = self.system_profile_schema()
+        definition = schema["$defs"]["SystemProfile"]
+
+        processed_data = deepcopy(raw_data)
```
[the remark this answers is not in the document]
Glutexo: This is a good point. Immutability is nice when the language is optimized for object copying. This is not Python’s case. I‘ll take a look.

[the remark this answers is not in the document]
Glutexo: I examined deeper (pun intended) the uses of _deepcopy_:

- [In the _pre-load_ function](https://github.com/Glutexo/insights-host-inventory/blob/c37e9f69a71a0c1c46075732eb75820f2c0784fe/app/models.py#L444) it is not necessary as long as the input object is not mutated. Unmarshalling itself then creates a safe copy.
- [In the key filtering function](https://github.com/Glutexo/insights-host-inventory/blob/c37e9f69a71a0c1c46075732eb75820f2c0784fe/app/models.py#L69), this is the most complicated case. If the value is described by the schema, it is traversed and in case of an object, filtered by keys. Each time, a new list/dictionary is created by a comprehension expression. The _deepcopy_ is then called only on leaves, which in case of immutable scalars returns the same object and is thus cheap. In case of arbitrary objects/arrays (only disk devices options in our case), real _deepcopy_ is performed, which is comparable to the recursive schema-aware call of ours.
- [In case of the type coercion function](https://github.com/Glutexo/insights-host-inventory/blob/c37e9f69a71a0c1c46075732eb75820f2c0784fe/app/models.py#L72), this is necessary for the wrapped Connexion function not to mutate the original object. Here it would be possible to omit it if we called in on the result of the filter key function that creates deeply copied object. That would however make the two calls inconsistent and thus possible error-prone.

As a conclusion, I’d refrain to make those two function consistent in that they are going to mutate the existing object. Then only one clearly visible _deepcopy_ call would be used. Although I would like to prevent mutation as much as possible, here the other approach makes sense too.

app/models.py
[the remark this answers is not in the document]
Glutexo: I gave this some more thought and hopefully came up with a solution. Several ways I tried were unnecessarily verbose and introduced only a lot of syntactic clutter. Finally I achieve these with only a slight increase in the LOC number:

- Magic strings coming from the JSON Schema DSL appear only once each. 
   - Schema keys are used only as a clear dict-key reference in initialization constructs.
   - Type names are listed only in a enum-like type map.
- The whole filter selection dependent on various schema dictionary information is concentrated in a single try-except clause. Fallback is not repeating.
- As a side effect, the filters are represented by Python types.

I hope this makes the code not only efficient, but also a good example for any of our code readers. Any feedback is appreciated!

Glutexo:
Tests need to be fixed because of changed syntax.

Glutexo:
Updated the unit tests. Found one bug in them, fixed. Added one more test case.

app/models.py
```diff
@@ -45,6 +55,45 @@ def _time_now():
     return datetime.now(timezone.utc)
 
 
+class FilterKeys:
+    def __new__(cls, schema):
+        try:
+            schema_type = schema["type"]
+            key_filter = FilterKeys.TYPE_MAP[schema_type]
+            return key_filter(schema)
+        except KeyError:
```
jharting, line 64: Let's not use exceptions for flow control.

app/models.py
```diff
@@ -45,6 +55,45 @@ def _time_now():
...
+            return FilterKeys.NoFilter(schema)
+
+    class ObjectFilter:
```
jharting, line 67: How are these objects better than just having functions for object/array/nofilter?

jharting (changes requested):

app/models.py
Glutexo: Python encourages this antipattern, calling it an [EAFP principle](https://docs.python.org/3/glossary.html#term-eafp), and its own standard library heavily uses it. It’s more or less officially the standard pythonic way. It’s nicely manifested by the existence of the _else_ part of _try/except/else_ block. More references: [\[1\]](https://jeffknupp.com/blog/2013/02/06/write-cleaner-python-use-exceptions/), [\[2\]](https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/), [\[3\]](https://www.pythonstudio.us/learning-3/a-final-note-on-pythonic-exception-handling.html).

But since you made this a requested change, I‘ll change it to LBYL. We’re already doing much worse with exceptions like using _except Exception_. We can live with this.

app/models.py
Glutexo: If you look at it from the other side, how are functions better than the objects?

The objects provide at least tiny namespaces that couple the two parts together. The _init_ part checks that it’s possible to use this filter and loads the necessary data from the schema. The _call_ part does the actual filtering. Also, classes can prevent function nesting, which may be less readable due to excessive indentation and unintuitive return values.

But again, since this is a requested change, I’ll try to come up with a way that doesn’t use classes. In Python, classes are not that much different from any other objects. Functions themselves are objects, just as dicts and other scalars.

app/models.py
jharting: @Glutexo 
> This common Python coding style assumes the existence of valid keys or attributes and catches exceptions if the assumption proves false.

Well, does that apply in our situation, though? Do we assume the existence of valid keys? We know for sure that in order to process a single host message that except clause will be entered dozens if not hundreds of time.
Also take a look at https://docs.python.org/3/faq/design.html#how-fast-are-exceptions

Glutexo: The documentation says
> A try/except block is extremely efficient if no exceptions are raised.

The only place where this exception is raised and caught is actually the infamous _/DiskDevice/options_ object that doesn’t define _properties_. In all other cases, the _except_ block is not triggered at all.

I did some benchmarking, deserializing 2,000 hosts. The refactored version is in total about 100ms slower. There were a few runs when the refactored version ran faster than the original one.
- The original procedural version with conditionals. Only modified to mutate the passed object.
   ```
   Sum: 3491.9967651367188ms
   Sum: 3478.938579559326ms
   Sum: 3504.4939517974854ms
   Sum: 3426.2311458587646ms
   Sum: 3596.139669418335ms
   ```
- The refactored version using types and exception handling.
   ```
   Sum: 3570.7499980926514ms
   Sum: 3647.9811668395996ms
   Sum: 3569.4141387939453ms
   Sum: 3652.5745391845703ms
   Sum: 3661.5259647369385ms
   ```

For comparison, I add two more measurements. Each of them contains only one of the two possible slowdown factors.

- Uses types, but does not use exceptions.
   ```
   Sum: 3608.1724166870117ms
   Sum: 3581.0811519622803ms
   Sum: 3634.829044342041ms
   Sum: 3566.1404132843018ms
   Sum: 3586.2221717834473ms
   ```
- Uses exceptions, but does not use types.
   ```
   Sum: 3550.719976425171ms
   Sum: 3508.180618286133ms
   Sum: 3527.4999141693115ms
   Sum: 3531.621217727661ms
   Sum: 3595.9036350250244ms
   ```

It looks like both changes decrease the performance a little. The exceptions slightly more than the types. That proves your fear was correct, although in a very small scale. Is it worth considering? What the difference would be in the production cluster?

jharting: > The only place where this exception is raised and caught is actually the infamous /DiskDevice/options object that doesn’t define properties. In all other cases, the except block is not triggered at all.

Really? What about the scalar properties?

Glutexo: 🤔 You are right!

Glutexo:
I gave it some more work and came up with a new refactoring

app/models.py
```diff
@@ -45,6 +57,55 @@ def _time_now():
...
 
+class SystemProfileNormalization:
+    SOME_ARBITRARY_STRING = "property"
+
+    class Schema(namedtuple("Schema", ("type", "properties", "items"))):
+        Types = Enum("SchemaTypes", ("array", "object"))
+
+        @property
+        def type(self):
+            type_str = self._asdict()["type"]
```
Glutexo, line 68: This is necessary to avoid recursion by accessing `self.type`.

Glutexo:
_Simple is better than complex._ A fresh look gave me a possibility to simplify stuff that required an explanatory comment.

app/models.py
Glutexo: Renamed to _schema_type_ to remove this shenanigan.

app/models.py
```diff
@@ -45,6 +57,55 @@ def _time_now():
...
+            return self.Types[type_str] if type_str and type_str in self.Types.__members__.keys() else None
+
+    Schema.__new__.__defaults__ = (None,) * len(Schema._fields)
```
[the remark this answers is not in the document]
Glutexo: Removed in favor of a simple _dict.get_.

jharting (approved):

Glutexo:
Awesome, thanks! 🙇🏻‍♂️

PR #701, opened 2020-08-21
Kafka producer additions

bastilian:
This makes a few adjustments to the Kafka producer to generate hosts to be created in the inventory.

It allows to set the number of hosts as well as the account they should be created for.
It also uses the random uuid to not have them all be named the same.

### DIFF
--- a/utils/kafka_producer.py
+++ b/utils/kafka_producer.py
@@ -8,7 +8,8 @@ from ttictoc import TicToc
 
 HOST_INGRESS_TOPIC = os.environ.get("KAFKA_HOST_INGRESS_TOPIC", "platform.inventory.host-ingress")
 BOOTSTRAP_SERVERS = os.environ.get("KAFKA_BOOTSTRAP_SERVERS", "localhost:29092")
-NUM_HOSTS = os.environ.get("NUM_HOSTS", 1)
+NUM_HOSTS = int(os.environ.get("NUM_HOSTS", 1))
+
 
 def main():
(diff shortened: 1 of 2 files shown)


bastilian:
I'm not really sure what the test failures about, since there were no changes to the Pipfile.

utils/kafka_producer.py
```diff
@@ -8,9 +8,7 @@
 
 HOST_INGRESS_TOPIC = os.environ.get("KAFKA_HOST_INGRESS_TOPIC", "platform.inventory.host-ingress")
 BOOTSTRAP_SERVERS = os.environ.get("KAFKA_BOOTSTRAP_SERVERS", "localhost:29092")
-
-NUM_HOSTS = 1
-
+NUM_HOSTS = os.environ.get("NUM_HOSTS", 1)
```
jharting, line 11: The env var value will need to be parsed into an integer first for this to work

jharting (changes requested):
This fails because of linter rules. Run `make style` to fix the problems.

utils/kafka_producer.py
bastilian: AH! I'm a total Python noob. :) Thanks. I'll fix that.

bastilian:
@jharting Thanks for the hint! I made sure now that `make style` passes. I also shortened the uuid to six chars that should be enough.

jharting (approved):

jharting:
@bastilian merged, thanks!

PR #703, opened 2020-08-26
RBAC Test Improvements

tahmidefaz:


tahmidefaz:
This is ready for review now. :arrow_up:

tests/test_api_facts.py
```diff
@@ -158,6 +158,8 @@ def test_put_facts_with_RBAC_denied(subtests, mocker, api_put, db_create_host, e
 
             assert_response_status(response_status, 403)
 
+            assert db_get_host(host.id).facts != DB_NEW_FACTS
```
jharting, line 161: these would never be equal as on the left side you have all the namespaces whereas on the right side you have data in DB_FACTS_NAMESPACE only

```suggestion
            assert db_get_host(host.id).facts[DB_FACTS_NAMESPACE]["facts"] != DB_NEW_FACTS
```

jharting (changes requested):
Found a problem in one test. Moreover, these two operations are not tested at all:
* get_host_system_profile_by_id
* get_host_tag_count

tests/test_api_facts.py
tahmidefaz: Looks like we weren't using the `api_put` correctly either. Fixed this + the `api_put`. :+1:

tahmidefaz:
Just added tests for `get_host_tag_count` and `get_host_system_profile_by_id`. This one needs another round of review now.

jharting (approved):

PR #705, opened 2020-08-27
Make host delete chunk size configurable - RHCLOUD-8272

tahmidefaz:


lib/host_delete.py
```diff
@@ -9,12 +10,11 @@
 from lib.metrics import delete_host_processing_time
 
 __all__ = ("delete_hosts",)
-CHUNK_SIZE = 1000
 
 
 def delete_hosts(select_query, event_producer, interrupt=lambda: False):
     while select_query.count():
-        for host in select_query.limit(CHUNK_SIZE):
+        for host in select_query.limit(inventory_config().host_delete_chunk_size):
```
jharting, line 17: This relies on the app context which is not available when reaper runs. As a result this change breaks reaper. We may need to pass the value as a parameter instead.

jharting (changes requested):

lib/host_delete.py
Glutexo: It may be worth thinking about some “global” configuration context for values common to all environments.

tahmidefaz:
Finally added that test, any suggestion on improvements?

jharting:
@tahmidefaz if I undo all the changes in this PR except for the test it still passes. That may be an indication that the test is not very good.

jharting (approved):

PR #706, opened 2020-09-02
RHCLOUD-8121 Harden delivery guarantees of delete events

jharting:


app/queue/event_producer.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, config):
         self._kafka_producer = KafkaProducer(bootstrap_servers=config.bootstrap_servers, **config.kafka_producer)
         self.topics = {Topic.egress: config.host_egress_topic, Topic.events: config.event_topic}
 
-    def write_event(self, event, key, headers, topic):
+    def write_event(self, event, key, headers, topic, wait=False):
```
Glutexo, line 21: Making this a keyword-only argument may make calls more readable.

```suggestion
    def write_event(self, event, key, headers, topic, *, wait=False):
```

Also thinking about alternative names. Like _wait_for_kafka_, _blocking_ or _synchronous_. _wait_ is fine though.

app/queue/event_producer.py
```diff
@@ -33,6 +33,9 @@ def write_event(self, event, key, headers, topic):
             send_future.add_callback(message_produced, logger, event, key, headers)
             send_future.add_errback(message_not_produced, logger, self.topics[topic], event, key, headers)
 
+        if wait:
+            send_future.get()
```
Glutexo, line 37: Do we want to pass a timeout here? From a quick glance into the source code it looks like it does not fall back to any default value. Instead, it just waits for the thread and subsequently for Kafka to finish or give up.

lib/host_delete.py
```diff
@@ -25,7 +25,7 @@ def delete_hosts(select_query, event_producer, chunk_size, interrupt=lambda: Fal
                 event = build_event(EventType.delete, host)
                 insights_id = host.canonical_facts.get("insights_id")
                 headers = message_headers(EventType.delete, insights_id)
-                event_producer.write_event(event, str(host.id), headers, Topic.events)
+                event_producer.write_event(event, str(host.id), headers, Topic.events, True)
```
Glutexo, line 28: The boolean argument is rather cryptic. Although I agree that with a smart IDE it’s just a matter of pointing a cursor on the value.

```suggestion
                event_producer.write_event(event, str(host.id), headers, Topic.events, wait=True)
```

tests/helpers/mq_utils.py
```diff
@@ -16,12 +16,14 @@ def __init__(self):
...
         self.topic = None
+        self.wait = None
 
-    def write_event(self, event, key, headers, topic):
+    def write_event(self, event, key, headers, topic, wait=False):
         self.event = event
         self.key = key
         self.headers = headers
         self.topic = topic
+        self.wait = wait
```
Glutexo, line 26: Why are we storing the value? It is not used at all in the tests. Not even to mock the KafkaProducer behavior.

```suggestion
```

tests/helpers/mq_utils.py
```diff
@@ -16,12 +16,14 @@ def __init__(self):
         self.key = None
         self.headers = None
         self.topic = None
+        self.wait = None
```
Glutexo, line 19:
```suggestion
```

tests/test_culling.py
```diff
@@ -365,6 +366,8 @@ def test_reaper_shutdown_handler(
     created_hosts = db_get_hosts(created_host_ids)
     assert created_hosts.count() == host_count
 
+    event_producer_mock.write_event = mock.Mock()
```
Glutexo, line 369: If the write_event method is replaced like this, then there is no need to use the _event_producer_mock_ fixture. It would be better if the _MockEventProducer_ class would record more than one call, but we probably don’t want to do that in this pull request. The call itself is however pretty useless and after your changes in this particular test case, it brings no value at all.

If you pass a plain ol’ _Mock_ instead of the whole _event_producer_, you’ll get _event_producer.write_event_ mock for free.

```suggestion
    event_producer_mock = mock.Mock()
```

tests/test_culling.py
```diff
@@ -376,6 +379,7 @@ def test_reaper_shutdown_handler(
 
     remaining_hosts = db_get_hosts(created_host_ids)
     assert remaining_hosts.count() == 1
+    assert event_producer_mock.write_event.call_count == 2
```
Glutexo, line 382: This would still work.

Glutexo (commented):
Some comments. So far it looks good to me. The suggested replacement of MockEventProducer in the existing tests would be very nice to have. Without it, there would be useless code that would increase a wtf-level for anybody who looks as the code.

Dropping only as a comment, because I want to take a closer look at the new tests and manually verify the workings.

tests/test_culling.py
```diff
@@ -411,3 +415,42 @@ def assert_system_culling_data(response_host, expected_stale_timestamp, expected
...
+            inventory_config,
+            mock.Mock(),
+            db.session,
+            event_producer,
+            shutdown_handler=mock.Mock(**{"shut_down.return_value": False}),
+        )
+
+    remaining_hosts = db_get_hosts(created_host_ids)
+    assert remaining_hosts.count() == 1
+    assert event_producer._kafka_producer.send.call_count == 2
```
Glutexo, line 456: These can be even computed from the host count ant side effect list length. But the literals are more readable here.

app/queue/event_producer.py
Glutexo: From the source code it looks like _get_ raises an error on failure. This can be either a timeout error, or some other type of Kafka failure error. This error doesn’t get caught, bubbles up and eventually causes anything calling this with `wait=True` to fail. It is even tested by a new test case. Is this a correct behavior?

I am not sure whether this calls the errback, triggering our _message_not_produced_ logging function. If not, I believe it should be fixed. I’ll try to test it.

If errbacks are correctly called and it is correct to interrupt when the wait fails, wouldn’t it be more appropriate to use this mechanism instead of letting the exception to propagate outside? I thought that the EventProducer should contain the _KafkaProducer_ entirely. In this case it should at least re-raise its own exception, or even better, just report a failure. It’s the callers responsibility to behave accordingly.

I am thinking of two possible solutions:

- Pass in a callback that would be called with the Future as its argument. It could call _get()_ and thus trigger the wait. That would break the containment too, but at least the calling code would be in control of the flow.
- Pass in an additional errback that would trigger the wait and would be called on failure. In this case, the exception can be contained, yet processed by the errback – it is passed as an argument to the registered function. The caller can then use it to shut down.

Glutexo (commented):
Some more comments. I have some doubts on the exception handling. If the intended behavior is to interrupt the caller, then I’d approve it because it works as intended. I’d however strongly suggest to think about the approach.

tests/test_api_hosts_delete.py
```diff
@@ -238,6 +240,23 @@ def test_delete_hosts_chunk_size(
...
+    kafka_producer, event_producer, db_create_multiple_hosts, api_delete_host, mocker, db_get_hosts
+):
+    event_producer._kafka_producer.send.side_effect = (mocker.Mock(), mocker.Mock(**{"get.side_effect": KafkaError()}))
+
+    hosts = db_create_multiple_hosts(how_many=3)
+    host_id_list = [str(host.id) for host in hosts]
+
+    response_status, response_data = api_delete_host(",".join(host_id_list))
+
+    assert_response_status(response_status, expected_status=500)
```
Glutexo, line 253: Again, here. I think that letting the Kafka exception to bubble up and kill Flask is not the best thing we can do. If the error is caught and suggested, we can gracefullty return a _503 Service Unavailable_ response.

Glutexo (changes requested):
Found a bug. 🐛 If a _KafkaError_ is raised before the send Future is created and returned, Python’s _UnboundLocalError_ occurs. This is the [code branch](https://github.com/jharting/insights-host-inventory/blob/6eb98111e2e79b101f32c88593c2fbd4ec907ceb/app/queue/event_producer.py#L31) under `except KafkaError`.

The reason is that the _send_future_ variable is initialized only when _KafkaProducer.send_ returns a value. If an exception is raised in _send_, the _send_future_ name remains uninitialized. This is alright in the original code, because it correctly references the name only in the _else_ branch of the _try-except_ clause. The new _wait-get_ code is however outside of the block and thus gets run even in case of an exception.

Accessing an uninitialized variable raises an UnboundLocalError, which propagates all the way up and shuts down the whole Process. Thanks to the _except_ clause, the failure gets logged and thanks to the shutdown handlers, the Reaper is terminated gracefully.

The bug does not cause any real harm. It however causes the Reaper to stop reaping in cases where it would normally try to send the event for the next deleted host.

The same happens in the more expected case, when a _KafkaError_ is raised in the _Future.get_ call. There again, instead of reporting the error and moving onto the next record, the whole iteration over the hosts to delete stops, because the error bubbles up to the calling code. That is not correct. The only effect the  _wait_ parameter should have is to make the event emission, well, wait. If we want to control the iteration and add an ability to interrupt it, it should be done explicitly with awareness and not by accidentally not catching errors.

To fix the _UnboundLocalError_, but keeping the bad behavior, the _wait-get_ commands can be moved to the _else_ branch. To fix the unwanted interruption caused by leaking _KafkaErrors_, move the commands to the _try_ block instead. Think about the interruption then separately, either in this pull request, or in a follow-up one.

After fixing, tests should be changed accordingly.

Glutexo:
If the reasoning behind the immediate shutdown is to rollback a database delete, then the new code does not fix it. The DELETE transaction is [commited](https://github.com/jharting/insights-host-inventory/blob/6eb98111e2e79b101f32c88593c2fbd4ec907ceb/lib/host_delete.py#L39) before the emmision starts. An interruption by an uncaught exception can only result in more message not being produced for hosts that are already deleted. This doesn’t however apply if the delete chunk size is se to 1.

tests/helpers/mq_utils.py
jharting: For consistency so that all parameters are captured.

tests/test_culling.py
jharting: Agreed that literals are more readable

app/queue/event_producer.py
jharting: > Do we want to pass a timeout here? From a quick glance into the source code it looks like it does not fall back to any default value. Instead, it just waits for the thread and subsequently for Kafka to finish or give up.

Yes we want to wait for kafka to finish or give up. How log we wait before giving up is configured using kafka producer parameters.

jharting: > From the source code it looks like get raises an error on failure. This can be either a timeout error, or some other type of Kafka failure error. This error doesn’t get caught, bubbles up and eventually causes anything calling this with wait=True to fail. It is even tested by a new test case. Is this a correct behavior?

Yes, the goal is to turn the asynchronous task of sending an event into synchronous i.e. await the result of the operation. If an error happens it should bubble up to the caller.

tests/test_api_hosts_delete.py
jharting: That's no different than how we handle e.g. sqlalchemy errors - we let them bubble up and "kill Flask". We can start catching those errors but that's outside the scope of this PR.

jharting:
@Glutexo thanks for the review. I applied the suggestions and fixed UnboundLocalError.

> The bug does not cause any real harm. It however causes the Reaper to stop reaping in cases where it would normally try to send the event for the next deleted host.

> The same happens in the more expected case, when a KafkaError is raised in the Future.get call. There again, instead of reporting the error and moving onto the next record, the whole iteration over the hosts to delete stops, because the error bubbles up to the calling code. That is not correct. The only effect the wait parameter should have is to make the event emission, well, wait. If we want to control the iteration and add an ability to interrupt it, it should be done explicitly with awareness and not by accidentally not catching errors.

Yes, the goal is to minimize the amount of inconsistency that emerges when a host is deleted from the database but the corresponding event is not produced. Currently if say kafka is down reaper could easily remove thousands of hosts ignoring the fact that none of the delete events were actually produced. After this change the amount of inconsistency in a single reaper run / mass delete is reduced to one host.

As for exceptions "bubbling up" this PR just brings kafka on par with the database where we also don't catch/ignore errors coming from the database. If we want to change that we can in a follow-up PR.

Glutexo:
> After this change the amount of inconsistency in a single reaper run / mass delete is reduced to one host.

This is not completely true. If Kafka goes down when the events are being emitted, all hosts in the chunk are already deleted, the database transaction is committed and the messages are produced only for some of them. The smaller the chunk, the smaller the probability and the harm. The only way to reduce this down to a single host is to set the chunk size to 1 though.

Glutexo (approved):
Gave some more comments, but in general it looks good to me. I am concerned though about the different behavior between _wait_ and _non-wait_ calls in terms of letting an exception crash the process. To be consistent, the _non-wait_ calls should raise an exception too. See my inline comment.

Also I am missing unit tests for the new logic:

- _Future.get_ gets called when _wait_ flag is provided, doesn’t when not provided, _do not wait_ is default.
- _message_not_produced_ is called when _Future.get_ raises an exception.

Also some functional tests:

- _lib.host_delete.delete_hosts_ calls this with the wait flag on, _app.queue.handle_message_ and _api.host._emit_patch_event_ use the default.
- Using the _wait_ flag actually makes the call synchronous. This is the most important one.

A new feature has been added only the error handling is tested, not the feature itself. I am afraid that these tests will never be written and not approving the pull request because of them would only result in an argument and a further delay. So here is my ✅.

Manually tested some of the features and confirmed them working.

- The _wait_ flag really makes the message being produced before the next host gets deleted.
- The _UnboundLocalError_ no longer happens on _send_ failure.
- On _send_ failure, the exception bubbles up and interrupts the Reaper.

PR #707, opened 2020-09-02
inventory_config - reset the inventory config after each test function

jharting:


tests/fixtures/app_fixtures.py
```diff
@@ -28,4 +29,5 @@ def flask_app(new_flask_app):
 
 @pytest.fixture(scope="function")
 def inventory_config(flask_app):
-    return flask_app.config["INVENTORY_CONFIG"]
+    yield flask_app.config["INVENTORY_CONFIG"]
+    flask_app.config["INVENTORY_CONFIG"] = Config(RuntimeEnvironment.TEST)
```
Glutexo, line 33: This repeats logic from _app.create_app_. https://github.com/RedHatInsights/insights-host-inventory/blob/12eb2f048ae92bca830986c367c9bdca6dff17d3/app/__init__.py#L79

I think that a cleaner approach would be to backup the original object and then return it instead of re-creating it. Not only this repeats [the logic](https://github.com/RedHatInsights/insights-host-inventory/blob/12eb2f048ae92bca830986c367c9bdca6dff17d3/app/__init__.py#L79) from _app.create_app_, but the respective fixtures mutate the config object.


[_copy.copy_](https://docs.python.org/3.6/library/copy.html#copy.copy) can be used to achieve this.

```python
from copy import copy

def patch_inventory_config():
    def patch(**config_values):
        original = flask_app.config["INVENTORY_CONFIG"]
        patched = copy(flask_app.config["INVENTORY_CONFIG"])
        for key, value in config_values:
            setattr(patched, key, value)
        flask_app.config["INVENTORY_CONFIG"] = patched
        yield patched
        flask_app.config["INVENTORY_CONFIG"] = original
    return patch
```

tests/fixtures/graphql_fixtures.py
```diff
@@ -57,26 +57,18 @@ def _patch_xjoin_post(response, status=200):
 @pytest.fixture(scope="function")
 def query_source_xjoin(inventory_config):
     inventory_config.bulk_query_source = BulkQuerySource.xjoin
```
Glutexo, line 59: Example use of the proposed patch fixture. I am not sure if fixture yields can be stacked exactly like this. Just demonstrating the idea.

```python
def query_source_xjoin(patch_inventory_config):
    yield patch_inventory_config(bulk_query_source=BulkQuerySource.xjoin)
```

Glutexo (approved):
I can confirm this is working and thus approving. Yet I see the logic duplication and object state mutation as a fair amount of technical debt. Offered an alternative solution.

jharting:
@Glutexo thanks for the alternative implementation. It seems to be adding complexity for no real gain. I am going to go with the original implementation for now. We can always revisit if it ever creates problems.

PR #708, opened 2020-09-02
Configuring Kafka Producer- RHCLOUD-8249

tahmidefaz:


tahmidefaz:
As far as I could tell, we mock the actual Kafka Producer in tests. So, I am not sure if this was the ask from a testing perspective.

One other thing to note is that the `"acks"` parameter can actually either be a string or an integer.

jharting (changes requested):
@tahmidefaz if I start the consumer using `KAFKA_PRODUCER_ACKS=1 make run_inv_mq_service` the messages do not seem to be produced any longer

Glutexo:
@tahmidefaz Is this ready for review?

tahmidefaz:
@jharting That could be because `"acks"` expects an integer when it equates to `1`, I think I will add an intermediate function to handle that logic. I wished `"acks"` took the values of the same type instead of `[0,1,"all"]` :disappointed: 

@Glutexo Not yet. Need to make this change ^

tahmidefaz:
Just added a simple dictionary to map those values. This is ready for another look. :+1:

jharting (approved):
I added some tweaks and proper tests. Would appreciate review from @Glutexo

tests/test_config.py
```diff
@@ -0,0 +1,35 @@
...
+    assert config.kafka_producer[param] == 2020
+
+
+@pytest.mark.parametrize(
+    "param", ["retries", "batch_size", "linger_ms", "retry_backoff_ms", "max_in_flight_requests_per_connection"]
+)
+def test_kafka_producer_int_params_invalid(monkeypatch, param):
+    monkeypatch.setenv(f"KAFKA_PRODUCER_{param.upper()}", "abc")
+    with pytest.raises(ValueError):
+        Config(RuntimeEnvironment.TEST)
```
Glutexo, line 35: Dtto default values. There is already one [test](https://github.com/RedHatInsights/insights-host-inventory/blob/d656bf700a10bf6fca221737a959dc80050ed091/tests/test_unit.py#L271) for those in the aforementioned existing test case.

Glutexo (commented):
Some comments for @jharting on the configuration tests.

tests/test_config.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+
+from app.config import Config
+from app.environment import RuntimeEnvironment
+
+
+@pytest.mark.parametrize("value", [0, 1, "all"])
+def test_kafka_producer_acks(monkeypatch, value):
+    monkeypatch.setenv("KAFKA_PRODUCER_ACKS", f"{value}")
+    config = Config(RuntimeEnvironment.TEST)
```
Glutexo, line 10: This repeats many times for each test , it’s almost worth extracting to a function.

Glutexo (commented):
Assuming untested code is broken code, I couldn’t give ✅ here. Please test that the configuration values are actually used. Dropped some more comments. Otherwise it looks good to me! Great work, thanks!

app/config.py
```diff
@@ -105,6 +118,13 @@ def _build_db_uri(self, ssl_mode, hide_password=False):
             db_uri += f"?sslmode={self._db_ssl_mode}&sslrootcert={self._db_ssl_cert}"
         return db_uri
 
+    def _from_dict(self, dict, name, default):
```
Glutexo, line 121: The _dict_ here shadows an internal Python type name. That is not good.

app/config.py
```diff
@@ -105,6 +118,13 @@ def _build_db_uri(self, ssl_mode, hide_password=False):
...
+        value = dict.get(os.environ.get(name, default))
+
+        if value is None:
+            raise ValueError(f"{os.environ.get(name)} is not a valid value for {name}")
+        return value
```
Glutexo, line 126:
```suggestion
        raw = os.environ.get(name, default)
        try:
            return dict[name]
        except KeyError as error:
            raise ValueError(f"{raw} is not a valid value for {name}. Valid values are {dict.keys()}.") from error
```

It’s easier to forgive than to allow. This is more pythonic.

PR #709, opened 2020-09-08
add sap_sids support

beav:


shdunning (approved):
👍 lgtm

Glutexo:
Confirmed that not providing _sap_sids_ does not break anything on the Inventory side. The host is accepted and saved including the new field both through the MQ and HTTP APIs. The fields are present both in the MQ output as in the System Profile GET response. Hopefully this doesn’t break any clients that don’t expect this field in our payloads. 🤷🏻‍♂️

Glutexo:
@beav I see in the sister PR https://github.com/RedHatInsights/inventory-schemas/pull/3 that the value has a regex validation. May I ask you to create a follow-up PR that adds the regex to Marshmallow? Since #700 is not merged yet, the value would not get validated by the regex at all. Even after merging, the regex validation would not happen for HTTP POST. Also please consider adding a constraint for minimum length as the regex requires the string to be exactly three characters long.

PR #711, opened 2020-09-17
Revert RHCLOUD-8614 until validation failures are investigated

jharting:


Glutexo (approved):
Agreed. It may be dangerous to deploy to production. There is a lot of validation errors from Yupana and we don’t know what is causing them. It is possible that those would cause the consumer to crash too often.

PR #712, opened 2020-09-18
Grafana dashboard migration

tahmidefaz:


kruai (approved):
LGTM!

tahmidefaz:
One more :heavy_check_mark: , maybe from @ryandillinfelton ? We should push to get this merged sometime today :slightly_smiling_face:

utils/validate_dashboards.py
```diff
@@ -0,0 +1,55 @@
...
+
+seen_error = False
+
+for f in os.listdir("dashboards"):
+    for err in validate("dashboards/" + f):
+        seen_error = True
+        print(err)
+
+if seen_error:
+    sys.exit(1)
```
ryandillinfelton, line 55: I think It'd be nice to have some sort of success confirmation message when no errors are found on a file. But I won't hold up migration on that.

utils/validate_dashboards.py
```diff
@@ -0,0 +1,55 @@
...
+    print("Validating %s" % f)
+    if not f.endswith(".yaml") and not f.endswith(".yml"):
+        yield Error("Bad file name", f)
+        return
+
+    try:
+        with open(f) as fp:
+            y = yaml.safe_load(fp)
+
+        if not y["metadata"]["name"]:
```
ryandillinfelton, line 29: This throws an error when the name field doesn't exist, but not in the way you want. 

> Traceback (most recent call last):
>   File "utils/validate_dashboards.py", line 29, in validate
>     if not y["metadata"]["name"]:
> KeyError: 'name'

I think you should be using
```
if y["metadata"].get("name")
    yield Error("Resource name not found", f)
```
This will still cause a keyerror if metadata does not exist, so It could be worth doing a get check on that before you check if name exists.

utils/validate_dashboards.py
```diff
@@ -0,0 +1,55 @@
...
+            yield Error("Invalid number of keys in ConfigMap", f)
+
+        key = list(d.keys())[0]
+
+        if not key.endswith(".json"):
+            yield Error("Key does not end with .json: %s" % key, f)
+
+        json.loads(d[key])
+    except Exception as e:
+        yield Error(e.msg, f)
```
ryandillinfelton, line 45: e.msg does not work for ScannerErrors caused by bad yml. They do not have a `msg` field.

> AttributeError: 'ScannerError' object has no attribute 'msg'

`yield Error(e, f)` works for them, but might cause issues with other Exceptions

utils/validate_dashboards.py
tahmidefaz: just did :+1:

utils/validate_dashboards.py
```diff
@@ -0,0 +1,56 @@
...
+    try:
+        with open(f) as fp:
+            y = yaml.safe_load(fp)
+
+        if not y["metadata"]["name"]:
+            yield Error("Resource name not found", f)
+
+        d = y["data"]
+
+        if len(d) != 1:
```
ryandillinfelton, line 34: This doesn't seem to work. If I add 
```
general.json: |-
    {"hey": "a key"}
```
above or below the dashboard JSON in the data data field the validator doesn't throw any errors (unless the JSON is bad)

ryandillinfelton (changes requested):
I found some issues with the validation script not catching the issues it's looking for. Everything else looks good though

utils/validate_dashboards.py
tahmidefaz: Just added a conditional to check for this, but it still isn't perfect. It has to do with how the yaml parsing library is setup + how dictionaries work in python. For example, if python sees this dictionary `{"key": 1, "key": 2}`, it will shorten it to `{"key":2}`. So, whenever we have two `general.json`, it overwrites it to the latest one it sees. I think a big change like that will likely be caught during review, imo :slightly_smiling_face:

ryandillinfelton: Seems reasonable. If this check isn't doing anything I'd say let's just remove it rather than create a false sense of security. But that being said I think you're right that the odds of this problem surfacing at all is super low. 

I'll approve the PR now since we need it by tomorrow. If we decide want to remove this we can do a follow up change tomorrow.

ryandillinfelton (approved):

PR #716, opened 2020-09-24
Add temp directory to .gitignore

thearifismail:


gmcculloug:
@thearifismail When is this needed?  Maybe you could add to your local global gitignore file.

Please provide more details in the description for this.

gmcculloug:
Closing for now as I believe global gitignore is the correct approach.  If we find this is needed we can reopen.

PR #717, opened 2020-09-25
Add sap_system filter

ryandillinfelton:
Adds the ability to filter bulk host queries by the sap_system field when using xjoin as the data source.

[Jira](https://issues.redhat.com/browse/RHCLOUD-8776)

Because the filtering only works with xjoin as the data source the tests are really is just checking the filter query parameter validation, and that we reject requests to filter on fields we don't support filtering on.


ryandillinfelton:
Flake8 passes locally but detects problems when run here. I'll fix them, but does anyone know why that might be happening?

api/host_query_xjoin.py
```diff
@@ -106,7 +109,16 @@ def _params_to_order(param_order_by=None, param_order_how=None):
     return xjoin_order_by, xjoin_order_how
 
 
-def _query_filters(fqdn, display_name, hostname_or_id, insights_id, tags, staleness, registered_with):
+def _sap_system_filters(sap_system):
+    if sap_system == "nil":
+        return ({"NOT": {"OR": [{"spf_sap_system": {"is": True}}, {"spf_sap_system": {"is": False}}]}},)
```
kruai, line 114: This line looks like it has this logic (I'm shortening `spf_sap_system` to `spf`):
`!(spf || !spf)` => `!spf && spf` => `false`
Is this line intended to always return false, as long as `spf_sap_system` is defined?

api/host_query_xjoin.py
```diff
@@ -106,7 +109,16 @@ def _params_to_order(param_order_by=None, param_order_how=None):
...
+    elif sap_system == "not_nil":
+        return ({"OR": [{"spf_sap_system": {"is": True}}, {"spf_sap_system": {"is": False}}]},)
```
kruai, line 116: This line looks like it always returns true if `spf_sap_system` is defined. Is there simpler syntax for this?

kruai (commented):
I guess I really just have comments... I don't feel qualified to give this a full review just yet

api/host_query_xjoin.py
ryandillinfelton: I just pushed a change simplifying both :+1:

kruai (approved):

tests/test_api_hosts_get.py
```diff
@@ -35,6 +35,8 @@
 from tests.helpers.test_utils import minimal_host
 from tests.helpers.test_utils import now
 
+DATA_SOURCE_XJOIN_HEADER = {"x-rh-cloud-bulk-query-source": "xjoin"}
```
tahmidefaz, line 38: @ryandillinfelton I don't know much about x-join, tbh. Can you tell me a little bit about the implication of this header?

ryandillinfelton: This header tells the API to use xjoin search as a data source for queries instead of the database. In this case the `sap_system` filter is not supported on the `host_query_db.py` `get_host_list` function (it will return a 400), so adding this header to a request makes sure we're testing the `host_query_xjoin.py` which has logic for the filter. All of the deployed environments use xjoin as the data source by default now so we don't need to mirror the functionality with the database (according to Jozef)

ryandillinfelton: I just swapped it to use a fixture that does the same thing. Cleaner that way.

api/host_query_xjoin.py
```diff
@@ -136,5 +148,13 @@ def _query_filters(fqdn, display_name, hostname_or_id, insights_id, tags, stalen
     if registered_with:
         query_filters += ({"NOT": {"insights_id": {"eq": None}}},)
 
+    if filter:
+        if filter.get("system_profile"):
+            if filter["system_profile"].get("sap_system"):
```
tahmidefaz, line 153: Is doing so fail anything for us (no idea)? It might be a little more concise? I fine either ways, tbh :slightly_smiling_face: 
`if filter.get("system_profile").get("sap_system"):`

tahmidefaz (changes requested):
Can we add tests to check whether we are generating the correct graphql query?

ryandillinfelton:
@tahmidefaz I added the tests you requested :rocket:

api/host_query_db.py
```diff
@@ -51,6 +53,9 @@ def get_host_list(
     if registered_with:
         query = find_hosts_with_insights_enabled(query)
 
+    if filter:
+        flask.abort(400)
```
jharting, line 57: if we happen to run in db mode that's not really a client error. In /tags we abort with 503 https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/tag.py#L71

Also, in this situation might be better to abort before doing a db query, not after.

tahmidefaz (approved):

jharting (commented):
Looks good. Note that the filter should also go to /tags (https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/tag.py#L60) (not necessarily in the same PR).

ryandillinfelton:
> Looks good. Note that the filter should also go to /tags (https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/tag.py#L60) (not necessarily in the same PR).

Noted! I'll fold it in with the sap_sids filter PR when it's ready

thearifismail (approved):

api/host_query_xjoin.py
fstavela: This could result in an attribute error if `filter` didn't have "system_profile" key, because we would be trying to do `.get("sap_system")` on a `None`. What we could do, is set the defaults for the get method to empty dictionaries, like this:
`if filter.get("system_profile", {}).get("sap_system")`
But probably a better way to do this would be to use try/except:
```python3
try:
    if isinstance(filter["system_profile"]["sap_system"], str):
        query_filters += _sap_system_filters(filter["system_profile"]["sap_system"])
    else:
        query_filters += _sap_system_filters(filter["system_profile"]["sap_system"]["eq"])
except (TypeError, KeyError):
    pass
```
We would get rid of a lot of nested ifs this way, which would be cleaner for now, but I think this might make the code less expendable. Especially because we might need those nested ifs in the future, when we start using more `filter` and `system_profile` parameters.

PR #720, opened 2020-09-28
Ultra-Light Check-In

kruai:
This PR is meant to address [RHCLOUD-9412](https://issues.redhat.com/browse/RHCLOUD-9412).

This adds a PUT endpoint that accepts a list of canonical facts and a time offset (in minutes). Given the list of canonical facts, it should find one specific host, and update its staleness timestamp using the supplied time offset.

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -23,8 +23,6 @@ from app import inventory_config
 from app import Permission
 from app.auth import current_identity
 from app.config import BulkQuerySource
-from app.culling import _Config as CullingConfig
-from app.culling import Timestamps
 from app.exceptions import InventoryException
 from app.exceptions import ValidationException
 from app.logging import get_logger
@@ -449,14 +447,9 @@ def _build_paginated_host_tags_response(total, page, per_page, tags_list):
 @metrics.api_request_time.time()
 def host_checkin(body):
     facts = body.get("canonical_facts")
-    checkin_frequency = body.get("checkin_frequency") or 1440
-    config = CullingConfig(
-        stale_warning_offset_delta=timedelta(minutes=checkin_frequency),
-        culled_offset_delta=inventory_config().culling_culled_offset_delta,
-    )
-    timestamps = Timestamps(config)
+    staleness_offset = timedelta(minutes=body.get("checkin_frequency") or 1440)
     serialized_host, host_id, insights_id, updated = update_host_staleness(
-        current_identity.account_number, facts, timestamps
+        current_identity.account_number, facts, staleness_offset
     )
     if serialized_host:
         _emit_patch_event(serialized_host, host_id, insights_id)
--- a/api/host.py
+++ b/api/host.py
@@ -449,9 +449,9 @@ def _build_paginated_host_tags_response(total, page, per_page, tags_list):
 @metrics.api_request_time.time()
 def host_checkin(body):
     facts = body.get("canonical_facts")
-    staleness_offset = body.get("staleness_offset")
+    checkin_frequency = body.get("checkin_frequency")
     config = CullingConfig(
-        stale_warning_offset_delta=timedelta(minutes=staleness_offset),
+        stale_warning_offset_delta=timedelta(minutes=checkin_frequency),
         culled_offset_delta=inventory_config().culling_culled_offset_delta,
     )
     timestamps = Timestamps(config)
--- a/api/host.py
+++ b/api/host.py
@@ -1,4 +1,6 @@
+from datetime import datetime
 from datetime import timedelta
+from datetime import timezone
 from enum import Enum
 
 import connexion
@@ -44,8 +46,8 @@ from app.utils import Tag
 from lib.host_delete import delete_hosts
 from lib.host_repository import add_host
 from lib.host_repository import AddHostResult
+from lib.host_repository import find_existing_host
 from lib.host_repository import find_non_culled_hosts
-from lib.host_repository import update_host_staleness
 from lib.middleware import rbac
 
 
@@ -450,11 +452,13 @@ def _build_paginated_host_tags_response(total, page, per_page, tags_list):
 def host_checkin(body):
     facts = body.get("canonical_facts")
     staleness_offset = timedelta(minutes=body.get("checkin_frequency") or 1440)
-    serialized_host, host_id, insights_id, updated = update_host_staleness(
-        current_identity.account_number, facts, staleness_offset
-    )
-    if serialized_host:
-        _emit_patch_event(serialized_host, host_id, insights_id)
+    existing_host = find_existing_host(current_identity.account_number, facts)
+
+    if existing_host:
+        existing_host._update_stale_timestamp(datetime.now(timezone.utc) + staleness_offset, "checkin")
+        db.session.commit()
+        serialized_host = serialize_host(existing_host, staleness_timestamps(), EGRESS_HOST_FIELDS)
+        _emit_patch_event(serialized_host, existing_host.id, existing_host.canonical_facts.get("insights_id"))
         return flask_json_response(serialized_host, 201)
     else:
         return flask_json_response(
--- a/api/host.py
+++ b/api/host.py
@@ -1,3 +1,4 @@
+from datetime import timedelta
 from enum import Enum
 
 import connexion
@@ -22,6 +23,8 @@ from app import inventory_config
 from app import Permission
 from app.auth import current_identity
 from app.config import BulkQuerySource
+from app.culling import _Config as CullingConfig
+from app.culling import Timestamps
 from app.exceptions import InventoryException
 from app.exceptions import ValidationException
 from app.logging import get_logger
@@ -447,4 +450,13 @@ def _build_paginated_host_tags_response(total, page, per_page, tags_list):
 def add_host_checkin(body):
     facts = body.get("canonical_facts")
     staleness_offset = body.get("staleness_offset")
-    return update_host_staleness(current_identity.account_number, facts, staleness_offset)
+    config = CullingConfig(
+        stale_warning_offset_delta=timedelta(minutes=staleness_offset),
+        culled_offset_delta=inventory_config().culling_culled_offset_delta,
+    )
+    timestamps = Timestamps(config)
+    serialized_host, host_id, insights_id, updated = update_host_staleness(
+        current_identity.account_number, facts, timestamps
+    )
+    _emit_patch_event(serialized_host, host_id, insights_id)
+    return serialized_host
--- a/api/host.py
+++ b/api/host.py
@@ -449,7 +449,7 @@ def _build_paginated_host_tags_response(total, page, per_page, tags_list):
 @metrics.api_request_time.time()
 def host_checkin(body):
     facts = body.get("canonical_facts")
-    checkin_frequency = body.get("checkin_frequency")
+    checkin_frequency = body.get("checkin_frequency") or 1440
     config = CullingConfig(
         stale_warning_offset_delta=timedelta(minutes=checkin_frequency),
         culled_offset_delta=inventory_config().culling_culled_offset_delta,
--- a/api/host.py
+++ b/api/host.py
@@ -445,7 +445,6 @@ def _build_paginated_host_tags_response(total, page, per_page, tags_list):
 @rbac(Permission.WRITE)
 @metrics.api_request_time.time()
 def add_host_checkin(body):
-    # Get fields from body
     facts = body.get("canonical_facts")
     staleness_offset = body.get("staleness_offset")
     return update_host_staleness(current_identity.account_number, facts, staleness_offset)
--- a/api/host.py
+++ b/api/host.py
@@ -44,6 +44,7 @@ from lib.host_delete import delete_hosts
 from lib.host_repository import add_host
 from lib.host_repository import AddHostResult
 from lib.host_repository import find_non_culled_hosts
+from lib.host_repository import update_host_staleness
 from lib.middleware import rbac
 
 
@@ -438,3 +439,13 @@ def _build_serialized_tags(host_list, search):
 def _build_paginated_host_tags_response(total, page, per_page, tags_list):
     json_output = build_collection_response(tags_list, page, per_page, total)
     return flask_json_response(json_output)
+
+
+@api_operation
+@rbac(Permission.WRITE)
+@metrics.api_request_time.time()
+def add_host_checkin(body):
+    # Get fields from body
+    facts = body.get("canonical_facts")
+    staleness_offset = 60
+    return update_host_staleness(current_identity.account_number, facts, staleness_offset)
--- a/api/host.py
+++ b/api/host.py
@@ -458,5 +458,16 @@ def add_host_checkin(body):
     serialized_host, host_id, insights_id, updated = update_host_staleness(
         current_identity.account_number, facts, timestamps
     )
-    _emit_patch_event(serialized_host, host_id, insights_id)
-    return serialized_host
+    if serialized_host:
+        _emit_patch_event(serialized_host, host_id, insights_id)
+        return flask_json_response(serialized_host, 201)
+    else:
+        return flask_json_response(
+            {
+                "detail": "No hosts match the provided facts.",
+                "status": 400,
+                "title": "Bad Request",
+                "type": "about:blank",
+            },
+            status=400,
+        )
--- a/api/host.py
+++ b/api/host.py
@@ -447,7 +447,7 @@ def _build_paginated_host_tags_response(total, page, per_page, tags_list):
 @api_operation
 @rbac(Permission.WRITE)
 @metrics.api_request_time.time()
-def add_host_checkin(body):
+def host_checkin(body):
     facts = body.get("canonical_facts")
     staleness_offset = body.get("staleness_offset")
     config = CullingConfig(
--- a/api/host.py
+++ b/api/host.py
@@ -447,5 +447,5 @@ def _build_paginated_host_tags_response(total, page, per_page, tags_list):
 def add_host_checkin(body):
     # Get fields from body
     facts = body.get("canonical_facts")
-    staleness_offset = 60
+    staleness_offset = body.get("staleness_offset")
     return update_host_staleness(current_identity.account_number, facts, staleness_offset)
--- a/lib/host_repository.py
+++ b/lib/host_repository.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from enum import Enum
 
 from sqlalchemy import and_
@@ -15,6 +16,7 @@ from lib.db import session_guard
 
 __all__ = (
     "add_host",
+    "update_host_staleness",
     "canonical_fact_host_query",
     "canonical_facts_host_query",
     "create_new_host",
@@ -55,6 +57,19 @@ def add_host(input_host, staleness_offset, update_system_profile=True, fields=DE
             return update_existing_host(existing_host, input_host, staleness_offset, update_system_profile, fields)
         else:
             return create_new_host(input_host, staleness_offset, fields)
+
+
+def update_host_staleness(account_number, canonical_facts, staleness_offset):
+    """
+    Updates the staleness timestamp for a host with the specified account number
+    and canonical facts.
+    """
+
+    existing_host = find_host_by_canonical_facts(account_number, canonical_facts)
+    input_host = deepcopy(existing_host)
+    # TODO: Update timestamps
+    new_host = update_existing_host(existing_host, input_host, None, False, DEFAULT_FIELDS)
+    return new_host
 
 
 @metrics.host_dedup_processing_time.time()
--- a/lib/host_repository.py
+++ b/lib/host_repository.py
@@ -1,5 +1,3 @@
-from datetime import datetime
-from datetime import timezone
 from enum import Enum
 
 from sqlalchemy import and_
@@ -7,7 +5,6 @@ from sqlalchemy import or_
 
 from app import inventory_config
 from app.culling import staleness_to_conditions
-from app.culling import Timestamps
 from app.logging import get_logger
 from app.models import db
 from app.models import Host
@@ -18,7 +15,6 @@ from lib.db import session_guard
 
 __all__ = (
     "add_host",
-    "update_host_staleness",
     "canonical_fact_host_query",
     "canonical_facts_host_query",
     "create_new_host",
@@ -59,25 +55,6 @@ def add_host(input_host, staleness_offset, update_system_profile=True, fields=DE
             return update_existing_host(existing_host, input_host, staleness_offset, update_system_profile, fields)
         else:
             return create_new_host(input_host, staleness_offset, fields)
-
-
-def update_host_staleness(account_number, canonical_facts, staleness_offset):
-    """
-    Updates the staleness timestamp for a host with matching canonical facts.
-    """
-
-    existing_host = find_existing_host(account_number, canonical_facts)
-    if existing_host:
-        input_host = Host(
-            existing_host.canonical_facts,
-            stale_timestamp=datetime.now(timezone.utc) + staleness_offset,
-            reporter=existing_host.reporter,
-        )
-        return update_existing_host(
-            existing_host, input_host, Timestamps.from_config(inventory_config()), True, DEFAULT_FIELDS
-        )
-    else:
-        return None, None, None, None
 
 
 @metrics.host_dedup_processing_time.time()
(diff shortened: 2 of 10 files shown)


tahmidefaz:
@kruai since we are only updating a portion of the host data, I am thinking a `PATCH` maybe the correct method to use rather than a `POST`?

thearifismail:
@kruai tests?

kruai:
> @kruai since we are only updating a portion of the host data, I am thinking a `PATCH` maybe the correct method to use rather than a `POST`?

There was a lot of discussion on this, actually; the general consensus was to make it a `POST`. I think the reasoning was that we're adding a new check-in/heartbeat. 

> @kruai tests?
I'll add those once I actually have finished code lmao

jharting (changes requested):
@kruai don't forget to emit an `updated` event for any host state change

kruai:
> @kruai don't forget to emit an `updated` event for any host state change

Didn't realize that was a thing! Will do.



> is there a test attempting to cull a host with timedelta < 14 days?

There are so many tests it's hard for me to answer this, but I also don't know if that's relevant to this PR! Unless you're asking if I've added tests for my check-in endpoint that set the culling timestamp at less than 14 days... to which the answer is no

lphiri:
@jharting @kruai  - how would this behave if the canonical facts passed in did not match any hosts?
Would this create a new host entry?  I am assuming we do not want this to be backdoor for adding hosts via the REST API,
which was recently removed.

kruai:
> @jharting @kruai - how would this behave if the canonical facts passed in did not match any hosts?
> Would this create a new host entry? I am assuming we do not want this to be backdoor for adding hosts via the REST API,
> which was recently removed.

At the moment, this scenario errors out with my code. I can add a check to return a 400 response if a matching host isn't found. Does that work?

lphiri:
> > @jharting @kruai - how would this behave if the canonical facts passed in did not match any hosts?
> > Would this create a new host entry? I am assuming we do not want this to be backdoor for adding hosts via the REST API,
> > which was recently removed.
> 
> At the moment, this scenario errors out with my code. I can add a check to return a 400 response if a matching host isn't found. Does that work?

@kruai yes, returning a 400 works

kruai:
I need to debug to figure out why my my new test's requests are returning 400. I think it might have something to do with the Insights ID or Account atm

tahmidefaz (changes requested):
In addition to the test, I also have the following questions/suggestions:

* When I sent out the following invalid JSON object, I get a `500`. I think it should throw a `4xx` or something else
```
{
	"canonical_facts": {"insights_id": "3f01b55457674041b75e41829bcee1dc"}
}
```
* When I sent a request with `staleness_offset: 60`, the `staleness_warning_timestamp` shrunk, but the `culled_timestamp` grew to 1 hour in the future. Is this expected behavior?

tests/test_api_hosts_update.py
```diff
@@ -42,6 +46,29 @@ def test_update_fields(patch_doc, event_producer_mock, db_create_host, db_get_ho
...
+    expected_stale_timestamp = host.stale_timestamp + timedelta(minutes=staleness_offset)
+    response_status, response_data = api_put(build_host_checkin_url(), put_doc)
+
+    assert_response_status(response_status, expected_status=201)
+    record = db_get_host(created_host.id)
+
+    expected_stale_timestamp = original_timestamp + timedelta(minutes=staleness_offset)
+
+    assert record.stale_timestamp == expected_stale_timestamp
+
```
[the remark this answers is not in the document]
kruai: For sure! I'll re-request your review once I add that test and make some fixes. Thank you

kruai:
> * When I sent a request with `staleness_offset: 60`, the `staleness_warning_timestamp` shrunk, but the `culled_timestamp` grew to 1 hour in the future. Is this expected behavior?

I'm actually not sure what people are expecting for the other timestamps tbh. My goal was to push only the `stalness_timestamp` back, but I'm not aware of how this should affect other timestamps. How have these been handled in the past @tahmidefaz ?

ryandillinfelton:
> > * When I sent a request with `staleness_offset: 60`, the `staleness_warning_timestamp` shrunk, but the `culled_timestamp` grew to 1 hour in the future. Is this expected behavior?
> 
> I'm actually not sure what people are expecting for the other timestamps tbh. My goal was to push only the `stalness_timestamp` back, but I'm not aware of how this should affect other timestamps. How have these been handled in the past @tahmidefaz ?

The other timestamps are just a fixed offset away from the base `staleness_timestamp`. You can create the whole set of updated timestamps using the base one and the functions defined [here](https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/culling.py#L25). Here's an [example of them being updated](https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/queue/queue.py#L101).

api/host.py
```diff
@@ -438,3 +442,32 @@ def _build_serialized_tags(host_list, search):
...
     json_output = build_collection_response(tags_list, page, per_page, total)
     return flask_json_response(json_output)
+
+
+@api_operation
+@rbac(Permission.WRITE)
+@metrics.api_request_time.time()
+def host_checkin(body):
+    facts = body.get("canonical_facts")
+    staleness_offset = body.get("staleness_offset")
```
gmcculloug, line 452: @kruai I'm thinking this parameter should be something like `checkin_frequency` for the caller.  If we use it to determine the `staleness_offset` would be an internal decision for the app.  At the moment we may not use it to adjust to staleness_timestamp, but instead to keep track of the next expected checkin.  This allows a caller to use that information to determine how late a system is for a checking in.

Thoughts?

kruai: That works for me, especially if this value is also going to affect the other timestamps.

kruai:
> > > * When I sent a request with `staleness_offset: 60`, the `staleness_warning_timestamp` shrunk, but the `culled_timestamp` grew to 1 hour in the future. Is this expected behavior?
> > 
> > 
> > I'm actually not sure what people are expecting for the other timestamps tbh. My goal was to push only the `stalness_timestamp` back, but I'm not aware of how this should affect other timestamps. How have these been handled in the past @tahmidefaz ?
> 
> The other timestamps are just a fixed offset away from the base `staleness_timestamp`. You can create the whole set of updated timestamps using the base one and the functions defined [here](https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/culling.py#L25). Here's an [example of them being updated](https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/queue/queue.py#L101).

It took me a while to realize that only the staleness_timestamp is stored, and the other 2 timestamps are only generated while outputting (during serialization). I've changed it up so that I modify the staleness_timestamp and use the default config for output

tests/test_api_hosts_update.py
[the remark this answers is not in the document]
kruai: I assumed that defining a default value in `api.spec.yml` would make it supply the value in `host.py`, but I guess that was wrong. I'm now defaulting the value of checkin_frequency if it isn't passed in

tahmidefaz (changes requested):
I think there's still an issue with the timestamp fields. 
I had a host with the following timestamps:
```
"stale_timestamp": "2020-11-03T06:20:08+00:00",
"stale_warning_timestamp": "2020-11-10T06:20:08+00:00",
"culled_timestamp": "2020-11-17T06:20:08+00:00",
```
Then, I updated it with `staleness_offset: 60`. This updated the host to have: 
```
"stale_timestamp": "2020-11-03T07:20:08+00:00",
"stale_warning_timestamp": "2020-11-03T08:20:08+00:00",
"culled_timestamp": "2020-11-17T07:20:08+00:00",
```
The `culled_timestamp` value looks good, however, the `stale_warning_timestamp` is incorrect. I think it should be `"2020-11-10T07:20:08+00:00"`. Can you repro this? I am not sure where exactly you need to make this change. You might need to do some digging :slightly_smiling_face:

tahmidefaz (approved):
Ok, I think these changes are LGTM. However, I would feel much comfortable with at least 1 more :heavy_check_mark: before merging :smile:

kruai:
@jharting @ryandillinfelton @subpop @gravitypriest @lphiri @iphands: Any of you wanna weigh in before it's merged?

gmcculloug:
@jharting @subpop @gravitypriest Would really appreciate getting your review on this PR.  Thanks

subpop:
I don't feel I can give an overall LGTM as I am not familiar with the project codebase, but the API interface that the client will be interacting with LGTM.

thearifismail:
> There are so many tests it's hard for me to answer this, but I also don't know if that's relevant to this PR! Unless you're asking if I've added tests for my check-in endpoint that set the culling timestamp at less than 14 days... to which the answer is no

IMO, this test is relevant.  Since its relevance spreads beyond this PR, this PR should be not be blocked.

thearifismail (approved):
LGTM.  All tests executed successfully.

kruai:
I don't have the option to dismiss @jharting 's review requesting changes (or the ability to re-request a review), so I'm going to need @tahmidefaz or @ryandillinfelton to merge this

iphands (approved):
Testing

ryandillinfelton (approved):
LGTM

kruai:
@ryandillinfelton @tahmidefaz I'm not allowed to merge it... can you?
Or do we have to wait for @jharting to update his review?

gmcculloug:
@iphands The other week you were thinking that the `staleness` timestamp should just be bumped by the default internal (1 day) during the ultra-light check-in instead of the check-in frequency value.  Is that still what you are thinking you want?

If I'm reading the code, and PR description, correctly we are still using the [check-in frequency with an default of 1 day]( https://github.com/RedHatInsights/insights-host-inventory/pull/720/files#diff-a5c55b46cc9c536070cb56798c97adfaR452).

I had made reference to this previously in https://github.com/RedHatInsights/insights-host-inventory/pull/720#discussion_r500647020 which is easily missed.

tests/test_api_hosts_update.py
```diff
@@ -42,6 +46,41 @@ def test_update_fields(patch_doc, event_producer_mock, db_create_host, db_get_ho
...
+    put_doc = {
+        "canonical_facts": {"insights_id": f"{created_host.canonical_facts['insights_id']}"},
+        "checkin_frequency": checkin_frequency,
+    }
+
+    response_status, response_data = api_put(build_host_checkin_url(), put_doc)
+
+    assert_response_status(response_status, expected_status=201)
+    record = db_get_host(created_host.id)
+    expected_stale_timestamp = original_timestamp + timedelta(minutes=checkin_frequency)
```
jharting, line 65: I don't think this is correct. The stale_timestamp should not be set to previous stale_timestamp + checkin_frequency. e.g. if stale_timestamp is set to 1 month into the future and checkin_frequence is 60 minutes the outcome shouldn't be 1 month + 60 minutes. Instead, stale_timestamp should be set to now() + 60 minutes, i.e. if the host does not do another "ultralight" check-in 60 minutes from now it is considered stale.

api/host.py
```diff
@@ -440,3 +442,27 @@ def _build_serialized_tags(host_list, search):
...
+@rbac(Permission.WRITE)
+@metrics.api_request_time.time()
+def host_checkin(body):
+    facts = body.get("canonical_facts")
+    staleness_offset = timedelta(minutes=body.get("checkin_frequency") or 1440)
+    serialized_host, host_id, insights_id, updated = update_host_staleness(
+        current_identity.account_number, facts, staleness_offset
+    )
+    if serialized_host:
+        _emit_patch_event(serialized_host, host_id, insights_id)
```
jharting, line 457: The produced event is not correct. It is missing host tags, system_profile and request_id information.

jharting (changes requested):
I don't think that the implemented logic matches the expected behavior (see inline comment).

Moreover, I did not find any test coverage for the updated event emitted on host checkin. Manual testing revealed problems with the event payload.

lib/host_repository.py
```diff
@@ -57,6 +59,25 @@ def add_host(input_host, staleness_offset, update_system_profile=True, fields=DE
...
 
+def update_host_staleness(account_number, canonical_facts, staleness_offset):
+    """
+    Updates the staleness timestamp for a host with matching canonical facts.
+    """
+
+    existing_host = find_existing_host(account_number, canonical_facts)
+    if existing_host:
+        input_host = Host(
+            {"insights_id": existing_host.canonical_facts["insights_id"]},
```
jharting, line 70: what if the host does not have insights_id?

kruai: Oh, I thought they all had to have `insights_id`. Guess I'll need to revise this

api/host.py
kruai: I'm not sure what I'm doing wrong here... I'm using the function in the exact same way as `patch_by_id`. Looking into it, though

tests/test_api_hosts_update.py
kruai: Oh, okay. I can change it to do that

kruai: Updated

lib/host_repository.py
kruai: I'm now passing in the existing canonical facts instead of assuming that `insights_id` exists.

api/host.py
kruai: The only thing I could think of is to set "update_system_profile" to true, but I'm not sure how to test the produced events yet. Got any idea what I could be missing here?

lib/host_repository.py
```diff
@@ -57,6 +61,25 @@ def add_host(input_host, staleness_offset, update_system_profile=True, fields=DE
...
 
 
+def update_host_staleness(account_number, canonical_facts, staleness_offset):
+    """
+    Updates the staleness timestamp for a host with matching canonical facts.
+    """
+
+    existing_host = find_existing_host(account_number, canonical_facts)
+    if existing_host:
+        input_host = Host(
```
jharting, line 71: I think that rather than creating new Host() and then calling `update_existing_host` (which is used to update the entire host on upload) it would be better to do

```
existing_host._update_stale_timestamp()
db.session.commit()
```
i.e. update the field(s) you need and commit the session.

Then, not unlike in the PATCH case, serialize the representation and emit the event
https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/host.py#L306

jharting: Note that it might be a good idea to define a specific reporter for these ultralight checkins. (As otherwise we risk confusion as of why this host with reporter yupana has stale_timestamp that yupana would never set?)

kruai: OK... What value makes sense for this new reporter value, then? Should I just make one up, and call it "heartbeat" or "checkin" or something? I'd like @gmcculloug and @iphands opinions on this one

jharting: Yes, I think we should define a new reporter value. Either of the examples you used looks good to me.

kruai: I've made the requested changes. Afterwards, the function in host_repository didn't seem necessary, so I've consolidated the code into the function in host.py.
I'm marking this as "resolved" to show that I've addressed these concerns, but please un-resolve it if there's something I missed.

api/host.py
kruai: It now functions nearly identically to the PATCH call. Can you please un-resolve this if you're still seeing the issue?

jharting (approved):
Functionality-wise looks good to me. I made a few improvements to the tests.

kruai:
Awesome, thank you!! I'll resolve the merge conflict(s) and merge.

swagger/api.spec.yaml
```diff
@@ -714,6 +751,24 @@ components:
...
+      title: Check-in data
+      description: >-
+        Data required to create a check-in record for a host.
+      type: object
+      required:
+        - canonical_facts
+      properties:
+        canonical_facts:
+          description: A set of canonical_facts belonging to the host.
+          $ref: '#/components/schemas/Facts'
```
Glutexo, line 764: This is not _canonical facts_, but _facts_.

Glutexo (commented):
I found a bug in the schema. There is a field that should contain canonical facts, but it’s schema is defined as Host facts. The fact that this went unnoticed also points to a fact that this is not tested.

Glutexo (commented):
One more comment to the schema. The new _CreateCheckIn_ schema is the only place in the 
whole specification where Canonical Facts is explicitly mentioned. We don’t expose this concept; for the outside world, these are first class properties of a Host object. Thus I don’t think it is correct to break this concept in the new endpoint.

I’d say, just put them in the root CreateCheckIn schema. _Flat is better than nested._ I personally like nested more, and even here, we could just call it like _host_ instead of _canonical_facts_, but it would be inconsistent.

```yaml
CreateCheckIn:
  …
  properties:
    insights_id: …
    suscription_manager_id: …
    …
    checkin_frequency: …
```

Of course, my suggestion to DRY the Canonical Facts property set to an extracted schema and use it here still applies. It would require to use the _allOf_ JSON Schema combinator.

Glutexo (commented):
Is it correct to return _400 Bad Request_ when there is no matching host found? I’d expect _404 Not Found_ instead. I thought that _400 Bad Request_ should occur only if the request is malformed.

Gave this a second thought after reading https://stackoverflow.com/a/44915434/1307676. That use case is however slightly different from ours. Trying to wrap my head around our check-in, I imagine that the request consists of two parts: ⑴ matching the host and ⑵ writing the check-in data. No matter whether we consider this to be a new Check-in resource operation, or a direct Host update, it could be split into two requests: ⑴ would match the host and either redirect or bail out with _404 Not Found_. ⑵ would return _200 OK_ or _201 Created_ depending on the semantics.

api/host.py
```diff
@@ -441,3 +445,29 @@ def _build_serialized_tags(host_list, search):
...
+    facts = body.get("canonical_facts")
+    staleness_offset = timedelta(minutes=body.get("checkin_frequency") or 1440)
+    existing_host = find_existing_host(current_identity.account_number, facts)
+
+    if existing_host:
+        existing_host._update_stale_timestamp(datetime.now(timezone.utc) + staleness_offset, "checkin")
+        db.session.commit()
+        serialized_host = serialize_host(existing_host, staleness_timestamps(), EGRESS_HOST_FIELDS)
+        _emit_patch_event(serialized_host, existing_host.id, existing_host.canonical_facts.get("insights_id"))
+        return flask_json_response(serialized_host, 201)
```
rafaelcassau, line 463: I think return 200 is more accurate in this context, 201 is commonly used when a new resource was created.

Glutexo: That’s questionable. We may consider this being a new (virtual) check-in resource, not a direct modification of the Host. In that case, we should be using POST and not PUT. Even if we consider this to be a Host operation, PUT is still not appropriate – PATCH is.

api/host.py
```diff
@@ -441,3 +445,29 @@ def _build_serialized_tags(host_list, search):
...
+    else:
+        return flask_json_response(
+            {
+                "detail": "No hosts match the provided facts.",
+                "status": 400,
+                "title": "Bad Request",
+                "type": "about:blank",
+            },
+            status=400,
```
rafaelcassau, line 472: maybe 404 fits better than 400 in this scenario.

Glutexo: Great point! Please see my comment above and tell me what you think! https://github.com/RedHatInsights/insights-host-inventory/pull/720#pullrequestreview-519590906

api/host.py
rafaelcassau: yes, I'll argue about PATCH instead PUT because it just update a part of a host object instead of the whole object, I agree with you about semantic, I prefer to use PATCH but POST can be also considered.

rafaelcassau:
> Is it correct to return _400 Bad Request_ when there is no matching host found? I’d expect _404 Not Found_ instead. I thought that _400 Bad Request_ should occur only if the request is malformed.
> 
> Gave this a second thought after reading https://stackoverflow.com/a/44915434/1307676. That use case is however slightly different from ours. Trying to wrap my head around our check-in, I imagine that the request consists of two parts: ⑴ matching the host and ⑵ writing the check-in data. No matter whether we consider this to be a new Check-in resource operation, or a direct Host update, it could be split into two requests: ⑴ would match the host and either redirect or bail out with _404 Not Found_. ⑵ would return _200 OK_ or _201 Created_ depending on the semantics.

I agree with you, it's a hard trade off, my point is, I agree 404 is related with the resource, but I always prefer to return more meaningful status code to the client, from the HTTP verb perspective 404 is not the best choice, but I think 400 is too generic, according with mozilla definition maybe 422 can be a better choice, what do you think?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422

api/host.py
Glutexo: It all depends on whether we consider the check-in to be a new resource.

- If yes: then it should be POST. We **create** a new object that internally manifests as a host stale timestamp update.
- If no: it should be PATCH, because PATCH provides a subset of object information, whereas PUT replaces the entire object.

After a discussion with Greg we both think that the former is correct. This is not an immediate patch, because we don’t provide new Host stale timestamp. We provide some information that the Inventory **uses** to calculate the new timestamp.

Does this make sense?

rafaelcassau: > It all depends on whether we consider the check-in to be a new resource.
> 
> * If yes: then it should be POST. We **create** a new object that internally manifests as a host stale timestamp update.
> * If no: it should be PATCH, because PATCH provides a subset of object information, whereas PUT replaces the entire object.
> 
> After a discussion with Greg we both think that the former is correct. This is not an immediate patch, because we don’t provide new Host stale timestamp. We provide some information that the Inventory **uses** to calculate the new timestamp.
> 
> Does this make sense?

yes, make sense, in this case we need to change the verb and we can maintain with 201 status code.

Glutexo:
@rafaelcassau If we consider the Check-in to be bound to a specific Host, we can translate it (virtually or physically) into the match and update parts. Like that the Check-in create would be equivatlent to _POST /host/{host_id}/check-in_, in which case 404 Not Found would be the most appropriate choice.

I however agree that _422 Unprocessable Entity_ sounds nice and the description fits kind of. [Here](https://stackoverflow.com/a/44915434/1307676) though, one user says that this is a special-purpose status code introduced as a WebDAV extension and it is not thus a wise choice for a simple HTTP REST operation.

In conclusion, I’d go with

- _404 Not Found_ if Check-in is considered a new resource.
- _422 Unprocessable Entity_ or _400 Bad Request_ if Check-in is considered a patch on a Host.

As already mentioned above, I am more in favor of the former.

kruai:
Y'all are leaving so many useful comments which would have been great before this merged 2 weeks ago :sweat_smile: 
I'm going to have to respond to most of this once I'm out of meetings; however, on the topic of POST vs PATCH, I'd lean towards PATCH. I have to squint really hard to see this thing as a "create" operation, especially since it's explicitly only an update on the code side

kruai:
> One more comment to the schema. The new _CreateCheckIn_ schema is the only place in the
> whole specification where Canonical Facts is explicitly mentioned. We don’t expose this concept; for the outside world, these are first class properties of a Host object. Thus I don’t think it is correct to break this concept in the new endpoint.
> 
> I’d say, just put them in the root CreateCheckIn schema. _Flat is better than nested._ I personally like nested more, and even here, we could just call it like _host_ instead of _canonical_facts_, but it would be inconsistent.
> 
> ```yaml
> CreateCheckIn:
>   …
>   properties:
>     insights_id: …
>     suscription_manager_id: …
>     …
>     checkin_frequency: …
> ```
> 
> Of course, my suggestion to DRY the Canonical Facts property set to an extracted schema and use it here still applies. It would require to use the _allOf_ JSON Schema combinator.

I get what you mean - the host creation API is already deprecated, so I may as well delete the CreateHostIn schema, and just use the flat CreateCheckIn schema. Am I understanding you correctly?

rafaelcassau:
> @rafaelcassau If we consider the Check-in to be bound to a specific Host, we can translate it (virtually or physically) into the match and update parts. Like that the Check-in create would be equivatlent to _POST /host/{host_id}/check-in_, in which case 404 Not Found would be the most appropriate choice.
> 
> I however agree that _422 Unprocessable Entity_ sounds nice and the description fits kind of. [Here](https://stackoverflow.com/a/44915434/1307676) though, one user says that this is a special-purpose status code introduced as a WebDAV extension and it is not thus a wise choice for a simple HTTP REST operation.
> 
> In conclusion, I’d go with
> 
> * _404 Not Found_ if Check-in is considered a new resource.
> * _422 Unprocessable Entity_ or _400 Bad Request_ if Check-in is considered a patch on a Host.
> 
> As already mentioned above, I am more in favor of the former.

agree

PR #721, opened 2020-09-28
[WIP] Add delete handling for rbac-inventory

thearifismail:
Added delete handler to implement the CRUD methods for rbac-inventory config


lib/middleware.py
```diff
@@ -77,8 +77,8 @@ def modified_func(*args, **kwargs):
                 if (
                     rbac_permission["permission"] == Permission.ADMIN.value  # inventory:*:*
                     or rbac_permission["permission"] == Permission.HOSTS_ALL.value  # inventory:hosts:*
-                    or rbac_permission["permission"] == f"inventory:*:{permission_type}"  # inventory:*:(read | write)
-                    or rbac_permission["permission"] == required_permission.value  # inventory:hosts:(read | write)
+                    or rbac_permission["permission"] == f"inventory:*:{permission_type}"  # inventory:*:(read | write | delete)
+                    or rbac_permission["permission"] == required_permission.value  # inventory:hosts:(read | write| delete)
```
ryandillinfelton, line 81: I know I added these comment, but do y'all actually find them helpful or are they just clutter? I'm a bit partial to removing them, doesn't have to happen in the this PR though (sorry this is a little off-topic, just came to mind while I was reviewing).

tests/helpers/api_utils.py
```diff
@@ -73,6 +73,22 @@
...
+)
+
+DELETE_PROHIBITED_RBAC_RESPONSE_FILES = (
+    "utils/rbac-mock-data/inv-write-only.json",
+    "utils/rbac-mock-data/inv-none.json",
+    "utils/rbac-mock-data/inv-read-only.json",
+    "utils/rbac-mock-data/inv-read-write.json",
+    "utils/rbac-mock-data/inv-star-read.json",
+    "utils/rbac-mock-data/inv-star-write.json",
+    "utils/rbac-mock-data/inv-write-only.json"
```
ryandillinfelton, line 89: Write only is in here twice

lib/middleware.py
thearifismail: I found them helpful.  So no harm in leaving them in

kruai: Personally, it's a bit of both. It's helpful to me as a newcomer, but it's also more cluttered/harder on the eyes. I'm fine with it being removed

ryandillinfelton (changes requested):
Just one small change needed, otherwise LGTM.

I did want to comment on the coverage change though. I doubt it will be a problem, but we are no longer covering every permutation of permissions possible.

We should explore a solution to the problem in the future because I don't think it's practical to keep creating a bunch of mock response files for every possible combination of permissions a user could have. I don't think it's urgent to address though, it seems unlikely we'd have a bug that wouldn't get caught in these tests.

ryandillinfelton:
Try running `make style` and then enabling the pre-commit hook with `pre-commit install`

ryandillinfelton (approved):

ryandillinfelton:
Fix the style failure before this is merged, but it looks good outside of that

tests/helpers/api_utils.py
```diff
@@ -73,6 +73,21 @@
     "utils/rbac-mock-data/inv-star-read.json",
 )
```
tahmidefaz, line 74: `"utils/rbac-mock-data/inv-host-delete.json"` needs to be added to this list, as well as under `READ_PROHIBITED_RBAC_RESPONSE_FILES`

utils/rbac-mock-data/inv-host-delete.json
```diff
@@ -0,0 +1,19 @@
...
+      "previous": null,
+      "last": "/api/rbac/v1/access/?application=inventory&limit=1000&offset=0"
+  },
+  "data": [
+      {
+          "resourceDefinitions": [],
+          "permission": "inventory:hosts:delete"
+      }
+  ]
+}
```
tahmidefaz, line 19: `</nitpick>` might be better to rename this file as `inv-delete-only.json` to align with the other files in the directory, but it's not really a big issue. :slightly_smiling_face:

tahmidefaz (changes requested):
Just requested some changes. Overall, it looks good to me. Please run `make style` like @ryandillinfelton suggested :slightly_smiling_face:

lib/middleware.py
tahmidefaz: I agree that it's clutter. After some time, the variable names become intuitive enough that we don't really need those comments. :slightly_smiling_face:

kruai:
If the consensus ends up being that DELETE permission is the same as WRITE permission, do we still want to distinguish between them in Inventory?

thearifismail:
> If the consensus ends up being that DELETE permission is the same as WRITE permission, do we still want to distinguish between them in Inventory?

I don't think so.  For clarity, may be mention it somehwere.

tahmidefaz:
Closing this because we will be standardizing in `read` and `write` at the moment.

PR #722, opened 2020-09-29
Make tags into json string

BlakeHolifield:
I'd like to open this up for discussion. In the Kibana instance of the crcp01ue1  cluster, a large quantity (Kibana is showing 8400 / 11000) of the field mappings are being taken up by the input.* fields. By swapping to a json string, we can keep the data and limit the fields that are being created. Let me know your thoughts on the change here. I'll include the ticket below. 

https://issues.redhat.com/browse/RHCLOUD-9488

### DIFF
--- a/app/queue/queue.py
+++ b/app/queue/queue.py
@@ -108,7 +108,7 @@ def add_host(host_data):
                         "canonical_facts": input_host.canonical_facts,
                         "reporter": input_host.reporter,
                         "stale_timestamp": input_host.stale_timestamp.isoformat(),
-                        "tags": input_host.tags,
+                        "tags": json.dumps(input_host.tags),
                     }
                 },
             )


jharting:
@kruai @BumbleFeng has this been dealt with in https://github.com/RedHatInsights/kibana/pull/5 ?

kruai:
> @kruai @BumbleFeng has this been dealt with in [RedHatInsights/kibana#5](https://github.com/RedHatInsights/kibana/pull/5) ?

It looks like that should have worked, but the index is still showing `host.tags` fields. It would be great if we could keep the fields but not index them... I can't tell why the dynamic template isn't working, though :/

@jharting would using a JSON string break any existing functionality that we need? I don't know what the tags are used for on the logging/Kibana level

jharting:
@kruai it should not break anything. I think it would be cleaner to configure what not to index on the mapping level. This approach should work too, though.

BlakeHolifield:
Definitely agreed there @jharting. This is more of a quick fix to resolve some issues other folks are seeing from full mappings. We're working on a more sustainable fix going forward.

kruai (approved):

kruai:
Anybody else wanna weigh in before merge @tahmidefaz @ryandillinfelton @thearifismail ?

BumbleFeng:
> @kruai @BumbleFeng has this been dealt with in [RedHatInsights/kibana#5](https://github.com/RedHatInsights/kibana/pull/5) ?

Amazon Elasticsearch Service did not support flattened type, so all the filed inside host.tags will still be created.
It can still easily hit the limit even the filed limit has been increased to 4000.
`# curl -s -XGET https://vpc-kibana-elasticsearch-fcam2sdlsyzhszeojgwoxyquq4.us-east-1.es.amazonaws.com/cwl-2020.09.16/_mapping?pretty | jq '.["cwl-2020.09.16"]["mappings"]["properties"]["input_host"]["properties"]["tags"]' | grep type | wc -l
1730 
# curl -s -XGET https://vpc-kibana-elasticsearch-fcam2sdlsyzhszeojgwoxyquq4.us-east-1.es.amazonaws.com/cwl-2020.09.16/_mapping?pretty | grep type | wc -l 
2341
`
As you can see, Most fields are consumed by "input_host.tags"

tahmidefaz (approved):

PR #724, opened 2020-10-01
Add sap sids filter

ryandillinfelton:
Could use some comments/help with swagger validation. Current version is overly permissive and doesn't enforce filter parameter names (i.e. `?filter[bacon][beans]=true` does not get caught)

### DIFF
--- a/api/host_query_db.py
+++ b/api/host_query_db.py
@@ -30,15 +30,7 @@ def get_host_list(
     order_how,
     staleness,
     registered_with,
-<<<<<<< HEAD
-<<<<<<< HEAD
     filter,
-=======
-    filter
->>>>>>> c8ea837... fixes & more tests
-=======
-    filter,
->>>>>>> 73aeb16... style
 ):
     if filter:
         flask.abort(503)
--- a/api/host_query_db.py
+++ b/api/host_query_db.py
@@ -30,7 +30,15 @@ def get_host_list(
     order_how,
     staleness,
     registered_with,
+<<<<<<< HEAD
+<<<<<<< HEAD
     filter,
+=======
+    filter
+>>>>>>> c8ea837... fixes & more tests
+=======
+    filter,
+>>>>>>> 73aeb16... style
 ):
     if filter:
         flask.abort(503)
--- a/api/host_query_db.py
+++ b/api/host_query_db.py
@@ -1,5 +1,6 @@
 from uuid import UUID
 
+import flask
 from sqlalchemy import and_
 from sqlalchemy import or_
 
@@ -29,7 +30,11 @@ def get_host_list(
     order_how,
     staleness,
     registered_with,
+    filter,
 ):
+    if filter:
+        flask.abort(503)
+
     if fqdn:
         query = _find_hosts_by_canonical_fact("fqdn", fqdn)
     elif display_name:
--- a/api/host_query_xjoin.py
+++ b/api/host_query_xjoin.py
@@ -111,11 +111,18 @@ def _params_to_order(param_order_by=None, param_order_how=None):
 
 def _sap_system_filters(sap_system):
     if sap_system == "nil":
-        return ({"spf_sap_system": {"is": None}},)
+        return {"spf_sap_system": {"is": None}}
     elif sap_system == "not_nil":
-        return ({"NOT": {"spf_sap_system": {"is": None}}},)
+        return {"NOT": {"spf_sap_system": {"is": None}}}
     else:
-        return ({"spf_sap_system": {"is": (sap_system == "true")}},)
+        return {"spf_sap_system": {"is": (sap_system == "true")}}
+
+
+def build_sap_system_filters(sap_system):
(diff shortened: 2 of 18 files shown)


shdunning:
@ryandillinfelton is the issue that `?filter[bacon][beans]=true` isn't a valid system profile object?

ryandillinfelton:
> @ryandillinfelton is the issue that `?filter[bacon][beans]=true` isn't a valid system profile object?

Yes, anything other than
`?filter[system_profile][sap_system]` | `?filter[system_profile][sap_system][eq]`
`?filter[system_profile][sap_sids]` | `?filter[system_profile][sap_sids][eq]`
should be rejected

api/host_query_db.py
```diff
@@ -30,7 +30,15 @@ def get_host_list(
     order_how,
     staleness,
     registered_with,
+<<<<<<< HEAD
```
gmcculloug, line 33: @ryandillinfelton Looks like some merge conflict lines got pushed by mistake.

api/tag.py
```diff
@@ -83,18 +87,29 @@ def get_tags(
...
 
     if registered_with:
         variables["hostFilter"]["NOT"] = {"insights_id": {"eq": None}}
 
+    if filter:
+        if filter.get("system_profile"):
+            if filter["system_profile"].get("sap_system"):
+                hostfilter_and_variables += build_sap_system_filters(filter["system_profile"].get("sap_system"))
+            if filter["system_profile"].get("sap_sids"):
+                hostfilter_and_variables += build_sap_sids_filter(filter["system_profile"]["sap_sids"])
```
tahmidefaz, line 109: I wonder if we can bake this into a function, it seems repetitive. But, this is not a priority at this point, since we have that deadline coming up. Refactoring can wait :slightly_smiling_face:

ryandillinfelton: Might just cross that bridge when we add more system profile fields to filter. Or do it with the validation follow up.

tahmidefaz (approved):
Looks good to me! :+1:

kruai (approved):
I've been looking, but I can't find anything I'd change! The tests pass when I run it on my end, too. LGTM :+1:

gmcculloug:
@jharting Wouldn't mind having you take a look at this one.  Thanks.

api/host_query_xjoin.py
```diff
@@ -111,11 +111,32 @@ def _params_to_order(param_order_by=None, param_order_how=None):
...
+def build_sap_system_filters(sap_system):
+    if isinstance(sap_system, str):
+        return (_sap_system_filters(sap_system),)
+    elif sap_system.get("eq"):
+        return (_sap_system_filters(sap_system["eq"]),)
+
+
+def _sap_sids_filters(sap_sids):
+    sap_sids_filters = ()
+    for sap_sid in sap_sids.split(","):
```
jharting, line 130: This deviates from how querying for multiple values should look like, i.e.
```
?filter[system_profile][sap_sids][contains][]=E24&filter[system_profile][sap_sids][contains][]=E04
```

as defined in https://platform-docs.cloud.paas.psi.redhat.com/backend/inventory.html#using-system-profile-in-queries

ryandillinfelton: I'll keep looking into it, but I actually don't know if that's possible, I think swagger may have an incomplete implementation that doesn't allow for that array format. Right now if I submit a query like that the filter object only holds the first value in a dictionary {"": "val1"}. I think it might be related to the OpenAPI spec not knowing to expect an array, but I've also seen some things that make me think they might actually just not support the array format in deep objects. I'll update or add a reply if I find anything else out.

ryandillinfelton: It seems like the issue actually might lie with how connexion parses the query parameters. Looking into it now. Let me know if you or @Glutexo have any insight

karelhala: If you try to send an array on api docs the filter would look like (notice the missing `[]` after each contains)
```
?filter[system_profile][sap_sids][contains]=E24&filter[system_profile][sap_sids][contains]=E04
```
I've tried submitting an array from our ui code and the URL looks the same.

jharting: @ryandillinfelton right, looks like connexion does not parse this properly and a custom parser for the query param might be needed.

jharting: Actually it should be enough to extend https://github.com/zalando/connexion/blob/master/connexion/decorators/uri_parsing.py#L150 to support the array syntax. More specifically, it looks like the built-in parser picks the first value always https://github.com/zalando/connexion/blob/master/connexion/decorators/uri_parsing.py#L196

The logic should be changed so that when an array is provided (signalled by the `[]` empty key at the end of the query key) then all the values are used.

ryandillinfelton: Extended :partying_face:. Thank you for the help!

tahmidefaz (approved):
LGTM :+1:

PR #727, opened 2020-10-07
Add system_profile endpoints

ryandillinfelton:
Adds two new endpoints:
`system_profile/sap_system` and `system_profile/sap_sids`

the endpoints return respectively:
- all of the values of sap_system present on the account and the number of hosts with that value
- all of the values of sap_sids present on the account and the number of hosts with that value

example responses:
*/sap_system*
```
{
    "total": 2,
    "count": 2,
    "page": 1,
    "per_page": 50,
    "results": [
        {
            "value": false,
            "count": 1
        },
        {
            "value": true,
            "count": 2
        }
    ]
}
```

*/sap_sids*
```
{
    "total": 5,
    "count": 5,
    "page": 1,
    "per_page": 50,
    "results": [
        {
            "value": "AEB",
            "count": 2
        },
        {
            "value": "C2C",
            "count": 1
        },
        {
            "value": "KLL",
            "count": 1
        },
        {
            "value": "M1P",
            "count": 1
        },
        {
            "value": "XQC",
            "count": 1
        }
    ]
}
```

Please let me know if there is any missing test coverage, I believe Tahmid and I have covered everything we can in these tests (since xjoin-search doesn't run in the tests) but more coverage  is better if it's possible.

### DIFF
--- a/api/system_profile.py
+++ b/api/system_profile.py
@@ -5,6 +5,8 @@ from api import build_collection_response
 from api import flask_json_response
 from api import metrics
 from api.host import get_bulk_query_source
+from api.host_query_xjoin import build_sap_sids_filter
+from api.host_query_xjoin import build_sap_system_filters
 from api.host_query_xjoin import build_tag_query_dict_tuple
 from app import Permission
 from app.config import BulkQuerySource
@@ -68,7 +70,7 @@ def xjoin_enabled():
 @api_operation
 @rbac(Permission.READ)
 @metrics.api_request_time.time()
-def get_sap_system(tags=None, page=None, per_page=None, staleness=None, registered_with=None):
+def get_sap_system(tags=None, page=None, per_page=None, staleness=None, registered_with=None, filter=None):
     if not xjoin_enabled():
         flask.abort(503)
 
@@ -80,12 +82,23 @@ def get_sap_system(tags=None, page=None, per_page=None, staleness=None, register
             "OR": list(staleness_filter(staleness))
         }
     }
+    hostfilter_and_variables = ()
 
     if tags:
-        variables["hostFilter"]["AND"] = build_tag_query_dict_tuple(tags)
+        hostfilter_and_variables = build_tag_query_dict_tuple(tags)
 
     if registered_with:
         variables["hostFilter"]["NOT"] = {"insights_id": {"eq": None}}
+
+    if filter:
+        if filter.get("system_profile"):
+            if filter["system_profile"].get("sap_system"):
+                hostfilter_and_variables += build_sap_system_filters(filter["system_profile"].get("sap_system"))
+            if filter["system_profile"].get("sap_sids"):
+                hostfilter_and_variables += build_sap_sids_filter(filter["system_profile"]["sap_sids"])
+
+    if hostfilter_and_variables != ():
+        variables["hostFilter"]["AND"] = hostfilter_and_variables
 
     response = graphql_query(SAP_SYSTEM_QUERY, variables)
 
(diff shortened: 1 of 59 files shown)


ryandillinfelton:
@jharting can I get some confirmation that everything is set up as expected?

jharting:
> @jharting can I get some confirmation that everything is set up as expected?

In xjoin or this PR?

ryandillinfelton:
> > @jharting can I get some confirmation that everything is set up as expected?
> 
> In xjoin or this PR?

In this PR. Like are the endpoints set up as expected and are all of the filters we want to support for them supported.

api/system_profile.py
```diff
@@ -0,0 +1,131 @@
...
+@rbac(Permission.READ)
+@metrics.api_request_time.time()
+def get_sap_sids(tags=None, page=None, per_page=None, staleness=None, registered_with=None):
+    if not xjoin_enabled():
+        flask.abort(503)
+
+    limit, offset = pagination_params(page, per_page)
+
+    variables = {
+        "hostFilter": {
```
jharting, line 111: This should also forward the filter (system profile) parameters. I.e. if you already selected one sap_sid the response from this operation should reflect the selection (same was as /tags has tags filtering parameter)

ryandillinfelton: Thanks, I'll add that in

swagger/api.spec.yaml
```diff
@@ -366,6 +366,64 @@ paths:
...
+      responses:
+        '200':
+            description: sap_system values and counts for account
+            content:
+              application/json:
+                schema:
+                  $ref: '#/components/schemas/SystemProfileSapSystemOut'
+        '400':
+            description: Invalid request.
+        '404':
```
kruai, line 424: Is 404 the appropriate status for this? Since it's a valid page but the query params are invalid, I would have assumed a 400 or something. Maybe even an empty 2xx that tells you what the last valid page is in the headers

kruai: Also, is it able to return 404? I don't see that defined anywhere

kruai (commented):
Just had a couple comments... I also think it might be cleaner to move SAP_SYSTEM_QUERY and SAP_SIDS_QUERY to another file, if we have one for constants. If not, then no worries

swagger/api.spec.yaml
ryandillinfelton: I interpret the 404 here to be more of a 'page not found' deal. It's not that the parameter is *wrong* per se, but rather that the page that was asked for cannot be returned. It is also consistent with other endpoints like `/tags` which is what I based this spec addition on. The 404 is returned in [`check_pagination`](https://github.com/RedHatInsights/insights-host-inventory/pull/727/files/8a3e5fbe082084ce34489203e3a1372cebbdb82d#diff-1f71ab25cb7b15604b8d8545edb6c375R127)

kruai: OK. Still feels kinda weird, but as long as it's consistent

tahmidefaz (approved):
Tested it locally, looks good to me. Didn't see any bugs :+1:

ryandillinfelton:
> Just had a couple comments... I also think it might be cleaner to move SAP_SYSTEM_QUERY and SAP_SIDS_QUERY to another file, if we have one for constants. If not, then no worries

That's just where the other ones were. I agree we should consider moving them in the future if we need to add more though, it's getting a little crowded

ryandillinfelton:
the `filter` parameter should now work on this endpoint. I added two tests to test_xjoin to check that. Also the diff on this probably looks wacky right now since I had to merge the sap_sids branch in, I'll rebase this once that one is in master, so

NOTE: merge after #724 !

gmcculloug:
Thanks @ryandillinfelton.   👍 on the PR description.

PR #728, opened 2020-10-09
RHCLOUD-9897 fix race condition in PATCH operation

jharting:


api/host.py
```diff
@@ -303,10 +303,11 @@ def patch_by_id(host_id_list, body):
 
     for host in hosts_to_update:
         host.patch(validated_patch_host_data)
-        serialized_host = serialize_host(host, staleness_timestamps(), EGRESS_HOST_FIELDS)
-        _emit_patch_event(serialized_host, host.id, host.canonical_facts.get("insights_id"))
 
-    db.session.commit()
+        if db.session.is_modified(host):
```
kruai, line 307: I'm having trouble understanding this line addition... Was the race condition caused by one thread trying to DELETE a host, but another was trying to PATCH it even though there were no differences to patch?
Does the race condition not occur when the PATCH actually has changes to apply?

jharting: The race condition was caused by the patch event being emitted regardless of whether the db operation succeeded or not. I.e. the API would emit an update event even for a host that has been removed since it was read. Hence, I moved `db.session.commit()` in front of the event emitting code so that the event is emitted only if the db transaction succeeds. That mitigates the race condition.

While at it I also noticed that the event is emitted even when the update operation does not actually modify any field. Therefore, I also added the conditional clause that prevents that from happening.

gmcculloug:
@kruai @ryandillinfelton @tahmidefaz Please review

kruai (approved):

tahmidefaz (approved):
LGTM :+1:

PR #729, opened 2020-10-09
Update Kibana URL for Prod env

gmcculloug:


tahmidefaz (changes requested):
I tried navigating to the new kibana URLs but the logs didn't show up for me. It seems to me that the URL format for the new kibana is different from the old one.

tahmidefaz (approved):
On a deeper look, I think this looks good to me. The issue was that I was missing a query parameter URL, but it gets added later on in the dashboard. The two kibana just behaves differently when a query isn't provided. :+1:

kruai:
idk, shouldn't the URLs work immediately? I think we should also modify the URL paths so that they load correctly

kruai (changes requested):
idk, shouldn't the URLs work immediately? I think we should also modify the URL paths so that they load correctly

kruai (approved):

kruai:
I updated my review to "Approve"; @tahmidefaz explained that the URLs don't work because the query is appended to it later. I was under the impression that they should already work. My mistake :)

tahmidefaz:
@kruai we should be fine, you can't really break what's already broken :smile:

kruai:
Yeah, I just figured that if we're making a PR to change it, it makes sense to fix it instead of keeping it broken! But yeah I'll merge it

PR #731, opened 2020-10-13
Add addtional SAP facts to system profile

CodeHeeler:
Pending schema PR: https://github.com/RedHatInsights/inventory-schemas/pull/25

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -389,6 +389,15 @@ class SystemProfileSchema(MarshmallowSchema):
             validate=[marshmallow_validate.Length(max=3), marshmallow_validate.Regexp(regex=r"^[A-Z][A-Z0-9]{2}$")]
         )
     )
+    sap_instance_number = fields.Str(
+        validate=[marshmallow_validate.Length(max=2), marshmallow_validate.Regexp(regex=r"(?!99)(?!98)^[0-9]{2}$")]
+    )
+    sap_version = fields.Str(
+        validate=[
+            marshmallow_validate.Length(max=22),
+            marshmallow_validate.Regexp(regex=r"^[0-9].[0-9]{2}.[0-9]{3}.[0-9]{2}.[0-9]{10}$"),
+        ]
+    )
     tuned_profile = fields.Str(validate=marshmallow_validate.Length(max=256))
 
 
--- a/swagger/system_profile.spec.yaml
+++ b/swagger/system_profile.spec.yaml
@@ -245,6 +245,18 @@ $defs:
           example: "H2O"
           maxLength: 3
           pattern: '^[A-Z][A-Z0-9]{2}$'
+      sap_instance_number:
+        type: string
+        description: The instance number of the SAP HANA system
+        example: '42'
+        maxLength: 2
+        pattern: '(?!99)(?!98)^[0-9]{2}$'
+      sap_version:
+        type: string
+        description: The version of the SAP HANA lifecycle management program
+        example:  '1.00.122.04.1478575636'
+        maxLength: 22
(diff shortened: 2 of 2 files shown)


kruai (approved):
LGTM, but would probably be safe to wait for an additional review from a veteran :)

thearifismail:
Question... What's the source of`maxLength` values for `sap_sid`, `sap_instance_number`, and `sap_version_info`?  How do I know that the maxLengths in `system_profile_spec` are correct, i.e. is there no chance of `sap_instance_number` being 3 char long?

CodeHeeler:
> @CodeHeeler, Question... What's the source of`maxLength` values for `sap_sid`, `sap_instance_number`, and `sap_version_info`? How do I know that the maxLengths in `system_profile_spec` are correct, i.e. is there no chance of `sap_instance_number` being 3 char long?

The source for the sap_instance_number is here:
https://help.sap.com/viewer/2c1988d620e04368aa4103bf26f17727/2.0.05/en-US/ae305ab18bad42efa7b067e7cc82b846.html

The source for the version number is an example from Bernd Finger in Red Hat and the example here as this is the output for --version:
https://answers.sap.com/questions/161894/hana-version-check.html

The source for the sid's being 3 is here:
https://help.sap.com/viewer/2c1988d620e04368aa4103bf26f17727/2.0.05/en-US/7672838aa9c44c7ab9d10188531669f3.html?q=sids

Anything longer is incorrect, so we can either make a longer length for cushion for incorrect values or we can validate here and correct reporters that are not using correct values, as with any of the other facts.

thearifismail:
> Anything longer is incorrect, so we can either make a longer length for cushion for incorrect values or we can validate here and correct reporters that are not using correct values, as with any of the other facts.

No don't need to make longer as it has lower potential for bugs.  Was just curious because `maxLength` for `sap_sid` and `sap_instane_number` appeared too small.  If adjustments is needed in future, it can be dealt with then.

LGTM

thearifismail (approved):

ryandillinfelton (approved):

swagger/system_profile.spec.yaml
```diff
@@ -245,6 +245,18 @@ $defs:
           example: "H2O"
           maxLength: 3
           pattern: '^[A-Z][A-Z0-9]{2}$'
+      sap_instance_number:
+        type: integer
+        description: The instance number of the SAP HANA system
+        example: '42'
+        maxLength: 2
+        pattern: '(?!99)(?!98)^[0-9]{2}$'
```
ryandillinfelton, line 253: Tahmid had a good improvement suggestion for this field https://github.com/RedHatInsights/inventory-schemas/pull/25/files#r509548114

app/models.py
```diff
@@ -389,6 +389,15 @@ class SystemProfileSchema(MarshmallowSchema):
             validate=[marshmallow_validate.Length(max=3), marshmallow_validate.Regexp(regex=r"^[A-Z][A-Z0-9]{2}$")]
         )
     )
+    sap_instance_number = fields.Int(
+        validate=[marshmallow_validate.Length(max=2), marshmallow_validate.Regexp(regex=r"(?!99)(?!98)^[0-9]{2}$")]
```
ryandillinfelton, line 393: Using the [Range validation](https://www.programcreek.com/python/example/108225/marshmallow.validate.Range) method could make things cleaner here

ryandillinfelton (changes requested):

ryandillinfelton (approved):

ryandillinfelton (approved):

tahmidefaz (approved):

PR #735, opened 2020-10-27
Ultra-Light Check-In Restructuring (Phase 1)

kruai:
This PR restructures the [Ultra-Light Check-In](https://issues.redhat.com/browse/RHCLOUD-9749) code to match what we agreed upon for Phase 1. The input schema has been restructured, and the culling timestamps are no longer modified; instead, only the "modified" date on the Host record gets updated for Phase 1. Once @iphands gets some more solidified requirements, we can update this with Phase 2's functionality (TBD).

Due to the restructure, this PR also addresses these bugs: 
- https://issues.redhat.com/browse/RHCLOUD-10223
- https://issues.redhat.com/browse/RHCLOUD-10224
- https://issues.redhat.com/browse/RHCLOUD-10229 (and consequently removed the Host creation endpoint and tests)
- https://issues.redhat.com/browse/RHCLOUD-10385

After deleting the Host-creation tests, the only tests left in test_api_hosts_create.py were some unrelated System Profile tests; I moved these to test_system_profile.py.


kruai:
@subpop @lphiri I know we discussed having a "staleness offset" value while we were planning this, but is there any harm in just supplying a staleness timestamp like you do on host creation? This PR standardizes on the existing schema for creating a host, so it's accepting a timestamp instead of an offset value.

Glutexo:
cc this question to @jharting  and @iphands .

Glutexo:
May I ask to split the schema fixes from the timestamps change? The former is clear, it’s just a bug in the specification and it is an easy fix. The latter is a change in semantics and may require some more discussion and waiting time not to break reporters. 🙇🏻‍♂️

jharting:
> is there any harm in just supplying a staleness timestamp like you do on host creation?

I think it's cleaner to let the client provide a "hint" on when the next check-in should be expected (which inventory then uses to compute the next timestamp) instead of giving the client the authority to define the stale timestamp directly.

Glutexo:
One more comment to the schema format. Please see my newest comment: https://github.com/RedHatInsights/insights-host-inventory/pull/720#pullrequestreview-519429624

subpop:
> > is there any harm in just supplying a staleness timestamp like you do on host creation?
> 
> I think it's cleaner to let the client provide a "hint" on when the next check-in should be expected (which inventory then uses to compute the next timestamp) instead of giving the client the authority to define the stale timestamp directly.

Agreed. During host creation, the timestamp is actually set by puptoo. In this case, the timestamp value is coming directly from the client. A clever client/user could manipulate the code to set the timestamp to whatever they want.

kruai:
> May I ask to split the schema fixes from the timestamps change? The former is clear, it’s just a bug in the specification and it is an easy fix. The latter is a change in semantics and may require some more discussion and waiting time not to break reporters. 🙇🏻‍♂️

Can you clarify what you mean by the timestamps change? If you're talking about the way staleness_timestamps is calculated, that was already merged in #720...

Glutexo:
@kruai Please split the changes to the API – HTTP verbs, response codes, request schemas – from the changes to the logic – calculating the timestamp instead of taking it from the request.

kruai:
> @kruai Please split the changes to the API – HTTP verbs, response codes, request schemas – from the changes to the logic – calculating the timestamp instead of taking it from the request.

OK, gotcha. Because of the restructuring that came from our latest conversation on Ultra-Light Check-In, that timestamp logic will no longer be a part of this (until Phase 2, once the requirements have been re-solidified). I'll unmark this as Draft once I've got everything in order for those changes

api/host.py
```diff
@@ -451,22 +451,21 @@ def _build_paginated_host_tags_response(total, page, per_page, tags_list):
...
 def host_checkin(body):
-    facts = body.get("canonical_facts")
-    staleness_offset = timedelta(minutes=body.get("checkin_frequency") or 1440)
-    existing_host = find_existing_host(current_identity.account_number, facts)
+    canonical_facts = deserialize_canonical_facts(body)
+    existing_host = find_existing_host(current_identity.account_number, canonical_facts)
 
     if existing_host:
-        existing_host._update_stale_timestamp(datetime.now(timezone.utc) + staleness_offset, "checkin")
+        existing_host.modified_on = datetime.now(timezone.utc)
```
gmcculloug, line 458: Is this the same as calling `datetime.utcnow()`.  More a curiosity on my part as I see both used in this code base.

Glutexo: From the documentation ([link](https://docs.python.org/3.6/library/datetime.html#datetime.datetime.utcnow)):

> This […] returns the current UTC date and time, as a **naive** datetime object. An **aware** current UTC datetime can be obtained by calling `datetime.now(timezone.utc)`.

api/host.py
```diff
@@ -451,22 +451,21 @@ def _build_paginated_host_tags_response(total, page, per_page, tags_list):
...
         return flask_json_response(serialized_host, 201)
     else:
         return flask_json_response(
             {
-                "detail": "No hosts match the provided facts.",
-                "status": 400,
-                "title": "Bad Request",
+                "detail": "No hosts match the provided canonical facts.",
+                "status": 404,
+                "title": "Not Found",
```
Glutexo, line 468: Thanks!

app/serialization.py
```diff
@@ -71,6 +71,10 @@ def deserialize_host_mq(raw_data):
     return deserialize_host(raw_data, MqHostSchema)
 
 
+def deserialize_canonical_facts(raw_data):
+    return _deserialize_canonical_facts(raw_data)
```
Glutexo, line 75: Don’t we want to run this through Marshmallow? I know that in the long run, we are trying to remove these Python schemas, but now it would be good to do this consistently. It should be easy to extract the canonical fact fields from _BaseHostSchema_ and use it here and make _BaseHostSchema_ inherit them.

swagger/api.spec.yaml
```diff
@@ -59,60 +59,8 @@ paths:
...
-                  value:
-                    status: 400
-                    title: Invalid Request
-                    detail: Could not process request
-                    host: Input host data
-        '400':
-          description: Invalid request.
   /hosts/checkin:
-    put:
+    post:
```
Glutexo, line 63: Awesome!

Glutexo (commented):
Gave some suggestions, found some nice things which pleased me. 😃 I need to go run a Python lecture today, but I’ll check the rest tomorrow.

api/host.py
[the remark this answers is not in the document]
kruai: Will do!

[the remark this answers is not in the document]
kruai: OK, moved it inside the Host definition

app/serialization.py
kruai: OK! I haven't done any inheritance with Python yet, but this is my chance to learn. Might take me a bit longer than the other fix, though :)

Glutexo: My idea is to have a canonical facts schema like

```python
class CanonicalFactsSchema(MarshmallowSchema):
    insights_id = fields.Str(validate=verify_uuid_format)
    …
```

and then it should be as simple as

```python
class BaseHostSchema(CanonicalFactsSchema):
    display_name = fields.Str(validate=marshmallow_validate.Length(min=1, max=200))
    …
```

Glutexo (changes requested):
Second part. As repeated a few times in the in-line comments. Please don’t remove the create/update host by POST in this pull request. It is a big functional change that should be tested and merged separately. Thanks!

tests/test_api_hosts_update.py
```diff
@@ -48,46 +46,43 @@ def test_update_fields(patch_doc, event_producer_mock, db_create_host, db_get_ho
...
     assert_response_status(response_status, expected_status=201)
     record = db_get_host(created_host.id)
 
-    assert (record.stale_timestamp > expected_stale_timestamp) and (
-        record.stale_timestamp < expected_stale_timestamp + timedelta(seconds=1)
-    )
+    assert record.modified_on > updated_time
 
-    assert event_producer_mock.key == str(host.id)
+    assert event_producer_mock.key == str(created_host.id)
```
Glutexo, line 64: This is already checked by the _assert_patch_event_is_valid_ method below. This line is thus redundant.

```suggestion
```

tests/test_api_hosts_update.py
```diff
@@ -48,46 +46,43 @@ def test_update_fields(patch_doc, event_producer_mock, db_create_host, db_get_ho
...
     assert_patch_event_is_valid(
-        host, event_producer_mock, "123456", event_datetime_mock, host.display_name, record.stale_timestamp, "checkin"
+        created_host,
+        event_producer_mock,
+        "123456",
+        event_datetime_mock,
+        created_host.display_name,
+        record.stale_timestamp,
```
Glutexo, line 71: It’s strange to pass the stale timestamp and not the reporter. Both values fall back to the host’s value. We may however want to check that these values remain untouched. This can be easily achieved by removing both arguments.

```suggestion
```

tests/test_api_hosts_update.py
```diff
@@ -48,46 +46,43 @@ def test_update_fields(patch_doc, event_producer_mock, db_create_host, db_get_ho
...
+        build_host_checkin_url(), post_doc, extra_headers={"x-rh-insights-request-id": "123456"}
     )
 
     assert_response_status(response_status, expected_status=201)
     record = db_get_host(created_host.id)
 
-    assert (record.stale_timestamp > expected_stale_timestamp) and (
-        record.stale_timestamp < expected_stale_timestamp + timedelta(seconds=1)
-    )
+    assert record.modified_on > updated_time
```
Glutexo, line 62: Let’s add checks that _stale_timestamp_ and _reporter_ didn’t change.

tests/test_api_hosts_update.py
```diff
@@ -48,46 +46,43 @@ def test_update_fields(patch_doc, event_producer_mock, db_create_host, db_get_ho
...
+        created_host,
+        event_producer_mock,
+        "123456",
+        event_datetime_mock,
+        created_host.display_name,
+        record.stale_timestamp,
     )
 
 
 @pytest.mark.system_culling
```
Glutexo, line 75: Since this does not change the stale timestamp any more, it is no longer related to system culling. The test above doesn’t have this mark, so let’s remove it from this one too.

```suggestion
```

tests/test_api_hosts_update.py
```diff
@@ -48,46 +46,43 @@ def test_update_fields(patch_doc, event_producer_mock, db_create_host, db_get_ho
...
-    created_host = db_create_host(host)
+def test_checkin_no_matching_host(event_producer_mock, db_create_host, db_get_host, api_post):
+    created_host = db_create_host()
 
-    put_doc = {
-        "canonical_facts": {"insights_id": f"nomatch_{created_host.canonical_facts['insights_id']}"},
-        "checkin_frequency": 60,
-    }
+    post_doc = created_host.canonical_facts
+    post_doc["insights_id"] = f"nomatch_{created_host.canonical_facts['insights_id']}"
```
Glutexo, line 80: This is not correct. Because with the prefix, the value is no longer a valid UUID and thus does not pass validation. Let’s generate a random UUID instead.

```suggestion
    post_doc["insights_id"] = generate_uuid()
```

tests/test_api_hosts_update.py
```diff
@@ -48,46 +46,43 @@ def test_update_fields(patch_doc, event_producer_mock, db_create_host, db_get_ho
...
+
+    response_status, response_data = api_post(
+        build_host_checkin_url(), post_doc, extra_headers={"x-rh-insights-request-id": "123456"}
+    )
 
-    response_status, response_data = api_put(build_host_checkin_url(), put_doc)
     assert_response_status(response_status, expected_status=400)
```
Glutexo, line 86: You correctly changed the response status code to _404 Not Found_ in this case. This test is now wrong and passes only because of the mistake above.

```suggestion
    assert_response_status(response_status, expected_status=404)
```

tests/test_system_profile.py
```diff
@@ -106,3 +112,57 @@ def test_get_system_profile_sap_sids_with_RBAC_bypassed_as_system(
...
+
+    assert response_data["count"] == expected_count
+    assert response_data["total"] == expected_total
+
+
+@pytest.mark.parametrize("invalid_host_id", ["notauuid", "922680d3-4aa2-4f0e-9f39-38ab8ea318bb,notuuid"])
+def test_get_system_profile_with_invalid_host_id(api_get, invalid_host_id):
+    response_status, response_data = api_get(f"{HOST_URL}/{invalid_host_id}/system_profile")
+
+    assert_error_response(response_data, expected_title="Bad Request", expected_status=400)
```
Glutexo, line 168: This is completely unrelated to the ultralight check-ins. Please remove and create a separate PR if necessary.

Glutexo (changes requested):
I can’t approve this. There is a mistake in the test and also there is unrelated code. Apart from the _/hosts_ POST removal there is a bunch of unrelated System Profile and SAP tests.

api/host.py
```diff
@@ -451,22 +449,21 @@ def _build_paginated_host_tags_response(total, page, per_page, tags_list):
...
                 "type": "about:blank",
             },
             status=400,
```
Glutexo, line 469: You claim status _404 Not Found_ in the JSON object, but the status of the response is still _400 Bad Request_. Both should be _404 Not Found_.

Glutexo (changes requested):
- Found one more bug with the status code if host is not matcher.
- Confirmed working by manual testing. Only the aforementioned status code is wrong.

tests/test_api_hosts_update.py
```diff
@@ -48,46 +46,43 @@ def test_update_fields(patch_doc, event_producer_mock, db_create_host, db_get_ho
         assert getattr(record, key) == patch_doc[key]
 
 
-@pytest.mark.parametrize("checkin_frequency", [1, 60, 1440])
-@pytest.mark.system_culling
-def test_checkin(checkin_frequency, event_datetime_mock, event_producer_mock, db_create_host, db_get_host, api_put):
-    host = db_host()
-    created_host = db_create_host(host)
+def test_checkin(event_datetime_mock, event_producer_mock, db_create_host, db_get_host, api_post):
```
Glutexo, line 49: The test is failing on system with a non-UTC time zone. Please see my comment below.

Glutexo (changes requested):
Found a failing test.

tests/test_api_auth.py
```diff
@@ -4,7 +4,6 @@
 from app.auth.identity import Identity
 from tests.helpers.api_utils import build_token_auth_header
 from tests.helpers.api_utils import HOST_URL
-from tests.helpers.test_utils import set_environment
```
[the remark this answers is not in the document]
thearifismail: > If `with set_environment({}):` is not removed, this needs to remain too.

this removal is valid because `tests/test_api_path` does not have any "with set_environment".

tests/test_api_hosts_update.py
thearifismail: > The test is failing on system with a non-UTC time zone. Please see my comment below.

Just ran this test successfully.

app/serialization.py
kruai: That's what I did, right? Or am I misreading this comment?

app/serialization.py
```diff
@@ -71,6 +72,14 @@ def deserialize_host_mq(raw_data):
     return deserialize_host(raw_data, MqHostSchema)
 
 
+def deserialize_canonical_facts(raw_data):
+    try:
+        validated_data = CanonicalFactsSchema(strict=True).load(raw_data).data
+    except ValidationError as e:
+        raise ValidationException(str(e.messages)) from None
```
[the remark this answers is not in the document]
kruai: Sorry, what does DRYing mean in this context (and is it worth holding up the PR for)?

kruai:
@Glutexo I separated out the host creation endpoint, which got merged. I've also addressed your comments - how does it look now?

swagger/api.spec.yaml
```diff
@@ -76,19 +76,19 @@ paths:
...
               $ref: '#/components/schemas/CreateCheckIn'
       responses:
         '201':
-          description: Successfully updated host's timestamps.
+          description: Successfully checked in Host.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/BulkHostOut'
+                $ref: '#/components/schemas/CreateHostOut'
```
ryandillinfelton, line 91: It looks like `BulkHostOut` isn't used by anything now (and I guess wasn't before)

swagger/api.spec.yaml
```diff
@@ -705,14 +705,70 @@ components:
       description: >-
         Data required to create a check-in record for a host.
       type: object
-      required:
-        - canonical_facts
       properties:
-        canonical_facts:
-          description: A set of canonical_facts belonging to the host.
-          $ref: '#/components/schemas/Facts'
+        insights_id:
```
ryandillinfelton, line 709: This is as much a comment to you as it is to future me to verify. Should there be more validation here on string lengths and what not, and shouldn't there be some required fields here?

app/serialization.py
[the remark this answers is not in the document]
Glutexo: I mean, create a function that would be reused in the two places where schemas are used to validate and unmarshall the incoming data. But it’s not worth holing up the PR.

app/serialization.py
Glutexo: Yes, it is. Perfect!

tests/test_api_hosts_update.py
Glutexo: @thearifismail That is strange. Datetimes retrieved from the database by SQLAlchemy are converted to timezone aware Python objects in the local timezone. Is it possible that you run it in a container or something that make Python assume UTC as local timezone?

tests/fixtures/api_fixtures.py
```diff
@@ -20,6 +20,16 @@ def _api_create_or_update_host(
     return _api_create_or_update_host
 
 
+@pytest.fixture(scope="function")
+def api_post(flask_client):
+    def _api_post(
+        url, host_data, query_parameters=None, extra_headers=None, auth_type="account_number", identity_type="User"
+    ):
+        return do_request(flask_client.post, url, host_data, query_parameters, extra_headers, auth_type, identity_type)
```
Glutexo, line 28: This is the only request fixture that passes in the _auth_type_ parameter. It passes in a default value though, thus it is redundant. Removing the parameter would make this function nicely aligned with its sister ones: _api_put_ and _api_put_. They are also very repetitive, they may be worth DRYING.

```suggestion
    def _api_post(url, host_data, query_parameters=None, extra_headers=None, identity_type="User"):
        return do_request(flask_client.post, url, host_data, query_parameters, extra_headers, identity_type)
```

Glutexo (approved):
LGTM

I offered a few small refactoring tips, but they don’t have to hold up this pull request. Feel free to implement them though if you feel like it.

Thanks for the changes. Everything now works correctly. Great work! 🙇🏻‍♂️🙏🏻

swagger/api.spec.yaml
kruai: Good catch! I've removed it.

swagger/api.spec.yaml
kruai: If this is a concern, I think we might want to open a Jira card for it! In this PR, I just took the schema that existed for host creation and repurposed it for this new endpoint. I can say that it does do the same amount of validation in the API using Marshmallow. What do you think?

ryandillinfelton: Since we're not using any of the parameters to update the database I think it's fine to not validate them. Worst case if they're incorrect it just won't be able to find the host. As for required fields, again, if they send nothing they'll find nothing. I don't see why it should be a problem to not specify any required fields.

ryandillinfelton (approved):
LGTM

Glutexo:
@ryandillinfelton 

> Since we're not using any of the parameters to update the database I think it's fine to not validate them. Worst case if they're incorrect it just won't be able to find the host. As for required fields, again, if they send nothing they'll find nothing. I don't see why it should be a problem to not specify any required fields.

Technically you are right, but validating the input parameters does make some sense. One of the most important ones is that it support easy non-defensive unmarshalling. Now we use Marshallow for that, but we plan to use purely JSONSchema for that. When that is implemented, marking a field as UUID format does not only validate it, but provide us with a UUID Python object.

That being said, I would not remove any validation rules if there is not a very good reason for that.

Glutexo (commented):
I missed a response code 400 in the OpenAPI specification. Please fix the code and its description. Thanks!

https://github.com/RedHatInsights/insights-host-inventory/blob/0456f4c59196e37aacff36c450f28695bcbceb88/swagger/api.spec.yaml#L97

PR #736, opened 2020-10-29
RHCLOUD-10247: Allow for case insensitivity on Inventory name search

aleccohan:
Modified xjoin filter to allow for case insensitive Inventory name search.

JIRA:  https://issues.redhat.com/browse/RHCLOUD-10247

### DIFF
--- a/app/xjoin.py
+++ b/app/xjoin.py
@@ -12,7 +12,7 @@ from app import REQUEST_ID_HEADER
 from app import UNKNOWN_REQUEST_ID_VALUE
 from app.culling import staleness_to_conditions
 
-__all__ = ("graphql_query", "pagination_params", "staleness_filter", "string_contains", "url")
+__all__ = ("graphql_query", "pagination_params", "staleness_filter", "string_contains", "string_contains_lc", "url")
 
 logger = getLogger("graphql")
 outbound_http_metric = outbound_http_response_time.labels("xjoin")
@@ -54,6 +54,10 @@ def staleness_filter(staleness):
 
 def string_contains(string):
     return {"matches": f"*{string}*"}
+
+
+def string_contains_lc(string):
+    return {"matches_lc": f"*{string}*"}
 
 
 def url():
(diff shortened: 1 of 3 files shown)


tests/test_xjoin.py
```diff
@@ -145,8 +145,8 @@ def test_query_variables_hostname_or_id_uuid(mocker, query_source_xjoin, graphql
             "filter": (
                 {
                     "OR": (
-                        {"display_name": {"matches": f"*{hostname_or_id}*"}},
-                        {"fqdn": {"matches": f"*{hostname_or_id}*"}},
+                        {"display_name": {"matches_lc": f"*{hostname_or_id}*"}},
+                        {"fqdn": {"matches_lc": f"*{hostname_or_id}*"}},
```
jharting, line 149: sadly, the fqdn operation does not support lower-case operands https://github.com/RedHatInsights/xjoin-search/blob/master/docs/schema.md#hostfilter so only `matches` can be used

app/xjoin.py
```diff
@@ -53,7 +53,7 @@ def staleness_filter(staleness):
 
 
 def string_contains(string):
-    return {"matches": f"*{string}*"}
+    return {"matches_lc": f"*{string}*"}
```
jharting, line 56: The `matches_lc` operator only compares whatever argument it receives with a lower-case value of stored value. (i.e. `lower(stored) == "*{string}*"`) For the comparison to work always the argument should be converted to lowercase also so that we get `lower(stored) == "*{lower(string)}*"`

jharting (changes requested):

app/xjoin.py
shdunning: @jharting crap. fqdn and display_name both use `string_contains`. i wonder if we should either:
a. make a `string_contains_lc` variant of this func
b. just add `to_lower = false` param and have `string_contains` return either the `matches` or `matches_lc` based on the bool val?
c. something else i hadn't thought of

jharting: @aleccohan Actually disregard my comment - the `*_lc` operators can be used directly i.e. no need to convert the argument to lowercase.

@shdunning yes we'll need a way to distinguish which variant to use. Both a) and b) look good to me

jharting (approved):

ryandillinfelton (approved):
LGTM :rocket:

kruai (approved):
LGTM!

PR #740, opened 2020-11-04
Remove host-creation endpoint, tests, etc.

kruai:
Addresses this Jira: https://issues.redhat.com/browse/RHCLOUD-10520
Removes the host-creation API endpoint, plus tests, and code that is no longer used as a result.


thearifismail (approved):
LGTM

tests/test_system_profile.py
```diff
@@ -106,3 +112,57 @@ def test_get_system_profile_sap_sids_with_RBAC_bypassed_as_system(
     response_status, response_data = api_get(url, identity_type="System")
 
     assert_response_status(response_status, 200)
+
+
+def test_get_system_profile_RBAC_allowed(mocker, subtests, api_get, db_create_host, enable_rbac):
```
ryandillinfelton, line 117: Why is adding these test part of this PR?

ryandillinfelton (changes requested):
I don't think the test_system_profile.py test additions should be part of this PR

tests/test_system_profile.py
kruai: They're not being added - they're being moved from `test_api_hosts_create.py` (they should have been in `test_system_profile.py` to begin with imo)

ryandillinfelton: Oh I see, they were in the wring file before. Switching to approve :+1:

ryandillinfelton (approved):

tahmidefaz (approved):

Glutexo:
I am glad this is no longer part of #735.

PR #741, opened 2020-11-05
RHCLOUD-8122 - Add support for compensating events.

thearifismail:
At the moment, I can think of only a couple of tests from unit tests.  From system/end-to-end test perspective, will work with QE regarding what and how to test.

https://issues.redhat.com/browse/RHCLOUD-8122


ryandillinfelton:
You've got some pre-commit failures. Try installing the pre-commit hook if you haven't yet https://github.com/RedHatInsights/insights-host-inventory#contributing

host_synchronizer.py
```diff
@@ -0,0 +1,116 @@
...
+from lib.handlers import register_shutdown
+from lib.handlers import ShutdownHandler
+from lib.host_delete import delete_hosts
+from lib.host_synchronize import synchronize_hosts
+from lib.host_repository import stale_timestamp_filter
+from lib.metrics import delete_host_count
+from lib.metrics import delete_host_processing_time
+from lib.metrics import synchronize_fail_count
+
+import json
```
kruai: Why? I don't see it used anywhere

kruai:
Just a note while I'm reviewing... Thanks for putting the Jira # in the title, but I think it would be more useful to have a link to it inside the description :smile:

thearifismail:
> Just a note while I'm reviewing... Thanks for putting the Jira # in the title, but I think it would be more useful to have a link to it inside the description 😄

🤣 🤣 🤣  thanks for reminding my a college professor of mine, he told me "Arif, no matter how perfect your are, I must point out something just to  make sure that I read it"

lib/host_synchronize.py
```diff
@@ -0,0 +1,57 @@
...
+def _should_synchronize(host):
+    # TODO: Check if this is still applicable when hosts are simply read from 
+    # 
+    # DB and NOT deleted by this session.
+    # This process of checking for an already deleted host relies
+    # on checking the session after it has been updated by the commit()
+    # function and marked the deleted hosts as expired.  It is after this
+    # change that the host is called by a new query and, if deleted by a
+    # different process, triggers the ObjectDeletedError and is not emited.
+    return not instance_state(host).expired
```
kruai, line 52: The TODO needs to be addressed and removed. This method might not need to exist, depending on how the TODO is resolved

lib/host_synchronize.py
```diff
@@ -0,0 +1,57 @@
...
+
+from app.queue.queue import EGRESS_HOST_FIELDS
+from app.culling import _Config as CullingConfig
+from datetime import timedelta
+
+__all__ = ("synchronize_hosts",)
+
+def synchronize_hosts(select_query, event_producer, chunk_size, interrupt=lambda: False):
+    start = 0
+    print("Total number: {}".format(select_query.count()))
```
kruai, line 20: Don't commit print statements

lib/host_synchronize.py
```diff
@@ -0,0 +1,57 @@
...
+    while select_query.offset(start).limit(chunk_size).count():
+        host_list = select_query.offset(start).limit(chunk_size)
+        for host in host_list:
+            host_id     = host.id
```
kruai: This var is unnecessary... you can just use host.id

lib/metrics.py
```diff
@@ -21,3 +21,7 @@
     "inventory_delete_host_commit_seconds", "Time spent deleting hosts from the database"
 )
 host_reaper_fail_count = Counter("inventory_reaper_fail_count", "The total amount of Host Reaper failures.")
+
+# synchronization counter
+synchronize_host_count = Counter("inventory_synchronize_host_count", "The total amount of hosts synchronized")
+synchronize_fail_count = Counter("inventory_synchronize_fail_count", "The total amount of synchronization failures.")
```
kruai, line 27:
```suggestion
synchronize_fail_count = Counter("inventory_synchronize_fail_count", "The total amount of synchronization failures")
```

lib/host_synchronize.py
```diff
@@ -0,0 +1,57 @@
...
+    # 
+    # DB and NOT deleted by this session.
+    # This process of checking for an already deleted host relies
+    # on checking the session after it has been updated by the commit()
+    # function and marked the deleted hosts as expired.  It is after this
+    # change that the host is called by a new query and, if deleted by a
+    # different process, triggers the ObjectDeletedError and is not emited.
+    return not instance_state(host).expired
+
+def _staleness_timestamps():
```
kruai, line 55: This function has already been defined in host_query.py. I'd recommend importing it from there.

lib/host_synchronize.py
```diff
@@ -0,0 +1,57 @@
...
+    cullingConfig = CullingConfig(stale_warning_offset_delta=timedelta(days=7), culled_offset_delta=timedelta(days=14))
```
kruai, line 56: ... just as a side note, if you did need to keep this definition, I'd ask that we not use those hardcoded values; instead, we should stick to the values provided in `inventory_config()`

kruai (changes requested):
A lot of this appears to be copied+pasted directly from host_reaper.py, host_delete.py, etc. Before I continue reviewing, can you please refactor a bit so there isn't a bunch of duplicated code?

host_synchronizer.py
kruai: Looks like this got resolved with your linting fixes.

lib/host_synchronize.py
thearifismail: Already gone!!!

lib/host_synchronize.py
thearifismail: removed .

host_synchronizer.py
```diff
@@ -0,0 +1,116 @@
...
+
+def _prometheus_job(namespace):
+    return f"{PROMETHEUS_JOB}-{namespace}" if namespace else PROMETHEUS_JOB
+
+
+def _excepthook(logger, type, value, traceback):
+    logger.exception("Host synchronizer failed", exc_info=value)
+
+
+# TODO question, do we need a counter
```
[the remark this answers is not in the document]
thearifismail: removed already!

host_synchronizer.py
```diff
@@ -0,0 +1,116 @@
...
+def run(config, logger, session, event_producer, shutdown_handler):
+
+    query = session.query(Host)
+
+    update_count = 1
+    events = synchronize_hosts(query, event_producer, 1000, shutdown_handler.shut_down)
+    for host_id, synchronize in events:
+        if synchronize:
+            logger.info("Synchronized host: %s", host_id)
+            logger.info("Number of hosts synchronized: {}".format(update_count))
```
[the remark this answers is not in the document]
thearifismail: Done.!

lib/host_synchronize.py
thearifismail: same problem as the comment above.   Replaced hard-coded values in the method call with pre-defined constants.

lib/host_synchronize.py
thearifismail: Removed a while back.  Should not see any more

host_synchronizer.py
```diff
@@ -0,0 +1,112 @@
...
+from lib.metrics import delete_host_count
+from lib.metrics import delete_host_processing_time
+from lib.metrics import synchronize_fail_count
+
+__all__ = ("main", "run")
+
+PROMETHEUS_JOB = "inventory-synchronizer"
+LOGGER_NAME = "inventory_synchronizer"
+COLLECTED_METRICS = (
+    delete_host_count,
```
jharting, line 34: These delete metrics are not used. synchronize_host_count is missing.

lib/host_synchronize.py
```diff
@@ -0,0 +1,62 @@
...
+__all__ = ("synchronize_hosts",)
+
+WARN_DAYS_AFTER = 7
+DELETE_DAYS_AFTER = 14
+
+
+def synchronize_hosts(select_query, event_producer, chunk_size, interrupt=lambda: False):
+    start = 0
+    print("Total number: {}".format(select_query.count()))
+    while select_query.offset(start).limit(chunk_size).count():
```
jharting, line 24: Looks wasteful to issue a count request first to see if there is more data if the data is going to be read anyway. Moreover, the set of hosts may change in between the count query and the select query. Wouldn't it be better to just fetch the data and if the number of rows is 0 then exit?

lib/host_synchronize.py
```diff
@@ -0,0 +1,62 @@
...
+            yield host.id, sync_up
+        start += chunk_size
+
+        # forced stop if needed.
+        if interrupt():
+            return
+
+
+def _should_synchronize(host):
+    # TODO: Check if this is still applicable when hosts are simply read from
```
jharting, line 46: As mentioned in the comment I don't think this applies here.

jharting (changes requested):

host_synchronizer.py
thearifismail: Fixed!!!

lib/host_synchronize.py
thearifismail: Done!

lib/host_synchronize.py
thearifismail: Gone!!!

lib/host_synchronize.py
thearifismail: I tried using the one you referenced but got `out of scope` error.  This is the only way got it working.  This time using CONST instead of hard coding values in the function call.

lib/host_synchronize.py
kruai: I still see it here (which means it wasn't removed). I also see it on your branch when I pull it.

tests/test_synchronizer.py
```diff
@@ -0,0 +1,77 @@
...
+    with pytest.raises(KafkaError):
+        host_synchronizer_run(
+            inventory_config,
+            mock.Mock(),
+            db.session,
+            event_producer,
+            shutdown_handler=mock.Mock(**{"shut_down.return_value": False}),
+        )
+
+    assert event_producer._kafka_producer.send.call_count == 2
```
kruai, line 77: I'm assuming this should be 2 because `host_count` is 3. To make this easier to understand for future devs, I'd recommend referencing host_count instead of hardcoding `2`:
```suggestion
    assert event_producer._kafka_producer.send.call_count == (host_count - 1)
```

kruai (changes requested):
This is looking a lot better! I'd feel more comfortable getting another review on this as well, but it's getting there :)

thearifismail:
@kruai, it was `RuntimeError: Working outside of application context` that I had gotten.  The console output is included at the end of this comment.  Curious to see how you got it working.

The reason I followed the approach, I took, was that `host_syncronizer` was expected to get triggered manually or run as an ad-hoc openshift job.  Irregardless of how a TimeStamp object is created, the `serialize` method updates it with:
```python
def serialize_host(host, staleness_timestamps, fields=DEFAULT_FIELDS):
    if host.stale_timestamp:
        stale_timestamp = staleness_timestamps.stale_timestamp(host.stale_timestamp)
        stale_warning_timestamp = staleness_timestamps.stale_warning_timestamp(host.stale_timestamp)
        culled_timestamp = staleness_timestamps.culled_timestamp(host.stale_timestamp)
    else:
        stale_timestamp = None
        stale_warning_timestamp = None
        culled_timestamp = None
```

Console log when using `staleness_timestamps()` from `host_query`, which runs in the context of a flask app.
```log
[2020-11-17 10:29:40,681] [6564] [4426395072] [sqlalchemy.engine.base.Engine] [INFO] ROLLBACK
[2020-11-17 10:29:40,684] [6564] [4426395072] [inventory.inventory_synchronizer] [ERROR] Host synchronizer failed
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/aarif/.vscode/extensions/ms-python.python-2020.11.358366026/pythonFiles/lib/python/debugpy/__main__.py", line 45, in <module>
    cli.main()
  File "/Users/aarif/.vscode/extensions/ms-python.python-2020.11.358366026/pythonFiles/lib/python/debugpy/../debugpy/server/cli.py", line 430, in main
    run()
  File "/Users/aarif/.vscode/extensions/ms-python.python-2020.11.358366026/pythonFiles/lib/python/debugpy/../debugpy/server/cli.py", line 267, in run_file
    runpy.run_path(options.target, run_name=compat.force_str("__main__"))
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/runpy.py", line 263, in run_path
    pkg_name=pkg_name, script_name=fname)
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "host_synchronizer.py", line 109, in <module>
    main(logger)
  File "host_synchronizer.py", line 99, in main
    run(config, logger, session, event_producer, shutdown_handler)
  File "<decorator-gen-7>", line 2, in run
  File "/Users/aarif/.local/share/virtualenvs/insights-host-inventory-408tqVB7/lib/python3.6/site-packages/prometheus_client/context_managers.py", line 23, in wrapped
    return func(*args, **kwargs)
  File "host_synchronizer.py", line 67, in run
    for host_id, synchronize in events:
  File "/Users/aarif/Documents/dev-ws/insights/insights-host-inventory/lib/host_synchronize.py", line 32, in synchronize_hosts
  File "/Users/aarif/Documents/dev-ws/insights/insights-host-inventory/api/host_query.py", line 29, in staleness_timestamps
    return Timestamps.from_config(inventory_config())
  File "/Users/aarif/Documents/dev-ws/insights/insights-host-inventory/app/__init__.py", line 71, in inventory_config
    return current_app.config["INVENTORY_CONFIG"]
  File "/Users/aarif/.local/share/virtualenvs/insights-host-inventory-408tqVB7/lib/python3.6/site-packages/werkzeug/local.py", line 347, in __getattr__
    return getattr(self._get_current_object(), name)
  File "/Users/aarif/.local/share/virtualenvs/insights-host-inventory-408tqVB7/lib/python3.6/site-packages/werkzeug/local.py", line 306, in _get_current_object
    return self.__local()
  File "/Users/aarif/.local/share/virtualenvs/insights-host-inventory-408tqVB7/lib/python3.6/site-packages/flask/globals.py", line 52, in _find_app
    raise RuntimeError(_app_ctx_err_msg)
RuntimeError: Working outside of application context.

This typically means that you attempted to use functionality that needed
to interface with the current application object in some way. To solve
this, set up an application context with app.app_context().  See the
documentation for more information.
[2020-11-17 10:29:41,200] [6564] [4426395072] [inventory.lib.handlers] [INFO] Closing producer
[2020-11-17 10:29:41,204] [6564] [4426395072] [inventory.lib.handlers] [INFO] Closing database
[2020-11-17 10:29:41,205] [6564] [4426395072] [inventory.lib.handlers] [INFO] Pushing metrics
Error in atexit._run_exitfuncs:
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/urllib/request.py", line 1349, in do_open
    encode_chunked=req.has_header('Transfer-encoding'))
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/http/client.py", line 1272, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/http/client.py", line 1318, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/http/client.py", line 1267, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/http/client.py", line 1038, in _send_output
    self.send(msg)
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/http/client.py", line 976, in send
    self.connect()
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/http/client.py", line 948, in connect
    (self.host,self.port), self.timeout, self.source_address)
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/socket.py", line 724, in create_connection
    raise err
  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/socket.py", line 713, in create_connection
    sock.connect(sa)
ConnectionRefusedError: [Errno 61] Connection refused

During handling of the above exception, another exception occurred:

  File "/Users/aarif/.pyenv/versions/3.6.11/lib/python3.6/urllib/request.py", line 1351, in do_open
    raise URLError(err)
urllib.error.URLError: <urlopen error [Errno 61] Connection refused>
(insights-host-inventory) bash-3.2$ ```

**Comment by kruai** [comment]
@thearifismail it turns out that how I got it working was "run the wrong file" :) 
However, I still don't think these values should be hardcoded. If the inventory configuration starts using different values for its culling config, this synchronizer will overwrite the staleness, stale warning, and culled timestamps. It should pull them from this config here: https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/config.py#L88-L95

That said, I don't know how to do that at the moment. I didn't realize app context would be an issue here. There's gotta be a way to use that same config file, though.

**Comment by kruai** [comment]
@thearifismail I got it working. In host_synchronizer.py, update your `synchronize_hosts` call to this:
```
events = synchronize_hosts(query, event_producer, 1000, config, shutdown_handler.shut_down)
```
Then, in host_synchronize.py, update `synchronize_hosts` to accept that new config argument, and update the host serialization to:
```
serialized_host = serialize_host(host, Timestamps.from_config(config), EGRESS_HOST_FIELDS)
```

**Comment by thearifismail** [comment]
@kruai implemented your recommendation.

**Review by kruai — DISMISSED** [review]
It looks good to me now! We should definitely get another review before merging, though.

File: lib/host_synchronize.py [hunk 14]
```diff
@@ -0,0 +1,33 @@
...
+from app.serialization import serialize_host
+from lib.metrics import synchronize_host_count
+
+__all__ = ("synchronize_hosts",)
+
+
+def synchronize_hosts(select_query, event_producer, chunk_size, config, interrupt=lambda: False):
+    start = 0
+    while True:
+        host_list = select_query.offset(start).limit(chunk_size)
```
jharting, line 16: Does this return the entries in some deterministic order? If not we should perhaps add an explicit ORDER BY clause. If it does implicitly it might be good to mention that in a comment

jharting: Also, I am wondering if offset-based pagination is stable enough for this possibly long-running script. For example, if a host gets delete while this is running it could cause some other hosts to be missed by the script (as the pages all shift).

lib/host_synchronize.py
```diff
@@ -0,0 +1,33 @@
...
+        if host_list.count() <= 0:
+            break
+        for host in host_list:
+            serialized_host = serialize_host(host, Timestamps.from_config(config), EGRESS_HOST_FIELDS)
+            event = build_event(EventType.updated, serialized_host)
+            insights_id = host.canonical_facts.get("insights_id")
+            headers = message_headers(EventType.updated, insights_id)
+            # incase of a failed update event, event_producer logs the message.
+            event_producer.write_event(event, str(serialized_host), headers, Topic.events, wait=True)
```
jharting, line 25: I don't think it's necessary to wait=True here - i.e. if we fail to produce a message we can always re-run this script.

lib/host_synchronize.py
```diff
@@ -0,0 +1,33 @@
...
+            synchronize_host_count.inc()
+
+            yield host.id, True
```
jharting, line 28: If we never yield False then this second return value is probably not needed?

thearifismail: Gone!

lib/host_synchronize.py
thearifismail: Gone!

lib/host_synchronize.py
thearifismail: @jharting, I could not find a doc explicitly describing any deterministic order used by `SQLAlchemy.Query.offset and limit`.  For deterministic implementation, we can use:
```python
def synchronize_hosts(select_query, event_producer, chunk_size, config, interrupt=lambda: False):
    start = 0
    ordered_list = select_query.order_by(Host.created_on)
    while True:
        host_list = ordered_list.offset(start).limit(chunk_size)
        if host_list.count() <= 0:
        ...
        ...
        ...
```

This, I fear, will increase the chances of running out of memory because `Query.order_by() is understood to bring all records from the host table in one go.   This fear is supported by 
https://groups.google.com/g/sqlalchemy/c/a8BfpLC_M-w?pli=1, which also states 
`...never SELECT more rows than you'd like to have in memory at once. This means if you are paging through a million rows, use limiting / windowing / criteria to ensure you only get a few hundred at a time. The naive approach to this is to use LIMIT/OFFSET, however OFFSET has the issue that the database has to scan through all the previous rows, limiting
performance. `

In case of one time, manual synch/restore job, performance should not be a concern.  The statement about scanning previous rows for each scan gives me confidence that offset and limits work properly. Furthermore, I have been testing the code with 2,115 records and today with 3,003 records and never saw any wrong count.

With this in mind, if you still have doubts about not using `ORDER_BY`, I can implement it.

thearifismail:
test_synchronizer test may fail.  Will investigate tomorrow.

lib/host_synchronize.py
jharting: > This, I fear, will increase the chances of running out of memory because Query.order_by() is understood to bring all records from the host table in one go.
Not if you combine it with limit()

> however OFFSET has the issue that the database has to scan through all the previous rows, limiting
performance
yes, I think that's a major concern given that we need to paginate over the entire database

> The statement about scanning previous rows for each scan gives me confidence that offset and limits work properly. Furthermore, I have been testing the code with 2,115 records and today with 3,003 records and never saw any wrong count.

Well, as long as the entire script runs in an read-isolated transaction then you're correct. Otherwise it is not guaranteed to be consistent. Unless you tried to do a concurrent delete in your testing the problem would not show up

I went ahead and switched to keyset-based pagination in https://github.com/thearifismail/insights-host-inventory/pull/1 This approach will be consistent (even without transaction isolation) and does not suffer from performance problems on high offsets

thearifismail: Thanks. The key-based pagination is merged into the rhc-8122 branch.

tahmidefaz (approved):
LGTM! We need to work on the e2e/app-sre deploy configs. :+1:

thearifismail (commented):
@jharting Can you please review again and approve it so that can merge the code?

jharting (approved):

kruai (approved):

ryandillinfelton (approved):

PR #742, opened 2020-11-06
Fix HTTP status for /checkin endpoint in API spec

kruai:
Addresses @Glutexo 's final comment in this Jira: https://issues.redhat.com/browse/RHCLOUD-10385
It just updates the API spec to expect a 404 instead of a 400.


kruai:
Oops - I opened this for the wrong branch. Sorry about that

PR #750, opened 2020-11-24
Regenerate lockfile

Glutexo:
Pytest uses _importlib_metadata_ that got moved to _importlib.metadata_ in newer Python versions. importlib-metadata has been added as a dependency, but it is not included in Pipfile.lock. Regenerated the lock file so that it works with Python 3.8 as well as with Python 3.6.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -180,6 +180,14 @@
             ],
             "version": "==2.8"
         },
+        "importlib-metadata": {
+            "hashes": [
+                "sha256:590690d61efdd716ff82c39ca9a9d4209252adfe288a4b5721181050acbd4175",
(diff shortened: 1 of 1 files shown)


Glutexo:
Why are the checks failing? When I click on Details, all steps are green.

Glutexo:
Oh, I see. `ModuleNotFoundError: No module named 'importlib_resources'` in pre-commit. Okay. Let’s fix this, but not now.

thearifismail:
Yes, I left it on the back burner to deal with later thinking you were done for the day

thearifismail (approved):

ryandillinfelton (approved):

kruai (approved):
LGTM!

Glutexo:
Thanks for the fix @thearifismail!

PR #751, opened 2020-11-24
Add all packages not in installed_packages

CodeHeeler:
A way to have all packages passed along and not only the max of multi values provided by installed_packages without breaking those dependent on the current implementation of installed_packages.

Dependent on:
https://github.com/RedHatInsights/insights-puptoo/pull/69


kruai (approved):
This looks fine to me, though I wonder if we could define a regex "pattern" for both this field and `installed_packages`... Even something really loose like `^.+:.+$`, just to make sure it's using the right format. That makes the assumption that all packages follow the format shown in the "example" field, though

CodeHeeler:
> This looks fine to me, though I wonder if we could define a regex "pattern" for both this field and `installed_packages`... Even something really loose like `^.+:.+$`, just to make sure it's using the right format. That makes the assumption that all packages follow the format shown in the "example" field, though

Created an issue to track this works separately as we'll want to be careful with changes effecting installed_packages.
https://issues.redhat.com/browse/HBISPSC-27

Let's leave this PR as is.  Thanks for the suggestion though!

kruai:
I appreciate you creating a ticket to track that. This PR looks good to me besides that.

CodeHeeler:
Updated the name to installed_packages_delta as per today's discussion.

kruai (approved):

thearifismail (approved):
LGTM

ryandillinfelton (approved):

PR #737, opened 2020-11-01
Add config via clowder

kylape:

PR #752, opened 2020-11-30
Add ClowdApp deployment template

kylape:
This is a separate commit so the app-interface MR to get the ephemeral
envs can be set up so PR #737 can pass its PR check.

### DIFF
--- a/deployment.yaml
+++ b/deployment.yaml
@@ -1,0 +1,286 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: insights-host-inventory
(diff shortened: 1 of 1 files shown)


kruai:
I'm guessing it's expected that the PR build fails for this too?

kylape:
yep, it's okay that the app-sre jenkins build is failing.

kruai:
I don't see anything wrong with this PR, and I don't think anyone else is taking a look at it... I'm gonna merge to unblock #737

PR #754, opened 2020-12-01
operating_system fact should be dict not list

CodeHeeler:
The os major/minor/name PR [0] just merged in puptoo and ci is having issues with it.  We suspect it's due to the use of List not Dict in the Marshmallow portion of the schema.  Could replace with fields.Dict or omit entirely.  Any feedback on this is appreciated, we'd like to get a fix out rather quickly.  Thanks!

Update: it appears to be due to the use of a bare reference instead of inline definition

[0] https://github.com/RedHatInsights/insights-puptoo/pull/82

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -372,7 +372,7 @@ class SystemProfileSchema(MarshmallowSchema):
     bios_version = fields.Str(validate=marshmallow_validate.Length(max=100))
     bios_release_date = fields.Str(validate=marshmallow_validate.Length(max=50))
     cpu_flags = fields.List(fields.Str(validate=marshmallow_validate.Length(max=30)))
-    operating_system = fields.Nested(OperatingSystemSchema)
+    operating_system = fields.Nested(OperatingSystemSchema())
     os_release = fields.Str(validate=marshmallow_validate.Length(max=100))
     os_kernel_version = fields.Str(validate=marshmallow_validate.Length(max=100))
     arch = fields.Str(validate=marshmallow_validate.Length(max=50))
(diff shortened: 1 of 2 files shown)


CodeHeeler:
@tahmidefaz @SteveHNH @ryandillinfelton The os major/minor/name work we did recently has a bug, please take a look when you can.

app/models.py
```diff
@@ -372,7 +372,7 @@ class SystemProfileSchema(MarshmallowSchema):
     bios_version = fields.Str(validate=marshmallow_validate.Length(max=100))
     bios_release_date = fields.Str(validate=marshmallow_validate.Length(max=50))
     cpu_flags = fields.List(fields.Str(validate=marshmallow_validate.Length(max=30)))
-    operating_system = fields.List(fields.Nested(OperatingSystemSchema()))
+    operating_system = fields.Nested(OperatingSystemSchema)
```
ryandillinfelton, line 375: I think this should be 
```suggestion
    operating_system = fields.Nested(OperatingSystemSchema())
```

ryandillinfelton (changes requested):

app/models.py
kruai: I'm actually not clear on this... [the docs](https://marshmallow.readthedocs.io/en/3.0/nesting.html) make it look like you don't use the parens, but some examples use parens when supplying a param

thearifismail (approved):
LGTM

app/models.py
gmcculloug: The use of parens is a little inconsistent elsewhere in this file, but for this section it seems to always pass "()", so might be better to include them.

@CodeHeeler @tahmidefaz Are there sufficient existing tests for this change?

kruai: I pushed the parentheses change; we can revert this if we decide against it later

kruai (approved):

ryandillinfelton (approved):
LGTM now!

tahmidefaz (approved):
LGTM!

swagger/system_profile.spec.yaml
```diff
@@ -182,7 +160,25 @@ $defs:
         type: array
       operating_system:
         type: object
```
RonnyPfannschmidt, line 162: this type line needs to go

PR #764, opened 2020-12-09
Fedramp logging changes [RHC-10358]

ryandillinfelton:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-10358).

Right now the PR is mainly concerning adding the Access control or flow controls rules invoked information to all logged messages (where rules may have been invoked) and insuring that we log every successful and unsuccessful action.
I wanted to get opinions specifically related to how I am determining which access rules have been invoked. Right now it is determining which was used in the log message based on the identity type. While this is how we determine which access control rule to use I worry that this could create some disconnect between the rule actually applied and the rule logged.

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [x] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


kruai (dismissed):
I think this looks good! I appreciate the refactoring; this should keep our logging a little more consistent. I'd like this to get another review because of its size, but I believe it addresses the applicable parts of the Jira card.

api/host.py
```diff
@@ -176,7 +182,7 @@ def get_host_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=
         query = query.order_by(*order_by)
     query_results = query.paginate(page, per_page, True)
 
-    logger.debug("Found hosts: %s", query_results.items)
+    log_host_list_get_succeded(logger, query_results.items)
```
kruai, line 185: Based on the details in the Set of Auditable Events, this isn't necessary, is it? If we need to log reads/GETs, then there are more places we'll need to add this

ryandillinfelton: Looking at the document I believe we need to log access requests. 
auditableEvents

kruai: Ah, I mistakenly read it to mean access-related changes. You're right, though. In that case, I think there are some more places where we access data. Wouldn't we need to do this when getting tags, sap_system, etc?

ryandillinfelton: Yeah, you're right, thanks for the catch. I'm adding those in now.

api/host.py
```diff
@@ -21,6 +21,12 @@
 from app.auth import current_identity
 from app.config import BulkQuerySource
 from app.exceptions import InventoryException
+from app.instrumentation import log_host_deleted
+from app.instrumentation import log_host_list_get_failed
+from app.instrumentation import log_host_list_get_succeded
```
thearifismail, line 26: Shouldn't `get_host_list_succeeded` be logged for the `get_host_list` method?

api/host.py
```diff
@@ -21,6 +21,12 @@
...
+from app.instrumentation import log_host_not_deleted
```
thearifismail, line 27: Shouldn't there be a log entry for successful deletion?

api/host.py
```diff
@@ -226,7 +232,7 @@ def patch_by_id(host_id_list, body):
     hosts_to_update = query.all()
 
     if not hosts_to_update:
-        logger.debug("Failed to find hosts during patch operation - hosts: %s", host_id_list)
+        log_patch_host_failed(logger, host_id_list)
```
thearifismail, line 235: `log_patch_host_failed` should be applicable when one or more hosts are available but the patching fails.  The way it is now, a failure message will be logged even when not hosts are found and the patching attempt is not made

app/instrumentation.py
```diff
@@ -38,6 +43,78 @@ def message_not_produced(logger, topic, value, key, headers, error):
     event_producer_failure.labels(event_type=headers["event_type"], topic=topic).inc()
```
thearifismail, line 45: The name `message_produced` is not clear, how about `message_event_produced` and `message_event_not_produced`

app/instrumentation.py
```diff
@@ -38,6 +43,78 @@ def message_not_produced(logger, topic, value, key, headers, error):
...
 
 
+def _control_rule():
+    if hasattr(g, "access_control_rule"):
```
thearifismail, line 47: "access_control_rule" being used for "RBAC" only.

Isn't one needed for automated accessed/updates, like for `satellite` or `system`?

app/instrumentation.py
```diff
@@ -38,6 +43,78 @@ def message_not_produced(logger, topic, value, key, headers, error):
...
+
+def log_host_not_deleted(logger, host_id):
+    logger.info(
+        "Hostidentity %s already deleted. Delete event not emitted.", host_id, extra={"access_rule": _control_rule()}
+    )
+
+
+# get host
+def log_host_list_get_succeded(logger, results_list):
+    logger.debug("Found hosts: %s", results_list, extra={"access_rule": _control_rule()})
```
thearifismail, line 66: what about logger.info() or logger.warning()?

app/instrumentation.py
```diff
@@ -38,6 +43,78 @@ def message_not_produced(logger, topic, value, key, headers, error):
...
+
+
+def log_host_list_get_failed(logger):
+    logger.debug("hosts not found", extra={"access_rule": _control_rule()})
+
```
thearifismail, line 71: same question about logging info or warnings?

app/instrumentation.py
```diff
@@ -38,6 +43,78 @@ def message_not_produced(logger, topic, value, key, headers, error):
...
+                "stale_timestamp": input_host.stale_timestamp.isoformat(),
+                "tags": json.dumps(input_host.tags),
+            },
+            "access_rule": _control_rule(),
+        },
+    )
+
+
+def log_add_update_host_success(logger, add_result, host_data, output_host):
+    metrics.add_host_success.labels(add_result.name, host_data.get("reporter", "null")).inc()  # created vs updated
```
thearifismail, line 92: No metrics needed for failures?

app/instrumentation.py
```diff
@@ -1,3 +1,8 @@
+import json
```
thearifismail, line 1: Some  methods have logger.debug() others have logger.info().  What is reason for using one and not the other?  For failures, shouldn't there be logger.error()

api/host.py
ryandillinfelton: log_host_deleted exists, but I can see how the naming there is inconsistent.

app/instrumentation.py
ryandillinfelton: I don't know if that change is in the scope of this PR.

app/instrumentation.py
ryandillinfelton: Right now we don't have access control rules for system identities. Once the authenticated system changes are in I will need to follow up with some more logging changes to track when the system authentication is being used.

app/queue/queue.py
```diff
@@ -99,36 +102,15 @@ def add_host(host_data):
...
-                        "account": input_host.account,
-                        "display_name": input_host.display_name,
-                        "canonical_facts": input_host.canonical_facts,
-                        "reporter": input_host.reporter,
-                        "stale_timestamp": input_host.stale_timestamp.isoformat(),
-                        "tags": json.dumps(input_host.tags),
-                    }
-                },
-            )
+            log_add_host_attempt(logger, input_host)
```
thearifismail, line 105: `log_delete_host_attempt`?

Is the word `attempt` necessary?

lib/middleware.py
```diff
@@ -69,6 +70,10 @@ def modified_func(*args, **kwargs):
             if current_identity.identity_type != CHECKED_TYPE:
                 return func(*args, **kwargs)
 
+            # track that RBAC is being used to control access
+            g.access_control_rule = "RBAC"
```
thearifismail, line 74: Is RBAC the only access mode that needs `access_control_rule` (ACR)?  ACR also needed when inventory hosts are accessed via automated ways?

app/instrumentation.py
ryandillinfelton: I kept the log level from the existing alert and just moved it to a new function/file. Now that you mention it though I think I should confirm that our debug logs go to Kibana and change the log level to one that does if they don't.

ryandillinfelton: Just verified out debug logs do go to Kibana.

app/instrumentation.py
ryandillinfelton: I only moved existing metrics. Adding more metrics is outside of the scope of this PR.

app/queue/queue.py
ryandillinfelton: Because it's logging the attempt not the outcome

thearifismail (approved):

api/host.py
kruai: OK, it sounds like we should standardize on the format "log_{foo}_{verb}_{succeeded/failed}". Does that work?

api/host.py
```diff
@@ -21,6 +21,12 @@
 from app.auth import current_identity
 from app.config import BulkQuerySource
 from app.exceptions import InventoryException
+from app.instrumentation import log_get_host_list_failed
+from app.instrumentation import log_get_host_list_succeded
+from app.instrumentation import log_host_deleted
+from app.instrumentation import log_host_not_deleted
+from app.instrumentation import log_patch_host_failed
```
thearifismail, line 28: Based on @kruai comment, this should be `log_patch_host_failed` should be `log_host_patch_failed`.  The latter sounds better

kruai: I was actually just suggesting based on what I saw on that specific diff; we don't have to adhere to that too strictly. I think it's just the "deleted"/"not_deleted" ones that should be updated to "delete_succeeded" and "delete_failed"

thearifismail: Whatever pattern is decided and agreed up should be stuck to

app/instrumentation.py
```diff
@@ -101,6 +128,9 @@ def log_add_update_host_success(logger, add_result, host_data, output_host):
     )
 
 
+# def log_add_update_host_failure(logger, add_result, host_data, output_host):
```
thearifismail, line 131: Delete the comment out code, if done with debugging.

app/instrumentation.py
```diff
@@ -101,6 +128,9 @@ def log_add_update_host_success(logger, add_result, host_data, output_host):
...
+#     pass
```
thearifismail, line 132: Delete the comment out code, if done with debugging.

thearifismail (commented):
LGTM

api/host.py
ryandillinfelton: I updated the names to match the established pattern

kruai (approved):
LGTM!

thearifismail (approved):

PR #767, opened 2020-12-10
Add dependencies

kylape:


kruai:
Should that config list the apps that must be running for HBI to work properly? Because if so, I'd say xjoin & strimzi should be on there. HBI can also technically function with or without ingress, puptoo, or engine AFAIK

kylape:
sorry @kruai this was opened just to test out the dependency deployment functionality of bonfire. I agree that we are missing dependencies and also that the added dependencies are not true requirements for HBI.

bsquizz:
[test]

kruai:
@kylape Sorry this has been sitting around for so long - we're now going through open PRs that have fallen by the wayside.
If we still need these dependencies added, can you remove the changes to `pr_check.sh`? I'd be fine with merging it then

kruai:
I'm going to close this because of how out-of-date it is (sorry)! If we still need to make a change, let me know and we can open another PR.

PR #771, opened 2020-12-11
Use Draft 7 json schema

tahmidefaz:
## Overview

This PR is being created to address the following Jiras:

- [RHCLOUD-9855](https://issues.redhat.com/browse/RHCLOUD-9855).

- [RHCLOUD-8639](https://issues.redhat.com/browse/RHCLOUD-8639)

- [RHCLOUD-10815](https://issues.redhat.com/browse/RHCLOUD-10815)

This PR updates the Prance library to use the recent features that we contributed to it. 
This also allows us to use the same draft 7 version on the schema we have in inventory-schemas.

## Secure Coding Checklist

- [x] Input Validation
~~- [ ] Output Encoding~~
~~- [ ] Authentication and Password Management~~
~~- [ ] Session Management~~
~~- [ ] Access Control~~
~~- [ ] Cryptographic Practices~~
~~- [ ] Error Handling and Logging~~
~~- [ ] Data Protection~~
~~- [ ] Communication Security~~
~~- [ ] System Configuration~~
~~- [ ] Database Security~~
~~- [ ] File Management~~
~~- [ ] Memory Management~~
~~- [ ] General Coding Practices~~


tahmidefaz:
Opening this as a WIP for now. I am on the fence about if we need to write a custom validator for `x-propertyNames`. I am thinking, even if we write one, how can we test it? We can't create hosts through REST anymore and the REST interface never uses the system_profile schema (unless I am missing something).

Plus, we will also need to figure out how to update the `NestedObject`.

tahmidefaz:
The MQ has a custom validator that somewhat checks for the empty keys, which is equivalent to `x-propertyNames`.

app/custom_validator.py
```diff
@@ -0,0 +1,29 @@
...
+from jsonschema.exceptions import ValidationError
+from jsonschema.validators import extend
+
+
+def validate_property_names(validator, prop_length, instance, schema):
+    if not validator.is_type(instance, "object"):
+        return
+
+    if len(min(instance.keys())) < prop_length:
+        yield ValidationError(f"{instance!r} key length is less than {prop_length!r}")
```
tahmidefaz, line 12: I just wanted to mention that this validator isn't being used atm, since we don't have a way to update system_profile via REST. The MQ has it's own custom validator to check for empty keys.

tahmidefaz:
This is ready for review now. I would appreciate some :eyes: on this before we go on vacation mode

app/__init__.py
```diff
@@ -83,7 +86,12 @@ def create_app(runtime_environment):
     connexion_app = connexion.App("inventory", specification_dir="./swagger/", options=connexion_options)
 
     # Read the swagger.yml file to configure the endpoints
-    parser = ResolvingParser(SPECIFICATION_FILE, resolve_types=RESOLVE_FILES)
+    parser = ResolvingParser(
+        SPECIFICATION_FILE,
+        resolve_types=RESOLVE_FILES | RESOLVE_INTERNAL,
+        resolve_method=TRANSLATE_EXTERNAL,
+        recursion_limit_handler=lambda x, y, z: {},
```
kruai, line 93: Why define a handler if it's just an empty function? Also, shouldn't we set a recursion_limit if we plan to use a handler?

app/custom_validator.py
```diff
@@ -0,0 +1,29 @@
...
+from jsonschema import Draft7Validator
+from jsonschema.exceptions import ValidationError
+from jsonschema.validators import extend
+
+
+def validate_property_names(validator, prop_length, instance, schema):
+    if not validator.is_type(instance, "object"):
+        return
+
+    if len(min(instance.keys())) < prop_length:
```
kruai, line 11: Can you explain this logic? To me, it looks like it's getting the length of the alphabetically-first key in `instance.keys()`, and throwing an error if it's less than `prop_length`. I'm not sure what `prop_length` represents, or why the key needs to be longer than it.

kruai (changes requested):
I've got a couple questions, which may or may not lead to changes being required

app/__init__.py
tahmidefaz: the `recursion_limit` gets set to `1` by default within `prance`, but I just added it for visibility :)

app/custom_validator.py
tahmidefaz: Good catch! :tada: 
I just updated it! Now, `prop_length` is basically the value for `x-propertyNames.minLength`. The purpose of this function is to check if any of the keys in the object (`instance`) is of length less than `x-propertyNames.minLength`

app/__init__.py
tahmidefaz: Internally within `prance`, `recursion_limit_handler` isn't empty by default. That's you we need a function that returns something empty when the recursion_limit is reached. Doing so tells it to simply skip the recursive reference.

Glutexo:
fixing fix 😂

app/custom_validator.py
Glutexo: Can’t we just translate _x-propertyNames_ to _propertyNames_ and run a full validation of the System Profile object using directly _jsonschema.validate_ with Draft 7?

app/custom_validator.py
```diff
@@ -0,0 +1,30 @@
...
+def validate_property_names(validator, property_names, instance, schema):
+    if not validator.is_type(instance, "object"):
+        return
+
+    prop_length = property_names["minLength"]
+    if any(len(key) < prop_length for key in instance.keys()):
+        yield ValidationError(f"{instance!r} key length is less than {prop_length!r}")
+
+
+Draft7RequestValidator = extend(Draft7Validator, {"x-propertyNames": validate_property_names})
```
Glutexo, line 16: What about changing all x-propertyNames in the schema to propertyNames and let the Draft7RequestValidator to do the work instead of replicating its logic by the `any(len(key) < prop_length for key in instance.keys())` condition?

Glutexo (changes requested):

app/custom_validator.py
tahmidefaz: Doing a full translation of _x-propertyNames_ to _propertyNames_ might not be straightforward as connexion has a built-in open api validator and having _propertyNames_ in an open api specification is invalid. Instead, I just updated this to use the same validation function of _propertyNames_ for _x-propertyNames_

app/custom_validator.py
tahmidefaz: answered above. :+1:

ryandillinfelton (dismissed):
Looks good to me and seems to behave correctly in my testing.

app/__init__.py
thearifismail: I like specifying `recursion_limit: 1` despite being the default value already specified.  For clarity, add a comment "1 is the default value for recursion_limit"

thearifismail: What happens when `recursion_limit_handler` not provided?  The default value for `recursion_limit` is 1.  Shouldn't it stop automatically after recursing once? Can recursion_limit check be disabled?

app/__init__.py
```diff
@@ -83,7 +86,13 @@ def create_app(runtime_environment):
     connexion_app = connexion.App("inventory", specification_dir="./swagger/", options=connexion_options)
 
     # Read the swagger.yml file to configure the endpoints
-    parser = ResolvingParser(SPECIFICATION_FILE, resolve_types=RESOLVE_FILES)
+    parser = ResolvingParser(
+        SPECIFICATION_FILE,
+        resolve_types=RESOLVE_FILES | RESOLVE_INTERNAL,
+        resolve_method=TRANSLATE_EXTERNAL,
+        recursion_limit=1,
```
thearifismail, line 93: I like specifying `recursion_limit: 1` despite being the default value already specified.  For clarity, add a comment "1 is the default value for recursion_limit"

swagger/system_profile.spec.yaml
```diff
@@ -1,7 +1,17 @@
 ---
 $id: system_profile.spec.yaml
-$schema: http://json-schema.org/draft-04/schema#
+$schema: http://json-schema.org/draft-07/schema#
 $defs:
+  NestedObject:
+    description: An arbitrary object that doesn’t allow empty string keys.
+    type: object
+    x-propertyNames:
```
thearifismail, line 8: does it have to be a plural?  Can it be `x-propertyName`?

thearifismail (dismissed):
Minor comments otherwise LTGM

swagger/system_profile.spec.yaml
```diff
@@ -1,7 +1,17 @@
 ---
 $id: system_profile.spec.yaml
-$schema: http://json-schema.org/draft-04/schema#
+$schema: http://json-schema.org/draft-07/schema#
```
Glutexo, line 3: I wonder whether this is really correct. The schema is not Draft 7 and that’s why we can’t use _propertyNames_ properly. It’s still Draft 4, although a superset of one because of our custom _x-properyNames_.

app/__init__.py
tahmidefaz: `recursion_limit` is only there to check the number of recursions arising from recursive schema definitions. It does not, however, tells us what to do when that `recursion_limit` is reached. That's where `recursion_limit_handler` comes into play. It lets us choose what to do when that `recursion_limit` is reached. 
When `recursion_limit_handler` isn't provided, `prance` will use its own default `recursion_limit_handler` which throws a [prance `ResolutionError` every time `recursion_limit` is reached.](https://github.com/jfinkhaeuser/prance/blob/master/prance/util/resolver.py#L27) That's not really something we want here. Whenever that `recursion_limit` is reached, we just want to skip that reference the second time and move on to resolving the other ones.
No, the `recursion_limit` check is internal to prance and cannot be disabled.

swagger/system_profile.spec.yaml
tahmidefaz: The JSON schema field that we are trying to replace here is `propertyNames`, so, I think it will be best to have some consistency here and use `x-propertyNames`. 
Resolving this conversation, feel free to unresolve if you think otherwise. :+1:

swagger/system_profile.spec.yaml
```diff
@@ -275,8 +286,8 @@ $defs:
           pattern: '^[A-Z][A-Z0-9]{2}$'
       sap_instance_number:
         type: string
-        description: The instance number of the SAP HANA system
-        example: '42'
+        description: The instance number of the SAP HANA system (a two-digit number between 00 and 97)
+        example: '03'
```
Glutexo, line 290: This change does not seem to belong here.

swagger/system_profile.spec.yaml
```diff
@@ -293,12 +304,12 @@ $defs:
       selinux_current_mode:
         type: string
         enum: [enforcing, permissive, disabled]
-        description: The current SELinux mode, either enforcing, permissive, or disabled
         maxLength: 10
+        description: The current SELinux mode, either enforcing, permissive, or disabled
```
Glutexo, line 308: This just moves the description around. It does not really belong here.

swagger/system_profile.spec.yaml
```diff
@@ -293,12 +304,12 @@ $defs:
...
         example: 'enforcing'
       selinux_config_file:
         type: string
         enum: [enforcing, permissive, disabled]
-        description: The SELinux mode provided in the config file
         maxLength: 10
+        description: The SELinux mode provided in the config file
```
Glutexo, line 314: Just as above, just moving around, does not belong here.

swagger/system_profile.spec.yaml
```diff
@@ -165,6 +175,7 @@ $defs:
           maxLength: 30
         type: array
       operating_system:
+        description: Object for OS details
```
Glutexo, line 178: Does this comment addition really belong here?

Glutexo (changes requested):

tahmidefaz:
^ @Glutexo this was an almost exact copy of the schema we have in [RedHatInsights/inventory-schemas](https://github.com/RedHatInsights/inventory-schemas). They may have been one or two lines off due to some recent merges in inventory-schemas. I just updated it to be an exact copy of what's in the `inventory-schemas` repo.

Glutexo (changes requested):
My biggest concern here is that there are no tests. The only tests that verify the nested objects are for the MQ API, because they test the Python code that validate the disk options. There are no such tests for the HTTP API and for the new changes. These changes extend external libraries and the code is non-trivial.

Also now with the Draft7 in, we can drop the arbitrary Python code for the MQ, consolidating the validation. This can be done in a follow-up PR, not to block this one.

tahmidefaz:
> My biggest concern here is that there are no tests. The only tests that verify the nested objects are for the MQ API, because they test the Python code that validate the disk options. There are no such tests for the HTTP API and for the new changes. These changes extend external libraries and the code is non-trivial.

Can you elaborate a little bit on what kind of tests might be good to include here? With the merging of #740 there isn't really a way to add/update SystemProfile fields via the HTTP API anymore.

> Also now with the Draft7 in, we can drop the arbitrary Python code for the MQ, consolidating the validation. This can be done in a follow-up PR, not to block this one.

Yes, we can do this in a separate PR :+1:

tahmidefaz:
@Glutexo please check my comments above ^

swagger/system_profile.spec.yaml
tahmidefaz: This is an exact replica of the one in `inventory-schemas`. In my opinion, this is Draft 7 as well as Draft 4 since I think Draft 7 is also compatible with Draft 4. Since we are adding a `Draft7Validator` with this PR, there's no harm in using `draft-07` as we might add more Draft 7 features in the coming months.

swagger/system_profile.spec.yaml
tahmidefaz: Resolving this and the other ones as this is an exact replica of the one in `inventory-schemas`

Glutexo:
@tahmidefaz The System Profile schema is used in GET /hosts/{host_ids}/system_profile. If the schema definition is not correct and does not match whatever has been accepted and stored in the database, response body validation would fail.

That means we need tests that verify that a System Profile that uses these Disk Options and stuff is not rejected by the Connexion response body validator. And possibly on the other hand tests verifying that an invalid one would be rejected. Even though an invalid payload should not even be there in the database, there are ways how this can happen:

- Incompatible schema evolution. This is the most probable scenario.
- Manual data manipulation. This is somehow alleviated by AppSRE.
- Incompatibility of the arbitrary Python code validation and the JSON Schema validation.

And in the future, we should even think about what to do if such a discrepancy really happens. Now, a whole response is scrapped and for GET of multiple system profiles, this can be fatal.

app/custom_validator.py
```diff
@@ -0,0 +1,20 @@
+from connexion.decorators.validation import RequestBodyValidator
+from jsonschema import Draft7Validator
+from jsonschema.validators import extend
+
+
+Draft7RequestValidator = extend(Draft7Validator, {"x-propertyNames": Draft7Validator.VALIDATORS.get("propertyNames")})
```
Glutexo, line 6: I wonder whether we should stick that much to Draft 7. I am afraid that it would lead to more and more discrepancies with OpenAPI in the future. We can always bring a specific features from newer drafts if necessary, just like we did with _propertyNames_.

OpenAPI documentation [says](https://swagger.io/docs/specification/data-models/keywords/):

> OpenAPI 3.0 uses an extended subset of JSON Schema Specification Wright Draft 00 (aka Draft 5) to describe the data formats

JSON Schema documentation [says](https://json-schema.org/specification-links.html#draft-5):

> Draft 5 was primarily a cleanup of Draft 4 and continued to use the Draft 4 meta-schemas.

It sounds like we should stick to Draft 4 and downgrade the externalized schema version. The same applies to the MQ ingestion that uses Draft 7 as a default. This does not necessarily need to be done in this PR. Draft7 at least match with the MQ API.

———

The _Draft7RequestValidator_ name looks a bit fishy to me as it is not a pure Draft 7 validator. It is more our custom validator, although based on Draft 7. Also we don’t use it to validate request Body

app/custom_validator.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class CustomRequestBodyValidator(RequestBodyValidator):
```
Glutexo, line 9: We don’t need to validate request with this, but responses instead. There is no endpoint now that would contain the System Profile in its request. There is only _GET /hosts/{host_ids}/system_profile_ that has it in its response.

swagger/system_profile.spec.yaml
Glutexo: Well, maybe we should downgrade the version there in the externalized schema. We reference it from an OpenAPI specification and that supports only Draft 5 (or 4 if you will). Even though we can force Connexion to actually use Draft 7 for validation, we can’t use its features because we expose the specification and thus it needs to be compliant. That was even the case how we found out that _propertyNames_ are not valid. That simply means that we can’t use more Draft 7 features. We may cherry-pick some, document them and implement them manually, but that should be only the last resort. Even the _x-propertyNames_ hack is pretty dubious.

Glutexo (commented):
Some more comments on Draft 4/5 vs Draft 7.

Glutexo (changes requested):
Confirmed not working. 😞 I retrieved a host with `{"disk_devices": {"options": {"": "invalid"}}}` and the response was not validated by the extended validator. Instead of 500 Internal Server Error I got simply _200 OK_. That’s what I wanted tests for.

I think that the reason is that the custom validator validates request and not responses as mentioned in my earlier comment. I did not find a quick way how to make that working, but https://connexion.readthedocs.io/en/latest/response.html#custom-validator should help.

Glutexo:
One more question, @tahmidefaz. How did you verify that this pull request works? Maybe I missed something. How can I reproduce it?

thearifismail:
Stepan,
This is Arif.  Tahmid is off today and gone until after the New Year.

tahmidefaz:
> Confirmed not working.  I retrieved a host with `{"disk_devices": {"options": {"": "invalid"}}}` and the response was not validated by the extended validator. Instead of 500 Internal Server Error I got simply _200 OK_. That’s what I wanted tests for.
> 
> I think that the reason is that the custom validator validates request and not responses as mentioned in my earlier comment. I did not find a quick way how to make that working, but https://connexion.readthedocs.io/en/latest/response.html#custom-validator should help.

Ah, I see! It doesn't support that yet. I will be updating this PR to fix that. :+1:

Glutexo:
Just adding that we don’t need to validate the System Profile in requests at all. The Host creation via POST has been removed. The only place where the schema is used by Connexion is in the GET Host System Profile response. That means we need to patch the validation only for responses and not for requests.

Glutexo:
@tahmidefaz Please ping me when there is an updated version ready for review. Thanks!
@ryandillinfelton Do you still remember how did you test that? Yours and my testing yielded different results, so I am curious where is the cause. Finding it may help Tahmid fix the PR.

app/custom_validator.py
tahmidefaz: I agree. Switched to using a Draft 4 validator instead. I will work on downgrading the externalized schema version to draft 4. Using a draft 7 validator here was introducing some unnecessary complexities. The system_profile schema is fully compatible with draft 7 and draft 4, therefore using a draft 4 validator to avoid unnecessary work sounds good to me. :+1:

tahmidefaz:
This is ready for another review. I swapped out the initial use of a draft 7 validator in favor of connexion's own draft 4 response validator as I found it to be simpler to deal with than a new draft 7 one. Connexion customizes its draft 4 validators by a decent amount and porting those customizations to draft 7 will surely add to the complexity. I found the draft 7 version of the scheme that's in the inventory-schemas repo to be fully backward compatible with draft 4 and, to me, it does not look like we will be needing inclusive draft 7 features. So, downgrading the system_profile schema and using a draft 4 validator here seems good to me here. :+1: 
@Glutexo

tahmidefaz:
Changes from draft04 -> draft06: [here](https://json-schema.org/draft-06/json-schema-release-notes.html)
Changes from draft06 -> draft07: [here](https://json-schema.org/draft-07/json-schema-release-notes.html)

kruai (dismissed):

kruai:
@tahmidefaz is something blocking this PR? A review from @Glutexo ?

tahmidefaz:
I just updated it to use a Draft 7 validator instead. Since `propertyNames` was introduced in Draft 7, we cannot modify and use connexion's native Draft 4 validator. In my latest commit, I customized the Draft7Validator by copying over the [custom validators from connexion's internal json_schema validator](https://github.com/zalando/connexion/blob/1ed14fbde5be3d144760fec4982e90064f151e88/connexion/json_schema.py#L109) and merging them with the Draft7Validator.

@kruai I just updated it. This needs some fresh :eyes:

swagger/system_profile.spec.yaml
tahmidefaz: I think this needs to be draft 7. Check my comment below.

Glutexo:
@tahmidefaz Is it not possible to extend the document Draft4Validator a with the propertyNames field validator from Draft7?

```
Validator = extend(Draft4Validator, {"x-propertyNames": Draft7Validator.VALIDATORS.get("propertyNames")})

```

Glutexo:
> Since `propertyNames` was introduced in Draft 7, we cannot modify and use connexion's native Draft 4 validator. In my latest commit, I customized the Draft7Validator by copying over the [custom validators from connexion's internal json_schema validator](https://github.com/zalando/connexion/blob/1ed14fbde5be3d144760fec4982e90064f151e88/connexion/json_schema.py#L109) and merging them with the Draft7Validator.

Experimentally proven not true. See my inline comments.

app/custom_validator.py
```diff
@@ -0,0 +1,39 @@
...
+
+    return custom_validators
+
+
+def get_custom_validator():
+    custom_validators = get_draft4_custom_validators()
+    CustomDraft7Validator = extend(
+        Draft7Validator, {**custom_validators, "x-propertyNames": Draft7Validator.VALIDATORS.get("propertyNames")}
+    )
+    return CustomDraft7Validator
```
Glutexo, line 25: All this is unnecessarily complex and not entirely correct. Earlier, we already agreed that the Schema is not Draft 7, but rather Draft 4 (standard for OpenAPI) with our custom extensions.

All we need to add to _Draft4ResponseValidator_ is the _propertyNames_ validator from Draft 7 hooked to _x-propertyNames_. That can be simply achieved by extending the original one as the previous solution did.

```suggestion
CustomDraft4ResponseValidator = extend(
    Draft4ResponseValidator, {"x-propertyNames": Draft7Validator.VALIDATORS.get("propertyNames")}
)
```

app/custom_validator.py
```diff
@@ -0,0 +1,39 @@
...
+
+class CustomResponseValidator(ResponseValidator):
+    """
+    Extends the connexion ResponseValidator with x-propertyNames
+    See https://connexion.readthedocs.io/en/latest/request.html#custom-validators
+    See https://github.com/zalando/connexion/blob/master/examples/swagger2/enforcedefaults/enforcedefaults.py
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, validator=get_custom_validator(), **kwargs)
```
Glutexo, line 36:
```suggestion
        super().__init__(*args, validator=CustomDraft4ResponseValidator, **kwargs)
```

swagger/system_profile.spec.yaml
Glutexo: Sorry, I am still not convinced. See my code suggestions.

Glutexo (changes requested):

swagger/system_profile.spec.yaml
Glutexo:
```suggestion
$schema: http://json-schema.org/draft-04/schema#
```

app/__init__.py
```diff
@@ -83,7 +86,13 @@ def create_app(runtime_environment):
     connexion_app = connexion.App("inventory", specification_dir="./swagger/", options=connexion_options)
 
     # Read the swagger.yml file to configure the endpoints
-    parser = ResolvingParser(SPECIFICATION_FILE, resolve_types=RESOLVE_FILES)
+    parser = ResolvingParser(
+        SPECIFICATION_FILE,
+        resolve_types=RESOLVE_FILES | RESOLVE_INTERNAL,
+        resolve_method=TRANSLATE_EXTERNAL,
+        recursion_limit=1,  # 1 is the default for recursion_limit, adding it for clarity
+        recursion_limit_handler=lambda x, y, z: {},
```
Glutexo, line 94: I finally understand how and why this works. Dumping it here for reference.

Prance’s [dereferencing iterator](https://github.com/jfinkhaeuser/prance/blob/977f13f4b16e4c475e6799c550e8dfbdcf3a7e83/prance/util/resolver.py#L118) traverses the specification starting at [the beginning of _api.spec.yaml_](https://github.com/RedHatInsights/insights-host-inventory/blob/78f45fb61aa4f7d941bf9082737d28dcd2a3b699/swagger/api.spec.yaml#L1).

tahmidefaz:
> @tahmidefaz Is it not possible to extend the document Draft4Validator a with the propertyNames field validator from Draft7?
> 
> ```python
> Validator = extend(Draft4Validator, {"x-propertyNames": Draft7Validator.VALIDATORS.get("propertyNames")})
> ```

@Glutexo this is similar to my previous approach which I abandoned because of consistency (e.g. I wanted to avoid using Draft7Validator features in Draft4Validator). I guess it makes sense to not overcomplicate this PR than it needs to. I will revert my commits to that approach. :+1:

app/__init__.py
Glutexo: @tahmidefaz If you have an idea where to put this so it’s better persisted and more easily to be found, please do so. Thanks!

tahmidefaz:
@Glutexo just done reverting + updating. This should be at what you suggested earlier. Please take a look. :+1:

Glutexo:
> I wanted to avoid using Draft7Validator features in Draft4Validator
I understand. The good thing is that the feture is just a very plain function. It can be even imported directly from _jsonschema._validators.propertyNames_. But doing that from Draft7Validator nicely conveys the message that we cherry-picked a single feature. Like \_\_future\_\_ in Python.

Glutexo (approved):
Thanks for the quick update!

LGTM

kruai (approved):

app/__init__.py
tahmidefaz: I think this is quite visible here. I will copy this over if I think of a better place for this. :+1:

thearifismail (approved):
LGTM

PR #774, opened 2021-01-06
Create system_profile validator endpoint

kruai:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-10580).
The goal is the provide an endpoint that does the following:
- Takes a branch name and a timeframe as path parameters
- Fetches the System Profile spec from the specified branch in the inventory-schemas repo
- Aggregates the Kafka messages that were produced within the supplied timeframe
- Runs validation on the messages's host data for both the specified branch and the master branch
- Responds with a pass/fail summary, grouped by reporter

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [x] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


kruai:
Don't review this quite yet - there's a lot of organization & cleanup I want to work on. I'll remove "WIP" when it's ready!

kruai:
I'm trying to test this, but the Disk Device options are failing:
```
[ERROR] {'system_profile': ["System profile does not conform to schema.
True is not of type 'object'

Failed validating 'type' in schema['properties']['disk_devices']['items']['properties']['options']['additionalProperties']:
    {'additionalProperties': {'$ref': '#/$defs/NestedObject'},
     'description': 'An arbitrary object that doesn’t allow empty string '
                    'keys.',
     'propertyNames': {'minLength': 1},
     'type': 'object'}

On instance['disk_devices'][0]['options']['ro']:
    True"]}
```

I think our spec for NestedObject might be incorrect. What do you think @tahmidefaz ?

api/system_profile.py
```diff
@@ -171,3 +174,11 @@ def get_sap_sids(search=None, tags=None, page=None, per_page=None, staleness=Non
     return flask_json_response(
         build_collection_response(data["sap_sids"]["data"], page, per_page, data["sap_sids"]["meta"]["total"])
     )
+
+
+@api_operation
+@rbac(Permission.READ)
+@metrics.api_request_time.time()
+def validate_schema(repo_branch="master", days=14):
```
tahmidefaz, line 182: @kruai  Got a question: Should it be `days=14`? I thought we were only looking back 1 day :thinking:

kruai: I don't have concrete requirements for that. You can of course configure this value, but we can change the default if necessary. What do you think @gmcculloug ?

tahmidefaz (changes requested):
It might be better to have some tests for this. What do you think?

lib/system_profile_validate.py
```diff
@@ -0,0 +1,85 @@
...
+        self.fail_count = 0
+
+
+def _validate_host_list(host_list, repo_branch):
+    system_profile_spec = safe_load(
+        get(
+            "https://raw.githubusercontent.com/RedHatInsights/inventory-schemas/"
+            f"{repo_branch}/schemas/system_profile/v1.yaml"
+        ).content.decode("utf-8")
+    )
```
tahmidefaz, line 31: I lot of the updates will be coming in from `forks` and not `branches`. This url might make fetching the files from forks difficult (or impossible, even). :thinking:

tahmidefaz (changes requested):

lib/system_profile_validate.py
ryandillinfelton: I think it just needs to be able to take a github username in and swap it with RedHatInsights (i.e.https://raw.githubusercontent.com/tahmidefaz/inventory-schemas/master/schemas/system_profile/v1.yaml)

kruai: Hmm, that does complicate things. Would be great if we could ask people to just make branches instead of forking, but that's not reasonable. I'll add an optional `fork` parameter that defaults to RedHatInsights

ryandillinfelton: If it's not too much trouble I think adding the optional fork parameter is ideal. We could open the repo up more to allow anyone to make a branch, but I could see requiring branches leading to some inconvenience down the line.

thearifismail: Shouldn't then `RedHatInsights` be replaced by a variable say `fork`?

kruai: Yeah that's the plan

swagger/api.spec.yaml
```diff
@@ -406,6 +406,30 @@ paths:
...
+      security:
+        - ApiKeyAuth: []
+      parameters:
+        - in: query
+          name: repo_branch
+          schema:
+            type: string
+          description: The branch of the schema repo to validate against
+          required: true
+        - in: query
```
ryandillinfelton, line 424: Odds are no one will try this, but the service 500s when you send days=0

lib/system_profile_validate.py
thearifismail: I personally feel at ease using `forks`, which was pretty common among Cloudform team member.  How about 
```python
def _validate_host_list(host_list, repo_branch, fork='RedHatInsights'):
```

api/system_profile.py
ryandillinfelton: 14 days would insure that all non-stale hosts work with the new schema, but I worry that it's a lot of data to check. 1-2 days should give us a similar amount of assurance with significantly less data. My vote is to have it default low until we know the performance.

lib/system_profile_validate.py
```diff
@@ -0,0 +1,85 @@
...
+def validate_sp_for_branch(config, repo_branch="master", days=14):
+    consumer = KafkaConsumer(
+        group_id=config.host_ingress_consumer_group,
+        bootstrap_servers=config.bootstrap_servers,
+        api_version=(0, 10, 1),
+        value_deserializer=lambda m: m.decode(),
+        **config.kafka_consumer,
+    )
+
+    tp = TopicPartition(config.host_ingress_topic, 0)
```
thearifismail, line 58: I am assuming `0` is the first partition.  Could there be multiple partitions and the `host_ingress_topic` getting posted to a second or third partition?

ryandillinfelton (changes requested):
Looks good and seems to work in my testing so far. I would just like fork support and a fix to the 500 bug I outlined above.

swagger/api.spec.yaml
thearifismail: @ryandillinfelton  do you mean 400, bad request for 0 days

thearifismail (commented):
Shouldn't there be tests for system_profile_validate for testing bad branch, fork, and primarily good and bad fields in the submitted schema?

swagger/api.spec.yaml
kruai: Yep, I defined a minimum for that field which should fix the issue.

lib/system_profile_validate.py
kruai: There's only one partition for all of our topics. I remember trying to make this code work for `n` number of partitions, but ran into some implementation issues. I'll give it another shot, though.

api/system_profile.py
kruai: I set the default to 1 day!

lib/system_profile_validate.py
kruai: I've parameterized the fork.

lib/system_profile_validate.py
kruai: @thearifismail Since our topics all have 1 partition, and we don't even know if this is going to be a long-term solution, I think just reading from this one partition should work fine. I'll make a note for myself to follow up and verify that it's the same case for all of our environments, including Prod. Is that okay with you?

ryandillinfelton: afaik we only use one partition in prod too. There was some talk about increasing that though, but we can address that down the line if it ever changes.

thearifismail:
Will system_profile validation run as a cronjob?

kruai:
> Will system_profile validation run as a cronjob?

That's the intent for step 2. The goal is to get this API endpoint merged so people can at least validate on their own, and then create a cronjob that scans for PRs and leaves validation summary comments

api/system_profile.py
```diff
@@ -181,3 +184,11 @@ def get_sap_sids(search=None, tags=None, page=None, per_page=None, staleness=Non
     return flask_json_response(
         build_collection_response(data["sap_sids"]["data"], page, per_page, data["sap_sids"]["meta"]["total"])
     )
+
+
+@api_operation
+@rbac(Permission.READ)
+@metrics.api_request_time.time()
```
ryandillinfelton, line 191: Is this the right metric for this? If this endpoint takes a while to produce a response when it has a huge amount of data I don't want it to trip any alarms we have for "normal" API requests that are more latency dependent.

tahmidefaz: Yes, agree. We might want to add a new metric in `api/metrics.py` :+1:

swagger/api.spec.yaml
```diff
@@ -406,6 +406,36 @@ paths:
...
+          name: repo_branch
+          schema:
+            type: string
+          description: The branch of the inventory-schemas repo to use
+          required: true
+        - in: query
+          name: days
+          schema:
+            type: integer
+            minimum: 0
```
ryandillinfelton, line 433: This can be added along with the tests but I wanted to note it. What happens if someone asks for more days of data than kafka can provide?

kruai: It doesn't matter. There's a consumer method I'm using that calculates the partition offset based on a timestamp (which is determined by the `days` param). If it asks for more days of data than are available, it just returns all the messages available.

api/system_profile.py
kruai: Oh, good point. I'll add that new metric.

kruai: OK, added!

api/system_profile.py
```diff
@@ -80,7 +83,7 @@ def xjoin_enabled():
 
 @api_operation
 @rbac(Permission.READ)
-@metrics.api_request_time.time()
+@metrics.schema_validation_time.time()
```
ryandillinfelton, line 86: I think you replaced the metric in the wrong spot

kruai: Ugh, you're right. Sorry about that. Should have taken a deep breath first

tahmidefaz (approved):
LGTM :+1:

ryandillinfelton (approved):
:rocket: LGTM

thearifismail (approved):
LGTM

PR #776, opened 2021-01-08
Fix reaper logging bug

ryandillinfelton:
## Overview

This changes how we log the access rule invoked for delete events to prevent the reaper from crashing when it tries to produce event success/failure log messages.

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -21,6 +21,7 @@ from app import Permission
 from app.auth import current_identity
 from app.config import BulkQuerySource
 from app.exceptions import InventoryException
+from app.instrumentation import get_control_rule
 from app.instrumentation import log_get_host_list_failed
 from app.instrumentation import log_get_host_list_succeeded
 from app.instrumentation import log_host_delete_failed
@@ -154,7 +155,7 @@ def delete_by_id(host_id_list):
             query, current_app.event_producer, inventory_config().host_delete_chunk_size
         ):
             if deleted:
-                log_host_delete_succeeded(logger, host_id)
+                log_host_delete_succeeded(logger, host_id, get_control_rule)
                 tracker_message = "deleted host"
             else:
                 log_host_delete_failed(logger, host_id)
--- a/api/host.py
+++ b/api/host.py
@@ -155,7 +155,7 @@ def delete_by_id(host_id_list):
             query, current_app.event_producer, inventory_config().host_delete_chunk_size
         ):
             if deleted:
-                log_host_delete_succeeded(logger, host_id, get_control_rule)
+                log_host_delete_succeeded(logger, host_id, get_control_rule())
                 tracker_message = "deleted host"
             else:
                 log_host_delete_failed(logger, host_id)
--- a/app/instrumentation.py
+++ b/app/instrumentation.py
@@ -43,7 +43,7 @@ def message_not_produced(logger, topic, value, key, headers, error):
     event_producer_failure.labels(event_type=headers["event_type"], topic=topic).inc()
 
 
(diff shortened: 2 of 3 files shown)


app/instrumentation.py
```diff
@@ -43,58 +43,58 @@ def message_not_produced(logger, topic, value, key, headers, error):
...
-    logger.info("Deleted host: %s", host_id, extra={"access_rule": _control_rule()})
+def log_host_delete_succeeded(logger, host_id, control_rule):
+    logger.info("Deleted host: %s", host_id, extra={"access_rule": control_rule})
 
 
-def log_host_delete_failed(logger, host_id):
+def log_host_delete_failed(logger, host_id, control_rule):
     logger.info(
-        "Hostidentity %s already deleted. Delete event not emitted.", host_id, extra={"access_rule": _control_rule()}
+        "Hostidentity %s already deleted. Delete event not emitted.", host_id, extra={"access_rule": control_rule}
```
tahmidefaz, line 60: Do we have to do `control_rule()` here to execute the `get_control_rule()` function?

tahmidefaz (changes requested):
Asked a question below, please take a look

app/instrumentation.py
ryandillinfelton: I changed the log_host_delete functions to take the control_rule as a parameter so that `reaper` can just pass a string in to avoid the error.

tahmidefaz: Gotcha! Thanks!

tahmidefaz (dismissed):
LGTM! Also confirmed locally that this fixes the reaper :+1: 
Let's get back our slack channel! :muscle:

app/instrumentation.py
```diff
@@ -43,58 +43,58 @@ def message_not_produced(logger, topic, value, key, headers, error):
...
         return g.access_control_rule
     else:
         return "None"
 
 
 # delete host
-def log_host_delete_succeeded(logger, host_id):
-    logger.info("Deleted host: %s", host_id, extra={"access_rule": _control_rule()})
+def log_host_delete_succeeded(logger, host_id, control_rule):
+    logger.info("Deleted host: %s", host_id, extra={"access_rule": control_rule})
```
kruai, line 55: Instead of changing `_control_rule()` to `get_control_rule()`... why not make this `control_rule` an optional parameter, and set it to `_control_rule()` if a value isn't supplied? Not a blocker by any means; it just seemed like a lot of changes to support making that function public

api/host.py
```diff
@@ -154,7 +155,7 @@ def delete_by_id(host_id_list):
             query, current_app.event_producer, inventory_config().host_delete_chunk_size
         ):
             if deleted:
-                log_host_delete_succeeded(logger, host_id)
+                log_host_delete_succeeded(logger, host_id, get_control_rule)
```
kruai, line 158: Wouldn't `get_control_rule` need to be `get_control_rule()`?

kruai (changes requested):

kruai (approved):

kruai:
@tahmidefaz give it one last look and then let's merge this thing!

tahmidefaz (approved):
Looks good! Let's take back what's ours! :fist_raised:

PR #755, opened 2020-12-01
RHC 8586: Limit CRUD operations to system owners when cert-auth used

thearifismail:
https://issues.redhat.com/browse/RHCLOUD-8586

This PR limits CRUD operation on hosts by the host owners only when the account owner is a system with "auth_type: cert-auth". This limitation does not apply to satellite accounts and the regular users.

This PR changed the way "Identity" was used and hosts were queried.  Now if the identity auth_type="cert_auth" and the idenity_type="System", then a filter is added to the quewry.  

Thanks to changes to identity, its effect is pretty pervasive. For now, please take a look at these files:
1. api/host.py
2. api/host_query_db.py
3. app/auth/identity.py
4. lib/host_repository.py

**Status** as January , 2021

1. Two tests failing, which are expected to get fixed by creating MockIdentity and adding "assert_called_once_with" method.  The failing tests are:
> 1. tests/test_api_host_get::test_query_using_fqdn_not_subset_match
> 2. tests/test_api_host_get::test_query_using_insights_id_not_subset_match
2. Need to pass in an "Identity" object to "lib.host_repository.add_host()" method.  Any test using "lib.host_repository.add_host()" is passing because manually using
```yaml
        USER_IDENTITY["account_number"] = input_host.account
        identity = Identity(USER_IDENTITY)
        existing_host = find_existing_host(identity, input_host.canonical_facts)
```

> Issue fixed.

**Questions**:
1. Should adding a host check for "owner_id"? If it does not match the `cn` for `auth_type: cert-auth`, it should not be added.
> Added dummy identity to `handle_message` in `app/queue/queue.py`

2. If a host addition is rejected, how to communicate back to the account owner?

> Not applicable anymore

@jharting , can you please review this PR? Thanks.

PR #778, opened 2021-01-12
Set test identity account number to the one in payload

thearifismail:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-8586).

This PR fixed a bug in https://github.com/RedHatInsights/insights-host-inventory/pull/755.  The bug was that `find_existing_host` was using `test` as the account number but should be using the one received in the payload from kafka message.

## Secure Coding Practices Checklist GitHub Link

- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #780, opened 2021-01-12
rhc-8586: Adding authenticate systems for CRUD operations

thearifismail:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-8586).

This PR fixed a bug in https://github.com/RedHatInsights/insights-host-inventory/pull/755.  The bug was that `find_existing_host` was using `test` as the account number but should be using the one received in the payload from kafka message.

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


thearifismail:
@ryandillinfelton comment from #778, "Could you add a test to insure that the bug this is fixing is fixed? I'd like to have it to insure we don't accidentally reintroduce it in the future. A test insuring that a created host has the right account ID should be enough.

thearifismail:
https://github.com/RedHatInsights/insights-host-inventory/pull/780#issuecomment-759086647

The new functionality is getting touched by https://github.com/thearifismail/insights-host-inventory/blob/20bcd73b8016d073fafbeb48a5b05881e9ef769d/tests/test_mq_service.py#L181

thearifismail:
@ryandillinfelton, the test you're looking for is already provided https://github.com/thearifismail/insights-host-inventory/blob/4fe5f32ea9fd299a1d38d2d682a1a8e2b61cfc71/tests/test_mq_service.py#L181

Couple of new tests to update host.account has been added.  There are a number of tests present for testing message handling by queue.

thearifismail:
Added [test_handle_message_with_different_account](https://github.com/thearifismail/insights-host-inventory/blob/246fc9fdd73cfe6906f0fe8b397bcdd0711a82d9/tests/test_mq_service.py#L1023), which updates identity.

kruai (changes requested):
I'd like one of your new tests to be a bit more precise, and I would also like to see a test that fails if yesterday's scenario happens again. That is to say, if you call `mq_create_or_update_host()` on a host twice, it should not create a host the second time. 

One way to do this is to call `db_get_hosts()` on the host's ID at the beginning of the test (it should return 0), in-between the `mq_create_or_update_host()` calls (it should return 1), and at the end of the test (it should still return 1). That test would make me feel more at ease with merging this.

kruai:
@tahmidefaz @ryandillinfelton does the new test suggestion in my review make sense to you two?

ryandillinfelton:
> @tahmidefaz @ryandillinfelton does the new test suggestion in my review make sense to you two?

Yes, I like that idea for a test. I agree that we should have a precise test for the scenario.

tests/test_mq_service.py
```diff
@@ -1013,3 +1018,31 @@ def test_invalid_string_is_found_in_list_item(obj):
...
+
+
+def test_handle_message_with_different_account(mocker, flask_app, subtests):
+    mocker.patch("app.queue.queue.build_event")
+    add_host = mocker.patch("app.queue.queue.add_host", return_value=(mocker.MagicMock(), None, None, None))
+
+    operation_raw = "🧜🏿‍♂️"
+    operation_escaped = json.dumps(operation_raw)[1:-1]
+
+    # account_name = "hbi_test_aaccount"
```
ryandillinfelton, line 1030: nitpicky, but small typo in the comment here.

tests/test_mq_service.py
```diff
@@ -1013,3 +1018,31 @@ def test_invalid_string_is_found_in_list_item(obj):
 def test_other_values_are_ignored(value):
     _validate_json_object_for_utf8(value)
     assert True
+
+
+def test_handle_message_with_different_account(mocker, flask_app, subtests):
```
ryandillinfelton, line 1023: This test is overbuilt for it's supposed to be testing. We already test that unicode is not damaged, that doesn't need to be included in a test checking that hosts created through MQ recieve the right account value.

thearifismail: I have removed the extra lines but the intent of this test is test the updated identity NOT the unicode.

tests/test_mq_service.py
thearifismail: Gone!!!

tests/test_api_hosts_get.py
```diff
@@ -975,3 +973,22 @@ def test_get_hosts_sap_sids(patch_xjoin_post, api_get, subtests, query_source_xj
...
 #     assert_response_status(implicit_response_status, 400)
 #     assert_response_status(eq_response_status, 400)
+
+
+def test_host_account_using_mq(mq_create_or_update_host, api_get, db_get_host):
+    host = minimal_host(subscription_manager_id=generate_uuid())
+    host.account = "dummy"
+
+    created_host = mq_create_or_update_host(host)
+    assert db_get_host(created_host.id)
```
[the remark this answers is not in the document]
thearifismail: An attempt to add an existing host does not fail; rather returns the same host with an updated timestamp updated.  So to test the scenario mentioned in the comment:

1.  Created "created_host" and "same_host" using the same "fqdn", i.e.  the same host dictionary object.
2. Set `created_host.updated = same_host.updated`
3. Compared the two hosts using `deepequals` for equality.

tests/test_api_hosts_get.py
```diff
@@ -975,3 +973,31 @@ def test_get_hosts_sap_sids(patch_xjoin_post, api_get, subtests, query_source_xj
...
+    host.account = "dummy"
+
+    created_host = mq_create_or_update_host(host)
+    assert db_get_host(created_host.id).account == "dummy"
+
+    # verify that the two hosts vars are pointing to the same resource.
+    same_host = mq_create_or_update_host(host)
+
+    # update_existing_host() resturns the same host but with updated timestamp.
+    created_host.updated = same_host.updated
```
kruai, line 989: Why are you manually setting created_host's "updated" field to the value in `same_host`?

tests/test_api_hosts_get.py
```diff
@@ -975,3 +973,31 @@ def test_get_hosts_sap_sids(patch_xjoin_post, api_get, subtests, query_source_xj
...
+
+    assert created_host.id == same_host.id
```
kruai, line 991: I'm not sure this is a meaningful assertion. If everything's working as expected, the IDs would be equal; however, if the add/exist functionality didn't work correctly and created a new host, it would still have that ID, right?
I think that to fully validate the issue we saw yesterday, we really need to use something like `db_get_hosts()` to check how many records exist for a certain ID.

kruai (changes requested):

tests/test_api_hosts_get.py
thearifismail: @kruai, I should have removed `assert created_host.id == same_host.id`.  

you asked, "if the add/exist functionality didn't work correctly and created a new host, it would still have that ID, right?".

The answer is No because each new host gets a unique ID.  Since for both hosts, the same `fqdn` is used, inventory returns the same host.
No, new host has a new `id`, generated by inventory host.  

I can take out `assert created_host.id == same_host.id` if it is causing confusion.

tests/test_api_hosts_get.py
kruai: Oh, I'm guessing it's so you can compare the `__dict__` for each... I don't know what that would tell us, though, since you added the exact same host twice

tests/test_api_hosts_get.py
kruai: Oh, I see what you mean. In that case, asserting that the new and old IDs are the same is a good test, as it tells us that it updated a host instead of creating a new one. Thanks for clarifying!

thearifismail: Updated test to use `get_db_hosts()` and then compared the number of hosts before and after the second attempted to create the same host.

tests/test_mq_service.py
ryandillinfelton: Thanks. I think you could go further though, it doesn't seem like you need the test to loop when there is only one message.

tests/test_api_hosts_get.py
```diff
@@ -975,3 +973,35 @@ def test_get_hosts_sap_sids(patch_xjoin_post, api_get, subtests, query_source_xj
 
 #     assert_response_status(implicit_response_status, 400)
 #     assert_response_status(eq_response_status, 400)
+
+
+def test_host_account_using_mq(mq_create_or_update_host, api_get, db_get_host, db_get_hosts):
```
ryandillinfelton, line 978: I feel like this test belongs more in test_mq than here. You also don't use the `api_get` function in this test.

kruai (approved):

tests/test_api_hosts_get.py
```diff
@@ -975,3 +973,35 @@ def test_get_hosts_sap_sids(patch_xjoin_post, api_get, subtests, query_source_xj
...
+    second_batch = db_get_hosts([created_host.id, same_host.id])
+
+    # update_existing_host() resturns the same host but with updated timestamp.
+    created_host.updated = same_host.updated
+
+    assert created_host.__dict__ == same_host.__dict__
+    assert len(first_batch.all()) == len(second_batch.all())
+
+
+def test_host_account_using_db(db_create_host, db_get_host):
```
ryandillinfelton, line 999: This test should also move to the test_mq file

tests/test_api_hosts_get.py
```diff
@@ -975,3 +973,35 @@ def test_get_hosts_sap_sids(patch_xjoin_post, api_get, subtests, query_source_xj
...
+    # verify that the two hosts vars are pointing to the same resource.
+    same_host = mq_create_or_update_host(host)
+
+    second_batch = db_get_hosts([created_host.id, same_host.id])
+
+    # update_existing_host() resturns the same host but with updated timestamp.
+    created_host.updated = same_host.updated
+
+    assert created_host.__dict__ == same_host.__dict__
+    assert len(first_batch.all()) == len(second_batch.all())
```
ryandillinfelton, line 996: This is right, and I'm happy with what the test accomplishes now. Correct me if I'm missing something, but I think you could just compare the id's of both created hosts to check the same thing.

thearifismail: @kruai had reservation about comparing just the `id`.  https://github.com/RedHatInsights/insights-host-inventory/pull/780#discussion_r556801483

Since comparing `__dict__` includes `id`, I took it out.

tests/test_api_hosts_get.py
thearifismail: There were three more `mq_create_or_update_host` tests in this file, so I left it in here because it also used `get_db_host()`.  So I left it in here.  

Since @kruai already approved it, I think we should leave it here for now and can move it in a following PR.  On my system, I have moved the test to the `test_mq_service.py` and it executed successfully.

tests/test_api_hosts_get.py
kruai: @thearifismail you'll see later in that comment chain that I agreed that you only needed to test the IDs and not the other fields. I don't mind either way between comparing the IDs and comparing the whole dicts at this point

thearifismail: Fidelity-wise, it is better to compare all params to not have any questions.  I am sorry, I did not see your comment about IDs.  

Is it good for approval by both of you, particularly @ryandillinfelton?

tests/test_api_hosts_get.py
kruai: Feels unfair to use my approval to disregard requested changes from someone else (that's why we have multiple reviewers). If you've already moved it and it's passing, why not just push that change?

tests/test_api_hosts_get.py
ryandillinfelton: I'm happy enough with how it works to approve.

thearifismail: Moved `test_host_account_using_mq` to the mq tests file per @ryandillinfelton suggestion.

ryandillinfelton (approved):

PR #785, opened 2021-01-15
system identity validation fix

ryandillinfelton:
## Overview

This PR improves out identity validation to prevent system identities lacking the required fields.

## Secure Coding Checklist

- [x] Input Validation
- [ ] Output Encoding
- [x] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/auth/identity.py
+++ b/app/auth/identity.py
@@ -34,6 +34,7 @@ class Identity:
         elif obj:
             self.is_trusted_system = False
             self.account_number = obj["account_number"]
+            self.auth_type = obj["auth_type"]
 
             # This check may change to if "auth_type" in obj.keys()
             # and then basic-auth and cert-auth.
@@ -56,9 +57,19 @@ class Identity:
 
     def _asdict(self):
         if self.identity_type == "User":
-            return {"account_number": self.account_number, "type": self.identity_type, "user": self.user.copy()}
+            return {
+                "account_number": self.account_number,
+                "type": self.identity_type,
+                "auth_type": self.auth_type,
+                "user": self.user.copy(),
+            }
         if self.identity_type == "System":
-            return {"account_number": self.account_number, "type": self.identity_type, "system": self.system.copy()}
+            return {
+                "account_number": self.account_number,
+                "type": self.identity_type,
+                "auth_type": self.auth_type,
+                "system": self.system.copy(),
+            }
 
     def __eq__(self, other):
         return self.account_number == other.account_number
@@ -79,7 +90,7 @@ def validate(identity):
         if not identity.account_number:
             raise ValueError("The account_number is mandatory.")
 
-        elif identity.identity_type == "System":
+        elif identity.identity_type == "System" and identity.auth_type != "classic-proxy":
             if not identity.system["cert_type"]:
                 raise ValueError("The cert_type field is mandatory.")
             if not identity.system["cn"]:
--- a/app/auth/identity.py
+++ b/app/auth/identity.py
@@ -56,9 +56,9 @@ class Identity:
 
     def _asdict(self):
         if self.identity_type == "User":
-            return {"account_number": self.account_number, "type": self.identity_type, "user": self.user}
+            return {"account_number": self.account_number, "type": self.identity_type, "user": self.user.copy()}
         if self.identity_type == "System":
-            return {"account_number": self.account_number, "type": self.identity_type, "system": self.system}
+            return {"account_number": self.account_number, "type": self.identity_type, "system": self.system.copy()}
 
     def __eq__(self, other):
         return self.account_number == other.account_number
--- a/app/auth/identity.py
+++ b/app/auth/identity.py
@@ -34,7 +34,11 @@ class Identity:
         elif obj:
             self.is_trusted_system = False
             self.account_number = obj["account_number"]
-            self.auth_type = obj["auth_type"]
+
+            if obj.get("auth_type"):
+                self.auth_type = obj["auth_type"]
+            else:
+                self.auth_type = None
 
             # This check may change to if "auth_type" in obj.keys()
             # and then basic-auth and cert-auth.
--- a/app/auth/identity.py
+++ b/app/auth/identity.py
@@ -78,3 +78,9 @@ def validate(identity):
         # Ensure the account number is present.
         if not identity.account_number:
             raise ValueError("The account_number is mandatory.")
+
+        elif identity.identity_type == "System":
+            if not identity.system["cert_type"]:
+                raise ValueError("The cert_type field is mandatory.")
+            if not identity.system["cn"]:
+                raise ValueError("The cn field is mandatory.")
--- a/app/auth/identity.py
+++ b/app/auth/identity.py
@@ -91,7 +91,7 @@ def validate(identity):
             raise ValueError("The account_number is mandatory.")
 
         elif identity.identity_type == "System" and identity.auth_type != "classic-proxy":
-            if not identity.system["cert_type"]:
+            if not identity.system.get("cert_type"):
                 raise ValueError("The cert_type field is mandatory.")
-            if not identity.system["cn"]:
+            if not identity.system.get("cn"):
                 raise ValueError("The cn field is mandatory.")
--- a/lib/host_repository.py
+++ b/lib/host_repository.py
@@ -187,7 +187,7 @@ def stale_timestamp_filter(gt=None, lte=None):
 def update_query_for_owner_id(identity, query):
     # kafka based requests have dummy identity for working around the identity requirement for CRUD operations
     # TODO: 'identity.auth_type is not 'classic-proxy' is a temporary fix. Remove when workaround is no longer needed
-    logger.info("identity auth type: %s", identity.auth_type)
+    logger.debug("identity auth type: %s", identity.auth_type)
     if (
         identity
         and identity.identity_type == "System"
--- a/lib/host_repository.py
+++ b/lib/host_repository.py
@@ -188,8 +188,6 @@ def update_query_for_owner_id(identity, query):
     # kafka based requests have dummy identity for working around the identity requirement for CRUD operations
     # TODO: 'identity.auth_type is not 'classic-proxy' is a temporary fix. Remove when workaround is no longer needed
     logger.info("identity auth type: %s", identity.auth_type)
-    # if identity.auth_type == "classic-proxy":
-    #     return query
     if (
         identity
         and identity.identity_type == "System"
--- a/lib/host_repository.py
+++ b/lib/host_repository.py
@@ -186,7 +186,16 @@ def stale_timestamp_filter(gt=None, lte=None):
 
 def update_query_for_owner_id(identity, query):
     # kafka based requests have dummy identity for working around the identity requirement for CRUD operations
-    if identity and identity.identity_type == "System" and identity.system["cert_type"] == "system":
+    # TODO: 'identity.auth_type is not 'classic-proxy' is a temporary fix. Remove when workaround is no longer needed
+    logger.info("identity auth type: %s", identity.auth_type)
+    # if identity.auth_type == "classic-proxy":
+    #     return query
+    if (
+        identity
+        and identity.identity_type == "System"
+        and identity.auth_type != "classic-proxy"
+        and identity.system["cert_type"] == "system"
+    ):
         return query.filter(and_(Host.system_profile_facts["owner_id"].as_string() == identity.system["cn"]))
     else:
         return query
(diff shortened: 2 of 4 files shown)


tests/test_api_auth.py
```diff
@@ -62,6 +80,18 @@ def test_validate_valid_system_identity(flask_client):
     assert 200 == response.status_code  # OK
 
 
+def test_validate_valid_invalid_system_identities(flask_client, subtests):
```
thearifismail, line 83: Rename the test to `test_invalid_system_identities` or `test_validate_invalid_system_identities` is more appropriate

thearifismail (changes requested):
Observing a strange behavior not seen before.  Using `pytest`, `pytest -vv` or `pytest --cov=.` fails 11 tests.  BUT, no such failures are encountered when only the files containing the failed tests `pytest -vv tests/test_api_facts.py` OR only the failing tests are executed directly, e.g `pytest -vv tests/test_api_facts.py::test_put_facts_with_RBAC_bypassed_as_system`.  

[pytest_results.txt](https://github.com/RedHatInsights/insights-host-inventory/files/5827083/pytest_results.txt)
[executed_just_failed_tests_and_passed.txt](https://github.com/RedHatInsights/insights-host-inventory/files/5827085/executed_just_failed_tests_and_passed.txt)

RonnyPfannschmidt:
Im awake by pure luck and this look like something with the constant or its keys is wrong, examining that data and code that changes it might be helpful

I'll take a closer look tommorow morning when I get back to the computer

RonnyPfannschmidt:
To be more specific, it may be necessary to examine if some code is missing a copy/deepcopy and modifies things in place, without any deeper investigation this looks like a accidental constant change which is now a unfortunate test dependency may be the cause

tests/test_api_auth.py
```diff
@@ -8,6 +8,24 @@
 from tests.helpers.api_utils import USER_IDENTITY
 
 
+def invalid_identities(identity_type):
```
RonnyPfannschmidt, line 11: this code breaks the `SYSTEM_IDENTITY` global as system comes as a reference not as a copy

after this case runs once `SYSTEM_IDENTITY`  is no longer usable

RonnyPfannschmidt (changes requested):
i recommend to create altered copies of SYSTEM_IDENTITY which then use altered copies of the `system` subkey

tests/test_api_auth.py
kruai: I think `test_invalid_system_identities` is more appropriate here.

tests/test_api_auth.py
kruai: Good catch! You'll need to clone the dictionary when it gets passed in @ryandillinfelton

app/auth/identity.py
```diff
@@ -78,3 +78,9 @@ def validate(identity):
         # Ensure the account number is present.
         if not identity.account_number:
             raise ValueError("The account_number is mandatory.")
+
+        elif identity.identity_type == "System":
+            if not identity.system["cert_type"]:
+                raise ValueError("The cert_type field is mandatory.")
+            if not identity.system["cn"]:
+                raise ValueError("The cn field is mandatory.")
```
kruai, line 86: Are our stakeholders cool with us erroring out in these scenarios? Just want to make sure, because we'll still end up with a situation like Friday where HBI was throwing a bunch of errors

kruai (changes requested):

tests/test_api_auth.py
kruai: Just to clarify, I mean `dict.deepcopy()`

tests/test_api_auth.py
```diff
@@ -8,6 +8,24 @@
...
+    if identity_type == "System":
+        return (
+            Identity(SYSTEM_IDENTITY)._asdict()["system"].pop("cert_type", None),
```
RonnyPfannschmidt, line 14: another issue, as identities are being constructed from the return value, i believe that using the pop as return value is completely incorrect as pop returns the element it took out of a dict

RonnyPfannschmidt: note that this wont show up as  problem since the incorrect data will cause authentication errors just the "same",

i would recommend to validate the error message to ensure each breakage is handled as one would hope

app/auth/identity.py
thearifismail: @asa who are the stakeholders on your mind?

thearifismail: fixed the issue by updating `_asdict()` method to return copies of user as `identity.user.copy()` and system as `identity.system.copy()`
https://github.com/RedHatInsights/insights-host-inventory/blob/368a976d08f09b92f15e0eabf20a4ddf9508bd9d/app/auth/identity.py#L59

tests/test_api_auth.py
thearifismail: fixed the issue by updating `_asdict()` method to return copies of user as `identity.user.copy()` and system as `identity.system.copy()`

thearifismail (approved):
LGTM.  @ryandillinfelton agreed with the change.

app/auth/identity.py
kruai: @thearifismail I'm mostly referring to @lphiri, as he said he was going to discuss the errors we saw on Friday with @coderbydesign. It could be that this change is valid, but that we need to keep this change from going into Production until the corresponding 3scale change has been deployed.

Keith and Lindani, does that sound correct?

tests/test_api_auth.py
kruai: @thearifismail I believe this comment still has yet to be addressed

ryandillinfelton: I'm addressing this now. Going to request changes to block till it's ready

app/auth/identity.py
thearifismail: Thanks Asa.  I think we need to run by the producers of all three reporters.

kruai: I'm not so sure; it seems like this functionality provides the correct desired state, but there's an issue with 3scale or insights-client that gave us some invalid identity headers. I think following up with Lindani and maybe Keith is the best first step to take on this one

tests/test_api_auth.py
ryandillinfelton: Thanks for the catch Ronny. I've manually validated that the identity headers are now incorrect in the right way. The API responds the same way when the identity header is invalid so I haven't made the test aware of if it's being marked invalid for the correct reason. I'll explore using mocks to verify, but I'm confident that the test is working as intended right now, even if it's less robust than it ideally would be.

RonnyPfannschmidt (dismissed):

ryandillinfelton:
I'm marking this as on hold since merging it will likely make it more difficult to add a work around for requests from insights classic. We can revisit this once the work around is sorted out.

app/auth/identity.py
```diff
@@ -78,3 +78,9 @@ def validate(identity):
         # Ensure the account number is present.
         if not identity.account_number:
             raise ValueError("The account_number is mandatory.")
+
+        elif identity.identity_type == "System":
+            if not identity.system["cert_type"]:
+                raise ValueError("The cert_type field is mandatory.")
```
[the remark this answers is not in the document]
ryandillinfelton: Noting that we've elected to handle this in a follow up

lib/host_repository.py
```diff
@@ -186,7 +186,16 @@ def stale_timestamp_filter(gt=None, lte=None):
 
 def update_query_for_owner_id(identity, query):
     # kafka based requests have dummy identity for working around the identity requirement for CRUD operations
-    if identity and identity.identity_type == "System" and identity.system["cert_type"] == "system":
+    # TODO: 'identity.auth_type is not 'classic-proxy' is a temporary fix. Remove when workaround is no longer needed
+    logger.info("identity auth type: %s", identity.auth_type)
+    # if identity.auth_type == "classic-proxy":
```
thearifismail, line 191: remove comment the commented `if` and `return` statements.

thearifismail (approved):
Besides one minor comment, LGTM

lib/host_repository.py
ryandillinfelton: Fixed

thearifismail (approved):
LGTM

app/auth/identity.py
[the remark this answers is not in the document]
kruai: Follow-up Jira bug: https://issues.redhat.com/browse/RHCLOUD-11801

app/auth/identity.py
```diff
@@ -78,3 +78,9 @@ def validate(identity):
         # Ensure the account number is present.
         if not identity.account_number:
             raise ValueError("The account_number is mandatory.")
+
+        elif identity.identity_type == "System":
+            if not identity.system["cert_type"]:
```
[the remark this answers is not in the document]
kruai: Good catch. This should probably be addressed before merging @ryandillinfelton

app/auth/identity.py
thearifismail: We had not expected to `cn` missing from `Identity.idenity_type == system`; now the case is handled when the query is checked and updated for `owner_id`.  Kudos to @ryandillinfelton for adding and updating tests.  Identity checks will get simpler than the current ones when a uniform identity structure is suppported.

kruai (dismissed):

tahmidefaz (approved):
Let's go!!

ryandillinfelton:
Worried about the assumption that the identity will have an auth-type. It's becoming clear that there are several internal teams making their own identity headers in different ways and that doesn't see to be a safe assumption or a good reason to reject the header IMO.

kruai:
> Worried about the assumption that the identity will have an auth-type. It's becoming clear that there are several internal teams making their own identity headers in different ways and that doesn't see to be a safe assumption or a good reason to reject the header IMO.

What if we only check for the auth-type if it's from insights-client? Would that help resolve it?

thearifismail:
> > Worried about the assumption that the identity will have an auth-type. It's becoming clear that there are several internal teams making their own identity headers in different ways and that doesn't see to be a safe assumption or a good reason to reject the header IMO.
> 
> What if we only check for the auth-type if it's from insights-client? Would that help resolve it?

Yes set `auth_type` when present, like the `cn` and `cert_type`.

tahmidefaz:
> > > Worried about the assumption that the identity will have an auth-type. It's becoming clear that there are several internal teams making their own identity headers in different ways and that doesn't see to be a safe assumption or a good reason to reject the header IMO.
> > 
> > 
> > What if we only check for the auth-type if it's from insights-client? Would that help resolve it?
> 
> Yes set `auth_type` when present, like the `cn` and `cert_type`.

Is doing so going to be safe? It's probably ok? Just wanted to make sure. :slightly_smiling_face:

ryandillinfelton:
> > > > Worried about the assumption that the identity will have an auth-type. It's becoming clear that there are several internal teams making their own identity headers in different ways and that doesn't see to be a safe assumption or a good reason to reject the header IMO.
> > > 
> > > 
> > > What if we only check for the auth-type if it's from insights-client? Would that help resolve it?
> > 
> > 
> > Yes set `auth_type` when present, like the `cn` and `cert_type`.
> 
> Is doing so going to be safe? This is probably ok? Just wanted to make sure.

I believe it is. We weren't requiring it before, and afaik it's only used for the insights-classic work around.

tests/test_api_auth.py
```diff
@@ -14,8 +38,10 @@ def valid_identity(identity_type):
     """
     if identity_type == "User":
         return Identity(USER_IDENTITY)
-    else:
+    elif identity_type == "System":
         return Identity(SYSTEM_IDENTITY)
+    elif identity_type == "Classic":
```
thearifismail, line 43: There is no `identity_type: "Classic"`.  For `"auth_type": "classic-proxy"`, `type` is `System`.  This means to the tests will have to adjust to something like: 
```code
    return Identity(random.choice[INSIGHTS_CLASSIC_IDENTITY, SYSTEM_IDENTITY)
```

asa:
Can you all take me off your shoutouts?  I am not the asa you are looking for! =). Hi Asa, cool to meet another one, but getting these updates clogs my inbox!

tests/test_api_auth.py
```diff
@@ -14,8 +38,10 @@ def valid_identity(identity_type):
...
+        return Identity(INSIGHTS_CLASSIC_IDENTITY)
 
 
 def valid_payload(identity_type):
```
thearifismail, line 47: Since there is no `identity_type`, this method needs an additional argument to create `INSIGHTS_CLASSIC_IDENTITY`

kruai (approved):

thearifismail (approved):
Looks good.  I like a lot the new create_identity_payload.

PR #787, opened 2021-01-19
Ubi8

kruai:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-11366).
This is only Part 1 of the work for this Jira card. It changes our dev Dockerfile to use the appropriate UBI8 image, and updates the Jenkinsfile to use postgres 10.6 (the one we use in Stage and Prod).

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/dev.dockerfile
+++ b/dev.dockerfile
@@ -1,15 +1,11 @@
-FROM registry.centos.org/centos/python-36-centos7
+FROM registry.access.redhat.com/ubi8/python-36
 
-USER root
-RUN yum -y install centos-release-scl-rh
-RUN yum -y install rh-postgresql10-postgresql
-RUN sed -i 's/source scl_\(.*\)$/source scl_\1 rh-postgresql10/' /opt/app-root/etc/scl_enable
 USER 1001
 
 WORKDIR /opt/app-root/src
 COPY . .
 
-RUN scl enable rh-python36 "pip install --upgrade pip"
+RUN pip install --upgrade pip
 RUN pip install pipenv
 RUN pipenv install --system --dev
 
--- a/dev.dockerfile
+++ b/dev.dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/python-36
+FROM registry.access.redhat.com/ubi8/python-36:1-123.16099467941
 
 USER 1001
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -39,7 +39,7 @@ def runStages() {
         ),
         containerTemplate(
             name: 'postgres',
-            image: 'postgres:9.6',
+            image: 'postgres:10.6',
             ttyEnabled: true,
             envVars: [
                 containerEnvVar(key: 'POSTGRES_USER', value: 'insights'),
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -39,7 +39,7 @@ def runStages() {
         ),
         containerTemplate(
             name: 'postgres',
-            image: 'postgres:10.6',
+            image: 'postgres:12.5',
             ttyEnabled: true,
             envVars: [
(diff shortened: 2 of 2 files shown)


tahmidefaz (approved):
LGTM

dev.dockerfile
```diff
@@ -1,15 +1,11 @@
-FROM registry.centos.org/centos/python-36-centos7
+FROM registry.access.redhat.com/ubi8/python-36
```
gmcculloug, line 1: Happy to see this.  I wonder if we should be using a tag like we do with other ubi images [like sources](https://github.com/RedHatInsights/sources-api/blob/master/Dockerfile#L1).

@lindgrenj6 What's the best practice here?

lindgrenj6: Usually best practice is to use a tag - that way it's easy for infosec to track vulnerabilities. Also that way there is always an artifact showing the upgrade to mitigate any CVEs that come along. 

Red Hat images are really good about re-tagging latest (as you can see here: [ubi8/python-36](https://catalog.redhat.com/software/containers/ubi8/python-36/5c839d99d70cc51dd4c42610)) e.g. the docker tag `1-123.16099467941` was pushed, re-tagged to `1` which is the "rolling" v1 image, then re-tagged to `latest` as well. 

_but_ 

Looking at the repo there I would recommend tagging with `1-123.16099467941` since that's the latest one that won't be re-taggeed. This provides more control about what gets upgraded rather than the image changing on a rebuild. 

The tag is ugly, but it looks like this might be a one-off thing, the next one might be `1-124`. This can be something that just gets upgraded either every other sprint or just when necessary, either would work.

kruai: Thanks for the info! I've updated it to use that tag.

kruai:
I've marked this one as "on hold" for now, because we'll need this to be a coordinated change with the RDS upgrade. The UBI images all use postgres 12.5, so it makes sense to move to that version now. However, if we push this change, we won't be able to safely push to Prod until the RDS instances are also running postgres 12.x.

Once we have a schedule prepared, we can merge this and thoroughly test a few days before the RDS instance is upgraded.

tahmidefaz:
@kruai It might be better to mark this as a WIP PR in the meantime.

dev.dockerfile
```diff
@@ -1,15 +1,11 @@
-FROM registry.centos.org/centos/python-36-centos7
+FROM registry.access.redhat.com/ubi8/python-36:1-123.16099467941
```
thearifismail, line 1: What's the plan for moving up to `Python V3.8-x`?

thearifismail (changes requested):
Shouldn't `yum update` or `yum update all` be in the docker file?

dev.dockerfile
lindgrenj6: This: https://catalog.redhat.com/software/containers/ubi8/python-38/5dde9cacbed8bd164a0af24a

registry.access.redhat.com/ubi/python-38:xxx

lindgrenj6:
@thearifismail typically that is recommended against for prod containers, just for the reasons I mentioned before where the running versions of packages can change out from underneath you between builds. It makes it harder for infosec to track what problems are in each image. 

It's better to just consistently update the tag every couple weeks/month/whenever it updates. That way there is a trackable artifact that shows exactly when upgrades happen.

kruai:
Closing this, because future work on using UBI will likely use a different image. We should still be able to reference the changes made in this PR if needed.

PR #792, opened 2021-01-22
Update build pipeline -- phase one

BlakeHolifield:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-11665).
This PR will:
* Build and deploy to latest and qa only on a push to master. 
* Update the Clowdapp template (deployment.yaml) to use the new podSpec inside deployments
* Remove `source build_deploy.sh` from pr_check
* Build and tag a pr_check hash without pushing to latest or qa 
* Copy `dev.dockerfile` to `Dockerfile` for build process -- Discuss rename

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -39,9 +39,8 @@ source build.sh
 source deploy_ephemeral_env.sh
 
 # Need to make a dummy results file to make tests pass
-cd ../..
-mkdir -p artifacts
-cat << EOF > artifacts/junit-dummy.xml
+mkdir -p $WORKSPACE/artifacts
+cat << EOF > ${WORKSPACE}/artifacts/junit-dummy.xml
 <testsuite tests="1">
     <testcase classname="dummy" name="dummytest"/>
 </testsuite>
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -31,11 +31,11 @@ IQE_FILTER_EXPRESSION=""
 # We'll take it from here ...
 # ---------------------------
 
-source build_deploy.sh
 
 CICD_URL=https://raw.githubusercontent.com/RedHatInsights/bonfire/master/cicd
 curl -s $CICD_URL/bootstrap.sh -o bootstrap.sh
 source bootstrap.sh  # checks out bonfire and changes to "cicd" dir...
+source build.sh
 source deploy_ephemeral_env.sh
 
 # Need to make a dummy results file to make tests pass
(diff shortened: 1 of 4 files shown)


pr_check.sh
```diff
@@ -31,17 +31,16 @@ IQE_FILTER_EXPRESSION=""
 # We'll take it from here ...
 # ---------------------------
 
-source build_deploy.sh
 
 CICD_URL=https://raw.githubusercontent.com/RedHatInsights/bonfire/master/cicd
 curl -s $CICD_URL/bootstrap.sh -o bootstrap.sh
 source bootstrap.sh  # checks out bonfire and changes to "cicd" dir...
+source build.sh
```
kruai, line 38: It looks like you're adding this before we even have a `build.sh` file. Won't this break our current PR checks?

kruai (changes requested):
I just have one issue with this PR so far, but I'm also not sure how to review `deployment.yaml`. There's so much config, and it's hard to verify that all of it (and its structure) is correct. When you review these configs, do you generally run them through a validation script, or copy&paste them into a dummy OpenShift project?

pr_check.sh
BlakeHolifield: The files we're sourcing such as `bootstrap.sh` and `build.sh` are part of the bonfire cicd tools we're using. It's not explicit, but that `source bootstrap.sh` is doing a git clone of bonfire and then cd'ing to the /cicd directory where `build.sh` and `deploy_ephemeral.sh` live. We curl once and get the smoke tools essentially. Source code; https://github.com/RedHatInsights/bonfire/blob/master/cicd/bootstrap.sh

On that note, the `build.sh` is running to make a new tag, and the next line `source deploy_ephemeral.sh` is using that tag to spin up and roll out the deployments from the deployment yaml. If a rollout times out, the build will fail. This one looked good according to ole man Jenkins: https://ci.ext.devshift.net/blue/organizations/jenkins/RedHatInsights-insights-host-inventory-pr-check/detail/RedHatInsights-insights-host-inventory-pr-check/310/pipeline#log-1133

kruai: Oh, I see. Just wanted to make sure that making this change would not only work for the clowder integration, but also continue working for our current PR pipeline. Thanks for the clarification!

kruai (approved):

thearifismail (approved):

PR #793, opened 2021-01-25
Run unit tests in pr_check.sh

BlakeHolifield:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-11665).
These changes will
* Run unit tests in PR checks using Clowder
* Add a `port` option for the Postgresql URI
* Lock prance to 20.0 and lock sqlalchemy-utils to 0.36.7 (due to a bug in 36.8)
* Pipe the db creds into the `set_environment` method

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -22,6 +22,7 @@ class Config:
         self._db_user = cfg.database.username
         self._db_password = cfg.database.password
         self._db_host = cfg.database.hostname
+        self._db_port = cfg.database.port
         self._db_name = cfg.database.name
         if cfg.database.rdsCa:
             self._db_ssl_cert = cfg.rds_ca()
@@ -50,6 +51,7 @@ class Config:
         self._db_user = os.getenv("INVENTORY_DB_USER", "insights")
         self._db_password = os.getenv("INVENTORY_DB_PASS", "insights")
         self._db_host = os.getenv("INVENTORY_DB_HOST", "localhost")
+        self._db_port = os.getenv("INVENTORY_DB_PORT", 5432)
         self._db_name = os.getenv("INVENTORY_DB_NAME", "insights")
         self.rbac_endpoint = os.environ.get("RBAC_ENDPOINT", "http://localhost:8111")
         self.host_ingress_topic = os.environ.get("KAFKA_HOST_INGRESS_TOPIC", "platform.inventory.host-ingress")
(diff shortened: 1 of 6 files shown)


pr_check.sh
```diff
@@ -31,19 +120,16 @@ IQE_FILTER_EXPRESSION=""
 # We'll take it from here ...
 # ---------------------------
 
-
 CICD_URL=https://raw.githubusercontent.com/RedHatInsights/bonfire/master/cicd
 curl -s $CICD_URL/bootstrap.sh -o bootstrap.sh
 source bootstrap.sh  # checks out bonfire and changes to "cicd" dir...
 source build.sh
 source deploy_ephemeral_env.sh
+# source smoke_test.sh
```
thearifismail, line 128: How is smoke test triggered?

thearifismail (approved):
LGTM.  All tests executed using `pytest` executed successfully.

pr_check.sh
BlakeHolifield: For now, smoke tests are run via the `deploy_ephemeral_env.sh`. We have work to do on getting the iqe tests to run in the eph environment with Clowder.

app/config.py
```diff
@@ -50,6 +51,7 @@ def non_clowder_config(self):
         self._db_user = os.getenv("INVENTORY_DB_USER", "insights")
         self._db_password = os.getenv("INVENTORY_DB_PASS", "insights")
         self._db_host = os.getenv("INVENTORY_DB_HOST", "localhost")
+        self._db_port = os.getenv("INVENTORY_DB_PORT", 5432)
```
tahmidefaz, line 54: Got a tiny question! Do we know with complete certainty that the DB runs on this port? I know it's typical for psql, but wanted to check. Otherwise, we might need to make some changes to our deployment configs.

thearifismail: @tahmidefaz have you observed a case where port `5432` is not available?

BlakeHolifield: So that will fetch the `INVENTORY_DB_PORT` if it exists. If it isn't set, it will set the var to 5432. The default behavior for before this change was to use port 5432. It's hardcoded into the sqlalchemy library unless you pass it in like we do here. Does that answer your question?

tahmidefaz: Got it, thanks!

tahmidefaz (approved):

PR #794, opened 2021-01-25
Copy Saas-Templates to Code repo

BlakeHolifield:
## Overview

This PR will mirror the saas-templates repo templates. In order to get CI / QA hosted in app-interface


kylape:
Instead of moving deployment.yaml, can you only copy it for now?  That way we can change the ref in app-interface and then delete it in a follow-up PR.

pr_check.sh
```diff
@@ -100,11 +100,7 @@ python manage.py db upgrade
 pytest --cov=. --junitxml=junit.xml --cov-report html -sv
 
 mkdir -p $WORKSPACE/artifacts
-cat << EOF > ${WORKSPACE}/artifacts/junit-dummy.xml
-<testsuite tests="1">
-    <testcase classname="dummy" name="dummytest"/>
-</testsuite>
-EOF
+cp $WORKSPACE/junit.xml $WORKSPACE/artifacts/junit.xml
```
kylape, line 103: oh neat - is this populated from pre-commit?

BlakeHolifield: Their pytest is configured to produce this and drop it in `pwd` 
https://github.com/RedHatInsights/insights-host-inventory/pull/794/files#diff-46233f36c38a197d402638dc37fffcc351a00cd8d359b9de2c3923ae63008c4cR100

kruai:
@BlakeHolifield Is this still a change we need to make? If so, I can make a Jira to update our `deployment.yml` file with the most recent updates to saas-templates. We should probably close this PR either way, though

BlakeHolifield:
@kruai I think we can close this. We can revisit if needed, but this one is pretty far behind and I'd like to do it fresh if we need it.

PR #797, opened 2021-01-27
short circuit to avoid 500 when no cert type. 

ryandillinfelton:
## Overview

Changing the logic to short circuit and avoid a 50 error from classic identity since it actually has not cert type. Updated identity used in tests to reflect that there is not cert_type in classic identity. I want to verify that the system field even exists in these identities before we merge this since it seems odd that they would send us an empty field, and it's emptyness could lead to more problems.

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/host_query_xjoin.py
+++ b/api/host_query_xjoin.py
@@ -90,8 +90,8 @@ def get_host_list(
     current_identity = get_current_identity()
     if (
         current_identity.identity_type == "System"
-        and current_identity.system["cert_type"] == "system"
         and current_identity.auth_type != "classic-proxy"
+        and current_identity.system["cert_type"] == "system"
     ):
         all_filters += owner_id_filter()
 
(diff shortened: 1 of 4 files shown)


kruai:
Weird that the unit test check appears successful here. Looks like the unit test check hit an error which is causing the pr-head to fail

kruai (approved):
It looks like the unit test check just hit a db error at the end of the script for some reason, but all of the tests appear to have passed.

tahmidefaz (approved):
Is this the last one? :smirk_cat:

kruai:
Boy I sure hope so. I'm going to go ahead and merge this. We should really get some QA/Stage tests to validate this too

thearifismail (approved):

PR #799, opened 2021-01-28
Lowercase all tags namespaces

ryandillinfelton:
## Overview

**STATUS**: This PR is ready to be reviewed, but we will need to wait for some changes to be made to reporters before we can merge and deploy it. ~~I've added the on hold label for now, but don't let it dissuade you from reviewing while we wait :smile:.~~

This PR is being created to address These Jiras:
- [RHC-11644](https://issues.redhat.com/browse/RHCLOUD-11644)
- [RHC-11643](https://issues.redhat.com/browse/RHCLOUD-11643)

Features:
- [x] Changes all namespaces on tags to lowercase when a new host is created
- [x] Changes namespace in tag filter on get /hosts requests to lowercase automatically (case insensitive)
- [x] Namespace is an enumeration of two values (insights-client and satellite)
- [x] Reject and log invalid namespaces

Notes:
You may (hopefully) notice two tests missing. I had to remove two tests that were predicated on being able to create tags with no namespace. I do not believe this qualifies as a reduction in coverage since the use case is now explicitly disallowed.

For logging, I believe we already log all validation errors we encounter. Since this will be just another validation rule I don't believe there is any new work to be done here, unless there is a desire to add a new tracker to grafana or set up alerts related to the change.

## Secure Coding Checklist

- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


ryandillinfelton:
Note to self: make sure that this still accepts namespace "Satellite"

ryandillinfelton:
Note for checking in on this PR in the future. The fix that allows this change to be made in HBI is in satellite 6.9+. We need to wait until sufficiently few users are still using versions older that 6.9 to add this to HBI.

kruai:
@ryandillinfelton With the work you've been doing to make tags case-insensitive ([ESSNTL-570](https://issues.redhat.com/browse/ESSNTL-570)), is this PR still required? I'm going to close this for now, but please reopen it if it's actually needed.

ryandillinfelton:
> @ryandillinfelton With the work you've been doing to make tags case-insensitive ([ESSNTL-570](https://issues.redhat.com/browse/ESSNTL-570)), is this PR still required? I'm going to close this for now, but please reopen it if it's actually needed.

I think it would still be good to do. The case-insensitivity is only to get correct type-ahead behavior when looking for tags to use in a query. In the host query itself though the tags are still case sensitive so `/api/hosts?tags=sat/namespace=val` returns different hosts than `/api/hosts?tags=sat/namespace=val`

And we can't just always read case-insensitive and write case-sensitive. It will lead to issues since internally there are two namespaces: `Sat` and `sat` they can have overlapping key value pairs.

The issue was that one reporter was sending `Sat` and the other `sat` but they didn't send the same set of tags, so they would be continually contradicting one another. I think the issue is fixed in new versions of Satellite, but I don't know if the older versions of Satellite that would cause problems with this change, are still supported or not. I'll have to check with Lindani.

kruai:
OK, reopening this, then! I was just trying to clean up some of our PRs because we have some very old ones.

PR #805, opened 2021-02-03
RHCLOUD-11719 System Profile Top Level Sparse fields

tahmidefaz:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-11719).
This PR allows for the querying of xjoin-search to implement top-level sparse fieldsets for the `/hosts/{host_id_list}/system_profile` endpoint.

## Secure Coding Checklist

- [x] Input Validation
- [x] Output Encoding
- [x] Access Control


tahmidefaz:
I know this is a decent size PR, but the main changes are in `api/api_parsing`, `api/fields_parser` and `api/system_profile_host_list`.

ryandillinfelton:
Jotting down notes during walk through:
- why separate field parsing from parsing.py?
- have you made sure that the new field parsing doesn't cause unexpected behavior on filter
- Should we fall back to the DB?

Tests:
- Unit tests for new order_by (especially with weird characters that sorts differently on different platforms)
- Tests to insure new array support doesn't wack out filter payload

I'll edit this if it turns out some of these are handled when I review

tahmidefaz:
@ryandillinfelton 

> why separate field parsing from parsing.py?

I remember QE wanting to extract the fields to deep object parsing into a library in the future. That was my main reasoning behind separating it into a different file.

api/system_profile_host_list.py
```diff
@@ -0,0 +1,111 @@
...
+
+        limit, offset = pagination_params(page, per_page)
+
+        try:
+            order_by, order_how = hosts_order_by_params(order_by, order_how)
+        except ValueError as e:
+            flask.abort(400, str(e))
+
+        if fields.get("system_profile"):
+            host_ids = [{"id": {"eq": host_id}} for host_id in host_id_list]
```
ryandillinfelton, line 74: This causes a 500 if you give it any other string as the index i.e. `hosts/f{host_id_list}/system_profile?fields[foo]=bar`
since the request never gets made, and as a  result `response_list` is not defined when you try to return it.

I think you either need to some of the request logic outside of this block, or make the initial check on line 62 that decides between using the database and xjoin stricter.

ryandillinfelton (changes requested):
Most of it looks good, but there's a bug that needs fixing :(. I also think it would be good to add a test case that would catch the bug I identified to make sure it doesn't pop back up. Otherwise all the tests make sense to me (except the kinda nitpicky comment I left on one) and the coverage seems pretty good. The only other things I have are organizational/structure suggestions.

api/parsing.py
```diff
@@ -2,6 +2,8 @@
 
 from connexion.decorators.uri_parsing import OpenAPIURIParser
 
+from api.fields_parser import custom_fields_parser
```
RonnyPfannschmidt, line 5: from the code management perspective its entirely ok to put this into the parsing file for now

as its currently integrated in the deep object parsing, its not yet extract-able anyway

swagger/api.spec.yaml
```diff
@@ -238,7 +238,7 @@ paths:
         Find one or more hosts by their ID and return the id and system profile
         <br /><br />
         Required permissions: inventory:hosts:read
-      operationId: api.host.get_host_system_profile_by_id
```
RonnyPfannschmidt, line 241: the operation id change is a breaking api change

tests/helpers/api_utils.py
```diff
@@ -23,6 +23,9 @@
 HOST_URL = "/api/inventory/v1/hosts"
 TAGS_URL = "/api/inventory/v1/tags"
 SYSTEM_PROFILE_URL = "/api/inventory/v1/system_profile"
+HOST_LIST_SYSTEM_PROFILE_URL = (
+    "/api/inventory/v1/hosts/6e7b6317-0a2d-4552-a2f2-b7da0aece49d,22cd8e39-13bb-4d02-8316-84b850dc5136/system_profile"
```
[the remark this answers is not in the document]
tahmidefaz: If I can remember, I opt-out from using the `build_system_profile_url` function is because it expects host objects as the parameter value for `host_list_or_id`. Anyways, I have changed the tests to use a `minimal_host` instances :+1:

api/system_profile_host_list.py
tahmidefaz: I changed this to check for the `system_profile` key if a user provides the `fields` parameter. If it can't find `system_profile` in `fields`, we now throw a good ol' `400` on their face. 
Great catch! Thanks! :+1:

api/parsing.py
tahmidefaz: I tried to think of how we can separate it from the deep object parsing in a clean way, but I couldn't think of one. If you have any suggestions, feel free to let me know. Seems like we first have to define the parsing library before we can make it extractable.  Since it's not that extract-able atm, I have copied the function over to `api/parsing.py`.

swagger/api.spec.yaml
tahmidefaz: @RonnyPfannschmidt did you mentioned this for reference purposes?

RonnyPfannschmidt: if we have that spec deployed to prod as it was before, we cant change the name without putting in a deprecated version

ryandillinfelton (approved):

kruai (dismissed):
This looks good to me, though I'm curious about the point Ronny raised.

swagger/api.spec.yaml
kruai: @RonnyPfannschmidt Would just changing the operationId actually break anything? I thought that just pointed to the method used for that request. The new path param is optional, and the schema for the return value hasn't changed. What would break, exactly?

RonnyPfannschmidt:
the changed operation id changes method names in generated clients

api/system_profile_host_list.py
```diff
@@ -0,0 +1,114 @@
...
+
+@api_operation
+@rbac(Permission.READ)
+@metrics.api_request_time.time()
+def get_host_system_profile_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None, fields=None):
+    if fields:
+        if not fields.get("system_profile"):
+            flask.abort(400, status.HTTP_400_BAD_REQUEST)
+
+        if not xjoin_enabled():
```
thearifismail, line 67: This assumes that xjoin is always enabled.  If that's the assumption then it is not being enforced anywhere else.

api/host.py
```diff
@@ -184,25 +183,6 @@ def get_host_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=
     return flask_json_response(json_data)
 
 
-@api_operation
-@rbac(Permission.READ)
-@metrics.api_request_time.time()
-def get_host_system_profile_by_id(host_id_list, page=1, per_page=100, order_by=None, order_how=None):
```
thearifismail, line 190: `get_host_system_profile_by_id` should be placed back into `api/host.py` and redirect traffic the way `get_host_list` does

thearifismail (commented):
pytests and manual tests using postman worked.  That said, a couple of gaps were realized that have been reported in:

1. https://issues.redhat.com/browse/RHCLOUD-12387
2. https://issues.redhat.com/browse/RHCLOUD-12388

api/system_profile_host_list.py
tahmidefaz: Can you elaborate here a little bit? Yes, it should assume that xjoin is always enabled, otherwise, we throw a `5xx` on xjoin enabled functionalities.

tahmidefaz:
> pytests and manual tests using postman worked. That said, a couple of gaps were realized that have been reported in:
> 
> 1. https://issues.redhat.com/browse/RHCLOUD-12378
> 2. https://issues.redhat.com/browse/RHCLOUD-12388

@thearifismail I think you linked the wrong ticket for no. 1. Please update.

api/system_profile_host_list.py
thearifismail: Since it is an openAPI, it is a contract that HBI announced and should not change endpoints without proper announcements.

I think the request should come to `api/host/get_host_system_profile_by_id`, then it should check if `x-rh-cloud-bulk-query-source == xjoin`, then it should send to xjoin else, it should serve the request itself.

tahmidefaz: @thearifismail I am moving this to `api/host`. 

Question: If `xjoin` is down, and we receive a sparse fields request, how would HBI serve this request by itself?

ryandillinfelton: If xjoin is down and we get a request for system profile that uses the spare feildset query responding with a **500** is correct because part of our service is compromised. It's the same thing we do the the `/tags` endpoint which relies on xjoin and has no db equivalent. If the user specifies db as the data-source in the header of their request to `/tags` HBI responds with a **503**. If xjoin-search is down it responds with a plain **500** because it needs xjoin-search up to serve the response.

I believe what is currently set up in this PR is that queries that don't specify a sparse fieldset are fulfilled the same way they were before, directly querying the database. However if the user asks for a sparse fieldset we fulfill with xjoin. I don't see much point in duplicating the sparse fieldset functionality for the database since every environment defaults to xjoin as the data source, and we should expect xjoin to have reliable uptime.

kruai:
> pytests and manual tests using postman worked. That said, a couple of gaps were realized that have been reported in:
> 
> 1. https://issues.redhat.com/browse/RHCLOUD-12378
> 2. https://issues.redhat.com/browse/RHCLOUD-12388

@thearifismail For the first Jira, did you mean [this one](https://issues.redhat.com/browse/RHCLOUD-12387)? (CC @tahmidefaz )

kruai (approved):
Still LGTM! I think it's a good call to keep the same operationID, and just have the functionality not change if the `fields` param isn't provided.

api/system_profile_host_list.py
thearifismail: Agree after hearing the rationale this morning that `xjoin` is a MUST WORK service.

thearifismail (approved):
BTW for specifying a bogus parameter, the following was returned:

tahmidefaz:
@thearifismail 
> Question: Should it show a warning/error that the requested field does not exist and check the name or spelling?

Eventually, we would throw a `400` if a requested field does not exist in the schema, but as a first approach, we decided not to pursue it as part of this pr.

RonnyPfannschmidt:
jsut as a fyi - https://issues.redhat.com/browse/RHCLOUD-12512 tracks the throwing error on invalid fields

PR #810, opened 2021-02-16
Add per_reporter_staleness tracking [RHC-12221]

ryandillinfelton:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-12221).

This PR is to add a new column to the database for tracking when individual reporters last checked in a host. This PR does not contain any of the filtering work that this data will be used for. All this is/should be is a migration for adding the new jsonb column to the db, changes to host updating and creation to insure we are setting the values in this new column, and tests to make sure that's happening correctly.

The design document for the data formatting (and it's relation to the filters we will add later) is [here](https://docs.google.com/document/d/107ljikJHs5-mj7n2eUt3dIUlRO6R9_VQi3tssqGTM9Y/edit?usp=sharing).

I took some liberties with the final names, they're a little longer, but I think that helps avoid ambiguity. Let me know what you think when you review.

:raccoon: 

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -383,67 +383,6 @@ class NetworkInterfaceSchema(MarshmallowSchema):
     type = fields.Str(validate=marshmallow_validate.Length(max=18))
 
 
-class SystemProfileSchema(MarshmallowSchema):
-    owner_id = fields.Str(validate=verify_uuid_format)
-    rhc_client_id = fields.Str(validate=verify_uuid_format)
-    cpu_model = fields.Str(validate=marshmallow_validate.Length(max=100))
-    number_of_cpus = fields.Int()
-    number_of_sockets = fields.Int()
-    cores_per_socket = fields.Int()
-    system_memory_bytes = fields.Int()
-    infrastructure_type = fields.Str(validate=marshmallow_validate.Length(max=100))
-    infrastructure_vendor = fields.Str(validate=marshmallow_validate.Length(max=100))
-    network_interfaces = fields.List(fields.Nested(NetworkInterfaceSchema()))
-    disk_devices = fields.List(fields.Nested(DiskDeviceSchema()))
-    bios_vendor = fields.Str(validate=marshmallow_validate.Length(max=100))
-    bios_version = fields.Str(validate=marshmallow_validate.Length(max=100))
-    bios_release_date = fields.Str(validate=marshmallow_validate.Length(max=50))
-    cpu_flags = fields.List(fields.Str(validate=marshmallow_validate.Length(max=30)))
-    operating_system = fields.Nested(OperatingSystemSchema())
-    os_release = fields.Str(validate=marshmallow_validate.Length(max=100))
-    os_kernel_version = fields.Str(
-        validate=[marshmallow_validate.Length(max=20), marshmallow_validate.Regexp(regex=r"^\d+\.\d+\.\d+(\.\d+)?$")]
-    )
-    arch = fields.Str(validate=marshmallow_validate.Length(max=50))
-    kernel_modules = fields.List(fields.Str(validate=marshmallow_validate.Length(max=255)))
-    last_boot_time = fields.Str(validate=marshmallow_validate.Length(max=50))
-    running_processes = fields.List(fields.Str(validate=marshmallow_validate.Length(max=1000)))
-    subscription_status = fields.Str(validate=marshmallow_validate.Length(max=100))
-    subscription_auto_attach = fields.Str(validate=marshmallow_validate.Length(max=100))
-    katello_agent_running = fields.Bool()
-    satellite_managed = fields.Bool()
-    cloud_provider = fields.Str(validate=marshmallow_validate.Length(max=100))
-    rhsm = fields.Nested(RhsmSchema())
-    yum_repos = fields.List(fields.Nested(YumRepoSchema()))
-    dnf_modules = fields.List(fields.Nested(DnfModuleSchema()))
-    installed_products = fields.List(fields.Nested(InstalledProductSchema()))
-    insights_client_version = fields.Str(validate=marshmallow_validate.Length(max=50))
-    insights_egg_version = fields.Str(validate=marshmallow_validate.Length(max=50))
-    captured_date = fields.Str(validate=marshmallow_validate.Length(max=32))
-    installed_packages = fields.List(fields.Str(validate=marshmallow_validate.Length(max=512)))
-    installed_packages_delta = fields.List(fields.Str(validate=marshmallow_validate.Length(max=512)))
-    installed_services = fields.List(fields.Str(validate=marshmallow_validate.Length(max=512)))
-    enabled_services = fields.List(fields.Str(validate=marshmallow_validate.Length(max=512)))
-    sap_system = fields.Bool()
-    sap_sids = fields.List(
-        fields.Str(
-            validate=[marshmallow_validate.Length(max=3), marshmallow_validate.Regexp(regex=r"^[A-Z][A-Z0-9]{2}$")]
-        )
-    )
-    sap_instance_number = fields.Str(
-        validate=[marshmallow_validate.Length(max=2), marshmallow_validate.Regexp(regex=r"^[0-9]{2}$")]
-    )
-    sap_version = fields.Str(
-        validate=[
-            marshmallow_validate.Length(max=22),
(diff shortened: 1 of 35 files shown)


migrations/versions/3d06b0a8828f_add_per_reporter_staleness_column.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "3d06b0a8828f"
+down_revision = "33e0aca8516f"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column("hosts", sa.Column("per_reporter_staleness", postgresql.JSONB(astext_type=sa.Text()), nullable=True))
```
RonnyPfannschmidt, line 20: should we consider this as non-nullable with a default of a empty mapping ? 
that could ease some of the logic

app/models.py
```diff
@@ -258,6 +262,17 @@ def _update_stale_timestamp(self, stale_timestamp, reporter):
...
         self.reporter = reporter
 
+    def _update_per_reporter_staleness(self, stale_timestamp, reporter):
+        if self.per_reporter_staleness is None:
+            self.per_reporter_staleness = {}
+
+        if not self.per_reporter_staleness.get(reporter):
+            self.per_reporter_staleness[reporter] = {}
+
+        self.per_reporter_staleness[reporter]["stale_timestamp"] = stale_timestamp.isoformat()
```
RonnyPfannschmidt, line 272: this needs extra tooling inform sqlalchemy of the field update

`orm.attributes.flag_modified(self, "per_reporter_staleness")` at the end should do the trick

as we ensure `self.per_reporter_staleness[reporter]` is a mapping, i suggest to use a compact dict.update call

```python
self.per_reporter_staleness[reporter].update(
        stale_timestamp=stale_timestamp.isoformat(),
        last_check_in=datetime.now(timezone.utc).isoformat(),
        check_in_succeeded=True,
    )

File: tests/test_unit.py [hunk 3]
```diff
@@ -1579,16 +1594,6 @@ class HostUpdateStaleTimestamp(TestCase):
     def _make_host(self, **values):
         return Host(**{"canonical_facts": {"fqdn": "some fqdn"}, **values})
 
-    def test_updated_when_empty(self):
```
ryandillinfelton, line 1582: We can't create hosts without reporter or staleness timestamp anymore so I think this test is pretty pointless, and basically impossible to actually do.

migrations/versions/3d06b0a8828f_add_per_reporter_staleness_column.py
```diff
@@ -0,0 +1,26 @@
...
+    op.execute("UPDATE hosts SET per_reporter_staleness = '{}'::jsonb WHERE per_reporter_staleness is NULL;")
```
RonnyPfannschmidt, line 21: i believe its possible to use `text("'{}'::jsonb")` as a default

is there a structural difference between adding a null column + update and adding a default value?

ryandillinfelton: Nope! switched it over. I did however find that, while it will insure the the field is initialized for all the hosts in the db which I expect will be good for filtering queries based on the field, we still need the logic initializing the field in models.

ryandillinfelton:
Tests should now be passing :frog:

tests/test_models.py
```diff
@@ -321,11 +337,89 @@ def test_host_model_timestamp_timezones(db_create_host):
...
 
     host = Host(**values)
 
     with pytest.raises(DataError):
         db_create_host(host)
+
+
+def test_create_host_sets_per_reporter_staleness(db_create_host, models_datetime_mock):
+    stale_timestamp = models_datetime_mock + timedelta(days=1)
+    # per_reporter_timestamp = {"yupana": {"last_check_in": now(), "stale_timestamp": n}}
```
kruai, line 358: Commented-out code should be removed

tests/test_models.py
```diff
@@ -321,11 +337,89 @@ def test_host_model_timestamp_timezones(db_create_host):
...
+        "puptoo": {
+            "last_check_in": models_datetime_mock.isoformat(),
+            "stale_timestamp": stale_timestamp.isoformat(),
+            "check_in_succeeded": True,
+        }
+    }
+
+
+def test_update_per_reporter_staleness(db_create_host, models_datetime_mock):
+    # TODO: finish this test
```
kruai, line 375: Is it finished? If so, please remove this comment

kruai (changes requested):
I just have a couple notes on this one

tests/test_models.py
ryandillinfelton: yep! Just pushed a commit removing the things you pointed out :+1:

kruai (approved):
LGTM!

thearifismail (commented):
Needs a couple of tests with a host reported by more than one reporter and with different stalessness times

migrations/versions/3d06b0a8828f_add_per_reporter_staleness_column.py
```diff
@@ -0,0 +1,33 @@
...
+
+def upgrade():
+    op.add_column(
+        "hosts",
+        sa.Column(
+            "per_reporter_staleness",
+            postgresql.JSONB(astext_type=sa.Text()),
+            nullable=False,
+            default=dict,
+            server_default="{}",
```
RonnyPfannschmidt, line 27: i vaguely recall requiring a text element here wit the server side value of the data

tahmidefaz: @ryandillinfelton is this autogenerated?

ryandillinfelton: Not auto generated. Yeah, Ronny is right. Not sure what happened that it changed. I"ll change it back.

tahmidefaz:
Apart from the concern raised by Ronny, other changes look good to me :+1:

ryandillinfelton:
> Needs a couple of tests with a host reported by more than one reporter and with different stalessness times

That's what the test I added does. It reports twice with puptoo to test overwriting, then once with yupana to insure multiple reporters work, and a new reporter does not alter existing per_reporter data.

https://github.com/RedHatInsights/insights-host-inventory/pull/810/files#diff-0adb667cd397bd56edce10eec11e1b10821740a10d72bd148b09645fc9108968R412

migrations/versions/3d06b0a8828f_add_per_reporter_staleness_column.py
ryandillinfelton: wait, it's the same thing it was just collapsed. I'm not sure I follow @RonnyPfannschmidt

RonnyPfannschmidt: @ryandillinfelton i do recall that the jsonb form for a default would have to be used in a text element, im now crosschecking the docs, i may have missed a feature addition to make it simpler

RonnyPfannschmidt: it seems sqlalchemy handles the details

ryandillinfelton: Okay. So all good?

app/models.py
```diff
@@ -158,14 +159,15 @@ def __init__(
         system_profile_facts=None,
         stale_timestamp=None,
         reporter=None,
+        per_reporter_staleness=None,
```
RonnyPfannschmidt, line 162: im wondering, should this perhaps not be passed in `__init__` as a parameter and instead always be set to a new empty-dict (i recall that instance reconstruction in sqlalchemy skips the ctor)

RonnyPfannschmidt: (this is fine as followup)

kruai: @RonnyPfannschmidt would having it be set to `None` cause issues? And if so, should we also be doing that for `system_profile_facts`?

ryandillinfelton: I think the correct thing to do is just remove it since the values in per_reporter_staleness should always and only be determined by the values of `reporter` and `stale_timestamp` to avoid discrepancies.

thearifismail (approved):
LGTM

PR #815, opened 2021-02-18
Remove redundant System Profile validation; consolidate host schemas

kruai:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-12054).
This is a code-hardening PR meant to simplify our schema naming and usage. It also removes redundant marshmallow validation of System Profile, as it is already validated against the OpenAPI spec using the `SystemProfileNormalizer` method in the `HostSchema` class.

## Secure Coding Checklist

### DIFF
--- a/app/queue/events.py
+++ b/app/queue/events.py
@@ -8,7 +8,6 @@ from marshmallow import fields
 from marshmallow import Schema
 
 from app.logging import threadctx
-from app.models import SystemProfileSchema
 from app.models import TagsSchema
 from app.queue.metrics import event_serialization_time
 from app.serialization import serialize_canonical_facts
@@ -49,7 +48,7 @@ class HostSchema(Schema):
     culled_timestamp = fields.Str()
     reporter = fields.Str()
     tags = fields.List(fields.Nested(TagsSchema))
-    system_profile = fields.Nested(SystemProfileSchema)
+    system_profile = fields.Dict()
 
 
 class HostEventMetadataSchema(Schema):
--- a/app/queue/events.py
+++ b/app/queue/events.py
@@ -23,7 +23,7 @@ def hostname():
 
 
 # Schemas
-class HostSchema(Schema):
+class SerializedHostSchema(Schema):
     id = fields.UUID()
     display_name = fields.Str()
     ansible_host = fields.Str()
@@ -37,10 +37,6 @@ class HostSchema(Schema):
     ip_addresses = fields.List(fields.Str())
     mac_addresses = fields.List(fields.Str())
     external_id = fields.Str()
-    # FIXME:
-    # created = fields.DateTime(format="iso8601")
-    # updated = fields.DateTime(format="iso8601")
-    # FIXME:
     created = fields.Str()
     updated = fields.Str()
     stale_timestamp = fields.Str()
@@ -57,7 +53,7 @@ class HostEventMetadataSchema(Schema):
 
 class HostCreateUpdateEvent(Schema):
     type = fields.Str()
-    host = fields.Nested(HostSchema())
+    host = fields.Nested(SerializedHostSchema())
     timestamp = fields.DateTime(format="iso8601")
     platform_metadata = fields.Dict()
     metadata = fields.Nested(HostEventMetadataSchema())
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -20,6 +20,7 @@ from kafka.errors import KafkaError
 from api import api_operation
 from api.host_query_db import _order_how
 from api.host_query_db import params_to_order_by
+from api.parsing import custom_fields_parser
 from app import create_app
 from app.auth.identity import from_auth_header
 from app.auth.identity import from_bearer_token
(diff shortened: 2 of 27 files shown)


app/queue/events.py
```diff
@@ -24,7 +23,7 @@ def hostname():
 
 
 # Schemas
-class HostSchema(Schema):
+class SerializedHostSchema(Schema):
```
kruai, line 26: Do we actually need this schema definition? It's really just a schema that validates the serialization of the Host object, which already has a schema...

thearifismail: Please help me understand the issue if I am off.  I did a bit of digging to understand your question.  I see the host is validated before adding it to inventory.  Once saved the some event is produced and the host is serialized again for adding the host to an event.  I guess it is good to validate the host again to avoid accidental modifications before adding it to an event

thearifismail (dismissed):
LGTM

kruai:
@thearifismail Yeah, that's pretty much what I was talking about. We had 3 different Host schemas, and I was hoping I could consolidate them into 1. I combined the HTTP and MQ host schemas, but couldn't figure out how to combine it with the one in `events.py`.

As far as I understand, this schema has to have the exact same structure as the other host schemas, except every field has to be a string because it's serialized. I'd love to get rid of this schema because it feels like duplication, and it also seems kinda redundant since we already validate the Host schema before we serialize it... but I understand if we should keep it so that we're doubly sure the events we produce have the exact right format.

app/queue/events.py
kruai: @thearifismail Yeah, that's pretty much what I was talking about. We had 3 different Host schemas, and I was hoping I could consolidate them into 1. I combined the HTTP and MQ host schemas, but couldn't figure out how to combine it with the one in events.py.

As far as I understand, this schema has to have the exact same structure as the other host schemas, except every field has to be a string because it's serialized. I'd love to get rid of this schema because it feels like duplication, and it also seems kinda redundant since we already validate the Host schema before we serialize it... but I understand if we should keep it so that we're doubly sure the events we produce have the exact right format.

ryandillinfelton:
>Please help me understand the issue if I am off. I did a bit of digging to understand your question. I see the host is validated before adding it to inventory. Once saved the some event is produced and the host is serialized again for adding the host to an event. I guess it is good to validate the host again to avoid accidental modifications before adding it to an event

Yeah I think that's it. I don't mind the redundancy too much, but I'm partial to removing it. It's another thing we have to update when we add new fields or adjust the data we send in messages which is inconvenient. If we're worried about sending malformed messages I think the better approach is to make sure we have enough test covering it, rather than checking every single message.

utils/kafka_consumer.py
```diff
@@ -17,11 +14,6 @@ def msg_handler(parsed):
     print("inside msg_handler()")
     print("type(parsed):", type(parsed))
     print("parsed:", parsed)
-    # id_ = parsed["id"]
```
ryandillinfelton, line 20: Did you mean to commit these changes? I figured we kept these commented out to make testing easier if you want to change the consumer's behavior quickly?

app/queue/events.py
ryandillinfelton: Yeah I think that's it. I don't mind the redundancy too much, but I'm partial to removing it. It's another thing we have to update when we add new fields or adjust the data we send in messages which is inconvenient. If we're worried about sending malformed messages I think the better approach is to make sure we have enough test covering it, rather than checking every single message.

utils/kafka_consumer.py
kruai: Oh, I didn't realize that. I removed them because in general we shouldn't have commented-out code, but it's a dev util so it doesn't really matter. I'll revert it

app/queue/events.py
kruai: I agree, though having a schema like this is probably the simplest way to test it as well. I think that for now, we should stick with renaming the schema for the sake of clarity, and come back to this later if we decide we want to change our approach to message validation (since it's kinda out-of-scope for this PR)

thearifismail: How about removing the HostSchema from `app/queue/events.py` and creating a  `serialize_host_mq()` like `deserialize_host_mq().  This way there is a single source of truth.  IMO, `ensuring we have enough tests` are bound to get forgotten in the heat of moment.

ryandillinfelton: I agree with Asa, I think we may as well just leave it as it is now and create Jira to follow up and at a later date to remove the schema and make sure all the same bases are covered by test and create any tests that are missing.

tests/test_unit.py
```diff
@@ -853,28 +849,12 @@ def test_with_tags(self):
...
                 "tags": tags,
-            },
-            MqHostSchema,
+            }
         )
 
         self.assertIs(Host, type(host))
         self.assertEqual(tags, host.tags)
 
-    def test_without_tags(self):
```
ryandillinfelton, line 863: Why was this test removed?

ryandillinfelton: Oh, I suppose it only applies to the HTTP schema

ryandillinfelton (approved):

tests/test_unit.py
kruai: I was going off of our discussion [here](https://ansible.slack.com/archives/CQFKM031T/p1613597215163200), where it sounded like we could remove the tests that specifically validated hosts added via HTTP. This one, for instance, is validating that the tags are removed for just such a host, right?

tahmidefaz (approved):
Looks good! Also did a local test to verify that once these changes are merged in, we only need to update the system_profile schema yaml when adding/updating fields in system_profile :tada:

tests/test_unit.py
ryandillinfelton: oh, I see when I deleted the comment from my review it was removed here too. Yeah since it's just related to testing the HTTP schema it's fine to remove.

PR #816, opened 2021-02-19
Fix bonfire breaking changes

BlakeHolifield:
## Overview

Bonfire has new features and improvments that break syntax. To solve this and add some quality of life fixes, I've updated the pr_check to do the following:

* Break out unit tests to a separate file
* Install bonfire via pip and pin to a known working version for the syntax we are using
* Update the bonfire commands to use new syntax
* Wait for all deployments to come up using bonfire -- this times out after 5 minutes and the PR check will exit after. No more running 700 tests that all fail :)

There are a few more improvements coming, like jobs and smoke testing. This should get your PRs in a better place for ephemeral envs and remain stable until we add more features later on.


thearifismail (approved):

unit_test.sh
```diff
@@ -0,0 +1,105 @@
...
+
+oc port-forward svc/${APP_NAME}-db $RANDOM_PORT:5432 &
+BG_PID=$!
+trap "killbg" EXIT ERR SIGINT SIGKILL TERM
+trap "nsrelease" EXIT ERR SIGINT SIGKILL TERM
+
+echo "DB migration"
+python manage.py db upgrade
+echo "Running pytest"
+pytest --cov=. --junitxml=junit.xml --cov-report html -sv
```
kruai, line 101: I realize this was originally in pr_check.sh, but is naming this after JUnit intentional? Seems weird since we're using pytest

unit_test.sh
```diff
@@ -0,0 +1,105 @@
...
+}
+
+function random_unused_port { local port=$(shuf -i 2000-65000 -n 1)
+    netstat -lat | grep $port > /dev/null
+    if [[ $? == 1 ]] ; then
+        RANDOM_PORT=$port
+    else
+        random_unused_port
+    fi
+}
```
kruai, line 41: This function's unused. Is this being added for later use?

unit_test.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+
+APP_NAME="host-inventory"  # name of app-sre "application" folder this component lives in
+IMAGE="quay.io/cloudservices/insights-inventory"
+BG_PID=1010101
+RANDOM_PORT=62212
```
kruai, line 6: This variable is defined in both pr_check.sh and unit_test.sh. It doesn't appear to be used in pr_check.sh.

pr_check.sh
```diff
@@ -9,104 +9,12 @@ BG_PID=1010101
 RANDOM_PORT=65000
 export LC_ALL=en_US.utf-8
 export LANG=en_US.utf-8
-
+export APP_ROOT=$(pwd)
+export WORKSPACE=${WORKSPACE:-$APP_ROOT}  # if running in jenkins, use the build's workspace
+export IMAGE_TAG=$(git rev-parse --short=7 HEAD)
+export GIT_COMMIT=$(git rev-parse HEAD)
```
kruai, line 15: These variables are either used in unit_test.sh and not this one (`APP_ROOT`/`WORKSPACE`, `IMAGE_TAG`), or not used at all (`GIT_COMMIT`). Can `APP_ROOT`, `WORKSPACE` and `IMAGE_TAG` be defined in unit_test.sh instead, for the sake of readability? Or are they also used in deploy_ephemeral_env.sh?

unit_test.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+
+APP_NAME="host-inventory"  # name of app-sre "application" folder this component lives in
+IMAGE="quay.io/cloudservices/insights-inventory"
+BG_PID=1010101
```
kruai, line 5: Since you're redefining this variable here, you'll want to delete the one in pr_check.sh

kruai (changes requested):
Had a few comments/questions

unit_test.sh
BlakeHolifield: I'm using what was in the Jenkinsfile in master: https://github.com/RedHatInsights/insights-host-inventory/blob/9d076a1f5a8194ffab956da55179cbb3bd765096/Jenkinsfile#L78

We can for sure change it if you'd like though!

unit_test.sh
BlakeHolifield: Nope that needs to be deleted. Good catch

unit_test.sh
BlakeHolifield: Another good catch. Cleaning up.

pr_check.sh
BlakeHolifield: Yep! The `export` vars are being using inside `deploy_ephemeral_env.sh` and `unit_test.sh`

unit_test.sh
BlakeHolifield: Deleting.

unit_test.sh
kruai: Oh... if we already had that, I guess it's fine. It feels wrong, but I really just wanted to make sure we weren't putting new incorrect naming into the script!

unit_test.sh
```diff
@@ -0,0 +1,94 @@
...
+# Wait on the App to be Ready. Do not run unit tests if you do not become Ready.
+bonfire namespace wait-on-resources $NAMESPACE || { echo 'App did not deploy properly' ; exit 1; }
+
+#
+# Grab DB creds
+#
+
+# Get a random port on the jenkins agent to forward
+echo "Getting a port to forward"
+# random_unused_port
```
kruai, line 70: This block should probably be deleted now

kruai (changes requested):
I'd like one more change but then I think it's good to go :+1:

unit_test.sh
BlakeHolifield: Doh! Fixed now.

kruai (approved):
Cool, looks good now :+1:

thearifismail (approved):

PR #820, opened 2021-02-23
Custom regex escape

tahmidefaz:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-10860).
This PR provides a custom regex escape method in HBI.

## Secure Coding Checklist

- [x] Input Validation
- [x] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


kruai (approved):
I think this looks good, but I just want to make sure... we only need to modify the `re.escape()` calls, right? I would assume we don't need to worry about our calls to `re.findall()`, `re.match()`, and `re.fullmatch()`

tahmidefaz:
> I think this looks good, but I just want to make sure... we only need to modify the `re.escape()` calls, right? I would assume we don't need to worry about our calls to `re.findall()`, `re.match()`, and `re.fullmatch()`

Yes, that's correct! Just the `re.escape()` calls

tests/test_unit.py
```diff
@@ -1984,5 +1985,18 @@ def test_custom_fields_parser(self):
...
+class CustomRegexMethodTestCase(TestCase):
+    def test_custom_regex_escape(self):
+        for regex_input, output in (
+            (".?well+", "\\.\\?well\\+"),
+            ("&[^abc]~", "\\&\\[^abc\\]\\~"),
+            ("some|*#thing", "some\\|\\*\\#thing"),
+            ('.?+*|{}[]()"\\#@&<>~', '\\.\\?\\+\\*\\|\\{\\}\\[\\]\\(\\)\\"\\\\\\#\\@\\&\\<\\>\\~'),
+            ("\\", "\\\\"),
+        ):
+            result = custom_escape(regex_input)
```
ryandillinfelton, line 1997: This works fine as is, but i think the cause of the test failure is easier to track down when you wrap each assert in a subtest. It gives the the test case it failed on in the header which I find to be helpful.
```suggestion
        with self.subTest(regex_input=regex_input):
            result = custom_escape(regex_input)
```

ryandillinfelton (dismissed):
Looks good. I'm cool with merging it as is but left a simple suggestion for a small test improvement. LMK if you choose to use the suggestion and I can re-approve quick :t-rex:

tests/test_unit.py
tahmidefaz: done! thanks! :+1:

ryandillinfelton (approved):
Looks :goat: to me!

thearifismail (approved):
LGTM.

PR #825, opened 2021-03-08
ESSNTL-746: prepare a alternate normalized openapi spec

RonnyPfannschmidt:
this helps the codegen and api compat

in this pr we implement bundling of the api specs using a alternate bunlder to avoid the bugs in the native bundlers

related https://issues.redhat.com/browse/RHCLOUD-11860

also related 
* https://github.com/jfinkhaeuser/prance/pull/82
* https://github.com/jfinkhaeuser/prance/issues/76
* https://github.com/jfinkhaeuser/prance/issues/77

- [x] verify generated code in the qe side on python
- [x] integrate the new bundled spec in serving data
- [x] feedback from @Glutexo

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -36,7 +36,9 @@ IDENTITY_HEADER = "x-rh-identity"
 REQUEST_ID_HEADER = "x-rh-insights-request-id"
 UNKNOWN_REQUEST_ID_VALUE = "-1"
 
-SPECIFICATION_FILE = join(SPECIFICATION_DIR, "api.spec.yaml")
+# temporary replaced for RHCLOUD-12800 - correct normalization of the bundled spec
+# SPECIFICATION_FILE = join(SPECIFICATION_DIR, "api.spec.yaml")
+SPECIFICATION_FILE = join(SPECIFICATION_DIR, "openapi.json")
 SYSTEM_PROFILE_SPECIFICATION_FILE = join(SPECIFICATION_DIR, "system_profile.spec.yaml")
 SYSTEM_PROFILE_BLOCK_LIST_FILE = join(SPECIFICATION_DIR, "system_profile_block_list.yaml")
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -36,7 +36,7 @@ IDENTITY_HEADER = "x-rh-identity"
 REQUEST_ID_HEADER = "x-rh-insights-request-id"
 UNKNOWN_REQUEST_ID_VALUE = "-1"
 
-# temporary replaced for RHCLOUD-12800 - correct normalization of the bundled spec
+# temporary replaced for ESSNTL-746 - correct normalization of the bundled spec
 # SPECIFICATION_FILE = join(SPECIFICATION_DIR, "api.spec.yaml")
 SPECIFICATION_FILE = join(SPECIFICATION_DIR, "openapi.json")
 SYSTEM_PROFILE_SPECIFICATION_FILE = join(SPECIFICATION_DIR, "system_profile.spec.yaml")
--- a/check_schemas.py
+++ b/check_schemas.py
@@ -1,0 +1,23 @@
+import json
+import sys
+from difflib import unified_diff
+
+import yaml
+
+from app.models import SPECIFICATION_DIR
+from lib.translating_parser.parser import TranslatingParser
+
+OAPI = SPECIFICATION_DIR + "openapi.json"
+
+with open(OAPI) as fp:
+    oapi_data = yaml.dump(json.load(fp))
+
+openapi = TranslatingParser(OAPI)
+openapi.parse()
+
+# spec = TranslatingParser(SPECIFICATION_DIR + "api.spec.yaml")
+# spec.parse()
+
+
+cdiff = unified_diff(oapi_data.splitlines(True), openapi.yaml().splitlines(True), fromfile="oapi", tofile="prance")
+sys.stdout.writelines(cdiff)
--- a/check_schemas.py
+++ b/check_schemas.py
@@ -12,12 +12,16 @@ OAPI = SPECIFICATION_DIR + "openapi.json"
 with open(OAPI) as fp:
     oapi_data = yaml.dump(json.load(fp))
 
-openapi = TranslatingParser(OAPI)
-openapi.parse()
 
-# spec = TranslatingParser(SPECIFICATION_DIR + "api.spec.yaml")
-# spec.parse()
+def make_diff(specfile, output_filename):
+    parser = TranslatingParser(SPECIFICATION_DIR + specfile)
+    cdiff = unified_diff(
+        oapi_data.splitlines(True), parser.yaml().splitlines(True), fromfile="oapi", tofile="output_filename"
+    )
+    sys.stdout.writelines(cdiff)
 
 
-cdiff = unified_diff(oapi_data.splitlines(True), openapi.yaml().splitlines(True), fromfile="oapi", tofile="prance")
-sys.stdout.writelines(cdiff)
+if __name__ == "__main__":
+    make_diff("openapi.json", "prace_reads_bundle")
+    # todo: reenable this once the prance bug about inclusion is resolved
+    # make_diff("api.spec.yaml", "prace_reads_spec")
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -607,8 +607,9 @@ components:
       style: deepObject
       explode: true
       schema:
-        type: object
-        description: System profile field values to filter with
+        $ref:  '#/components/schemas/SystemProfileNestedObject'
+       # type: object
+       # description: System profile field values to filter with
     fields_param:
       in: query
       name: fields
@@ -617,8 +618,9 @@ components:
       style: deepObject
       explode: true
       schema:
-        type: object
-        description: System profile field values to fetch
+        $ref:  '#/components/schemas/SystemProfileNestedObject'
+        # type: object
+        # description: System profile field values to fetch
 
   schemas:
     SystemProfileSapSystemOut:
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -608,8 +608,6 @@ components:
       explode: true
       schema:
         $ref:  '#/components/schemas/SystemProfileNestedObject'
-       # type: object
-       # description: System profile field values to filter with
     fields_param:
       in: query
       name: fields
@@ -619,8 +617,6 @@ components:
       explode: true
       schema:
         $ref:  '#/components/schemas/SystemProfileNestedObject'
-        # type: object
-        # description: System profile field values to fetch
 
   schemas:
     SystemProfileSapSystemOut:
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -707,8 +707,6 @@ components:
       type: array
       items:
         type: string
-        nullable: False
-      nullable: false
     InsightsId:
       description: >-
         An ID defined in /etc/insights-client/machine-id. This field is
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -745,6 +745,7 @@ components:
       type: array
       items:
         type: string
+        nullable: False
       nullable: false
     InsightsId:
       description: >-
@@ -1142,6 +1143,20 @@ components:
     PerPage:
       type: integer
       description: The number of items to return per page
+    SystemProfile:
+      $ref: 'system_profile.spec.yaml#/$defs/SystemProfile'
+    SystemProfileNestedObject:
+      $ref: 'system_profile.spec.yaml#/$defs/NestedObject'
+    SystemProfileDiskDevice:
+      $ref: 'system_profile.spec.yaml#/$defs/DiskDevice'
+    SystemProfileYumRepo:
+      $ref: 'system_profile.spec.yaml#/$defs/YumRepo'
+    SystemProfileDnfModule:
+      $ref: 'system_profile.spec.yaml#/$defs/DnfModule'
+    SystemProfileInstalledProduct:
+      $ref: 'system_profile.spec.yaml#/$defs/InstalledProduct'
+    SystemProfileNetworkInterface:
+      $ref: 'system_profile.spec.yaml#/$defs/NetworkInterface'
   responses:
     PageOutOfBounds:
       description: Requested page is outside of the range of available pages
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -466,7 +466,6 @@ paths:
         '403':
             description: Forbidden
 
-
 components:
   securitySchemes:
     BearerAuth:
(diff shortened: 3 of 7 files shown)


swagger/api.spec.yaml
```diff
@@ -1096,6 +1072,18 @@ components:
...
+    SPF_DiskDevice:
+      $ref: 'system_profile.spec.yaml#/$defs/DiskDevice'
+    SPF_YumRepo:
+      $ref: 'system_profile.spec.yaml#/$defs/YumRepo'
+    SPF_DnfModule:
+      $ref: 'system_profile.spec.yaml#/$defs/DnfModule'
+    SPF_InstalledProduct:
+      $ref: 'system_profile.spec.yaml#/$defs/InstalledProduct'
+    SPF_NetworkInterface:
+      $ref: 'system_profile.spec.yaml#/$defs/NetworkInterface'
```
Glutexo, line 1086: These objects are not referenced from anywhere in _api.spec.yaml_.

If you need them to nudge your code generator in the right direction, can’t you add them programmatically?

Glutexo (approved):
Wouldn’t it be possible to add the helper schema references using a Python code somewhere between loading the file and processing the schema, instead of adding them to the API file and then storing somehow compiled version?

Glutexo:
Didn’t intend to approve. I wanted to just comment.

swagger/openapi.json
```diff
@@ -0,0 +1,2086 @@
+{
```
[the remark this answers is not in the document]
RonnyPfannschmidt: the unbundling lib we currently use has bugs and jsut went unmaintained, we have to figure a solution

this is to manually manage unbundleing with correct references

swagger/api.spec.yaml
RonnyPfannschmidt: this pretty much nudges a standard compliant tool/verification tool-chain into the right direction,
as a intermediate solution i would consider that better than a custom python hack, i haven't figured yet how to do the de-ref correctly - when i made a script, the first iteration had some really dumb mistakes, so this helps to go to working

those names pretty much prevent in-lining of the sub-types used in the system profile spec

swagger/api.spec.yaml
```diff
@@ -716,6 +712,7 @@ components:
       type: array
       items:
         type: string
+        nullable: False
       nullable: false
```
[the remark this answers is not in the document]
RonnyPfannschmidt: this was added by the verification tool i added to pre-commit, its perhaps due to the surrounding null-able
im crosschecking now

swagger/api.spec.yaml
```diff
@@ -843,7 +840,7 @@ components:
           $ref: "#/components/schemas/BiosUuid"
         ip_addresses:
           $ref: "#/components/schemas/IpAddresses"
-        Fqdn:
+        fqdn:
```
kruai, line 843: FYI, this was fixed by #826. You probably need to update this branch with master

RonnyPfannschmidt: @kruai fixed, we should have a little discuss on the implication of the nested validation removal, but it practically was never implemented as far as i understood

kruai (commented):
I'm still a bit unclear on the direction this will take us in. My main concern is regarding `swagger/openapi.json`: is it a requirement for us to switch our API spec from yaml to json?

RonnyPfannschmidt:
@kruai no, the json is a autogenerated and pre-commit enforced de-bundle while we resolve the issue with broken and unmaintained prasec

We will work with the yaml

swagger/system_profile.spec.yaml
```diff
@@ -9,11 +9,8 @@ $defs:
     # to validate x-propertyNames as propertyNames
     x-propertyNames:
       minLength: 1
-    additionalProperties:
-      oneOf:
-        - "$ref": "#/$defs/NestedObject"
-        - not:
-            type: object
+    additionalProperties: {} # todo bring back nested validation
+
```
tahmidefaz, line 13: Question: How are we handling this `NestedObject`? By the current schema definition, `NestedObject` can have an unspecified number of keys and each key definition must have a length of at least 1. A key can also be other `NestedObjects`, and so on.

RonnyPfannschmidt: As far as I understood, the incorrect unbundleing broke this , I'll add a unittest that demonstrate the behaviour later

kruai (dismissed):
Based on our discussion... this makes sense for the interim until the Prance issue is resolved, even though I don't like the duplication. Do you have a Jira card to track the follow-up work to remove the JSON, etc.?

swagger/openapi.json
```diff
@@ -0,0 +1,2062 @@
...
+        }
+      },
+      "SPF_NestedObject": {
+        "type": "object",
+        "description": "An arbitrary object that does not allow empty string keys.",
+        "x-propertyNames": {
+          "minLength": 1
+        },
+        "additionalProperties": {}
+      },
```
tahmidefaz, line 1904: Correct me if I am wrong but I think this means that it will fail on something like:
`{"disk_device":{"options": {"": "something"}}}`

But, it will pass on something like:
`{"disk_device": {"options": {"key": {"": "empty_key"}}}}`

I think that diverges from our current behavior e.g. all keys in `disk_device.options` and in its nested objects should have a length of at least 1.

tahmidefaz: For adding host via MQ, this won't matter as this gets re-verified with another function, but trying to fetch a host that has
`{"disk_device": {"options": {"key": {"": "empty_key"}}}}` will not throw a `5xx` anymore as this is not invalid anymore per this new definition, whereas it should throw a `5xx` in this situation

tahmidefaz (changes requested):

tahmidefaz:
Note to self: we need to update our HBI unit tests to catch the aforementioned failure :memo:

swagger/openapi.json
Glutexo: `{"disk_device": {"options": {"key": {"": "empty_key"}}}}` should not pass!

Glutexo: I’ll double check.

Glutexo: I tested with a host entry like this 

```sql
INSERT INTO public.hosts (id, account, display_name, created_on, modified_on, facts, tags, canonical_facts, system_profile_facts, ansible_host, stale_timestamp, reporter, per_reporter_staleness) VALUES ('e0a2bcc2-3c68-4a7e-adf9-9af38e343f3e', 'test', 'test', '2021-05-07 17:35:39.397000', '2021-05-07 17:35:43.006000', '{}', '{}', '{"insights_id": "8b75c23f-8149-499e-848a-1638295ccb1a"}', '{"arch": "x86-64", "owner_id": "1b36b20f-7fa0-4454-a6d2-008294e06378", "cpu_flags": ["flag1", "flag2"], "cpu_model": "Intel(R) Xeon(R) CPU E5-2690 0 @ 2.90GHz", "yum_repos": [{"name": "repo1", "enabled": true, "base_url": "http://rpms.redhat.com", "gpgcheck": true}], "os_release": "Red Hat EL 7.0.1", "bios_vendor": "Turd Ferguson", "bios_version": "1.0.0uhoh", "disk_devices": [{"type": "ext3", "label": "home drive", "device": "/dev/sdb1", "options": {"ro": true, "key": {"": "empty_key", "C": "D"}, "uid": "0"}, "mount_point": "/home"}], "captured_date": "2020-02-13T12:16:00Z", "rhc_client_id": "044e36dc-4e2b-4e69-8948-9c65a7bf4976", "is_marketplace": false, "kernel_modules": ["i915", "e1000e"], "last_boot_time": "12:25 Mar 19, 2019", "number_of_cpus": 1, "cores_per_socket": 4, "enabled_services": ["ndb", "krb5"], "rhc_config_state": "044e36dc-4e2b-4e69-8948-9c65a7bf4976", "bios_release_date": "10/31/2013", "number_of_sockets": 2, "os_kernel_version": "3.10.0", "running_processes": ["vim", "gcc", "python"], "satellite_managed": false, "installed_products": [{"id": "123", "name": "eap", "status": "UP"}, {"id": "321", "name": "jbws", "status": "DOWN"}], "installed_services": ["ndb", "krb5"], "network_interfaces": [{"mtu": 1500, "name": "eth0", "type": "loopback", "state": "UP", "mac_address": "aa:bb:cc:dd:ee:ff", "ipv4_addresses": ["10.10.10.1"], "ipv6_addresses": ["2001:0db8:85a3:0000:0000:8a2e:0370:7334"]}], "infrastructure_type": "jingleheimer junction cpu", "selinux_config_file": "enforcing", "subscription_status": "valid", "system_memory_bytes": 1024, "insights_egg_version": "120.0.1", "selinux_current_mode": "enforcing", "infrastructure_vendor": "dell", "katello_agent_running": false, "insights_client_version": "12.0.12", "subscription_auto_attach": "yes"}', null, '2021-06-07 17:36:07.934000', 'reporter', '{}');
```

containing `"options": {"ro": true, "key": {"": "empty_key", "C": "D"}`.

`GET http://localhost:8081/api/inventory/v1/hosts/e0a2bcc2-3c68-4a7e-adf9-9af38e343f3e/system_profile` results in  an _500 Internal Server Error_ just as expected.

```json
{
  "detail": "{'': 'empty_key', 'C': 'D'} is not valid under any of the given schemas\n\nFailed validating 'oneOf' in schema['properties']['results']['items']['properties']['system_profile']['properties']['disk_devices']['items']['properties']['options']['additionalProperties']:\n    {'oneOf': [{'additionalProperties': <Recursion on dict with id=4436957248>,\n                'description': 'An arbitrary object that does not allow '\n                               'empty string keys.',\n                'type': 'object',\n                'x-propertyNames': {'minLength': 1},\n                'x-scope': ['']},\n               {'not': {'type': 'object'}}]}\n\nOn instance['results'][0]['system_profile']['disk_devices'][0]['options']['key']:\n    {'': 'empty_key', 'C': 'D'}",
  "status": 500,
  "title": "Response body does not conform to specification",
  "type": "about:blank"
}
```

Can you please verify? @tahmidefaz If you can’t reproduce, let’s hop on a call to find the cause.

swagger/system_profile_commit_id
```diff
@@ -0,0 +1 @@
+dd15575876c2869e1318919a71e5551fc3cd5374
```
kruai, line 1: Adding this file is a good call :+1:

swagger/api.spec.yaml
```diff
@@ -623,8 +607,9 @@ components:
       style: deepObject
       explode: true
       schema:
-        type: object
-        description: System profile field values to filter with
+        $ref:  '#/components/schemas/SystemProfileNestedObject'
+       # type: object
+       # description: System profile field values to filter with
```
kruai, line 612: Why comment instead of remove?

swagger/api.spec.yaml
```diff
@@ -633,8 +618,9 @@ components:
       style: deepObject
       explode: true
       schema:
-        type: object
-        description: System profile field values to fetch
+        $ref:  '#/components/schemas/SystemProfileNestedObject'
+        # type: object
+        # description: System profile field values to fetch
```
kruai, line 623: Same Q here

swagger/openapi.json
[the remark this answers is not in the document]
kruai: This file raises a few Qs. Does adding this mean we have to maintain 2 files now? Are we moving from YAML to JSON?

swagger/api.spec.yaml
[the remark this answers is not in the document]
RonnyPfannschmidt: removed

swagger/api.spec.yaml
```diff
@@ -716,6 +712,7 @@ components:
       type: array
       items:
         type: string
+        nullable: False
```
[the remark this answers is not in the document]
RonnyPfannschmidt: removed

swagger/system_profile.spec.yaml
RonnyPfannschmidt: undone

swagger/openapi.json
RonnyPfannschmidt: i beleive we don't follow up on this detail we included the details again

swagger/api.spec.yaml
RonnyPfannschmidt: also upstream issue

check_schemas.py
```diff
@@ -0,0 +1,23 @@
...
+OAPI = SPECIFICATION_DIR + "openapi.json"
+
+with open(OAPI) as fp:
+    oapi_data = yaml.dump(json.load(fp))
+
+openapi = TranslatingParser(OAPI)
+openapi.parse()
+
+# spec = TranslatingParser(SPECIFICATION_DIR + "api.spec.yaml")
+# spec.parse()
```
kruai, line 19: Should be removed instead of commented

kruai (changes requested):
Only had one minor request in the comments. My main concern is that the build is failing every time. The most recent error appears to be caused by the setuptools pinning:
`ERROR: Invalid requirement: 'pip setuptools<58'`

Once you get the builds passing, I'll approve this.

RonnyPfannschmidt:
moved to draft, need to run another api spec update

kruai (approved):
I think this is good now!

kruai (dismissed):

app/__init__.py
```diff
@@ -36,7 +36,9 @@
 REQUEST_ID_HEADER = "x-rh-insights-request-id"
 UNKNOWN_REQUEST_ID_VALUE = "-1"
 
-SPECIFICATION_FILE = join(SPECIFICATION_DIR, "api.spec.yaml")
+# temporary replaced for RHCLOUD-12800 - correct normalization of the bundled spec
```
ryandillinfelton, line 39: Might want to switch this to the ESSNTL card number if there is one

check_schemas.py
```diff
@@ -0,0 +1,27 @@
...
+    cdiff = unified_diff(
+        oapi_data.splitlines(True), parser.yaml().splitlines(True), fromfile="oapi", tofile="output_filename"
+    )
+    sys.stdout.writelines(cdiff)
+
+
+if __name__ == "__main__":
+    make_diff("openapi.json", "prace_reads_bundle")
+    # todo: reenable this once the prance bug about inclusion is resolved
+    # make_diff("api.spec.yaml", "prace_reads_spec")
```
ryandillinfelton, line 27: I think it'd be cool to add a script to `package.json` that runs this, but I won't block on it.

ryandillinfelton:
@RonnyPfannschmidt could you open a PR on `[inventory-schemas](https://github.com/RedHatInsights/inventory-schemas)` with the same changes you made to the base schema. We should merge it there first to make sure they stay in sync.

app/__init__.py
RonnyPfannschmidt:
```suggestion
# temporary replaced for ESSNTL-746 - correct normalization of the bundled spec
```

check_schemas.py
RonnyPfannschmidt: before prance is fixed, this may not be as interesting as we'd like it

RonnyPfannschmidt:
@ryandillinfelton the pagination changes only belong to our own api, no upstreaming needed

kruai (approved):

ryandillinfelton (approved):

PR #834, opened 2021-03-18
Removing all references of host-egress topic

tahmidefaz:
## Overview

This PR removes all references to the deprecated `platform.inventory.host-egress` topic.

## Secure Coding Checklist

### DIFF
--- a/app/queue/event_producer.py
+++ b/app/queue/event_producer.py
@@ -9,14 +9,14 @@ from app.logging import get_logger
 
 logger = get_logger(__name__)
 
-Topic = Enum("Topic", ("egress", "events"))
+Topic = Enum("Topic", ("events"))
 
 
 class EventProducer:
     def __init__(self, config):
         logger.info("Starting EventProducer()")
         self._kafka_producer = KafkaProducer(bootstrap_servers=config.bootstrap_servers, **config.kafka_producer)
-        self.topics = {Topic.egress: config.host_egress_topic, Topic.events: config.event_topic}
+        self.topics = {Topic.events: config.event_topic}
 
     def write_event(self, event, key, headers, topic, *, wait=False):
         logger.debug("Topic: %s, key: %s, event: %s, headers: %s", topic, key, event, headers)
--- a/utils/kafka_consumer.py
+++ b/utils/kafka_consumer.py
@@ -8,7 +8,7 @@ from kafka import KafkaConsumer
 # from app.models import SystemProfileSchema
 
 
-HOST_EGRESS_TOPIC = os.environ.get("KAFKA_EVENTS_TOPIC", "platform.inventory.events")
+HOST_EGRESS_TOPIC = os.environ.get("KAFKA_EVENT_TOPIC", "platform.inventory.events")
 HOST_INGRESS_GROUP = os.environ.get("KAFKA_HOST_INGRESS_GROUP", "inventory-mq")
 BOOTSTRAP_SERVERS = os.environ.get("KAFKA_BOOTSTRAP_SERVERS", "localhost:29092")
 
--- a/utils/kafka_consumer.py
+++ b/utils/kafka_consumer.py
@@ -8,7 +8,7 @@ from kafka import KafkaConsumer
 # from app.models import SystemProfileSchema
 
 
-HOST_EGRESS_TOPIC = os.environ.get("KAFKA_EVENT_TOPIC", "platform.inventory.events")
+EVENTS_TOPIC = os.environ.get("KAFKA_EVENT_TOPIC", "platform.inventory.events")
 HOST_INGRESS_GROUP = os.environ.get("KAFKA_HOST_INGRESS_GROUP", "inventory-mq")
 BOOTSTRAP_SERVERS = os.environ.get("KAFKA_BOOTSTRAP_SERVERS", "localhost:29092")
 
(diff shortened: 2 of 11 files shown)


kruai (changes requested):
If we're getting rid of the host-egress topic altogether, I think we also need to remove the references to `HOST_EGRESS_TOPIC`. They can all be changed to `EVENT_TOPIC` now, right?

tahmidefaz:
> If we're getting rid of the host-egress topic altogether, I think we also need to remove the references to `HOST_EGRESS_TOPIC`. They can all be changed to `EVENT_TOPIC` now, right?

Actually, we still use that environment variable in all our deployments: [here](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/host-inventory/deploy.yml#L86)

kruai:
> > If we're getting rid of the host-egress topic altogether, I think we also need to remove the references to `HOST_EGRESS_TOPIC`. They can all be changed to `EVENT_TOPIC` now, right?
> 
> Actually, we still use that environment variable in all our deployments: [here](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/host-inventory/deploy.yml#L86)

We do set the environment variable in the deployment templates, but we can remove it from the templates once we get rid of the refs in our code.

RonnyPfannschmidt (dismissed):

utils/kafka_consumer.py
```diff
@@ -8,7 +8,7 @@
 # from app.models import SystemProfileSchema
 
 
-HOST_EGRESS_TOPIC = os.environ.get("KAFKA_HOST_EGRESS_TOPIC", "platform.inventory.host-egress")
+HOST_EGRESS_TOPIC = os.environ.get("KAFKA_HOST_EGRESS_TOPIC", "platform.inventory.events")
```
thearifismail, line 11: do other teams/services use `KAFKA_HOST_EGRESS_TOPIC`?  Shouldn't this env var be named `KAFKA_HOST_EVENT_TOPIC`?

thearifismail (dismissed):
asked a question about a minor issue for clarity or consitency

utils/kafka_consumer.py
kruai: We have a var named `KAFKA_EVENT_TOPIC` that we use for events

tahmidefaz: Yeah, Asa had the same thought as well. Fixed it

tahmidefaz:
> > > If we're getting rid of the host-egress topic altogether, I think we also need to remove the references to `HOST_EGRESS_TOPIC`. They can all be changed to `EVENT_TOPIC` now, right?
> > 
> > 
> > Actually, we still use that environment variable in all our deployments: [here](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/host-inventory/deploy.yml#L86)
> 
> We do set the environment variable in the deployment templates, but we can remove it from the templates once we get rid of the refs in our code.

I see what you mean. Updated :+1:

kruai (changes requested):
Good progress, but there are still other references in the project to the host-egress topic. These should be changed to reference the events topic. Here are a few:
- https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/__init__.py#L52
- https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/config.py#L226
- https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/queue/event_producer.py#L19
- https://github.com/RedHatInsights/insights-host-inventory/blob/master/tests/test_unit.py#L1690

tahmidefaz:
@kruai Just removed all refs of `host-egress`. Lmk, if you see anything that I missed.

I also remembered that we introduced an additional environment variable called `KAFKA_SECONDARY_TOPIC_ENABLED` during the transition period to `platform.inventory.events` topic. I went ahead and removed that as well since we don't need it anymore.

kruai (approved):
I think you got them all!

kruai (changes requested):
Oops sorry, there's actually one more reference to the secondary-topic-enabled value:
https://github.com/RedHatInsights/insights-host-inventory/blob/remove-host-egress-string/tests/fixtures/mq_fixtures.py#L75

tahmidefaz:
> Oops sorry, there's actually one more reference to the secondary-topic-enabled value:
> https://github.com/RedHatInsights/insights-host-inventory/blob/remove-host-egress-string/tests/fixtures/mq_fixtures.py#L75

I did remove that, but forgot to save the file :man_facepalming:

kruai (approved):

app/queue/event_producer.py
```diff
@@ -9,14 +9,14 @@
 
 logger = get_logger(__name__)
 
-Topic = Enum("Topic", ("egress", "events"))
+Topic = Enum("Topic", ("events"))
```
ryandillinfelton, line 12: For all of the enums and things like this that are built around us having multiple egress topics it seems like we should switch them to just be singular since as far as I know there's no plan to have multiple again, and it doesn't seem all that likely that one will be made.

ryandillinfelton (approved):
This all looks like it'll work just fine so I'll approve. However I do feel like while you're at it you might as well adjust some of the code to only expect one topic.

app/queue/event_producer.py
tahmidefaz: @ryandillinfelton Good idea! I am working on it right now! If I push up my refactoring changes that you suggested here, it will make this PR quite large. Therefore, I am suggesting that we open a subsequent refactoring PR to address these changes. What's your opinion? (I am almost done, btw. Just trying to address a test failure)

tahmidefaz:
As promised, here is the follow-up refactoring PR! https://github.com/RedHatInsights/insights-host-inventory/pull/840 :slightly_smiling_face: 
@ryandillinfelton

PR #838, opened 2021-03-19
Add per reporter staleness filter

ryandillinfelton:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-xxxxx).
(A description of your PR's changes goes here.)

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


RonnyPfannschmidt (commented):
i'm getting the impression that the filter building is/qA very repetitive and handles None/nulls/missing all over the place

as a followup/side project i would like us to investigate expressing this in a more general compact way in order to remove those null/none/missing checks

cc @gmcculloug @kruai 

as we use tuples often and do check for empty/`()` often, and given that
```pycon
>>> () + (1,) + ()
(1,)
```

i believe changing the filters creation to something we could just add up without sprinkling conditionals all over the place  wil lhelp to make for compact more readable code

ryandillinfelton:
This is ready for review. I don't believe there is anything stopping us from merging it in, but there is still some work to be done on the xjoin side to transform the data as it goes into elasticsearch. All that to say, I think we can merge it in and just not tell anyone about the filters until they actually work.

api/host_query_xjoin.py
```diff
@@ -185,14 +185,55 @@ def _query_filters(fqdn, display_name, hostname_or_id, insights_id, tags, stalen
...
+    if system_profile.get("sap_sids"):
+        system_profile_filter += build_sap_sids_filter(system_profile["sap_sids"])
+
+    return system_profile_filter
+
+
+def _build_per_reporter_staleness_filter(per_reporter_staleness):
+    prs_dict_array = []
+
+    for reporter in per_reporter_staleness:
```
RonnyPfannschmidt, line 214:
```suggestion
    for reporter, props in per_reporter_staleness.items():
```
should drop duplicate sub-expressions nicely

kruai: I agree... then you can just use `props["stale_timestamp"]` instead of `per_reporter_staleness[reporter]["stale_timestamp"]`, etc etc

api/host_query_xjoin.py
```diff
@@ -185,14 +185,55 @@ def _query_filters(fqdn, display_name, hostname_or_id, insights_id, tags, stalen
...
+            prs_dict["check_in_succeeded"] = {
+                "is": per_reporter_staleness[reporter]["check_in_succeeded"].lower() == "true"
+            }
+
+        prs_dict_array.append({"per_reporter_staleness": prs_dict})
+
+    return ({"AND": prs_dict_array},)
+
+
+def build_prs_reporter_filter(sap_sids):
```
kruai, line 231: This function looks like it's specifically for `sap_sids`. Can you explain why the function is called `build_prs_reporter_filter`? I'm a little confused

tests/test_xjoin.py
```diff
@@ -1780,3 +1780,96 @@ def test_sp_sparse_xjoin_query_translation(
 
     assert response_status == 200
     graphql_sparse_system_profile_empty_response.assert_called_once_with(SYSTEM_PROFILE_QUERY, variables, mocker.ANY)
+
+
+def test_query_hosts_filter_per_reporter_staleness(
+    mocker, subtests, query_source_xjoin, graphql_query_empty_response, patch_xjoin_post, api_get
+):
+    queries_http = (
+        "filter[per_reporter_staleness][yupana][exists]",
```
kruai, line 1789: This will need to be updated if we're going with the `[exists]=true` syntax (along with the other instances in this file)

api/host_query_xjoin.py
kruai: Actually, is this used anywhere? If not, it should be removed

kruai (changes requested):
Had a few notes/questions

tests/test_xjoin.py
thearifismail: > This will need to be updated if we're going with the `[exists]=true` syntax (along with the other instances in this file)

or `nil`

api/host_query_xjoin.py
```diff
@@ -185,14 +185,55 @@ def _query_filters(fqdn, display_name, hostname_or_id, insights_id, tags, stalen
...
+                query_filters += per_reporter_staleness_filter
 
     logger.debug(query_filters)
     return query_filters
 
 
+def _build_system_profile_filter(system_profile):
+    system_profile_filter = tuple()
+
+    if system_profile.get("sap_system"):
```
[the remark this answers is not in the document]
kruai: This PR is only for per-reporter staleness filters, not system_profile fields. I'm on board with this new `_build_system_profile_filter` function to better separate the new filters that will be added soon

tests/test_xjoin.py
```diff
@@ -1780,3 +1780,96 @@ def test_sp_sparse_xjoin_query_translation(
...
+        "filter[per_reporter_staleness][yupana][exists]&filter[per_reporter_staleness][puptoo][exists]",
+        "filter[per_reporter_staleness][yupana][stale_timestamp][gte]=2021-03-18T21%3A21%3A00.935093%2B00%3A00",
```
thearifismail, line 1791: How about replacing hard-coded timestamps with dynamically generated ones using something `Time.now()`

RonnyPfannschmidt: i'm seeing that we use this timestamp in a lot of places both urlencoded and plain
i'd recommend to at least put it in a constant as currently it constitutes a magic number

api/host_query_xjoin.py
ryandillinfelton: removed it

api/host_query_xjoin.py
ryandillinfelton: Done :+1:

tests/test_xjoin.py
ryandillinfelton: I'm going with the existing way to query nil/not_nil and expanding it to work with strings aswell

tests/test_xjoin.py
ryandillinfelton: I think using time.now would be more hassle than it's worth. It could be wise to extract some timestamps into constants, but since we often need a mix of timestamp values I think an effort to do so should be it's own PR and not rolled into this one. This test is the only spot where this specific timestamp shows up at all, and I'd rather not open up the bag of worms of unifying it and other tests to a set of timestamps now.

api/host_query_xjoin.py
```diff
@@ -125,34 +126,44 @@ def _params_to_order(param_order_by=None, param_order_how=None):
...
+def _sap_sids_filters(field_name, sap_sids):
     sap_sids_filters = ()
     for sap_sid in sap_sids:
-        sap_sids_filters += ({"spf_sap_sids": {"eq": sap_sid}},)
+        sap_sids_filters += ({field_name: {"eq": sap_sid}},)
     return sap_sids_filters
 
 
+def build_filter(field_name, field_value, field_type, operation, filter_building_function):
+    if isinstance(field_value, field_type):
```
RonnyPfannschmidt, line 155: if we return a empty tuple when `field_value is None` i believe we can replace the conditionals for get with 
streamlined calls

api/host_query_xjoin.py
```diff
@@ -188,14 +199,42 @@ def _query_filters(fqdn, display_name, hostname_or_id, insights_id, tags, stalen
...
+
+    for reporter, props in per_reporter_staleness.items():
+        prs_dict = {"reporter": {"eq": reporter}}
+
+        if per_reporter_staleness[reporter].get("stale_timestamp"):
+            prs_dict["stale_timestamp"] = props["stale_timestamp"]
+        if props.get("last_check_in"):
+            prs_dict["last_check_in"] = props["last_check_in"]
+        if props.get("check_in_succeeded"):
+            prs_dict["check_in_succeeded"] = {"is": props["check_in_succeeded"].lower() == "true"}
```
RonnyPfannschmidt, line 232: it does seem to me, that this one looks like a boolean filter as well, would it make sense, or should this one stay specific

tests/test_xjoin.py
RonnyPfannschmidt: lets have a low prio  followup card then

tests/test_xjoin.py
ryandillinfelton: Actually, I think we need exists for this. Specifically because we can't send
`filter[per_reporter_staleness][yupana]=not_nil,filter[per_reporter_staleness][yupana][check_in_succeeded]=true`
without one overwritting the other in the resulting object. It would be possible to do some parsing trickery but I don't think it'd be wise. So I think just adding [exists] as on operation for checking if objects exists makes sense and doesn't clash with existing operations for other types.

ryandillinfelton:
Fiddling with existence queries. Hold off on reviewing until I say otherwise in another comment. I got ahead of myself.

api/host_query_xjoin.py
ryandillinfelton: I think I'll make it a boolean filter as well. My only hesitation is that the field is never really going to be nil, but I don't suppose that is really a problem

ryandillinfelton: Okay so it's actually bad to let it be nullable so I made a non_nullable boolean filter function.

ryandillinfelton:
Okay I believe this is ready for re-review.
Also, short comment. You may notice that the test is only for happy path and there is nothing enforcing the inputs. There is a similar issue to the system_profile filters where we cannot validate the parameters using jsonschema. I'll be doing a follow up PR to this and the incoming extra system profile filters that adds manual checking in.

tests/test_xjoin.py
thearifismail: Yes in the spirit of keep PR short.  here is [RHCLOUD-13225](https://issues.redhat.com/browse/RHCLOUD-13225) for followup.

tests/test_xjoin.py
thearifismail: Agree.  Let's get it working now.  Elegance can be added later.

kruai: > Agree. Let's get it working now. Elegance can be added later.

That feels like a dangerous mindset that can lead to lots of technical debt... but I agree that the `[exists]` operation is fine. I really just want to make sure we agree on one and keep it consistent throughout or app!

api/host_query_xjoin.py
```diff
@@ -125,34 +126,52 @@ def _params_to_order(param_order_by=None, param_order_how=None):
...
-def build_sap_system_filters(sap_system):
-    if isinstance(sap_system, str):
-        return (_sap_system_filters(sap_system),)
-    elif sap_system.get("eq"):
-        return (_sap_system_filters(sap_system["eq"]),)
+def _string_filter(field_name, field_value):
+    if field_value == NIL_STRING:
+        return ({field_name: {"eq": None}},)
+    elif field_value == NOT_NIL_STRING:
+        return ({"NOT": {field_name: {"eq": None}}},)
```
RonnyPfannschmidt, line 146: compared to bools, why do the nils not use is

ryandillinfelton: `Is` isn't supported on string fields in xjoin. This is the equivalent expression for strings and wildcard strings.

thearifismail:
> Agree. Let's get it working now. Elegance can be added later.

Oops, should have said, "Agree. Let's get it working now. Elegance can be added later `after the Summit`

kruai (approved):
LGTM!

api/host_query_xjoin.py
```diff
@@ -120,34 +121,52 @@ def _params_to_order(param_order_by=None, param_order_how=None):
...
+
 def build_sap_sids_filter(sap_sids):
-    if isinstance(sap_sids, list):
-        return _sap_sids_filters(sap_sids)
-    elif sap_sids.get("contains"):
-        return _sap_sids_filters(sap_sids["contains"])
+    return build_filter("spf_sap_sids", sap_sids, list, "contains", _sap_sids_filters)
+
+
+def build_check_in_succeeded_filter(check_in_succeeded):
```
thearifismail, line 168: Is `def build_check_in_succeeded_filter()` is still needed because it is not referenced anywhere?

api/host_query_xjoin.py
```diff
@@ -120,34 +121,52 @@ def _params_to_order(param_order_by=None, param_order_how=None):
...
-def _sap_sids_filters(sap_sids):
+def _sap_sids_filters(field_name, sap_sids):
     sap_sids_filters = ()
     for sap_sid in sap_sids:
-        sap_sids_filters += ({"spf_sap_sids": {"eq": sap_sid}},)
+        sap_sids_filters += ({field_name: {"eq": sap_sid}},)
     return sap_sids_filters
 
 
+def build_filter(field_name, field_value, field_type, operation, filter_building_function):
```
thearifismail, line 153: This function should be private because it is not used outside of this module.

thearifismail (commented):
Any plan replacing CONSTANTS NIL and NOT_NIL.  Theoretical correct but making the code more congested.

ryandillinfelton:
> Any plan replacing CONSTANTS NIL and NOT_NIL. Theoretical correct but making the code more congested.

What would you suggest I replace them with?

thearifismail:
> > Any plan replacing CONSTANTS NIL and NOT_NIL. Theoretical correct but making the code more congested.
> 
> What would you suggest I replace them with?

nothing; just remove `NIL` and `NOT_NIL` constants and use `nil` and `not_nil` directly.

kruai:
> > > Any plan replacing CONSTANTS NIL and NOT_NIL. Theoretical correct but making the code more congested.
> > 
> > 
> > What would you suggest I replace them with?
> 
> nothing; just remove `NIL` and `NOT_NIL` constants and use `nil` and `not_nil` directly.

That definitely feels like a step backward to me. What would we gain by doing this?

ryandillinfelton:
> > > > Any plan replacing CONSTANTS NIL and NOT_NIL. Theoretical correct but making the code more congested.
> > > 
> > > 
> > > What would you suggest I replace them with?
> > 
> > 
> > nothing; just remove `NIL` and `NOT_NIL` constants and use `nil` and `not_nil` directly.
> 
> That definitely feels like a step backward to me. What would we gain by doing this?

ryandillinfelton:
I always think that's the cancel comment button :facepalm:

ryandillinfelton:
> > > Any plan replacing CONSTANTS NIL and NOT_NIL. Theoretical correct but making the code more congested.
> > 
> > 
> > What would you suggest I replace them with?
> 
> nothing; just remove `NIL` and `NOT_NIL` constants and use `nil` and `not_nil` directly.

I felt like just using the raw string was bad practice since bugs could arise from a simple typo.

thearifismail:
> > > > Any plan replacing CONSTANTS NIL and NOT_NIL. Theoretical correct but making the code more congested.
> > > 
> > > 
> > > What would you suggest I replace them with?
> > 
> > 
> > nothing; just remove `NIL` and `NOT_NIL` constants and use `nil` and `not_nil` directly.
> 
> That definitely feels like a step backward to me. What would we gain by doing this?

IMO, the code should look cleaner with `nil` `not_nil` than`NIL` and `NOT_NIL`.  Constants are good when their values are subject to change, which is not the case here.  It is just a suggestion.

ryandillinfelton:
Conflicts fixed

api/host_query_xjoin.py
```diff
@@ -253,5 +255,26 @@ def _build_system_profile_filter(system_profile):
     return system_profile_filter
 
 
+def _build_per_reporter_staleness_filter(per_reporter_staleness):
+    prs_dict_array = []
```
RonnyPfannschmidt, line 259: lets just return `()` here if `per_reporter_staleness` is falsey

api/host_query_xjoin.py
```diff
@@ -221,6 +221,8 @@ def _query_filters(fqdn, display_name, hostname_or_id, insights_id, tags, stalen
     if filter:
         if filter.get("system_profile"):
             query_filters += _build_system_profile_filter(filter["system_profile"])
+        if filter.get("per_reporter_staleness"):
```
RonnyPfannschmidt, line 224: if we let the `_build_per_reporter_staleness_filter` function return `()` on falsey values, we can start to take the ifs out of the equation
ideal case is that eventually the composition of the filter is free of most conditions at this level

kruai:
@ryandillinfelton This is a feature we still want to implement, but this PR is pretty old. If this PR is still the way to go, what do we need in order for it to not be on hold? 
(Also if we still want to use this PR, please update the Jira link in the description because it's still using the placeholder :slightly_smiling_face:)

kruai:
Out of date. Replaced by #1115.

PR #840, opened 2021-03-22
Topic logic refactoring

tahmidefaz:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-xxxxx).
(A description of your PR's changes goes here.)

## Secure Coding Checklist

### DIFF
--- a/tests/test_mq_service.py
+++ b/tests/test_mq_service.py
@@ -12,7 +12,6 @@ from app import db
 from app.auth.identity import Identity
 from app.exceptions import InventoryException
 from app.exceptions import ValidationException
-from app.queue.event_producer import Topic
 from app.queue.queue import _validate_json_object_for_utf8
 from app.queue.queue import event_loop
 from app.queue.queue import handle_message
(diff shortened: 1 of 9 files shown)


tests/test_mq_service.py
```diff
@@ -122,33 +121,33 @@ def test_shutdown_handler(mocker, flask_app):
     assert handle_message_mock.call_count == 2
 
 
-def test_events_sent_to_correct_topic(mocker, flask_app):
-    host_id = generate_uuid()
-    insights_id = generate_uuid()
+# def test_events_sent_to_correct_topic(mocker, flask_app):
```
tahmidefaz, line 124: I remember adding this test when introducing multiple topic functionality. Do we still need to keep it? It's not easy to test this functionality based on the way I am refactoring here. Also, I am thinking, since we only have one egress topic, what purpose does this test serve?

tests/test_unit.py
```diff
@@ -1701,38 +1700,25 @@ def test_happy_path(self):
...
-                    self.topic_names[topic],
+                    self.topic_name,
                     key=host_id.encode("utf-8"),
                     value=event.encode("utf-8"),
                     headers=expected_encoded_headers(event_type, threadctx.request_id, host_id),
                 )
                 send.reset_mock()
 
-    # Insure that a ValueError exception is raised if a topic not defined in the Topic enum is used
-    def test_invalid_topic_causes_failure(self):
```
tahmidefaz, line 1727: Same thing as above

tests/test_mq_service.py
kruai: I think this test still makes sense; it's validating that the correct event type is created, and that the event is sent to the topic... so it's making sure that the logic doesn't change and it starts sending messages to the Ingress queue or something like that

tests/test_unit.py
kruai: Considering the changes made in your implementation, _this_ test can definitely be removed :laughing:

tests/test_mq_service.py
tahmidefaz: Ooo, I need to think of how to modify this test! It's quite difficult to test this functionality, unless I change the way things are laid out here :thinking:

tahmidefaz: @kruai Actually, after the refactoring, [this test](https://github.com/RedHatInsights/insights-host-inventory/blob/master/tests/test_unit.py#L1700) is testing the exact same thing. What do you think? :thinking:

kruai: Yeah, I think you're right... due to the naming, I think I'd prefer this test to get destroyed? But idc either way tbh

kruai (changes requested):
I think this is almost ready to merge. If you remove one of the redundant tests (as per my comment response), I'll LGTM it

tests/test_mq_service.py
tahmidefaz: Just removed it :slightly_smiling_face:

kruai (approved):
LGTM, once you deal with that new merge conflict!

kruai (approved):
LGTM!

ryandillinfelton (approved):

thearifismail (approved):
LGTM.

thearifismail:
Yes I approved based the expectation that the environment variables in inventory-mq deployment templates will be updated soon.

RonnyPfannschmidt (approved):

PR #843, opened 2021-03-23
Add handler for system_profile messages

kruai:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-12844).

Allows `inv_mq_service` to be configured with different topics, which informs how the message handler operates. If the current Kafka topic is the host-ingress one, it operates normally by adding/updating hosts. If the current Kafka topic is the system-profile one, it tells the message handler to update the system profile, rather than add/update the whole host. 

I implemented it this way for flexibility, and due to discussions about AI CoE potentially requiring a similar approach but for a new field. I also wanted to avoid copypasta as much as possible, since the message-handling part of this is very similar to what we already have.

## Secure Coding Checklist

- [x] Input Validation
- [x] Output Encoding
- [x] Data Protection
- [x] System Configuration

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -157,7 +157,6 @@ class Config:
         self.host_delete_chunk_size = int(os.getenv("HOST_DELETE_CHUNK_SIZE", "1000"))
         self.script_chunk_size = int(os.getenv("SCRIPT_CHUNK_SIZE", "1000"))
         self.sp_authorized_users = os.getenv("SP_AUTHORIZED_USERS", "tuser@redhat.com").split()
-        self.mq_restrict_to_field = os.getenv("MQ_RESTRICT_TO_FIELD", "ALL")
 
     def _build_base_url_path(self):
         app_name = os.getenv("APP_NAME", "inventory")
--- a/app/config.py
+++ b/app/config.py
@@ -43,7 +43,10 @@ class Config:
         self.additional_validation_topic = topic(
             os.environ.get("KAFKA_ADDITIONAL_VALIDATION_TOPIC", "platform.inventory.host-ingress-p1")
         )
-        self.system_profile_topic = topic("platform.system-profile")
+        self.system_profile_topic = topic(
+            os.environ.get("KAFKA_SYSTEM_PROFILE_TOPIC", "platform.inventory.system-profile")
+        )
+        self.kafka_consumer_topic = topic(os.environ.get("KAFKA_CONSUMER_TOPIC", "platform.inventory.host-ingress"))
         self.event_topic = topic("platform.inventory.events")
         self.payload_tracker_kafka_topic = topic("platform.payload-status")
 
@@ -59,7 +62,8 @@ class Config:
         self.additional_validation_topic = os.environ.get(
             "KAFKA_ADDITIONAL_VALIDATION_TOPIC", "platform.inventory.host-ingress-p1"
         )
-        self.system_profile_topic = os.environ.get("KAFKA_TOPIC", "platform.system-profile")
+        self.system_profile_topic = os.environ.get("KAFKA_SYSTEM_PROFILE_TOPIC", "platform.inventory.system-profile")
+        self.kafka_consumer_topic = os.environ.get("KAFKA_CONSUMER_TOPIC", "platform.inventory.host-ingress")
         self.bootstrap_servers = os.environ.get("KAFKA_BOOTSTRAP_SERVERS", "localhost:29092")
         self.event_topic = os.environ.get("KAFKA_EVENT_TOPIC", "platform.inventory.events")
         self.payload_tracker_kafka_topic = os.environ.get("PAYLOAD_TRACKER_KAFKA_TOPIC", "platform.payload-status")
@@ -219,7 +223,9 @@ class Config:
 
             if self._runtime_environment == RuntimeEnvironment.SERVICE:
                 self.logger.info("Kafka Host Ingress Topic: %s" % self.host_ingress_topic)
-                self.logger.info("Kafka Host Ingress Group: %s" % self.host_ingress_consumer_group)
+                self.logger.info("Kafka System Profile Topic: %s" % self.system_profile_topic)
+                self.logger.info("Kafka Consumer Topic: %s" % self.kafka_consumer_topic)
+                self.logger.info("Kafka Consumer Group: %s" % self.host_ingress_consumer_group)
                 self.logger.info("Kafka Events Topic: %s" % self.event_topic)
 
             if self._runtime_environment.event_producer_enabled:
--- a/app/config.py
+++ b/app/config.py
@@ -157,6 +157,7 @@ class Config:
         self.host_delete_chunk_size = int(os.getenv("HOST_DELETE_CHUNK_SIZE", "1000"))
         self.script_chunk_size = int(os.getenv("SCRIPT_CHUNK_SIZE", "1000"))
         self.sp_authorized_users = os.getenv("SP_AUTHORIZED_USERS", "tuser@redhat.com").split()
+        self.mq_restrict_to_field = os.getenv("MQ_RESTRICT_TO_FIELD", "ALL")
 
     def _build_base_url_path(self):
         app_name = os.getenv("APP_NAME", "inventory")
--- a/app/config.py
+++ b/app/config.py
@@ -219,17 +219,17 @@ class Config:
             self.logger.info("RBAC Timeout Seconds: %s", self.rbac_timeout)
 
         if self._runtime_environment == RuntimeEnvironment.SERVICE or self._runtime_environment.event_producer_enabled:
-            self.logger.info("Kafka Bootstrap Servers: %s" % self.bootstrap_servers)
+            self.logger.info("Kafka Bootstrap Servers: %s", self.bootstrap_servers)
 
             if self._runtime_environment == RuntimeEnvironment.SERVICE:
-                self.logger.info("Kafka Host Ingress Topic: %s" % self.host_ingress_topic)
-                self.logger.info("Kafka System Profile Topic: %s" % self.system_profile_topic)
-                self.logger.info("Kafka Consumer Topic: %s" % self.kafka_consumer_topic)
-                self.logger.info("Kafka Consumer Group: %s" % self.host_ingress_consumer_group)
-                self.logger.info("Kafka Events Topic: %s" % self.event_topic)
+                self.logger.info("Kafka Host Ingress Topic: %s", self.host_ingress_topic)
+                self.logger.info("Kafka System Profile Topic: %s", self.system_profile_topic)
+                self.logger.info("Kafka Consumer Topic: %s", self.kafka_consumer_topic)
+                self.logger.info("Kafka Consumer Group: %s", self.host_ingress_consumer_group)
+                self.logger.info("Kafka Events Topic: %s", self.event_topic)
 
             if self._runtime_environment.event_producer_enabled:
-                self.logger.info("Kafka Event Topic: %s" % self.event_topic)
+                self.logger.info("Kafka Event Topic: %s", self.event_topic)
 
         if self._runtime_environment.payload_tracker_enabled:
             self.logger.info("Payload Tracker Kafka Topic: %s", self.payload_tracker_kafka_topic)
--- a/app/instrumentation.py
+++ b/app/instrumentation.py
@@ -145,7 +145,7 @@ def log_add_host_failure(logger, host_data):
 # update system profile
 def log_update_system_profile_success(logger, host_data):
     metrics.update_system_profile_success.inc()
-    logger.info(f"System profile updated for host ID {host_data.get('id')}")
+    logger.info("System profile updated for host ID: %s", host_data.get("id"))
 
 
 def log_update_system_profile_failure(logger, host_data):
--- a/app/instrumentation.py
+++ b/app/instrumentation.py
@@ -149,7 +149,7 @@ def log_update_system_profile_success(logger, host_data):
 
 
 def log_update_system_profile_failure(logger, host_data):
-    logger.exception("Error updating system profile for host ", extra={"host": host_data})
+    logger.exception("Error updating system profile for host ID: %s", host_data.get("id"))
     metrics.update_system_profile_failure.labels("InventoryException").inc()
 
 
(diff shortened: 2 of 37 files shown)


inv_mq_service.py
```diff
@@ -7,13 +7,17 @@
...
+from app.queue.queue import add_host
 from app.queue.queue import event_loop
 from app.queue.queue import handle_message
+from app.queue.queue import update_system_profile
 from lib.handlers import register_shutdown
 from lib.handlers import ShutdownHandler
 
 logger = get_logger("mq_service")
 
+field_restriction_to_handler = {{"ALL": add_host}, {"SYSTEM_PROFILE": update_system_profile}}
```
RonnyPfannschmidt, line 19: this looks like it should be a single dict

RonnyPfannschmidt (changes requested):
i do wonder if instead of a new service with new restrictions and deployment env vars,
a new message type wouldn't fit better


then the system profile update message could update jsut the system profile,
additionally we wouldnt see the event/egress queue order issues as well

if we basically start a new mq service with a new queue that inter spares different types of inventory updates we are going to see the same temporal issues we saw with just having 2 input queues

i think we should have a new message type for limited updates, not a new queue/mq consumer

RonnyPfannschmidt:
i  just cross-read the reasoning for the separate topic/deployment in the jira and i can agree on that

RonnyPfannschmidt:
i have a alternative proposal - how about instead of making a new topic with a different "permission" having a new message for the "non-reporter" updates and actually restricting the messages specific topics are allowed to consume

so the default topic would not be allowed to consume the system profile updates and the system profile topic would only be allowed to consume the system profile updates

i'm currently pretty sure that a new message type will be better than a shift of function in multiple layers

inv_mq_service.py
kruai: Yep, you're right!

thearifismail:
@kruai I feel I am missing some piece of information but going ahead with asking questions and commenting on it.

how will `non-reporters` be authenticating to access a host before modifying? 
IOW, how will they gain access to a host?

Why there has to be a different topic?  How about a new event_type?  For this event type, use `update_system_profile` method by finding the desired host.
`
EVENT_TYPE_MAP = {
    EventType.created: host_create_update_event,
    EventType.updated: host_create_update_event,
    EventType.delete: host_delete_event,
    EventType.update_system_profile: update_system_profile,
}

def def update_system_profile(input_host, identity, staleness_offset, fields):
    ...
    ...
    ...
    return output_host, existing_host.id, insights_id, UpdateSystemProfileResult.updated
`

kruai:
> @kruai I feel I am missing some piece of information but going ahead with asking questions and commenting on it.
> 
> how will `non-reporters` be authenticating to access a host before modifying?
> IOW, how will they gain access to a host?
> 
> Why there has to be a different topic? How about a new event_type? For this event type, use `update_system_profile` method by finding the desired host.
> 
> ```
> EVENT_TYPE_MAP = {
>     EventType.created: host_create_update_event,
>     EventType.updated: host_create_update_event,
>     EventType.delete: host_delete_event,
>     EventType.update_system_profile: update_system_profile,
> }
> 
> def def update_system_profile(input_host, identity, staleness_offset, fields):
>     ...
>     ...
>     ...
>     return output_host, existing_host.id, insights_id, UpdateSystemProfileResult.updated
> ```

Non-reporters will need to supply identity via metadata, just like we're used to.
We decided to go with a separate topic so that an influx of host-ingress messages wouldn't cause kafka lag for system-profile updates, and vice-versa. Messages that come in on the system-profile topic would only be allowed to modify the system_profile field on hosts. On this Jira card, Fero and I were chatting about this and agreed that we should use another operation type... though whether we'll need another event type might need a little more analysis

thearifismail:
Okay avoiding Kafka lag time makes sense and having a dedicated topic for system_profile update is fine.  Do we expect that much traffic?  If someone has a host.id, why shouldn't she be able to update the general properties of a host?  Why such block?

kruai:
> Okay avoiding Kafka lag time makes sense and having a dedicated topic for system_profile update is fine. Do we expect that much traffic? 

We don't expect it at first, but we don't know how frequently we're going to get these requests in the future, and we don't want their updates to be delayed if puptoo/yupana suddenly pushes 100K messages to the queue. We're also preparing for the possibility that AI CoE may also use this topic (if we don't give them their own; TBD)

> If someone has a host.id, why shouldn't she be able to update the general properties of a host? Why such block?

We're trying to lock down access as much as is reasonable. It would be nice to fully trust any app on our platform, but we don't want some random app to accidentally change a host's canonical facts or something. We only want to grant limited access to these apps that want to push this data without updating the staleness or reporter data.

app/queue/events.py
```diff
@@ -15,7 +15,7 @@
 logger = logging.getLogger(__name__)
 
 
-EventType = Enum("EventType", ("created", "updated", "delete"))
+EventType = Enum("EventType", ("created", "updated", "delete", "system_profile_updated"))
```
jharting, line 18: The event type tells the downstream consumers how to deal with the given event (create a new entry, update existing one, delete old one). In this regard I don't think `system_profile_updated` is any different from `updated`. If you think downstream consumers may want to distinguish between different ways of host updates (e.g. host-ingress, system-profile or HTTP patch) it may be better to add that info in a specific event header.

jharting (changes requested):

app/queue/events.py
RonnyPfannschmidt: @kruai until we have a better handle on what we push out to consumers, i believe we should map the system profile update action to a normal update event

app/config.py
```diff
@@ -219,7 +223,9 @@ def log_configuration(self):
 
             if self._runtime_environment == RuntimeEnvironment.SERVICE:
                 self.logger.info("Kafka Host Ingress Topic: %s" % self.host_ingress_topic)
-                self.logger.info("Kafka Host Ingress Group: %s" % self.host_ingress_consumer_group)
+                self.logger.info("Kafka System Profile Topic: %s" % self.system_profile_topic)
```
RonnyPfannschmidt, line 226: short note on logging, lets replace all the `%` by `,`

app/instrumentation.py
```diff
@@ -142,6 +142,17 @@ def log_add_host_failure(logger, host_data):
     metrics.add_host_failure.labels("InventoryException", host_data.get("reporter", "null")).inc()
 
 
+# update system profile
+def log_update_system_profile_success(logger, host_data):
+    metrics.update_system_profile_success.inc()
+    logger.info("System profile updated for host", extra={"host": host_data})
```
RonnyPfannschmidt, line 148: for optional followup: would it help to make the host id part of the normal message?

app/models.py
```diff
@@ -320,7 +320,7 @@ def merge_facts_in_namespace(self, namespace, facts_dict):
             self.facts[namespace] = facts_dict
         orm.attributes.flag_modified(self, "facts")
 
-    def _update_system_profile(self, input_system_profile):
+    def update_system_profile(self, input_system_profile):
         logger.debug("Updating host's (id=%s) system profile", self.id)
         if not self.system_profile_facts:
             self.system_profile_facts = input_system_profile
```
RonnyPfannschmidt, line 326: followup critical question: i just took note, that the system profile update is a flat merge instead of a deep merge, is that a potential bug?

lib/host_repository.py
```diff
@@ -195,3 +203,29 @@ def update_query_for_owner_id(identity, query):
...
+
+            existing_host.update_system_profile(input_host.system_profile_facts)
+            db.session.commit()
+
+            metrics.update_host_count.inc()
+            logger.debug("Updated system profile for host:%s", existing_host)
+
+            output_host = serialize_host(existing_host, staleness_offset, fields)
+            insights_id = existing_host.canonical_facts.get("insights_id")
+            return output_host, existing_host.id, insights_id, AddHostResult.system_profile_updated
```
RonnyPfannschmidt, line 227: i believe at this point we should not return a `system_profile_updated` message until we have the consumers onboard

RonnyPfannschmidt (changes requested):
overall this comes together nicely,
the considerations for output messages need some more iteration and a followup after other teams give input on what they want to use

app/queue/events.py
kruai: OK, that's fine with me!

app/config.py
kruai: I can do that for my additions! If you think it's a big enough deal to change it for all of these strings, can you make a follow-up Jira for it?

lib/host_repository.py
kruai: Fine with me! Updated.

app/instrumentation.py
RonnyPfannschmidt: i vaguely recall tat i don't see host ids for those on the cli log, do we have a log processor that makes them accessible?

app/models.py
RonnyPfannschmidt: done

app/instrumentation.py
kruai: I'm not sure if it's super visible on the CLI log, but using this syntax causes the data to be outputted in the Kibana logs. You can see some examples of this [here](https://kibana.apps.crcp01ue1.o9m8.p1.openshiftapps.com/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))&_a=(columns:!(_source),filters:!(),index:'43c5fed0-d5ce-11ea-b58c-a7c95afd7a5d',interval:auto,query:(language:lucene,query:'@log_stream:%20%22inventory-mq%22%20AND%20@message:%20%22Host%20updated%22'),sort:!(!('@timestamp',desc))))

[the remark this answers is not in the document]
kruai: I was trying to keep it consistent with our add/update Host logging, but if we want to keep this log concise, that's fine with me. I'll make that change shortly

app/config.py
```diff
@@ -215,15 +219,17 @@ def log_configuration(self):
...
 
         if self._runtime_environment == RuntimeEnvironment.SERVICE or self._runtime_environment.event_producer_enabled:
-            self.logger.info("Kafka Bootstrap Servers: %s" % self.bootstrap_servers)
+            self.logger.info("Kafka Bootstrap Servers: %s", self.bootstrap_servers)
 
             if self._runtime_environment == RuntimeEnvironment.SERVICE:
-                self.logger.info("Kafka Host Ingress Topic: %s" % self.host_ingress_topic)
-                self.logger.info("Kafka Host Ingress Group: %s" % self.host_ingress_consumer_group)
-                self.logger.info("Kafka Events Topic: %s" % self.event_topic)
+                self.logger.info("Kafka Host Ingress Topic: %s", self.host_ingress_topic)
```
thearifismail, line 225: nitpick: should start using f"... {var}" instead of `%s, var`

app/config.py
```diff
@@ -59,7 +62,8 @@ def non_clowder_config(self):
         self.additional_validation_topic = os.environ.get(
             "KAFKA_ADDITIONAL_VALIDATION_TOPIC", "platform.inventory.host-ingress-p1"
         )
-        self.system_profile_topic = os.environ.get("KAFKA_TOPIC", "platform.system-profile")
+        self.system_profile_topic = os.environ.get("KAFKA_SYSTEM_PROFILE_TOPIC", "platform.inventory.system-profile")
+        self.kafka_consumer_topic = os.environ.get("KAFKA_CONSUMER_TOPIC", "platform.inventory.host-ingress")
```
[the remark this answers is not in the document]
kruai: This change follows the standards we set in the e2e PR: https://github.com/RedHatInsights/e2e-deploy/pull/2914
Basically, the `KAFKA_CONSUMER_TOPIC` tells the app which topic it should be assigned to. 

Setting `KAFKA_HOST_INGRESS_TOPIC` and `KAFKA_SYSTEM_PROFILE_TOPIC` allows the app to be aware of what the ingress and system profile topics are, so it can determine whether `KAFKA_CONSUMER_TOPIC == KAFKA_HOST_INGRESS_TOPIC`, `KAFKA_CONSUMER_TOPIC == KAFKA_SYSTEM_PROFILE_TOPIC`, etc.

app/config.py
kruai: I'd also be fine with that, but I just made that change to address @RonnyPfannschmidt 's earlier requested change. If y'all want these to change any more, please open up a Jira card, because that's unrelated to this PR's goal

RonnyPfannschmidt: full stop, this one is the valid logging syntax (no f strings there intentionally)

app/instrumentation.py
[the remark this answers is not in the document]
kruai: OK, updated

app/config.py
kruai: OK. In that case, I'm resolving this thread

app/instrumentation.py
kruai: As you and @thearifismail noted, it's most important that we log the Host ID for these transactions. I've changed the log to just output the message + the Host ID. My main concern with this is that it isn't great for auditing. Or is that not an issue?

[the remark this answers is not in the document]
kruai: Note: I've now updated `log_update_system_profile_failure` with this change as well.

thearifismail:
Will approve it after the update branch build completes.

app/instrumentation.py
```diff
@@ -142,6 +142,17 @@ def log_add_host_failure(logger, host_data):
...
 
 
+# update system profile
+def log_update_system_profile_success(logger, host_data):
+    metrics.update_system_profile_success.inc()
+    logger.info("System profile updated for host ID: %s", host_data.get("id"))
+
+
+def log_update_system_profile_failure(logger, host_data):
+    logger.exception("Error updating system profile for host ID: %s", host_data.get("id"))
```
thearifismail, line 152: For failures, the whole host_data should be logged.

kruai: OK. Reverted that change

tahmidefaz (approved):
Looks good! :+1:

thearifismail (approved):
My questions and comments have been addressed.

RonnyPfannschmidt (approved):

jharting (approved):

PR #845, opened 2021-03-24
Use test for account_number

tahmidefaz:
## Overview

Using `"test"` as the `account_number` in `payloads.py` allows us to keep using the header we have been using all along and also makes it easier to integrate with `xjoin-search`.

### DIFF
--- a/utils/payloads.py
+++ b/utils/payloads.py
@@ -9,7 +9,7 @@ from datetime import timezone
 # good
 IDENTITY = {
     "identity": {
-        "account_number": "sysaccount",
+        "account_number": "test",
         "type": "System",
         "auth_type": "cert-auth",
         "system": {"cn": "1b36b20f-7fa0-4454-a6d2-008294e06378", "cert_type": "system"},


kruai (approved):

utils/payloads.py
```diff
@@ -9,7 +9,7 @@
 # good
 IDENTITY = {
     "identity": {
-        "account_number": "sysaccount",
+        "account_number": "test",
```
thearifismail, line 12: `test` account number is used everywhere in tests.  `payloads` is used for interactive testing.  That's why using `sysaccount` makes it easier to diffrentiate who created the host.

thearifismail (commented):
Yes having this non-standard test-account name makes it easier to differentiate hosts created using different identity types, particularly during testing using Postman

utils/payloads.py
tahmidefaz: @thearifismail I was under the impression that those sort of "interactive testing" only needs to happen rarely. So, when needed, we can simply change this string to whatever we feel like.
As I mentioned in my PR description, using "test" here, we get to keep using the headers that's mentioned in our readme + while working with `xjoin-search`.

ryandillinfelton (approved):

utils/payloads.py
ryandillinfelton: Don't the hosts created during automated testing not exist after the tests are finished? I'm not sure I understand the advantage of having a separate account number for manual testing. Personally I always set the account number to `test` when i'm doing manual testing so it matches the hosts in xjoin-search and I don't have to swap headers between testing db and xjoin.

thearifismail: No. SQLAlchemy does not commit the hosts created by pytest testing.  I felt the need for it when creating hosts using both User and System identities.  It helps with keeping tab of which `apiKey` to use in Postman.

thearifismail: @tahmidefaz  yes manual testing is done rarely but happens quite frequently during implementation.

thearifismail:
Oops.  so much for fighting for it.

PR #847, opened 2021-03-25
Host List Sparse System Profile Response - RHCLOUD-12849

tahmidefaz:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-12849).
The goal of this PR is to enable fetching of some system_profile fields from the `/hosts` endpoint using sparse fields.

This PR does not intend on  changing the current response in `host_list` endpoint. Users will need to mention the `fields` query parameter if they want to fetch system_profile data through the `host_list` endpoint. If HBI is using the `db` and/or the user does not mention the `fields` query parameter, the response should remain same as it is today.

## Secure Coding Checklist

- [x] Input Validation
- [x] Output Encoding

### DIFF
--- a/api/host_query.py
+++ b/api/host_query.py
@@ -1,7 +1,6 @@
 from collections import namedtuple
 from enum import Enum
 
-from app import HostListResponse
 from app import inventory_config
 from app.culling import Timestamps
 from app.serialization import DEFAULT_FIELDS
@@ -14,15 +13,10 @@ OrderBy = Enum("OrderBy", ("display_name", "id", "modified_on"))
 OrderHow = Enum("OrderHow", ("ASC", "DESC"))
 Order = namedtuple("Order", ("by", "how"))
 
-type_mapper = {
-    HostListResponse.generic: DEFAULT_FIELDS,
-    HostListResponse.sparse_sp: DEFAULT_FIELDS + ("system_profile",),
-}
 
-
-def build_paginated_host_list_response(total, page, per_page, host_list, response_type=HostListResponse.generic):
+def build_paginated_host_list_response(total, page, per_page, host_list, additional_fields=tuple()):
     timestamps = staleness_timestamps()
-    json_host_list = [serialize_host(host, timestamps, type_mapper[response_type]) for host in host_list]
+    json_host_list = [serialize_host(host, timestamps, DEFAULT_FIELDS + additional_fields) for host in host_list]
     return {
         "total": total,
         "count": len(json_host_list),
--- a/api/host_query.py
+++ b/api/host_query.py
@@ -1,8 +1,8 @@
 from collections import namedtuple
 from enum import Enum
 
-from app import inventory_config
 from app import HostListResponse
+from app import inventory_config
 from app.culling import Timestamps
 from app.serialization import DEFAULT_FIELDS
 from app.serialization import serialize_host
@@ -14,14 +14,14 @@ OrderBy = Enum("OrderBy", ("display_name", "id", "modified_on"))
 OrderHow = Enum("OrderHow", ("ASC", "DESC"))
 Order = namedtuple("Order", ("by", "how"))
 
-type_mapper = {HostListResponse.generic: DEFAULT_FIELDS, HostListResponse.sparse_sp: DEFAULT_FIELDS+("system_profile",)}
+type_mapper = {
+    HostListResponse.generic: DEFAULT_FIELDS,
+    HostListResponse.sparse_sp: DEFAULT_FIELDS + ("system_profile",),
+}
 
 
 def build_paginated_host_list_response(total, page, per_page, host_list, response_type=HostListResponse.generic):
     timestamps = staleness_timestamps()
-    # if source == BulkQuerySource.xjoin:
-    #     json_host_list = [serialize_host(host, timestamps, DEFAULT_FIELDS + ("system_profile",)) for host in host_list]
-    # else:
     json_host_list = [serialize_host(host, timestamps, type_mapper[response_type]) for host in host_list]
     return {
         "total": total,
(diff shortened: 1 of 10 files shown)


api/host_query.py
```diff
@@ -12,10 +14,15 @@
...
 Order = namedtuple("Order", ("by", "how"))
 
+type_mapper = {
+    HostListResponse.generic: DEFAULT_FIELDS,
+    HostListResponse.sparse_sp: DEFAULT_FIELDS + ("system_profile",),
+}
 
-def build_paginated_host_list_response(total, page, per_page, host_list):
+
+def build_paginated_host_list_response(total, page, per_page, host_list, response_type=HostListResponse.generic):
```
kruai, line 23: I've been trying to determine how you could get this done without adding the enumeration + type mapper. I think a more elegant way to achieve this is to have this function accept an `additional_fields` set/list, and you pass in `{ "system_profile" }` as appropriate (it defaults to empty). What do you think?

tahmidefaz: It does clean things up a bit! I opted in to use a `tuple` instead because then we can do `DEFAULT_FIELDS + additional_fields` :+1:

tests/test_api_hosts_get.py
```diff
@@ -1070,3 +1071,23 @@ def test_validate_sp_sparse_fields_invalid_requests(query_source_xjoin, api_get)
 
         response_status, response_data = api_get(build_system_profile_url(hosts, query=query))
         assert response_status == 400
+
+
+def test_host_list_sp_fields_requested(patch_xjoin_post, query_source_xjoin, api_get):
+    patch_xjoin_post(response={"data": XJOIN_HOSTS_RESPONSE})
```
thearifismail, line 1077: `XJOIN_HOSTS_RESPONSE`  creates hosts data with timestamps set in the past and the values are `"2019-02-10T08:07:03.354307Z"`.  Not sure how much this helps when no valid data is returned by queries.   Better create a method which creates a valid timestamp to represent the moment in real time.

tests/test_api_hosts_get.py
```diff
@@ -1070,3 +1071,23 @@ def test_validate_sp_sparse_fields_invalid_requests(query_source_xjoin, api_get)
...
+    response_status, response_data = api_get(HOST_URL + "?fields[system_profile]=test_data,random")
+
+    assert response_status == 200
+
+    for host_data in response_data["results"]:
+        assert "system_profile" in host_data
```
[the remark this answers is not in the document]
tahmidefaz: Well, on HBI's side, we just sent out the request with the correct `xjoin-search` filters and we currently do not check whether any additional fields gets sent to us from `xjoin-search`. This currently applies to all responses HBI gets from `xjoin-search`.  Do you think we should have that logic in place?

Also, here, we are patching the `xjoin-search` response and my intention for this test was just to check that we do not sent out the `system_profile` key with our responses when the user does not include `fields` in their query param

tests/test_api_hosts_get.py
tahmidefaz: So, at present, once we receive data from `xjoin-search` we do not mutate any of the `value` (on some instance, we just mutate the `key` names). With `patch_xjoin_post`, replicates a response from `xjoin-search`.
Your suggestion to "create a method which creates a valid timestamp to represent the moment in real time" does sounds nice to be, but it also makes me wonder that it can be part of a test refactoring PR. What do you think?

thearifismail: In the spirit of keeping this PR short and sweet, a test refactoring PR sound good but I fear it may never get done.  I suggest creating a JIRA card and putting in here to resolve it.

thearifismail:
Cloning the branch and running `pytest tests/test_xjoin.py` fails 22 tests. Slacking the failure log to PR owner

thearifismail:
Yes pulling the latest code passed all the tests.

api/host_query_db.py
```diff
@@ -64,7 +65,7 @@ def get_host_list(
 
     log_get_host_list_succeeded(logger, query_results.items)
 
-    return query_results.items, query_results.total
+    return query_results.items, query_results.total, tuple()
```
ryandillinfelton, line 68: I'm not a fan of how ambiguous this return is. Would it be possible to return just `all_fields` instead of `additional_field`, then return just `DEFAULT_FIELDS` here and return `DEFAULT_FIELDS + additional_fields` from the xjoin get_host_list func.

If that's not possible It'd be nice to have at least a comment explaining why we're returning an empty tuple.

tests/test_api_hosts_get.py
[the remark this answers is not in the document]
thearifismail: @kruai  is right.  it needs an `assertion` using `test_data` because that's what is being used.  This should work:
`        assert "test_data" in host_data["system_profile"].keys()
`

api/host_query_db.py
tahmidefaz: I think returning `DEFAULT_FIELDS` breaks the code patterns even more when compared to similar functions within HBI. Also, I think this function gets used by other endpoints as well.
Instead of a comment, I used a descriptive variable name to make this less cryptic. Lmk, if this works for ya :+1:

tahmidefaz:
^ what is `cryption`? I meant to write `cryptic` :laughing:

api/host_query_db.py
ryandillinfelton: I think a comment on the definition could help still. But I think it's clear enough that it's not a problem at this point.

ryandillinfelton (approved):
Looks goood

thearifismail (approved):

kruai (approved):
LGTM

PR #863, opened 2021-04-08
Schema changes to support system-profile MQ service

kruai:
## Overview

This PR is being created to address [RHCLOUD-13461](https://issues.redhat.com/browse/RHCLOUD-13461), [RHCLOUD-13466](https://issues.redhat.com/browse/RHCLOUD-13466), and [RHCLOUD-13505](https://issues.redhat.com/browse/RHCLOUD-13505).
Separates out a functionality-restricted version of the Host schema and model, named LimitedHost. LimitedHost acts as a bean of sorts, and cannot be added/updated in the DB. Host extends LimitedHost, but requires the reporter and stale_timestamp fields, and has all the usual access.

I created `model_builder()` in LimitedHostSchema and HostSchema so that we could keep our generic Host-deserialization functionality, while having it generate the correct model type.

## PR Checklist

- [x] Tests: edge cases

## Secure Coding Checklist

- [x] Input Validation

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -123,7 +123,7 @@ class SystemProfileNormalizer:
             self.filter_keys(value, schema.items)
 
 
-class Host(db.Model):
+class LimitedHost(db.Model):
     __tablename__ = "hosts"
     # These Index entries are essentially place holders so that the
     # alembic autogenerate functionality does not try to remove the indexes
@@ -133,6 +133,40 @@ class Host(db.Model):
         Index("idxaccount", "account"),
         Index("hosts_subscription_manager_id_index", text("(canonical_facts ->> 'subscription_manager_id')")),
     )
+
+    def __init__(
+        self,
+        canonical_facts,
+        display_name=None,
+        ansible_host=None,
+        account=None,
+        facts=None,
+        tags=None,
+        system_profile_facts=None,
+    ):
+
+        if not canonical_facts:
+            raise InventoryException(
+                title="Invalid request", detail="At least one of the canonical fact fields must be present."
+            )
+
+        self.canonical_facts = canonical_facts
+
+        if display_name:
+            # Only set the display_name field if input the display_name has
+            # been set...this will make it so that the "default" logic will
+            # get called during the save to fill in an empty display_name
+            self.display_name = display_name
+        self._update_ansible_host(ansible_host)
+        self.account = account
+        self.facts = facts or {}
+        self.tags = tags or {}
+        self.system_profile_facts = system_profile_facts or {}
+
+    def _update_ansible_host(self, ansible_host):
+        if ansible_host is not None:
+            # Allow a user to clear out the ansible host with an empty string
+            self.ansible_host = ansible_host
 
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
     account = db.Column(db.String(10))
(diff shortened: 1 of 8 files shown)


app/models.py
```diff
@@ -160,6 +192,7 @@ def __init__(
         stale_timestamp=None,
         reporter=None,
     ):
+        super().__init__(canonical_facts, display_name, ansible_host, account, facts, tags, system_profile_facts)
```
ryandillinfelton, line 195: I think you're setting the canonical_facts before checking for their presence in the line below. Seems like the check should be done in limited_host or before calling super().__init__()

ryandillinfelton (approved):

app/models.py
ryandillinfelton: I'm leaving this here in case my further thinking was wrong. I'm approving since setting the value on the model shouldn't matter when the check failing will produce an exception and keep the model from making a change to the DB. Basically if I'm thinking about it right the order of assignment and check doesn't matter to the final outcome.

kruai: It's the same functionality, but I agree that it's better do those checks before calling super.init(). Updating now

kruai: BTW, I'm keeping that functionality out of LimitedHost, since it shouldn't require canonical facts to be instantiated

ryandillinfelton (approved):
:+1:

app/models.py
```diff
@@ -486,6 +494,18 @@ def _normalize_system_profile(normalize, data):
         normalize(system_profile)
         return {**data, "system_profile": system_profile}
 
+    @staticmethod
+    def build_model(data, canonical_facts, facts, tags):
```
thearifismail, line 498: From a reader's perspective, `build_host()` as the function name tells what it is does.  What model was the question came up upon seeing `build_model()`

thearifismail (commented):
LGTM except the func name change comment

app/models.py
kruai: I think `build_host()` is misleading, because one of them generates a LimitedHost, and the other generates a Host. The reason I named the function `build_model()` was to provide each Schema with a way to create its associated Model: LimitedHostSchema can generate a LimitedHost model object, and HostSchema can generate a Host model object

app/models.py
```diff
@@ -506,9 +526,33 @@ def system_profile_is_valid(self, system_profile):
                 raise MarshmallowValidationError(f"Empty key in /system_profile/disk_devices/{dd_i}/options.")
 
 
+class HostSchema(LimitedHostSchema):
+    stale_timestamp = fields.DateTime(required=True, timezone=True)
+    reporter = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=255))
+
+    @staticmethod
+    def build_model(data, canonical_facts, facts, tags):
```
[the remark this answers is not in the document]
kruai: I responded above - we can keep the conversation to that thread

app/models.py
thearifismail: Makes sense.  Thanks.

thearifismail (approved):

tahmidefaz (approved):
Looks good! :+1:

tests/test_unit.py
```diff
@@ -907,7 +907,9 @@ def test_with_all_fields(self, deserialize_canonical_facts, deserialize_facts, d
             "stale_timestamp": datetime.now(timezone.utc).isoformat(),
             "reporter": "some reporter",
         }
-        host_schema = Mock(**{"return_value.load.return_value.data": host_input})
+        host_schema = Mock(
```
RonnyPfannschmidt, line 910: @kruai @tahmidefaz it seems to me we have a pretty painful mocking pattern at hand here
should we take a look at a more painless one **after** summit?

kruai: I'm not opposed to it; the mocks in other places do it in method calls after the constructor. This one is quicker and more condensed, but the other one is much easier to read

PR #867, opened 2021-04-13
Edit pr_check to use common code

bsquizz:
Switches pr_check.sh and unit_test.sh to use common cicd scripts

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -28,22 +28,14 @@ IQE_FILTER_EXPRESSION=""
 
 # Get bonfire helper scripts
 CICD_URL=https://raw.githubusercontent.com/RedHatInsights/bonfire/master/cicd
-curl -s $CICD_URL/build.sh -o build.sh
-curl -s $CICD_URL/deploy_ephemeral_env.sh -o deploy_ephemeral_env.sh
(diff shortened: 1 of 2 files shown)


kruai (approved):

pr_check.sh
```diff
@@ -28,22 +28,14 @@ IQE_FILTER_EXPRESSION=""
...
-# source deploy_ephemeral_env.sh
-# source smoke_test.sh
-
-mkdir -p $WORKSPACE/artifacts
-cat << EOF > ${WORKSPACE}/artifacts/junit-dummy.xml
-<testsuite tests="1">
-    <testcase classname="dummy" name="dummytest"/>
-</testsuite>
-EOF
+# source $CICD_ROOT/deploy_ephemeral_env.sh
```
BlakeHolifield, line 40: Do we need to turn this back on if we're only spinning up a db? The previous `unit_test.sh` was bringing up the whole app.

bsquizz: That'll be uncommented when we start running the smoke test, deploy_ephemeral_db.sh does actually still deploy the ClowdApp (just without its extra dependencies)

BlakeHolifield (approved):

PR #872, opened 2021-04-14
platform_metadata fix for rhsm-conduit

kruai:
## Overview

This is a fix addressing rhsm-conduit's issues regarding not sending platform_metadata.
They weren't made aware that we were making platform_metadata a requirement, and we need to not break our support for them.


tahmidefaz (approved):

thearifismail (approved):
LGTM

PR #876, opened 2021-04-15
Add owner id filter

tahmidefaz:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-13172).
Adds filter for owner_id including. Also fixes a bug that fails on multiple strings filtering.

## PR Checklist

## Secure Coding Checklist

- [x] Input Validation
- [x] Output Encoding

### DIFF
--- a/api/host_query_xjoin.py
+++ b/api/host_query_xjoin.py
@@ -153,6 +153,13 @@ def _nullable_string_filter(field_name, field_value):
     return _nullable_wrapper(field_name, field_value, "eq", _string_filter)
 
 
+def _nullable_multiple_string_filter(field_name, field_value):
+    multiple_string_filter = ()
+    for value in field_value:
+        multiple_string_filter += _nullable_wrapper(field_name, value, "eq", _string_filter)
+    return ({"OR": multiple_string_filter},)
+
+
 def _string_filter(field_name, field_value):
     return ({field_name: {"eq": (field_value)}},)
 
@@ -177,6 +184,13 @@ def build_filter(field_name, field_value, field_type, operation, filter_building
         return filter_building_function(field_name, field_value)
     elif field_value.get(operation):
         return filter_building_function(field_name, field_value[operation])
+
+
+def build_filter_string_multiple(field_name, field_value, field_type, operation):
+    if isinstance(field_value, field_type) or isinstance(field_value[operation], field_type):
+        return build_filter(field_name, field_value, str, operation, _nullable_string_filter)
+    elif field_value.get(operation):
+        return build_filter(field_name, field_value[operation], list, operation, _nullable_multiple_string_filter)
 
 
 def build_sap_system_filter(sap_system):
@@ -238,8 +252,8 @@ def _build_system_profile_filter(system_profile):
             "spf_is_marketplace", system_profile["is_marketplace"], str, "eq", _nullable_boolean_filter
         )
     if system_profile.get("rhc_client_id"):
-        system_profile_filter += build_filter(
-            "spf_rhc_client_id", system_profile["rhc_client_id"], str, "eq", _nullable_string_filter
+        system_profile_filter += build_filter_string_multiple(
+            "spf_rhc_client_id", system_profile["rhc_client_id"], str, "eq"
         )
     if system_profile.get("insights_client_version"):
         system_profile_filter += build_filter(
@@ -249,7 +263,8 @@ def _build_system_profile_filter(system_profile):
             "eq",
             _nullable_wildcard_filter,
         )
-
+    if system_profile.get("owner_id"):
+        system_profile_filter += build_filter_string_multiple("spf_owner_id", system_profile["owner_id"], str, "eq")
     return system_profile_filter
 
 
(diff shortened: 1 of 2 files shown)


tahmidefaz:
To make you all's life a little easier, `api/host_query_xjoin.py ` is the place you want to start. The tests are almost an exact copy of the `rhc_client_id` since this filter is no different than that one. This PR also enables filtering based on multiple `owner_id` and `rhc_client_id`

api/host_query_xjoin.py
```diff
@@ -179,6 +186,13 @@ def build_filter(field_name, field_value, field_type, operation, filter_building
         return filter_building_function(field_name, field_value[operation])
 
 
+def build_filter_string_multiple(field_name, field_value, field_type, operation):
+    if isinstance(field_value, field_type) or isinstance(field_value[operation], field_type):
+        return build_filter(field_name, field_value, str, operation, _nullable_string_filter)
+    else:
```
ryandillinfelton, line 192: Since we have basically no validation of these from the api spec (and this recently bit Ronny) I'd like to at least get some decent validation coverage in the executive code for new filters. you don't have to go add checks to everything in the file, but right here could use one. Instead of assuming anything that doesn't fit the first case is a list, an explicit check that it's a list and an else that raises a validation exception would be nice.
```suggestion
    if isinstance(field_value, list) or isinstance(field_value[operation], list):
        return build_filter(field_name, field_value[operation], list, operation, _nullable_multiple_string_filter)
    else:
        raise ValidationException(f"field_name expected to be a string or array of strings, instead got {type(field_value)}")
```
something like that. The suggestion is not precise.

api/host_query_xjoin.py
```diff
@@ -238,8 +252,8 @@ def _build_system_profile_filter(system_profile):
             "spf_is_marketplace", system_profile["is_marketplace"], str, "eq", _nullable_boolean_filter
         )
     if system_profile.get("rhc_client_id"):
-        system_profile_filter += build_filter(
-            "spf_rhc_client_id", system_profile["rhc_client_id"], str, "eq", _nullable_string_filter
+        system_profile_filter += build_filter_string_multiple(
```
ryandillinfelton, line 255: TY :heart:

tests/test_xjoin.py
```diff
@@ -1647,6 +1647,124 @@ def test_query_hosts_filter_spf_rhc_client_id(
                 graphql_query_empty_response.reset_mock()
 
 
+def test_query_hosts_filter_spf_rhc_client_id_multiple(
```
ryandillinfelton, line 1650: :heart: :heart:

ryandillinfelton (changes requested):
Mostly looks good. I pointed out some improvements I'd like to see.

api/host_query_xjoin.py
```diff
@@ -179,6 +186,13 @@ def build_filter(field_name, field_value, field_type, operation, filter_building
         return filter_building_function(field_name, field_value[operation])
 
 
+def build_filter_string_multiple(field_name, field_value, field_type, operation):
```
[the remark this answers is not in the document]
tahmidefaz: I think the `field_type` was kind of implied by the function name. Removed. :+1:

api/host_query_xjoin.py
tahmidefaz: Added a `ValidationException` and added tests for it as well :+1:

tests/test_xjoin.py
```diff
@@ -1647,6 +1647,150 @@ def test_query_hosts_filter_spf_rhc_client_id(
...
+                },
+                mocker.ANY,
+            )
+            graphql_query_empty_response.reset_mock()
+
+
+def test_spf_owner_id_invalid_field_value(
+    mocker, query_source_xjoin, graphql_query_empty_response, patch_xjoin_post, api_get
+):
+    mocker.patch(
```
ryandillinfelton, line 1784: I don't understand what this patch is supposed to do

tahmidefaz: I added this test as a followup to one of your earlier concern that we should throw a `ValidationException` whenever `build_filter_string_multiple` receive something other than a string or an array. However, I didn't find an easy way to have it receive anything but a string or an array. Therefore, I am patching the return value of `api.parsing.customURIParser._make_deep_object` to intentionally make `build_filter_string_multiple` receive `{"owner_id": {"eq": None}}` (in other world, a `nulltype`) so that we can hit that `ValidationException` in `build_filter_string_multiple`. Does this clear it up a little?

ryandillinfelton: Yeah it clears it up. It does make me question if we even need to handle that if it basically can't happen.

ryandillinfelton: If you hit the endpoint with `?filter[system_profile][owner_id][foo][bar]` it should come through as a `dict` for those checks and fail. Though I think you'd want to catch those with another kind of check to make sure the operation is valid. But that could be a good way to catch invalid paths and not have them cause a `500`

tahmidefaz: Yes! I was scratching my head around it a bit, and my changes that I committed earlier actually catches it throws `ValidationExceptions` accordingly (I didn't even know)! So, I just went ahead and simplified my invalid tests :+1: 
Lmk, what you think!

ryandillinfelton (approved):
Sweet! Looking good now :+1:

ryandillinfelton (approved):

ryandillinfelton:
@thearifismail @kruai Monday thing, but we just need one more :+1: on this

kruai (approved):
LGTM!

tests/test_xjoin.py
kruai: @ryandillinfelton anything you wanna follow up on in this thread?

PR #882, opened 2021-04-20
Log cert_type errors, but don't raise an exception

kruai:
## Overview

This PR is being created to fix our cert_type errors and improve logging in prod.
We'll want to re-add the ValueError once @thearifismail has confirmed the list of possible values.

### DIFF
--- a/app/auth/identity.py
+++ b/app/auth/identity.py
@@ -116,7 +116,8 @@ def validate(identity):
             if not identity.system.get("cert_type"):
                 raise ValueError("The cert_type field is mandatory.")
             if identity.system.get("cert_type") not in CertType.__members__.values():
-                raise ValueError(f"The cert_type {identity.system.get('cert_type')} is invalid.")
+                # TODO: Raise ValueError once we solidify all cert_type values
+                logger.error("The cert_type %s is invalid.", identity.system.get("cert_type"))
             if not identity.system.get("cn"):
                 raise ValueError("The cn field is mandatory.")
             if not identity.auth_type:


tahmidefaz (dismissed):
Looks good! :+1:

app/auth/identity.py
```diff
@@ -116,7 +116,8 @@ def validate(identity):
             if not identity.system.get("cert_type"):
                 raise ValueError("The cert_type field is mandatory.")
             if identity.system.get("cert_type") not in CertType.__members__.values():
-                raise ValueError(f"The cert_type {identity.system.get('cert_type')} is invalid.")
+                # TODO: Raise ValueError once we solidify all cert_type values
+                logger.error("The cert_type %s is invalid.", identity.system.get("cert_type"))
```
thearifismail, line 123: IMO, should be `ValueError` and should go ahead and add `class CertType(str, Enum):
    HYPERVISOR = "hypervisor"
    RHUI = "rhui"
    SAM = "sam"
    SATELLITE = "satellite"
    SYSTEM = "system"`

app/auth/identity.py
```diff
@@ -31,6 +31,9 @@ class AuthType(str, Enum):
 
 
 class CertType(str, Enum):
+    HYPERVISOR = "hypervisor"
+    RHUI = "RHUI"
```
ryandillinfelton, line 35: I thought we were going to keep these all lowercase for comparisons?

kruai: Oh yeah, we did say that. Will fix

thearifismail:
> I thought we were going to keep these all lowercase for comparisons?

No, `RHUI` = RHUI` is correct because the incoming messages have "RHUI"

ryandillinfelton:
> > I thought we were going to keep these all lowercase for comparisons?
> 
> No, `RHUI` = RHUI` is correct because the incoming messages have "RHUI"

The idea is that once we start failing incorrect cert types again we would lowercase the incoming value for comparison so that simple typo on a reporters part wont cause a bug. I figure rhui vs RHUI is all the same as far as we're concerned.

ryandillinfelton (approved):

thearifismail (approved):

tahmidefaz (approved):

app/auth/identity.py
kruai: What do you mean? Raising a ValueError would undo the entire point of this PR...

thearifismail (approved):

PR #883, opened 2021-04-20
log reporter with cert type value errors (and other value errors)

ryandillinfelton:
## Overview

No Jira for this one. Just adds the reporter to the log we create when we get a value error (like cert_type) while trying to create/update a host. This is specifically to help figure out the cert type bug.


thearifismail (approved):
LGTM

kruai (approved):
LGTM!

PR #888, opened 2021-04-27
Adding kafka consumer ssl configs

tahmidefaz:
## Overview

This PR is part of the work needeed for [this Jira](https://issues.redhat.com/browse/RHCLOUD-10353).
This is a similar PR to that of RedHatInsights/insights-advisor-service#293
The Kafka Consumer API doc can be found [here](https://kafka-python.readthedocs.io/en/master/apidoc/KafkaConsumer.html).

## PR Checklist

- [x] Links to related PRs

## Secure Coding Checklist

- [x] Input Validation

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -102,6 +102,14 @@ class Config:
         self.prometheus_pushgateway = os.environ.get("PROMETHEUS_PUSHGATEWAY", "localhost:9091")
         self.kubernetes_namespace = os.environ.get("NAMESPACE")
 
+        self.kafka_consumer_ssl = {}
+        self.kafka_ssl_enabled = os.environ.get("KAFKA_SSL_ENABLED", "false").lower() == "true"
+        if self.kafka_ssl_enabled:
+            self.kafka_consumer_ssl = {
+                "security_protocol": os.environ.get("KAFKA_SECURITY_PROTOCOL", "SSL").upper(),
+                "ssl_cafile": os.environ.get("KAFKA_SSL_CAFILE", "/opt/certs/kafka-cacert"),
+            }
+
         # https://kafka-python.readthedocs.io/en/master/apidoc/KafkaConsumer.html#kafka.KafkaConsumer
         self.kafka_consumer = {
             "request_timeout_ms": int(os.environ.get("KAFKA_CONSUMER_REQUEST_TIMEOUT_MS", "305000")),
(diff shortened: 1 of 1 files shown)


app/config.py
```diff
@@ -102,6 +102,14 @@ def __init__(self, runtime_environment):
...
         self.kubernetes_namespace = os.environ.get("NAMESPACE")
 
+        self.kafka_consumer_ssl = {}
+        self.kafka_ssl_enabled = os.environ.get("KAFKA_SSL_ENABLED", "false").lower() == "true"
+        if self.kafka_ssl_enabled:
+            self.kafka_consumer_ssl = {
+                "security_protocol": os.environ.get("KAFKA_SECURITY_PROTOCOL", "SSL").upper(),
+                "ssl_cafile": os.environ.get("KAFKA_SSL_CAFILE", "/opt/certs/kafka-cacert"),
+            }
+
```
kruai, line 112: I think it would be better if these values were always set, instead of them only existing if `KAFKA_SSL_ENABLED == true`. If SSL shouldn't be enabled, you should just be able to set "security_protocol" to "PLAINTEXT" and "ssl_cafile" to None.

kruai: I'm also not sure we need both an env var to specify that SSL is enabled, and an additional env var to specify "SSL". I think it would be better to just use `KAFKA_SECURITY_PROTOCOL` and `KAFKA_SSL_CAFILE`

app/config.py
```diff
@@ -114,6 +122,7 @@ def __init__(self, runtime_environment):
             "max_poll_interval_ms": int(os.environ.get("KAFKA_CONSUMER_MAX_POLL_INTERVAL_MS", "300000")),
             "session_timeout_ms": int(os.environ.get("KAFKA_CONSUMER_SESSION_TIMEOUT_MS", "10000")),
             "heartbeat_interval_ms": int(os.environ.get("KAFKA_CONSUMER_HEARTBEAT_INTERVAL_MS", "3000")),
+            **self.kafka_consumer_ssl,
```
kruai, line 125: ... And the values would be set here, instead of doing a var expansion:

```suggestion
            "security_protocol": os.environ.get("KAFKA_SECURITY_PROTOCOL", "PLAINTEXT"),
            "ssl_cafile": os.environ.get("KAFKA_SSL_CAFILE"),
```

kruai (changes requested):

app/config.py
tahmidefaz: Great suggestion, it simplifies this PR even more. Thanks!

kruai (approved):
LGTM!

thearifismail (approved):
LGTM

PR #908, opened 2021-05-14
⬆️ Bump reorder-python-imports pre-commit hook to use --py37-plus

Glutexo:
Since we are now using Python 3.8.

_reorder-python-imports_ supports _--py37-plus_ as its maximum.
```
  --py37-plus           Remove obsolete future imports (generator_stop). implies: --py22-plus, --py23-plus, --py26-plus, --py3-plus
```

_pyupgrade_’s most recent support is still _--py36-plus_, so no change there.
```
  --py36-plus
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
   rev: v1.6.0
   hooks:
   - id: reorder-python-imports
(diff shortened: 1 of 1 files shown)


ryandillinfelton (approved):
Thanks! LGTM

PR #909, opened 2021-05-14
Generic filtering

ryandillinfelton:
## Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/RHCLOUD-12869).

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


ryandillinfelton:
Okay, so the test failure if from an issue with a grafana dashboard that is unchanged in this PR. I think @kruai is working on that fix, but just for the sake of clarity, all test tests and linting pass for this PR.

kruai:
FYI: The failure in pr-head is caused by an issue with the Grafana dashboard. PR #918 fixes this.

tahmidefaz (changes requested):
@ryandillinfelton We should improve the usability around having a `system_profile_block_list.yaml`. Right now, if the file does not exist or if the file is empty, it throws errors.

ryandillinfelton:
Whoops I must not have added it to the commit > @ryandillinfelton We should improve the usability around having a `system_profile_block_list.yaml`. Right now, if the file does not exist or if the file is empty, it throws errors.

api/filtering/filtering.py
```diff
@@ -0,0 +1,130 @@
...
+logger = get_logger(__name__)
+
+NIL_STRING = "nil"
+NOT_NIL_STRING = "not_nil"
+OR_FIELDS = ("owner_id", "rhc_client_id", "host_type")
+
+
+def _boolean_filter(field_name, field_value):
+    if not field_value.lower() in ("true", "false"):
+        raise ValidationException("invalid field value")
```
kruai, line 21: I had originally left a comment on each ValidationException, but that would have been overwhelming and redundant. Basically, I don't think these exceptions provide enough information. Instead of:
`ValidationException("invalid field value")`
I'd much prefer to see something like:
`ValidationException(f"{field_value} is an invalid value for the {field_type} field {field_name}")`

I'd love more details in the other similar exceptions that are thrown, like `invalid filter field`, `wrong type for filter`, `filter key is invalid`, etc. We should help the user out as much as we can if we fail their request

api/system_profile.py
```diff
@@ -174,12 +173,11 @@ def get_sap_sids(search=None, tags=None, page=None, per_page=None, staleness=Non
...
-        if filter.get("system_profile"):
-            if filter["system_profile"].get("sap_system"):
-                hostfilter_and_variables += build_sap_system_filter(filter["system_profile"].get("sap_system"))
-            if filter["system_profile"].get("sap_sids"):
-                hostfilter_and_variables += build_sap_sids_filter(filter["system_profile"]["sap_sids"])
+    for key in filter:
+        if key == "system_profile":
+            hostfilter_and_variables += build_system_profile_filter(filter["system_profile"])
+        else:
+            raise ValidationException("filter key is invalid")
```
kruai, line 180: This block is exactly the same as L119-123; can we refactor this, or is that overkill?
I realize that there are other commonalities between the functions, but since you didn't change anything beyond the filters, those can wait.

api/tag.py
```diff
@@ -108,10 +106,7 @@ def get_tags(
 
     if filter:
         if filter.get("system_profile"):
-            if filter["system_profile"].get("sap_system"):
-                hostfilter_and_variables += build_sap_system_filter(filter["system_profile"].get("sap_system"))
-            if filter["system_profile"].get("sap_sids"):
-                hostfilter_and_variables += build_sap_sids_filter(filter["system_profile"]["sap_sids"])
+            hostfilter_and_variables += build_system_profile_filter(filter["system_profile"])
```
kruai, line 109: Same thing here

tests/test_api_hosts_get.py
```diff
@@ -982,35 +982,35 @@ def test_get_hosts_with_satellite_identity(db_create_host, api_get):
 
 
 # DISABLED. Query validation will be added back in a future PR
```
kruai, line 984: Since it's no longer disabled, this should be removed

tests/test_xjoin.py
```diff
@@ -2462,3 +2462,486 @@ def test_query_variables_system_profile(
...
+        graphql_query_empty_response,
+        graphql_tag_query_empty_response,
+        graphql_system_profile_sap_system_query_empty_response,
+        # graphql_system_profile_sap_sids_query_empty_response
+    )
+    endpoint_url_builders = (
+        build_hosts_url,
+        build_tags_url,
+        build_system_profile_sap_system_url,
+        # build_system_profile_sap_sids_url
```
kruai, line 2573: Why do each of these have their final line commented out? If there's no particular reason, let's remove them

tests/test_xjoin.py
```diff
@@ -2462,3 +2462,486 @@ def test_query_variables_system_profile(
...
```
kruai, line 2735: same Q here

tests/test_xjoin.py
```diff
@@ -2462,3 +2462,486 @@ def test_query_variables_system_profile(
...
+    suffixes = (
+        # bad operation
+        "[foo]=bar",
+        "[bar][]=bar",
+        "[eq][foo]=bar",
+        "[is]=bar",
+        "[lt]=bar",
+        "[gt]=bar",
+        "[lte]=bar",
+        "[gte]=bar",
```
kruai, line 2931: Is this test's purpose to check invalid values, invalid operations, or both? If it's just for values, I think the invalid operations are redundant here

kruai (changes requested):
I like the structure you've implemented! There are a few things that could be refactored, and some errors that could be a bit more descriptive, but in general things are looking good

tests/test_xjoin.py
ryandillinfelton: Both. Since both are dependent on the type of the field I though it'd make sense to roll them up together

tests/test_xjoin.py
ryandillinfelton: Removed. It was related to getting one test working on all endpoints at once. Should have just removed the lines once I gave up on making that part of this PR.

tests/test_xjoin.py
kruai: Ah, I forgot that the allowed operations are also dependent on the field type. No issue here, then

api/system_profile.py
ryandillinfelton: I think factoring this out to a function would lead to a weird kind of pattern. I think I'd rather refactor the system_profile stuff more broadly at the next opportunity though.

api/tag.py
ryandillinfelton: I do need to add the proper check here

tahmidefaz (changes requested):

app/__init__.py
```diff
@@ -70,6 +77,62 @@ def inventory_config():
...
+        raise Exception("system profile spec is invalid")
+
+    # determine if the field uses a custom filter
+    if field_name in custom_filter_fields:
+        return field_name
+
+    # determine if the string field supports wildcard queries
+    if field_type == "string" and props.get("description"):
+        if "wildcard" in props["description"].lower():
+            return "wildcard"
```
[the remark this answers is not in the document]
ryandillinfelton: Adding fields to the schema technically doesn't break spec with JSONschema but it could cause compatibility issues down the line. There is no requirement on how tools should handle custom fields, so it's possible that tooling could the field out entirely. So I'm partial to re-purposing  some of the annotation fields like description. My hope also is that having it work this way will encourage better documentation, since users will want to know if wildcards are supported on a field.

app/__init__.py
```diff
@@ -70,6 +77,62 @@ def inventory_config():
...
+        return _get_field_filter(None, props["items"])
+
+    return field_type
+
+
+def process_system_profile_spec():
+    system_profile_blocklist = []
+    with open(SYSTEM_PROFILE_BLOCK_LIST_FILE) as fp:
+        system_profile_blocklist = yaml.safe_load(fp)["fields"]
+        print(f"block list fields: {system_profile_blocklist}")
```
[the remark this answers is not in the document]
ryandillinfelton: Yes, I'll swap it to an info level log

app/__init__.py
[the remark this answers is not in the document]
kruai: I must have missed this in my review! It feels sketchy and kinda dangerous doing things this way; IMO changing something's description should not affect the functionality of the app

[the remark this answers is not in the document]
ryandillinfelton: I'll explore using a custom field. Since we're parsing it ourselves in this instance it might not make any difference. My plan as is was just to create robust documentation and a checklist for all inventory-schema PRs to insure that people have considered this.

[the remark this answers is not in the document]
ryandillinfelton: If we end up able to use custom fields in the JSONschema. Would y'all prefer we use a marker like `indexed: false` instead of the current block list?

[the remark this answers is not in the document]
tahmidefaz: @ryandillinfelton I think we should be able to use custom fields in our JSONschema. We already have one (`x-propertyNames`). I don't have any preferences as to the use of a marker.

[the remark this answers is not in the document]
ryandillinfelton: I swapped both over to using custom tags in JSONschema.
So to mark a field as supporting wildcard you just add:
`x-wildcard: true`
and to block a field from being indexed you add:
x-indexed: false`
I have it default to assuming fields are indexed and don't support wildcards.

Do these seem like good and intuitive names, and more readable than the description approach?

api/system_profile.py
kruai: Yes, that's fine!

api/filtering/filtering.py
ryandillinfelton: I think I've addressed this. LMK if you see any more places it can be improved.

app/__init__.py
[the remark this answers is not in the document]
kruai: I'm cool with those names, at least :)

api/filtering/filtering.py
kruai: Looks good!

kruai (approved):
I think this looks good now!

ryandillinfelton:
> Looks good!  
> You also need to open a PR in `inventory-schemas` so that both versions of the system_profile schema are the same.

opened a PR on inventory schemas. Also includes some documentation on the steps to add a new field
https://github.com/RedHatInsights/inventory-schemas/pull/70

PR #911, opened 2021-05-19
ESSNTL-746: subtask - add inventory schemas updater

RonnyPfannschmidt:
* makefile based
* uses git clone
* supports pre-created checkout
* records the commit id we used

This PR is being created to address [ESSNTL-746](https://issues.redhat.com/browse/ESSNTL-746).

## PR Checklist

- [x] Tests: edge cases
- [x] Documentation, if this PR changes the way other services interact with host inventory
- [x] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


kruai (commented):
I'm very interested in this! If we went in this direction, would we have to manually run `update_schema`, or can we automate that as well?

RonnyPfannschmidt:
@kruai yes, this only makes the mechanics more foolproof for now

we still copy the file content so the submodules are not in general required
a followup would be a sheduled/triggered github action that updates the submodule data regular

.gitignore
```diff
@@ -9,5 +9,7 @@ tags
 .pytest_cache/
 /.sonar/sonar-scanner.properties
 /.scannerwork/
+.idea/
 
 coverage.xml
+.hypothesis/
```
kruai, line 15: Were these for some changes that weren't committed?

kruai (changes requested):

.gitignore
RonnyPfannschmidt: indeed, those where accidentally added while finishing the rebases, should i move them to a different pr

kruai: Yeah - if the related files are in the other PR, this change should be there instead

RonnyPfannschmidt: there are no related files .hypothesis was something visual studio code adds and .idea is from pycharm

kruai: OK. I haven't seen those files in mine, but I guess I'm not against this gitignore addition

kruai (dismissed):

tahmidefaz (dismissed):
Looks good! If we can make gh actions run `update_schema` periodically and open PRs, that'd be cool! :+1:

.gitignore
ryandillinfelton: I'm gonna disagree here. We should only have rules in the gitignore that relate directly to this specific project. IDE file extensions and such should be added to your own global gitignore rather than the gitignore of each project you contribute to.

ryandillinfelton (changes requested):
Sorry, a bit of a picky change request, but I'd like to keep up with best practices in this case. Everything else looks good to me though, so it should be an easy merge once this small fix is made.

ryandillinfelton:
I'm not convinced we need to add a submodule for this. If we want it to be automated through a github action already, and we aren't using the file from the submodule directly, why not update the file in the action itself? We could have the job checkout the schema repo and copy the latest version over.

RonnyPfannschmidt:
@ryandillinfelton however if we set the complete logic out in a action, it cant be nicely used/integrated

like for example tracking a branch of the upstream

also we loose the direct information about what commit a update was taken from

for me the key feature of the subrepo being possibly usable is that it is extremely easy, straightforward and well tested and works in all envs

kruai (approved):
LGTM! I especially think the SP commit ID is a good idea.

ryandillinfelton (approved):

PR #914, opened 2021-05-25
Make cert_type and auth_type case-insensitive

kruai:
## Overview

This PR is being created to address [RHCLOUD-14391](https://issues.redhat.com/browse/RHCLOUD-14391).
Some host-ingress messages are failing because the `cert_type` is set to "RHUI", while Identity allows only lowercase values. This PR changes `cert_type` and `auth_type` to be case-insensitive.

## PR Checklist

- [x] Keep PR title short, ideally under 72 characters
- [x] Descriptive comments provided in complex code blocks
- [x] Tests: validate optimal/expected output
- [x] Tests: validate exceptions and failure scenarios
- [x] Tests: edge cases
- [ ] Documentation, if this PR changes the way other services interact with host inventory
- [ ] Links to related PRs

## Secure Coding Practices Checklist GitHub Link

- https://github.com/RedHatInsights/secure-coding-checklist

## Secure Coding Checklist

- [x] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [x] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [x] General Coding Practices

PR #916, opened 2021-05-26
Fix AttributeError when `auth_type` is None

kruai:
## Overview

This PR fixes a bug that I introduced in #914. When an Identity has `auth_type == None`, the app hits an AttributeError.
This resolves that issue, and adds a test that verifies as much.


RonnyPfannschmidt (approved):

tahmidefaz (approved):
Looks good :+1:

ryandillinfelton (approved):

PR #919, opened 2021-06-02
Use /bin/sh -c on container startup commands

bsquizz:
This seems to fix the certifi import issues when running in kubernetes. I'm not sure why but I suspect there's an issue with the virtual environment when `python` is called directly.


thearifismail (approved):
lgtm

PR #941, opened 2021-06-28
Set display_name to FQDN on update, if it's set to the ID

kruai:
# Overview

This PR is being created to address [ESSNTL-824](https://issues.redhat.com/browse/ESSNTL-824).
If we update a Host and its old display_name is set to its ID, we should set the display_name to the FQDN.
This is one of the features that was discussed after a bunch of email exchanges. Without this change, the display_name is saved to the ID upon a Host's creation, so if the FQDN isn't specified at that time, it will never default.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -60,7 +60,7 @@ def _set_display_name_on_save(context):
     the id exists and can be used as the display_name if necessary.
     """
     params = context.get_current_parameters()
-    if not params["display_name"]:
+    if not params["display_name"] or params["display_name"] == str(params["id"]):
         return params["canonical_facts"].get("fqdn") or params["id"]
 
 
@@ -253,7 +253,7 @@ class Host(LimitedHost):
     def update_display_name(self, input_display_name):
         if input_display_name:
             self.display_name = input_display_name
-        elif not self.display_name:
+        elif not self.display_name or self.display_name == str(self.id):
             # This is the case where the display_name is not set on the
             # existing host record and the input host does not have it set
             if "fqdn" in self.canonical_facts:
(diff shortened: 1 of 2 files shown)


thearifismail (approved):
LGTM

app/models.py
```diff
@@ -60,7 +60,7 @@ def _set_display_name_on_save(context):
     the id exists and can be used as the display_name if necessary.
     """
     params = context.get_current_parameters()
-    if not params["display_name"]:
+    if not params["display_name"] or params["display_name"] == str(params["id"]):
         return params["canonical_facts"].get("fqdn") or params["id"]
```
ryandillinfelton, line 64: It still looks like the display name will be set to it's id as a fallback if there is no FQDN. I don't think that's necessarily a bad thing, but I'd like to be clear on if we're okay with the fact this won't get rid of every host with it's ID as it's name.

ryandillinfelton (approved):
I'm approving because I think it does what it's supposed to do. I left a comment clarifying my concern. If it is in fact no big deal then I'm happy to have this merged.

app/models.py
kruai: Yes, that's correct. Before this PR, this FQDN fallback would only happen upon Host creation, since the display_name is set to the ID afterwards. The Host ID is still a fine fallback for now, if the display_name and FQDN continue to not be sent

PR #945, opened 2021-07-01
Add separate metrics server for API

kruai:
# Overview

This PR is being created as part of [ESSNTL-761](https://issues.redhat.com/browse/ESSNTL-761).
It makes it so the metrics endpoint is separate from the other API traffic, and is configurable via Clowder.
I also added a small update to the MQ service's config to make its metrics path configurable as well (only the port was configured before).

## PR Checklist

- [x] Keep PR title short, ideally under 72 characters
- [ ] Descriptive comments provided in complex code blocks
- [ ] Tests: validate optimal/expected output
- [ ] Tests: validate exceptions and failure scenarios
- [ ] Tests: edge cases
- [ ] Documentation, if this PR changes the way other services interact with host inventory
- [ ] Links to related PRs

## Secure Coding Practices Documentation Reference

You can find documentation on this checklist [here](https://github.com/RedHatInsights/secure-coding-checklist).

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [x] General Coding Practices

PR #946, opened 2021-07-01
Hotfix: Remove endopint param from MQ metrics server command

kruai:
The MQ config was busted by my last PR: https://github.com/RedHatInsights/insights-host-inventory/pull/945
I incorrectly assumed that both `start_http_server()` calls would take the same params. I'm reverting that change here.


thearifismail (approved):
why is metricPath not needed?  Rather how is the prometheusServer

kruai:
> why is metricPath not needed? Rather how is the prometheusServer

`endpoint` is an optional param for `PrometheusMetrics.start_http_server()`, but not for `prometheus_client.start_http_server()`. I didn't realize they were 2 different calls.

PR #951, opened 2021-07-09
Inventory Pendo Syncher [ESSNTL-613]

tahmidefaz:
# Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/ESSNTL-613).

This PR implements the Pendo Syncher cron job to sync hosts per account data with pendo.

On a high level, this does the following two things:

- Runs a query on the HBI DB that is equivalent to the SQL: 
  - `SELECT account, count(id) from hosts GROUP BY account ORDER BY account;`
- Sends a POST request to Pendo with the request body formatted by the following syntax:
```
[{
  "accountId: "hbi-account-1",
  "values": {
    "hostCount": count
  }
},
{
  "accountId: "hbi-account-2",
  "values": {
    "hostCount": count
  }
}...]
```

The `hostCount` is a custom Pendo account metadata what we added using an admin account.

To successfully send a request to Pendo, the following environment variables with the correct values must be present:

- `PENDO_SYNC_ACTIVE`
- `PENDO_INTEGRATION_KEY`

Information on the endpoint used and other functionalities of the Pendo API can be [found here](https://developers.pendo.io/docs/?python#set-value-for-a-set-of-agent-or-custom-fields).

## PR Checklist

- [ ] Tests: edge cases
- [x] Links to helpful references

## Secure Coding Checklist

- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


lib/pendo_sync_helpers.py
```diff
@@ -0,0 +1,75 @@
...
+
+def _process_response(pendo_response, logger):
+    if pendo_response.status_code != 200:
+        raise Exception(f"Pendo responded with status {pendo_response.status_code}")
+
+    resp_data = pendo_response.json()
+
+    logger.info(
+        f"Pendo Data Sent. Total: {resp_data['total']} Updated: {resp_data['updated']}, Failed, {resp_data['failed']}"
+    )
```
ryandillinfelton, line 29: This spot could use a little layout touch up. Everything works fine, just a little cramped visually

lib/pendo_sync_helpers.py
```diff
@@ -0,0 +1,75 @@
...
+        _make_request(request_body, config, logger)
+
+
+def pendo_sync(select_query, limit, config, logger, interrupt=lambda: False):
+    query = select_query.group_by(Host.account).order_by(Host.account)
+    account_list = query.limit(limit).all()
+
+    while len(account_list) > 0 and not interrupt():
+        _pendo_update(account_list, config, logger)
+        account_list = query.filter(Host.account > account_list[-1].account).limit(limit).all()
```
ryandillinfelton, line 75: This note is also kind of to myself when I re-review. This looks like it'd work but I'm not sure it seems like kind of a strange way to do it. I need to dust off my SQL alchemy and take a closer look.

tahmidefaz: @ryandillinfelton this part is somewhat similar to what's being used in `host_synchronizer`. But if you can think of a better/simpler way to do this, let me know for sure and I will change it :+1:

pendo_syncher.py
```diff
@@ -0,0 +1,89 @@
...
+RUNTIME_ENVIRONMENT = RuntimeEnvironment.PENDO_JOB
+
+
+def _init_config():
+    config = Config(RUNTIME_ENVIRONMENT)
+    config.log_configuration()
+    return config
+
+
+def _init_db(config):
```
ryandillinfelton, line 38: Not sure I understand what the distinction is between functions that are put in the helpers file, vs functions put in this file. I understand why main and run are in here, it just looks like a few more could be factored into helpers if you're going to divide it anyways.

tahmidefaz: @ryandillinfelton right now, the main file initializing everything for the scripts, such as db, metrics etc. The functions in the `pendo_sync_helpers` currently does the actual work of querying, making the request etc. Does this clear it up a little? Maybe we shouldn't call it `pendo_sync_helpers`... :thinking:

lib/pendo_sync_helpers.py
tahmidefaz: @ryandillinfelton updated this to only label and show three number, e.g. `total`, `updated`, `failed`. Lmk, if this works for you!

pendo_syncher.py
ryandillinfelton: I would almost expect it to be the opposite. I'm a little partial to just combining stuff, especially since the helper functions are only going to be used in this one script for the foreseeable future. I don't think it would be too long either. I'm open to other ideas though if you find that disagreeable.

lib/pendo_sync_helpers.py
ryandillinfelton: I should've been more specific that I literally just wanted a newline in there to uncramp things. It's all looking good now though.

pendo_syncher.py
tahmidefaz: @ryandillinfelton sure thing! They should all be in the same file now

tests/test_pendo_syncher.py
```diff
@@ -0,0 +1,95 @@
...
+    mock_request_function_mock.assert_called_once_with(request_body, config, mock.ANY)
+
+
+def test_pendo_syncher_response_process(mocker, db_create_host):
+    db_create_host()
+
+    response_content = {"total": 1, "updated": 1, "failed": 0}
+
+    mock_response = MockResponseObject()
+    mock_response.status_code = 200
```
ryandillinfelton, line 35: This seems redundant since it's already set in the initialization

tests/test_pendo_syncher.py
```diff
@@ -0,0 +1,95 @@
...
+from pendo_syncher import run as pendo_syncher_run
+from tests.helpers.test_utils import MockResponseObject
+
+
+def test_pendo_syncher_request_body(mocker, db_create_host):
+    mock_request_function_mock = mocker.patch("pendo_syncher._make_request")
+
+    host = db_create_host()
+
+    request_body = [{"accountId": host.account, "values": {"hostCount": 1}}]
```
ryandillinfelton, line 18: Being picky, but I think this could be one line without sacrificing readability. I won't request changes for this alone though.

tests/test_pendo_syncher.py
```diff
@@ -0,0 +1,95 @@
...
+    request_session_post_mock = mocker.patch("pendo_syncher.Session.post")
+    request_session_post_mock.side_effect = mock_response
+
+    config = Config(RuntimeEnvironment.PENDO_JOB)
+    config.pendo_sync_active = True
+
+    mock_pendo_failure = mocker.patch("pendo_syncher.pendo_failure")
+    pendo_syncher_run(config, mock.Mock(), db.session, shutdown_handler=mock.Mock(**{"shut_down.return_value": False}))
+
+    mock_pendo_failure.assert_called_once()
```
ryandillinfelton, line 65: I'd like to see this check the error message that pendo_failure is passed to make sure that the failure is cause by the expected issue.

ryandillinfelton (changes requested):
Requesting changes because I'd like the tests to be a little more intentional. Otherwise everything looks good to me outside of a few nitpicks.

tests/test_pendo_syncher.py
tahmidefaz: @ryandillinfelton it was a little tricky, but I changes this to check for the failure message as cleanly as possible.

pendo_syncher.py
```diff
@@ -0,0 +1,157 @@
...
+    request_body = []
+    for account_info in account_list:
+        account_number, host_count = account_info
+
+        request_data = {"accountId": account_number, "values": {"hostCount": host_count}}
+        request_body.append(request_data)
+
+        logger.debug(f"Account Number: {account_number}, Host Count: {host_count}")
+    request_body = json.dumps(request_body)
+    if config.pendo_sync_active:
```
kruai, line 106: Shouldn't this be higher up? As it stands, this code will always do the processing even if the sync isn't active, but stop just short of actually sending the data. It might make sense for this logic to be in `main()`, `run()`, or `pendo_sync()`, unless I'm misunderstanding the variable's purpose.

kruai (changes requested):
I just have one note so far. There's also a lot of redundancy in the tests, and it would be great if we could refactor them, but I wouldn't want to hold it up on that.

pendo_syncher.py
tahmidefaz: @kruai I actually added this variable to do just that - stop just short of actually sending the data to pendo. I was thinking that initially we will initialize this cron job with `pendo_sync_active=False` and verify all the pre-processing is working great in the pods and then finally set it to `True`. I want to avoid sending bulk of invalid data to pendo at all cost, because doing so might make it very difficult for us to revert those changes. Lmk, what you think about it!

kruai: OK, I can understand that. I thought the purpose was similar to the `REAPER_SUSPEND` variable or whatever it was called. But yes, it makes sense to verify that everything's working first.

kruai (approved):
LGTM

tests/test_pendo_syncher.py
ryandillinfelton: I was hoping you could do it for all of the failure cases

tahmidefaz: Ah, you meant all of them!

tahmidefaz: @ryandillinfelton the other exception messages are different then this `403`, since a generic exception message won't be provided unlike the case we had here. Anyways, I added a exception string check for one additional test. The test for the `RetryFailure` doesn't provide an exception message by default, so, I refrained from adding a similar check for that one.

ryandillinfelton (approved):
LGTM!

kruai (approved):
Still LGTM!

PR #953, opened 2021-07-13
Upgrade Marshmallow to 3.8.0

kruai:
# Overview

This PR is being created to address [ESSNTL-890](https://issues.redhat.com/browse/ESSNTL-890).
It upgrades our version of Marshmallow to 3.8.0, and fixes many breaking changes that were introduced since version 2.x. This upgrade will make it simpler to validate certain field types without adding unnecessary custom code (such as IP addresses, for instance).

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -229,7 +229,7 @@ def _emit_patch_event(serialized_host, host_id, insights_id):
 @metrics.api_request_time.time()
 def patch_by_id(host_id_list, body):
     try:
-        validated_patch_host_data = PatchHostSchema(strict=True).load(body).data
+        validated_patch_host_data = PatchHostSchema().load(body)
     except ValidationError as e:
         logger.exception(f"Input validation error while patching host: {host_id_list} - {body}")
         return ({"status": 400, "title": "Bad Request", "detail": str(e.messages), "type": "unknown"}, 400)
--- a/app/models.py
+++ b/app/models.py
@@ -11,6 +11,7 @@ from flask_sqlalchemy import SQLAlchemy
 from jsonschema import RefResolver
 from jsonschema import validate as jsonschema_validate
 from jsonschema import ValidationError as JsonSchemaValidationError
+from marshmallow import EXCLUDE
 from marshmallow import fields
 from marshmallow import post_load
 from marshmallow import pre_load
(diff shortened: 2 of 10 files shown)


api/host.py
```diff
@@ -229,7 +229,7 @@ def _emit_patch_event(serialized_host, host_id, insights_id):
 @metrics.api_request_time.time()
 def patch_by_id(host_id_list, body):
     try:
-        validated_patch_host_data = PatchHostSchema(strict=True).load(body).data
```
ryandillinfelton, line 232: The strict removal looks good since it's implicit now. I don't follow the removal of `.data` though. It looks like .data still does the same thing. Can I ask for an explanation?

kruai: `.data` actually breaks, because a tuple is no longer returned. This change is documented here: https://marshmallow.readthedocs.io/en/stable/upgrading.html#schemas-are-always-strict

app/models.py
```diff
@@ -483,8 +493,7 @@ def validate_tags(self, tags):
 
     @staticmethod
     def _validate_tags_list(tags):
-        TagsSchema(many=True, strict=True).validate(tags)
-        return True
+        return TagsSchema(many=True).load(tags)
```
ryandillinfelton, line 496: This doesn't look right. It will still raise a validation error, but I don't think it will ever return `True` since `.load()` returns the data

app/queue/events.py
```diff
@@ -119,8 +119,8 @@ def build_event(event_type, host, **kwargs):
     with event_serialization_time.labels(event_type.name).time():
         build = EVENT_TYPE_MAP[event_type]
         schema, event = build(event_type, host, **kwargs)
-        result = schema(strict=True).dumps(event)
-        return result.data
+        result = schema().dumps(event)
+        return result
```
ryandillinfelton, line 123: Same question about removing `.data`

ryandillinfelton: OH NVM

api/host.py
ryandillinfelton: I figured it out on a comment below. resolving

ryandillinfelton:
Okay, Everything looks pretty good. If the one comment I left unresolved needs to be addressed then I'll request changes, if it's a false alarm then I'm ready to approve.

app/models.py
kruai: It turns out that no value needs to be thrown by the `validate_tags()` method anyway, so I don't think it matters. But for consistency's sake, I've modified this to return True again.

ryandillinfelton (approved):
LGTM

app/models.py
```diff
@@ -416,12 +417,18 @@ class FactsSchema(MarshmallowSchema):
 
 
 class TagsSchema(MarshmallowSchema):
+    class Meta:
```
thearifismail, line 420: What triggered adding this class?

kruai: Correct - I only wanted to add it as necessary. Adding this class's contents makes it so if fields are included in the object that aren't defined in the schema, they are simply ignored (which is the default in the older versions of Marshmallow). Without it, the new default behavior is to throw a validation error (not desired).

thearifismail (commented):
Apart from `Class Meta:` question, LGTM

thearifismail (approved):

PR #948, opened 2021-07-06
Casefold canonical facts

kruai:
# Overview

This PR is being created to address [ESSNTL-772](https://issues.redhat.com/browse/ESSNTL-772).
Some duplication is being caused by reporters not agreeing on capitalization. This PR does a casefold on the canonical facts fields upon deserialization; that's essentially a lowercase in most situations, and is designed to remove case differences from the equation. 

Since deserialization happens very frequently, I want to make the `_recursive_casefold` function as performant as possible. If you can think of a way to make it even faster, please let me know!

## PR Checklist

- [x] Keep PR title short, ideally under 72 characters
- [ ] Descriptive comments provided in complex code blocks
- [x] Tests: validate optimal/expected output
- [ ] Tests: validate exceptions and failure scenarios
- [ ] Tests: edge cases
- [ ] Documentation, if this PR changes the way other services interact with host inventory
- [ ] Links to related PRs

## Secure Coding Practices Documentation Reference

You can find documentation on this checklist [here](https://github.com/RedHatInsights/secure-coding-checklist).

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #956, opened 2021-07-19
Casefold canonical fact filters

kruai:
# Overview

This PR is being created to address [ESSNTL-893](https://issues.redhat.com/browse/ESSNTL-893).
My previous PR (#948) attempted to casefold the canonical facts, in order to eliminate case-sensitivity in our identifiers, and thus reduce duplication. However, I didn't account for our filter logic not casefolding as well, so this is meant to rectify that.

## PR Checklist

- [ ] Tests: edge cases
- [x] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -106,11 +106,11 @@ def get_host_list(
     try:
         host_list, total, additional_fields = get_host_list(
             display_name,
-            fqdn.casefold(),
+            fqdn.casefold() if fqdn else None,
             hostname_or_id,
-            insights_id.casefold(),
-            provider_id.casefold(),
-            provider_type.casefold(),
+            insights_id.casefold() if insights_id else None,
+            provider_id.casefold() if provider_id else None,
+            provider_type.casefold() if provider_type else None,
             tags,
             page,
             per_page,
--- a/api/host.py
+++ b/api/host.py
@@ -44,7 +44,6 @@ from app.serialization import serialize_host
 from app.serialization import serialize_host_system_profile
 from app.utils import Tag
 from lib.host_delete import delete_hosts
-from lib.host_repository import AddHostResult
 from lib.host_repository import find_existing_host
 from lib.host_repository import find_non_culled_hosts
 from lib.host_repository import update_query_for_owner_id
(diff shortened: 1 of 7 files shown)


api/host.py
```diff
@@ -59,13 +58,6 @@
 logger = get_logger(__name__)
 
 
-def _convert_host_results_to_http_status(result):
-    if result == AddHostResult.created:
-        return 201
-    else:
-        return 200
-
-
```
kruai, line 68: FYI, I removed this because it isn't used anywhere. Figured I'd clean up while I'm thinking about it.

api/host.py
```diff
@@ -114,11 +106,11 @@ def get_host_list(
     try:
         host_list, total, additional_fields = get_host_list(
             display_name,
-            fqdn,
+            fqdn.casefold() if fqdn else None,
             hostname_or_id,
```
ryandillinfelton, line 110: hostname_or_id here doesn't need to be casefolded?

kruai: I didn't want to break `display_name` filtering since that's still case-sensitive, so I casefolded the FQDN part specifically here: https://github.com/RedHatInsights/insights-host-inventory/blob/ad6073f3d353eefdfe5a6427a29108435c973790/api/host_query_db.py#L139

However, it looks like the xjoin implementation already compares `display_name` as lowercase; maybe I should just do the same thing here?

ryandillinfelton: It only works on xjoin because we made the search case insensitive there. I think you're right to leave this as is on the database since if we only lowercase the search term it will just cause a lot of queries to miss.

kruai: OK, resolving this thread then!

tests/test_xjoin.py
```diff
@@ -175,7 +175,7 @@ def test_query_variables_hostname_or_id_uuid(mocker, query_source_xjoin, graphql
 
 
 def test_query_variables_insights_id(mocker, query_source_xjoin, graphql_query_empty_response, api_get):
-    insights_id = generate_uuid()
+    insights_id = generate_uuid().upper()
```
ryandillinfelton, line 178: Why upper for just UUIDs? I suppose that's how I'm used to seeing them, but IIRC the last change just casefolded everything to lowercase right?

kruai: Correct. I made this one uppercase so that I could verify that it's casefolded later (L192).

ryandillinfelton: Oh my god this is a test :facepalm:

ryandillinfelton (approved):

thearifismail (approved):
LGTM

PR #958, opened 2021-07-20
return per_reporter_staleness data with host search results

thearifismail:
# Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/ESSNTL-641).

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/serialization.py
+++ b/app/serialization.py
@@ -35,6 +35,7 @@ DEFAULT_FIELDS = (
     "ansible_host",
     "facts",
     "reporter",
+    "per_reporter_staleness",
     "stale_timestamp",
     "stale_warning_timestamp",
     "culled_timestamp",
@@ -105,6 +106,8 @@ def serialize_host(host, staleness_timestamps, fields=DEFAULT_FIELDS):
         serialized_host["facts"] = _serialize_facts(host.facts)
     if "reporter" in fields:
         serialized_host["reporter"] = host.reporter
+    if "per_reporter_staleness" in fields:
+        serialized_host["per_reporter_staleness"] = host.per_reporter_staleness
     if "stale_timestamp" in fields:
         serialized_host["stale_timestamp"] = stale_timestamp and _serialize_datetime(stale_timestamp)
     if "stale_warning_timestamp" in fields:
--- a/tests/test_api_hosts_get.py
+++ b/tests/test_api_hosts_get.py
@@ -43,9 +43,14 @@ from tests.helpers.test_utils import SYSTEM_IDENTITY
 from tests.helpers.test_utils import USER_IDENTITY
 
 
-def test_query_all(mq_create_three_specific_hosts, api_get, subtests):
+def test_query_all(mq_create_three_specific_hosts, api_get, subtests, mocker):
     created_hosts = mq_create_three_specific_hosts
     expected_host_list = build_expected_host_list(created_hosts)
+
+    for host in expected_host_list:
+        host["per_reporter_staleness"] = {
+            host["reporter"]: {"check_in_succeeded": True, "last_check_in": mocker.ANY, "stale_timestamp": mocker.ANY}
+        }
 
     response_status, response_data = api_get(HOST_URL)
 
@@ -55,9 +60,14 @@ def test_query_all(mq_create_three_specific_hosts, api_get, subtests):
     api_base_pagination_test(api_get, subtests, HOST_URL, expected_total=len(created_hosts))
 
 
-def test_query_using_display_name(mq_create_three_specific_hosts, api_get):
+def test_query_using_display_name(mq_create_three_specific_hosts, api_get, mocker):
     created_hosts = mq_create_three_specific_hosts
     expected_host_list = build_expected_host_list([created_hosts[0]])
+
+    for host in expected_host_list:
+        host["per_reporter_staleness"] = {
+            host["reporter"]: {"check_in_succeeded": True, "last_check_in": mocker.ANY, "stale_timestamp": mocker.ANY}
+        }
 
     url = build_hosts_url(query=f"?display_name={created_hosts[0].display_name}")
     response_status, response_data = api_get(url)
@@ -67,9 +77,14 @@ def test_query_using_display_name(mq_create_three_specific_hosts, api_get):
     assert expected_host_list == response_data["results"]
 
 
-def test_query_using_fqdn_two_results(mq_create_three_specific_hosts, api_get):
+def test_query_using_fqdn_two_results(mq_create_three_specific_hosts, api_get, mocker):
     created_hosts = mq_create_three_specific_hosts
     expected_host_list = build_expected_host_list([created_hosts[0], created_hosts[1]])
+
+    for host in expected_host_list:
+        host["per_reporter_staleness"] = {
+            host["reporter"]: {"check_in_succeeded": True, "last_check_in": mocker.ANY, "stale_timestamp": mocker.ANY}
+        }
 
     url = build_hosts_url(query=f"?fqdn={created_hosts[0].fqdn}")
     response_status, response_data = api_get(url)
@@ -79,9 +94,14 @@ def test_query_using_fqdn_two_results(mq_create_three_specific_hosts, api_get):
     assert expected_host_list == response_data["results"]
 
 
-def test_query_using_fqdn_one_result(mq_create_three_specific_hosts, api_get):
+def test_query_using_fqdn_one_result(mq_create_three_specific_hosts, api_get, mocker):
     created_hosts = mq_create_three_specific_hosts
     expected_host_list = build_expected_host_list([created_hosts[2]])
+
+    for host in expected_host_list:
+        host["per_reporter_staleness"] = {
+            host["reporter"]: {"check_in_succeeded": True, "last_check_in": mocker.ANY, "stale_timestamp": mocker.ANY}
+        }
 
     url = build_hosts_url(query=f"?fqdn={created_hosts[2].fqdn}")
     response_status, response_data = api_get(url)
@@ -99,9 +119,14 @@ def test_query_using_non_existent_fqdn(api_get):
     assert len(response_data["results"]) == 0
 
 
-def test_query_using_display_name_substring(mq_create_three_specific_hosts, api_get, subtests):
+def test_query_using_display_name_substring(mq_create_three_specific_hosts, api_get, subtests, mocker):
     created_hosts = mq_create_three_specific_hosts
     expected_host_list = build_expected_host_list(created_hosts)
+
+    for host in expected_host_list:
+        host["per_reporter_staleness"] = {
+            host["reporter"]: {"check_in_succeeded": True, "last_check_in": mocker.ANY, "stale_timestamp": mocker.ANY}
+        }
 
     host_name_substr = created_hosts[0].display_name[:4]
 
--- a/tests/test_xjoin.py
+++ b/tests/test_xjoin.py
@@ -647,7 +647,7 @@ def test_query_variables_staleness_with_search(
     )
 
 
-def test_response_processed_properly(query_source_xjoin, graphql_query_with_response, api_get):
+def test_response_processed_properly(query_source_xjoin, graphql_query_with_response, api_get, mocker):
     response_status, response_data = api_get(HOST_URL)
 
     assert response_status == 200
@@ -672,6 +672,9 @@ def test_response_processed_properly(query_source_xjoin, graphql_query_with_resp
                 "insights_id": "a58c53e0-8000-4384-b902-c70b69faacc5",
                 "stale_timestamp": "2020-02-10T08:07:03.354307+00:00",
                 "reporter": "puptoo",
+                "per_reporter_staleness": {
+                    "puptoo": {"check_in_succeeded": True, "last_check_in": mocker.ANY, "stale_timestamp": mocker.ANY}
+                },
                 "rhel_machine_id": None,
                 "subscription_manager_id": None,
                 "bios_uuid": None,
@@ -696,6 +699,9 @@ def test_response_processed_properly(query_source_xjoin, graphql_query_with_resp
                 "insights_id": "17c52679-f0b9-4e9b-9bac-a3c7fae5070c",
                 "stale_timestamp": "2020-01-10T08:07:03.354307+00:00",
                 "reporter": "puptoo",
+                "per_reporter_staleness": {
+                    "puptoo": {"check_in_succeeded": True, "last_check_in": mocker.ANY, "stale_timestamp": mocker.ANY}
+                },
                 "rhel_machine_id": None,
                 "subscription_manager_id": None,
                 "bios_uuid": None,
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -1253,7 +1253,12 @@ class SerializationSerializeHostCompoundTestCase(SerializationSerializeHostBaseT
         }
         host = Host(**host_init_data)
 
-        host_attr_data = {"id": uuid4(), "created_on": datetime.utcnow(), "modified_on": datetime.utcnow()}
+        host_attr_data = {
+            "id": uuid4(),
+            "created_on": datetime.utcnow(),
+            "modified_on": datetime.utcnow(),
+            "per_reporter_staleness": host.per_reporter_staleness,
+        }
         for k, v in host_attr_data.items():
             setattr(host, k, v)
 
@@ -1281,6 +1286,7 @@ class SerializationSerializeHostCompoundTestCase(SerializationSerializeHostBaseT
             "culled_timestamp": self._timestamp_to_str(
                 self._add_days(host_init_data["stale_timestamp"], config.culled_offset_delta.days)
             ),
+            "per_reporter_staleness": host_attr_data["per_reporter_staleness"],
         }
         self.assertEqual(expected, actual)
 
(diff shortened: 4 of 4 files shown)


tests/test_api_hosts_get.py
```diff
@@ -43,10 +43,15 @@
...
+def test_query_all(mq_create_three_specific_hosts, api_get, subtests, mocker):
     created_hosts = mq_create_three_specific_hosts
     expected_host_list = build_expected_host_list(created_hosts)
 
+    for host in expected_host_list:
+        host["per_reporter_staleness"] = {
+            host["reporter"]: {"check_in_succeeded": True, "last_check_in": mocker.ANY, "stale_timestamp": mocker.ANY}
+        }
+
     response_status, response_data = api_get(HOST_URL)
```
kruai, line 55: This code block is repeated each time `build_expected_host_list()` is called. I would think we'd want `build_expected_host_list()` to return those fields as well, but I'm assuming they're not set by `mq_create_three_specific_hosts` (which uses `minimal_host`).

If all that makes sense, can we add the per-reporter staleness to the `minimal_host` implementation instead?

kruai (changes requested):

ryandillinfelton:
Have you tested this with xjoin? Because I don't believe we ingest per_reporter_staleness data into elasticsearch yet.

tests/test_api_hosts_get.py
thearifismail: Changing `build_expected_host_list()` requires more changes to other tests/test calls.  Moving the repetitive calls to a function looks much better than before.  Changing `minimal_host` to add `per_reporter_staleness` does not help because `add_host()` function adds them dynamically.

thearifismail:
> Have you tested this with xjoin? Because I don't believe we ingest per_reporter_staleness data into elasticsearch yet.

Not with `xjoin enabled`.  If hosts in the `xjoin elasticsearch` have `per_reporter_staleness`, then it should work because the change is implemented by adding `per_reporter_staleness` to `DEFAULT_FIELDS` in "serialization.py".  The change did required fixing a test in `test_xjoin.py`

ryandillinfelton:
> > Have you tested this with xjoin? Because I don't believe we ingest per_reporter_staleness data into elasticsearch yet.
> 
> Not with `xjoin enabled`. If hosts in the `xjoin elasticsearch` have `per_reporter_staleness`, then it should work because the change is implemented by adding `per_reporter_staleness` to `DEFAULT_FIELDS` in "serialization.py". The change did required fixing a test in `test_xjoin.py`

I don't suspect that this is written wrong or will break anything, but I still don't feel comfortable merging a pull request where the primary new feature is untested.

thearifismail:
@ryandillinfelton yes, let me test it out.  https://github.com/RedHatInsights/insights-host-inventory/pull/958#issuecomment-884441105

thearifismail:
@ryandillinfelton yes running with `xjoin-enabled` and got the `per_reporter_staleness` databack.  Here is the screenshot.
per_reporter_staleness_data_from_xjoin_search

thearifismail:
@kruai Thanks, yes adding `per_reporter_staleness` to `SerializedHostSchema` worked and simplified the code.

tests/test_xjoin.py
```diff
@@ -672,6 +672,9 @@ def test_response_processed_properly(query_source_xjoin, graphql_query_with_resp
                 "insights_id": "a58c53e0-8000-4384-b902-c70b69faacc5",
                 "stale_timestamp": "2020-02-10T08:07:03.354307+00:00",
                 "reporter": "puptoo",
+                "per_reporter_staleness": {
+                    "puptoo": {"check_in_succeeded": True, "last_check_in": mocker.ANY, "stale_timestamp": mocker.ANY}
+                },
```
kruai, line 677: This test was confusing for me to figure out, because I couldn't figure out where all the hardcoded values were coming from. However, I eventually found that it's using [this fixture](https://github.com/RedHatInsights/insights-host-inventory/blob/master/tests/fixtures/graphql_fixtures.py#L33), which causes `graphql_query_with_response` to return a specific dataset. The dataset is defined [here](https://github.com/RedHatInsights/insights-host-inventory/blob/master/tests/fixtures/graphql_fixtures.py#L33); you can add `per_reporter_staleness` to this dataset with some hardcoded dates, which would allow you to update this test without having to bring in the mocker.

tests/test_xjoin.py
```diff
@@ -696,6 +699,9 @@ def test_response_processed_properly(query_source_xjoin, graphql_query_with_resp
                 "insights_id": "17c52679-f0b9-4e9b-9bac-a3c7fae5070c",
                 "stale_timestamp": "2020-01-10T08:07:03.354307+00:00",
                 "reporter": "puptoo",
+                "per_reporter_staleness": {
+                    "puptoo": {"check_in_succeeded": True, "last_check_in": mocker.ANY, "stale_timestamp": mocker.ANY}
+                },
```
kruai, line 704: Same thing applies here ofc

kruai (changes requested):
I think it's looking good overall - I'm just requesting one more piece of simplification :)

tests/test_xjoin.py
thearifismail: This bamboozled me for quite sometime the other.  I had started off by adding `per_reporter_staleness` to the fixture, particularly before this line https://github.com/RedHatInsights/insights-host-inventory/blob/1ebd6199466e3713d8ad5a77ee02c7493658f501/tests/helpers/graphql_utils.py#L33 but a subsequent call leads to over-writing the per_reporter_staleness by https://github.com/RedHatInsights/insights-host-inventory/blob/1ebd6199466e3713d8ad5a77ee02c7493658f501/app/models.py#L298.

Changing the model starts a whole new discussion about expected input to create a host.  That's why I suggest leaving `mocker.ANY` in this test for now.

tests/test_xjoin.py
thearifismail: same as above.

tests/test_xjoin.py
kruai: What subsequent call? The tests validate xjoin response content... the response data isn't loaded into a Host model, so the code you linked would never be called

kruai: I don't think we should use `mocker.ANY` here because we *should* know the exact values to expect. However, I did see that the per-reporter staleness gets updated when a Host object is initialized, and I don't think that's correct ([this line in the init](https://github.com/RedHatInsights/insights-host-inventory/blob/1ebd6199466e3713d8ad5a77ee02c7493658f501/app/models.py#L223)). Because of that, it will always tell us that the reporter last checked in _right now_.

@ryandillinfelton Was this an intentional change, or would it be OK if we removed that call from the init?

thearifismail: Done.  mocker.ANYs gone.

tests/test_xjoin.py
thearifismail: Done.

tests/test_xjoin.py
```diff
@@ -647,7 +647,7 @@ def test_query_variables_staleness_with_search(
     )
 
 
-def test_response_processed_properly(query_source_xjoin, graphql_query_with_response, api_get):
+def test_response_processed_properly(query_source_xjoin, graphql_query_with_response, api_get, mocker):
```
kruai, line 650: `mocker` is no longer used in this test

tests/test_xjoin.py
```diff
@@ -672,6 +672,13 @@ def test_response_processed_properly(query_source_xjoin, graphql_query_with_resp
                 "insights_id": "a58c53e0-8000-4384-b902-c70b69faacc5",
                 "stale_timestamp": "2020-02-10T08:07:03.354307+00:00",
                 "reporter": "puptoo",
+                "per_reporter_staleness": {
+                    "puptoo": {
+                        "check_in_succeeded": True,
+                        "last_check_in": "2020-02-10T08:07:03.354307Z",
+                        "stale_timestamp": "2020-08-10T08:07:03.354307Z",
```
kruai, line 679: Awesome! The last thing I'm concerned about here is that these dates aren't in the same format as the other dates. All the others use `_deserialize_datetime()` to convert from ISO to a timezone-sensitive datetime.

I know it won't be as simple as passing the field into that function, since it's a different structure, but I do think these values need to be in the same format.

kruai (changes requested):
Nearly there! Just one more thing that needs to be fixed in order for the data to be usable from the front-end, I think.

tests/test_xjoin.py
kruai: Nice!

tests/helpers/graphql_utils.py
```diff
@@ -30,6 +30,13 @@
                 "facts": None,
                 "stale_timestamp": "2020-02-10T08:07:03.354307Z",
                 "reporter": "puptoo",
+                "per_reporter_staleness": {
+                    "puptoo": {
+                        "check_in_succeeded": True,
+                        "last_check_in": "2020-02-10T08:07:03.354307+00:00",
+                        "stale_timestamp": "2020-02-10T08:07:03.354307+00:00",
```
kruai, line 37: Does xjoin give us these values in this format? The other timestamps are in that ISO format that ends in "Z".

kruai (changes requested):
I don't think xjoin gives us the timestamps in that format. You'll likely need to convert the date values in `deserialize_host_xjoin()` and `serialize_host()`, in a way similar to `_deserialize_datetime()` and `_serialize_datetime()`, respectively.

thearifismail:
Yes, `xjoin-search` returned `per_reporter_staleness` data in this format.  Please see the screenshot included in a discussion above.  https://github.com/RedHatInsights/insights-host-inventory/pull/958#discussion_r677458090

tests/helpers/graphql_utils.py
kruai: @thearifismail confirmed that it does, so this change is correct.

app/models.py
```diff
@@ -220,7 +221,9 @@ def __init__(
         super().__init__(canonical_facts, display_name, ansible_host, account, facts, tags, system_profile_facts)
         self.stale_timestamp = stale_timestamp
         self.reporter = reporter
-        self._update_per_reporter_staleness(stale_timestamp, reporter)
+        self.per_reporter_staleness = per_reporter_staleness
+        if not per_reporter_staleness:
+            self._update_per_reporter_staleness(stale_timestamp, reporter)
```
kruai, line 226: My last Q is about this; this is better than what it was, but it may be best to remove these 2 lines.
With this code, if a Host object is created and that field isn't provided, it will have incorrect per-reporter staleness data. If we remove this code, the Host object would have None when that field isn't provided. What do you think @thearifismail @ryandillinfelton?

kruai: I gave it some more thought, and I think that this approach is acceptable. The `last_checkin` value might be wrong if the Host hasn't updated that field before, but that's the only scenario where that happens. That's probably preferable to it having no value at all.

kruai (dismissed):

thearifismail:
yes, let's wait for @ryandillinfelton input.  To have `last_checkin` available upon first query after the host creation, `per_reporter_staleness` MUST be set during creation.

ryandillinfelton:
> yes, let's wait for @ryandillinfelton input. To have `last_checkin` available upon first query after the host creation, `per_reporter_staleness` MUST be set during creation.

Is there concern that `per_reporter_staleness` is not set during host creation? I believe that it is, and assuming [this test](https://github.com/RedHatInsights/insights-host-inventory/blob/master/tests/test_models.py#L385) is working correctly it should be for sure. But I'd be happy to dig in further if you still have concerns.

thearifismail:
> > yes, let's wait for @ryandillinfelton input. To have `last_checkin` available upon first query after the host creation, `per_reporter_staleness` MUST be set during creation.
> 
> Is there concern that `per_reporter_staleness` is not set during host creation? I believe that it is, and assuming [this test](https://github.com/RedHatInsights/insights-host-inventory/blob/master/tests/test_models.py#L385) is working correctly it should be for sure. But I'd be happy to dig in further if you still have concerns.

No, it is not a concern.  Asa had questioned why to add `per_reporter_staleness` when the original payload did not provide it?     Subsequent testing revealed that it was needed upon "host creation" to update the "last_check_in".  

This is all done.  Can you please approved it?  Thanks.

swagger/api.spec.yaml
```diff
@@ -984,6 +984,11 @@ components:
               description: Reporting source of the host. Used when updating the stale_timestamp.
               type: string
               nullable: true
+            per_reporter_staleness:
+              description: >-
+                Reporting source of the last checkin status, stale_timestamp, and last_check_in.
+              type: object
+              nullable: true
```
RonnyPfannschmidt, line 991: for general consistency i propose this to be a more structured type and to externalize it a little to inform the codeegen type names

additionally i propose to **drop** nullable and always making this a object


so something like this untested example
```yaml
per_reporter_staleness: $ref "#/components/schemas/PerReporterStaleness"

components:
  schemas:
    PerReporterStaleness
      description: >-
                Reporting source of the last checkin status, stale_timestamp, and last_check_in.
      type: object
      additionalProperties:
         $ref: #/components/schemas/ReporterStalenessEntry"

     ReporterStalenessEntry:
        last_checking:
           type: datetime ?
        status:
           type: string
         stale_timestamp:
           type: timestamp

Comment (thearifismail) [reply swagger/api.spec.yaml:991]: Since this is `nice-to-have` and expands the work and the current implementation works as desired by the JIRA card, we should merge this PR and then create a new JIRA card to to create a structured `per_reporter_staleness` as suggested.

Comment (RonnyPfannschmidt) [reply swagger/api.spec.yaml:991]: one element we must change before deploying is the nullable,
as its not a required field, we ought to rather skip it than to have a additional invariant, allthough a empty dict is prefferable

Comment (kruai) [reply swagger/api.spec.yaml:991]: @thearifismail  I mean I guess it's technically a "nice-to-have" and not a hard requirement, but doesn't it make sense for that to be delivered with the feature, since it's being considered before it's merged? Otherwise we're intentionally accruing technical debt. I'm open to discussion, of course; I just don't get why we would intentionally wait to correct it.

@RonnyPfannschmidt is also correct that it doesn't make sense to be nullable; it would be preferable to return an empty dict for this data.

Comment (thearifismail) [reply swagger/api.spec.yaml]: My only fear is this PR's growth and I removed the `nullable: true`

Comment (kruai) [reply swagger/api.spec.yaml]: Yep! If you can make a follow-up Jira then I suppose we can finish it later

**Review by kruai — APPROVED** [review]

File: app/serialization.py [hunk 9]
```diff
@@ -74,6 +75,7 @@ def deserialize_host_xjoin(data):
         system_profile_facts=data["system_profile_facts"] or {},
         stale_timestamp=_deserialize_datetime(data["stale_timestamp"]),
         reporter=data["reporter"],
+        per_reporter_staleness=data.get("per_reporter_staleness"),
```
RonnyPfannschmidt, line 78: for the invariant, shouldnt this use `get(..., {})`

swagger/api.spec.yaml
thearifismail: Follow up JIRA https://issues.redhat.com/browse/ESSNTL-1032

app/serialization.py
thearifismail: Now it is `per_reporter_staleness=data["per_reporter_staleness"] or {},`

thearifismail: @RonnyPfannschmidt ^^^ Is it good now?

app/models.py
```diff
@@ -220,7 +221,9 @@ def __init__(
         super().__init__(canonical_facts, display_name, ansible_host, account, facts, tags, system_profile_facts)
         self.stale_timestamp = stale_timestamp
         self.reporter = reporter
-        self._update_per_reporter_staleness(stale_timestamp, reporter)
+        self.per_reporter_staleness = per_reporter_staleness
```
RonnyPfannschmidt, line 224: i think here is a missing None to dict transform missing

RonnyPfannschmidt (commented):
lovely, with the transformation to schema and no longer using get/defaults it looks way easier to follow

app/models.py
kruai: Wouldn't it be simpler to just change the default value? I'll leave a comment where I'm talking about

app/models.py
```diff
@@ -203,6 +203,7 @@ def __init__(
         system_profile_facts=None,
         stale_timestamp=None,
         reporter=None,
+        per_reporter_staleness=None,
```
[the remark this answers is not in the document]
RonnyPfannschmidt: @kruai we should never put a mutlable value into a default - its evaluated at declaration time, its a total trap ^^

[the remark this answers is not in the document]
kruai: OK! I need to learn why, but I'll trust your judgment on this.

kruai (approved):

app/models.py
kruai: nvm I was wrong

RonnyPfannschmidt (approved):
tested with the soon to be merged rest stale report tests :+1:

app/serialization.py
```diff
@@ -74,6 +75,7 @@ def deserialize_host_xjoin(data):
         system_profile_facts=data["system_profile_facts"] or {},
         stale_timestamp=_deserialize_datetime(data["stale_timestamp"]),
         reporter=data["reporter"],
+        per_reporter_staleness=data["per_reporter_staleness"] or {},
```
ryandillinfelton, line 78: I'm not gonna request changes since it's consistent with existing code and super picky, but I'd kinda like to expand the linter to require spaces around `=`.

ryandillinfelton (approved):

PR #968, opened 2021-07-30
Validate System Profile date-time fields

kruai:
# Overview

This PR is being created to address [ESSNTL-559](https://issues.redhat.com/browse/ESSNTL-559).
It enforces strict rfc3339 format on System Profile date-time fields, so that people can't put something like "June 17 last year" or "Mar 9 2021". It also updates the "valid system profile" test data to conform to this standard.

I've also checked various values for SP date-time fields in Kibana ([example](https://kibana.apps.crcp01ue1.o9m8.p1.openshiftapps.com/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-90d,to:now))&_a=(columns:!(_source),filters:!(),index:'43c5fed0-d5ce-11ea-b58c-a7c95afd7a5d',interval:auto,query:(language:kuery,query:'@log_group:%20%22host-inventory-prod%22%20and%20%22last_boot_time%22'),sort:!()))), and I haven't found any that are breaking the format yet.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


ryandillinfelton:
Generally the code changes look good and I think this accomplishes the timestamp verification correctly. I don't have any complaints about test coverage either. It does seem like this change has the potential of causing unexpected validation errors to crop up though. Have you explored that at all?

Also I'm seeing all the checks fail which is a bit concerning. Not sure if it's a jenkins issue or what, but I'll hold off approving until the cause it determined.

kruai:
I must have missed the test failures on this one! Pushed a fix.

kruai:
Um... for some reason, the machine trying to build this PR doesn't recognize what pipenv is. I haven't seen that happen before

thearifismail (commented):
Overall coding wise it LGTM, but I think it needs more tests to check valid formats.  Tests should check for various valid formats, like local time with time-of-the-day, GMT, local time offset, day-month-year, or month-day-year.

Is just month, day, year without time-of-day a valid formart?

kruai:
> Overall coding wise it LGTM, but I think it needs more tests to check valid formats. Tests should check for various valid formats, like local time with time-of-the-day, GMT, local time offset, day-month-year, or month-day-year.
> 
> Is just month, day, year without time-of-day a valid formart?

The only format we want to allow is RFC 3339, which is a profile of ISO 8601. So no, we don't want to allow loose formats like simply month-day-year etc :)

kruai:
> Generally the code changes look good and I think this accomplishes the timestamp verification correctly. I don't have any complaints about test coverage either. It does seem like this change has the potential of causing unexpected validation errors to crop up though. Have you explored that at all?
> 
> Also I'm seeing all the checks fail which is a bit concerning. Not sure if it's a jenkins issue or what, but I'll hold off approving until the cause it determined.

@ryandillinfelton  The only SP field that's defined as a date-time is `last_boot_time`. I've sifted through a bunch of values in Prod, and I don't see any invalid values. I had been concerned about `bios_release_date`, but that one is simply a string and not a date-time.

And you probably saw already, but we resolved the Jenkins issues :)

thearifismail (approved):
LGTM

app/models.py
```diff
@@ -555,7 +556,7 @@ def filter_system_profile_keys(self, data, **kwargs):
     @validates("system_profile")
     def system_profile_is_valid(self, system_profile):
         try:
-            jsonschema_validate(system_profile, self.system_profile_normalizer.schema)
+            jsonschema_validate(system_profile, self.system_profile_normalizer.schema, format_checker=FormatChecker())
```
ryandillinfelton, line 559: The reason I'm concerned here is that you don't initialize the format check with a list of formats (which also make me wonder how it is successfully catching the malformed timestamps since documentation says that it defaults to `None`). So I'm worried it'll make out checks on other non-timestamp fields stricter and maybe fail some fields we haven't been strict about in the past. I think ideally we would be explicit here and give the FormatChecker a list with only the timestamp type defined.

ryandillinfelton: @kruai in response to [your response to my question](https://github.com/RedHatInsights/insights-host-inventory/pull/968#issuecomment-895286261). Let me know if I've missinterpretted [this part of the docs](https://python-jsonschema.readthedocs.io/en/stable/validate/#jsonschema.FormatChecker) I trust you're more familiar with the actuality of the thing since you're actually working with it. I just don't want puptoo yelling at us :laughing:

kruai: Totally understand! It took a while for me to figure out how it works. It turns out that when you install specific validators (listed at the bottom of the doc page you sent), the `FormatChecker` class registers them by default. So, with strict-rfc3339 installed, the `date-time` validator uses that (instead of doing no validation for date-time).
To be clear, this only applies to the system_profile, and only for fields with `format: date-time`, which is what's registered for that rfc3339 package.

ryandillinfelton: Awesome! Good to know. I'll approve :+1:

ryandillinfelton (approved):

PR #976, opened 2021-08-09
Hotfix: Set PRS to {} if None or key doesn't exist

kruai:
Should fix the error caused by xjoin not returning `per_reporter_staleness` data.


thearifismail (approved):
LGTM

PR #977, opened 2021-08-10
Host dedup logic fix

kruai:
# Overview

This PR is being created to address [ESSNTL-1059](https://issues.redhat.com/browse/ESSNTL-1059).
It further tweaks the dedup logic to be more precise. Once it's determined that it can't find the host by elevated ID, the basic logic will now match INPUT_HOST to EXISTING_HOST if:
- EXISTING_HOST is in the user's account, AND
- none of INPUT_HOST's canonical facts have different values than EXISTING_HOST's
- at least one of INPUT_HOST's canonical facts is the same on EXISTING_HOST

## PR Checklist

- [x] Keep PR title short, ideally under 72 characters
- [x] Descriptive comments provided in complex code blocks
- [x] Tests: validate optimal/expected output
- [ ] Tests: validate exceptions and failure scenarios
- [ ] Tests: edge cases
- [ ] Documentation, if this PR changes the way other services interact with host inventory
- [ ] Links to related PRs

## Secure Coding Practices Documentation Reference

You can find documentation on this checklist [here](https://github.com/RedHatInsights/secure-coding-checklist).

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #978, opened 2021-08-10
Don't find a match if higher-prio elevated IDs are different

kruai:
# Overview

This PR is being created to address [ESSNTL-555](https://issues.redhat.com/browse/ESSNTL-555) and [ESSNTL-542](https://issues.redhat.com/browse/ESSNTL-542).
Makes it so that while searching for existing hosts by ID, no match is found if the higher-priority IDs have differing values.
Should be merged AFTER #977.

## PR Checklist

- [ ] Tests: edge cases
- [x] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


lib/host_repository.py
```diff
@@ -84,12 +84,16 @@ def find_existing_host_by_id(identity, host_id):
...
-        cf_value = canonical_facts.get(elevated_cf_name)
-        if cf_value:
-            existing_host = find_host_by_single_canonical_fact(identity, elevated_cf_name, cf_value)
-            if existing_host:
-                return existing_host
+    elevated_facts = {key: value for (key, value) in canonical_facts.items() if key in ELEVATED_CANONICAL_FACT_FIELDS}
+    if elevated_facts:
+        existing_host = find_host_by_multiple_canonical_facts(identity, elevated_facts)
+        if existing_host:
+            for elevated_field in (key for key in elevated_facts if key in existing_host.canonical_facts):
```
thearifismail, line 91: I think all elevated_canonical_fact (ECF) values should be compared before returning match or no-match because if all ECFs are present with only one with a different value but the other two have the same.  In that case, inventory may end up with two hosts with two identical ECFs.  In that case, the host should be rejected with message stating that a new host could not be created because an existing host has some of the identical canonical.

kruai: It sounds like what you're requesting is at odds with ESSNTL-555, unless I'm misunderstanding. 
As for the situation you mentioned with two hosts that have identical elevated IDs, I think my `test_elevated_id_priority` tests cover that situation, right? They create two hosts with one differing elevated ID, and expect dedup logic to work according to the elevated ID priority

thearifismail (approved):
LGTM

ryandillinfelton (approved):
We worked it out on Slack. LGTM now :+1:

PR #991, opened 2021-09-12
Delete duplicate hosts from HBI database

thearifismail:
# Overview

This PR is being created to address [ESSNTL-1221](https://issues.redhat.com/browse/ESSNTL-1221). 

For finding duplicate systems, first it uses "elevated_canonical_facts" in the order of their priority.  If multiple hosts are found, it holds the last updated host and remove the others.  When no elevated canonical_facts are found then the remaining canonical_facts are used to find duplicates.  The hosts, which are to be kept are added to a unique hosts list, the the unique hosts list compare to all hosts. The hosts not in the unique list are deleted..

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,167 @@
...
+
+# from app.logging import get_logger
+
+# logger = None
+
+__all__ = ("delete_duplicate_hosts",)
+
+# initialize a empty lists
+unique_list = []
+duplicate_list = []
```
kruai, line 18: Using these as static variables feels like a bit of a strange pattern to me. Was this based off of another job's implementation?

kruai: I'm also concerned that `unique_list` would become enormous using this approach

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,167 @@
...
+    logger.debug("find_host_by_multiple_elevated_canonical_facts(%s)", canonical_facts)
+
+    if canonical_facts.get("provider_id"):
+        if canonical_facts.get("subscription_manager_id"):
+            canonical_facts.pop("subscription_manager_id")
+        if canonical_facts.get("insights_id"):
+            canonical_facts.pop("insights_id")
+    elif canonical_facts.get("insights_id"):
+        if canonical_facts.get("subscription_manager_id"):
+            canonical_facts.pop("subscription_manager_id")
```
kruai, line 81: Can you explain why you're popping all but the highest-prio elevated ID? I'm a bit lost

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,167 @@
...
+    query = select_query
+    all_hosts = query.limit(chunk_size).all()
+
+    for host in all_hosts:
+        logger.info(f"Host ID: {host.id}")
+        logger.info(f"Canonical facts: {host.canonical_facts}")
+        elevated_facts = get_elevated_canonical_facts(host.canonical_facts)
+        logger.info(f"elevated canonical facts: {elevated_facts}")
+        if elevated_facts:
+            find_host_by_multiple_elevated_canonical_facts(elevated_facts, query, logger)
```
kruai, line 143: Definitely confusing to see this not return anything, and it's not clear that this "find" method is actually updating a static list defined at the top of the file. Why not instead define those 2 lists at the top of this function, and add/remove values based on the results of the function calls?

kruai (changes requested):
I have some questions and I'm a bit confused about the design patterns here. But this also definitely needs thorough tests to make sure we don't delete data that we're not supposed to delete

host_delete_duplicates.py
```diff
@@ -0,0 +1,105 @@
...
+
+__all__ = ("main", "run")
+
+PROMETHEUS_JOB = "duplicate_hosts_remover"
+LOGGER_NAME = "duplicate_hosts_remover"
+COLLECTED_METRICS = (
+    delete_duplicate_host_count,
+    event_producer_failure,
+    event_producer_success,
+    event_serialization_time,
```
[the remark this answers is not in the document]
thearifismail: They should've been removed.  Gone now.

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+# this function is called when no elevated canonical facts are present in the host
+def find_host_by_multiple_canonical_facts(canonical_facts, query, logger):
+    """
+    Returns all matches for a host containing given canonical facts
+    """
+    logger.debug("find_host_by_multiple_canonical_facts(%s)", canonical_facts)
+
+    for fact in canonical_facts:
```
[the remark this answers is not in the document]
thearifismail: This function is called when there is no elevated canonical fact in the host.  To ensure that every canonical_fact(CF) is looked at, each fact is queried by the loop.  If there are > 1 hosts, only the first host is saved because the query results are provided with descending update time.  

The only possibility that this fact may be a duplicate is if it is in a host which also has an elevated canonical fact.

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,167 @@
...
+    return elevated_facts
+
+
+def get_unelevated_canonical_facts(canonical_facts):
+    unelevated_facts = {key: canonical_facts[key] for key in CANONICAL_FACTS if key in canonical_facts.keys()}
+    return unelevated_facts
+
+
+def _delete_host(query, host):
+    delete_query = query.filter(Host.id == host["id"].hex)
```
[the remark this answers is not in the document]
thearifismail: The value of `host["id']` is an object of type `class 'uuid.UUID'`, which converted into a string by `.hex` and String is what is saved in DB

lib/host_remove_duplicates.py
thearifismail: Size of `unique_list` has been on my concern but at this time I want to get the logic right.  Since this job will be run only once and seldom after that, hopefully.  It may be okay.  We need to test the heck out before using it on the Prod cluster.

lib/host_remove_duplicates.py
thearifismail: Since highest priority CF takes precedence over others, the code checks for the highest CF. My understand is that same `insights_id` CF may be present in two different hosts with unique `provider_ids`, right? 

If there is no provider_id, then the host is checked for "insights_id", if there is no "insights_id" then the "subscription_manager_id" is used.

lib/host_remove_duplicates.py
thearifismail: I agree method and variable names should be changed for clarity, but for now I wanted to focus on logic to make sure that only the right hosts are deleted.

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,167 @@
...
+
+def unique(host):
+    # Use id and account to create unique hosts across all accounts.
+    unique_host = {"id": host.id, "account": host.account}
+    if unique_host not in unique_list:
+        unique_list.append(unique_host)
+
+
+# The order is important, particularly the first 3 which are elevated facts with provider_id being the highest priority
+CANONICAL_FACTS = ("fqdn", "satellite_id", "bios_uuid", "ip_addresses", "mac_addresses")
```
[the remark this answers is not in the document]
kruai: True, though that's kind of an odd one since it goes hand-in-hand with `provider_id`. Because a host can only have both or neither of these fields, I think it makes sense to leave that out here

lib/host_remove_duplicates.py
kruai: Every time it runs, though, we can expect it to contain every host in the database at once (excluding duplicates). That might cause the pod to run out of memory in the worst case; a better case being that it just runs extremely slowly.
And I agree about the testing, but it doesn't look like you've added any tests to this yet

lib/host_remove_duplicates.py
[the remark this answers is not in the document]
kruai: Now that `trim_extra_facts` isn't being called, this is only searching for one canonical fact at a time, and adding the first entry to the list of unique hosts each time. This means that the FQDN on the host is `localhost`, this function could put any other host with with FQDN `localhost` in the list of unique hosts. I'm not sure what this function is trying to do makes sense

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def unique(host):
+    # Use id and account to create unique hosts across all accounts.
+    unique_host = {"id": host.id, "account": host.account}
+    if unique_host not in unique_list:
+        unique_list.append(unique_host)
+
+
+# The order is important, particularly the first 3 which are elevated facts with provider_id being the highest priority
```
kruai, line 26: Minor, but order is no longer important for `CANONICAL_FACTS`

lib/host_remove_duplicates.py
kruai: I understand wanting to focus on logic, it's just harder to review when a function does something other than what the name implies. I do still think it would be much easier to read if `unique_list` and `duplicate_list` were scoped to this function instead of being static vars; then each of these functions could return the host values that should be added to each list

kruai (changes requested):
I'm going to have to review more in-depth again later, but I've requested some changes. We do need tests to go along with this to make sure your code is working as you intend

lib/host_remove_duplicates.py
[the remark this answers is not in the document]
ryandillinfelton: I know we talked about the possibility of provider_ids being the same, bur from different providers, but I'm pretty sure the ones we support right now all use different formats so I agree with Asa that we don't need to worry about the provider_type.

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+        (contains_no_incorrect_facts_filter(canonical_facts))
+        & (matches_at_least_one_canonical_fact_filter(canonical_facts))
+    )
+    return query
+
+
+# Get hosts by the highest elevated canonical fact present
+def find_host_by_elevated_canonical_facts(elevated_cfs, query, logger):
+    """
+    First check if multiple hosts are returned.  If they are then retain the with the highest
```
ryandillinfelton, line 56: Obviously not gonna block on this. Real nit-picking hours.

```suggestion
    First check if multiple hosts are returned.  If they are then retain the one with the highest
```

utils/duplicate_remover_clowd_job_invocation.yaml
```diff
@@ -0,0 +1,14 @@
+# Host synchronizer is run manually to synchronize host-invnenotry DB
```
ryandillinfelton, line 1: This a copy-paste casualty?

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+
+        acct_query = query.filter(Host.account == account)
+        host_list = acct_query.limit(chunk_size).all()
+        for host in host_list:
+            logger.info(f"Host ID: {host.id}")
+            logger.info(f"Canonical facts: {host.canonical_facts}")
+            elevated_cfs = get_elevated_canonical_facts(host.canonical_facts)
+            logger.info(f"elevated canonical facts: {elevated_cfs}")
+            if elevated_cfs:
+                hosts = find_host_by_elevated_canonical_facts(elevated_cfs, acct_query, logger)
```
ryandillinfelton, line 137: Since in either case you only use the 0th entry in hosts why not have `find_host_by_elevated_canonical_facts` use `.first()` and return just one host value?

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+                logger.info(f"regular canonical facts: {regular_cfs}")
+                if regular_cfs:
+                    hosts = find_host_by_regular_canonical_facts(regular_cfs, acct_query, logger)
+
+            unique(hosts[0])
+            logger.info(f"Unique hosts count: {len(unique_list)}")
+            logger.info(f"All hosts count: {len(host_list)}")
+
+        duplicate_list = []
+        for host in host_list:
```
ryandillinfelton, line 149: This can be done with a set operation. Just make sure the superset is on the left.
```suggestion
        duplicate_list = list(set(host_list) - set(unique_list))
```

ryandillinfelton: Just realized you transform the elements of host_list before comparison. You could maybe do that with a map, but at that point I think it's just a little too much reworking, and too little simplifying. Especially for something that already works.

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+        # delete duplicate hosts
+        while len(duplicate_list) > 0 and not interrupt():
+            for host in duplicate_list:
+                _delete_host(query, host)
+                duplicate_list.remove(host)
+                delete_duplicate_host_count.inc()
+
+                yield host["id"]
+                # load next chunk using keyset pagination
+        host_list = query.filter(Host.id > host_list[-1].id).limit(chunk_size).all()
```
ryandillinfelton, line 164: I don't think you're processing any more than the first chunk of hosts with this approach. You set host_list to the next chunk, but you're in the for(accounts) loop, so the next time you hit the for(host_list) loop you'll have a different account number and the host list will have been redefined. I think you need some kind of `while there are still chunks` loop. Maybe check if the query returns the full chunk size, and if not you're at the end of the hosts.

ryandillinfelton (changes requested):
Let me know if I'm mistaken, but I think I found a bug that will prevent the script from de-duplicating any accounts with more hosts than the chunk size. Otherwise the logic is looking good. Just had some minor suggestions.

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+    query = select_query
+    logger.info(f"Total number of hosts in inventory: {query.count()}")
+
+    distinct_accounts_query = query.distinct(Host.account)
+    logger.info(f"Total number of accounts in inventory: {distinct_accounts_query.count()}")
+
+    accounts = distinct_accounts_query.limit(chunk_size).all()
+    distinct_accounts = []
+    for acct in accounts:
+        distinct_accounts.append(acct.account)
```
RonnyPfannschmidt, line 113:
```suggestion
    distinct_accounts = [row.account for row in distinct_accounts_query.limit(chunk_size).all()]
  
```

lib/host_remove_duplicates.py
RonnyPfannschmidt, line 159: this line doesnt work correct unless there is a extra lop level

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+        # set uniquess within the account
+        unique_list = []
+
+        def unique(host):
+            unique_host = {"id": host.id, "account": host.account}
+            if unique_host not in unique_list:
+                unique_list.append(unique_host)
+
+        acct_query = query.filter(Host.account == account)
+        host_list = acct_query.limit(chunk_size).all()
```
RonnyPfannschmidt, line 125: use use https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.yield_per here here and introduce a extra lloop level while at it to enable the processing of all host lists at once

host_delete_duplicates.py
```diff
@@ -0,0 +1,97 @@
...
+def _excepthook(logger, type, value, traceback):
+    logger.exception("Host duplicate remover failed", exc_info=value)
+
+
+def run(config, logger, session, event_producer, shutdown_handler):
+
+    query = session.query(Host)
+
+    update_count = 0
+    events = delete_duplicate_hosts(query, config.script_chunk_size, logger, shutdown_handler.shut_down)
```
RonnyPfannschmidt, line 55: can we move the complete logging into the delete_duplicate_hosts and only return the update count here?

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+
+def matches_at_least_one_canonical_fact_filter(canonical_facts):
+    # Contains at least one correct CF value
+    # Correct value = contains key:value
+    # -> OR( *correct values )
+    filter_ = ()
+    for key, value in canonical_facts.items():
+        filter_ += (Host.canonical_facts.contains({key: value}),)
+
+    return or_(*filter_)
```
RonnyPfannschmidt, line 24: i recal lthat this should do the job as well
```suggestion
    return or_(Host.canonical_facts.contains({key: value}) for key, value in canonical_facts.items())
```

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def contains_no_incorrect_facts_filter(canonical_facts):
+    # Does not contain any incorrect CF values
+    # Incorrect value = AND( key exists, NOT( contains key:value ) )
+    # -> NOT( OR( *Incorrect values ) )
+    filter_ = ()
+    for key, value in canonical_facts.items():
+        filter_ += (
+            and_(Host.canonical_facts.has_key(key), not_(Host.canonical_facts.contains({key: value}))),  # noqa: W601
```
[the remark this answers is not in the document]
RonnyPfannschmidt: i do believe that a null-able lookup compared to a text would be more effective than the has_key and contains combination
i couldnt find the correct way to query this in my timebox

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+
+        duplicate_list = []
+        for host in host_list:
+            hostIdAccount = {"id": host.id, "account": host.account}
+            if hostIdAccount not in unique_list:
+                duplicate_list.append(hostIdAccount)
+        logger.info(f"Duplicate hosts count: {len(duplicate_list)}")
+
+        # delete duplicate hosts
+        while len(duplicate_list) > 0 and not interrupt():
```
RonnyPfannschmidt, line 151: just use the boolean check here

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+            for host in duplicate_list:
+                _delete_host(query, host)
```
RonnyPfannschmidt, line 153: should there be Kafka delete events?
also why not use a singe query for the duplicate list

lib/host_remove_duplicates.py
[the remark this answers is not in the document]
fstavela: Ah yeah, that's true. Okay, then there is no need to add `provider_type` here.

lib/host_remove_duplicates.py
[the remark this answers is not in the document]
kruai: Resolving since this is low prio. But if you figure it out, please unresolve and let me know bc we'll want to use that in the function this is based on

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+    logger.info(f"Total number of accounts in inventory: {distinct_accounts_query.count()}")
+
+    accounts = distinct_accounts_query.limit(chunk_size).all()
+    distinct_accounts = []
+    for acct in accounts:
+        distinct_accounts.append(acct.account)
+
+    for account in distinct_accounts:
+        # set uniquess within the account
+        unique_list = []
```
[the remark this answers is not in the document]
kruai: Yep, we brought that up yesterday. I'll change it

lib/host_remove_duplicates.py
kruai: Yeah, it doesn't look like the code in general does the chunking loop correctly

lib/host_remove_duplicates.py
kruai: Still an issue, but we've got bigger things to deal with. Resolving

lib/host_remove_duplicates.py
kruai: Added kafka delete events, and updated how we query and delete

lib/host_remove_duplicates.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    logger.debug("find_host_by_elevated_canonical_facts(%s)", elevated_cfs)
+
+    if elevated_cfs.get("provider_id"):
+        elevated_cfs.pop("subscription_manager_id", None)
+        elevated_cfs.pop("insights_id", None)
+    elif elevated_cfs.get("insights_id"):
+        elevated_cfs.pop("subscription_manager_id", None)
+
+    hosts = multiple_canonical_facts_host_query(elevated_cfs, query).order_by(Host.modified_on.desc()).all()
```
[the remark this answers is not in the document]
RonnyPfannschmidt: based on the errors seen, i wonder if we can fetch them in groups - and we should use first instead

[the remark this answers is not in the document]
kruai: @RonnyPfannschmidt I've gone back to using `all()`, and treating the first host returned as "unique" and the others as potential duplicates. What kind of groups do you mean we should fetch them in?

ryandillinfelton (approved):

kruai (approved):

PR #997, opened 2021-09-17
Fix job name

kruai:
This PR fixes the duplicate-hosts-remover job name, which currently has underscores (which aren't allowed).


ryandillinfelton (approved):
LGTM

PR #1002, opened 2021-09-20
Add memory limit specific to dedup job

kruai:
This just adds a new deployment parameter that controls the dedup job's memory limit, separate from the other pods.


thearifismail (approved):
LGTM

PR #1003, opened 2021-09-22
Script to sync event topic

kruai:
# Overview

This PR is being created to address [ESSNTL-1281](https://issues.redhat.com/browse/ESSNTL-1281).
Context in this thread: https://ansible.slack.com/archives/CADT032B0/p1632242392478200
This goes through the messages in the `platform.inventory.events topic` and syncs it with the HBI DB. This is to help get Cyndi sync'd again.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/queue/queue.py
+++ b/app/queue/queue.py
@@ -21,8 +21,6 @@ from app.instrumentation import log_add_host_attempt
 from app.instrumentation import log_add_host_failure
 from app.instrumentation import log_add_update_host_succeeded
 from app.instrumentation import log_db_access_failure
-from app.instrumentation import log_host_delete_failed
-from app.instrumentation import log_host_delete_succeeded
 from app.instrumentation import log_update_system_profile_failure
 from app.instrumentation import log_update_system_profile_success
 from app.logging import get_logger
@@ -40,7 +38,6 @@ from app.queue.events import operation_results_to_event_type
 from app.serialization import DEFAULT_FIELDS
 from app.serialization import deserialize_host
 from lib import host_repository
-from lib.host_delete import delete_hosts
 
 
 logger = get_logger(__name__)
@@ -166,17 +163,8 @@ def parse_operation_message(message):
     return parsed_operation
 
 
-def sync_event_message(message, session, event_producer, shutdown_handler):
-    if message["type"] == EventType.delete.name:
-        query = session.query(Host).filter((Host.account == message["account"]) & (Host.id == UUID(message["id"])))
-        # If the host exists in the DB, delete it.
-        if query.count():
-            for host_id, deleted in delete_hosts(query, event_producer, 1, shutdown_handler.shut_down):
-                if deleted:
-                    log_host_delete_succeeded(logger, host_id, "TOPIC-REBUILD")
-                else:
-                    log_host_delete_failed(logger, host_id, "TOPIC-REBUILD")
-    else:
+def sync_event_message(message, session, event_producer):
+    if message["type"] != EventType.delete.name:
         query = session.query(Host).filter(
             (Host.account == message["host"]["account"]) & (Host.id == UUID(message["host"]["id"]))
         )
--- a/app/queue/queue.py
+++ b/app/queue/queue.py
@@ -11,7 +11,7 @@ from sqlalchemy.exc import OperationalError
 
 from app import inventory_config
 from app import UNKNOWN_REQUEST_ID_VALUE
-from app.auth.identity import create_mock_identity_from_host
+from app.auth.identity import create_mock_identity_with_account
 from app.auth.identity import Identity
 from app.auth.identity import IdentityType
 from app.culling import Timestamps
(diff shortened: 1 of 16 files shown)


app/queue/queue.py
```diff
@@ -161,6 +166,32 @@ def parse_operation_message(message):
...
+        # If the host exists in the DB, delete it.
+        if query.count():
+            for host_id, deleted in delete_hosts(query, event_producer, 1, shutdown_handler.shut_down):
+                if deleted:
+                    log_host_delete_succeeded(logger, host_id, "TOPIC-REBUILD")
+                else:
+                    log_host_delete_failed(logger, host_id, "TOPIC-REBUILD")
+    else:
+        query = session.query(Host).filter(
+            (Host.account == message["host"]["account"]) & (Host.id == UUID(message["host"]["id"]))
```
ryandillinfelton, line 181: What's up with the different indices for accessing the account number here and on line 171
```
message["account"]
```
vs
```
message["host"]["account"]
```
Is there a difference between `delete` messages and `update/create` message?

kruai: Yes, it's in our docs. `updated` event: https://consoledot.pages.redhat.com/docs/dev/services/inventory.html#_updated_event
`delete` event: https://consoledot.pages.redhat.com/docs/dev/services/inventory.html#_delete_event

ryandillinfelton:
Okay, very rambly but I'm leaving this as a note to investigate tomorrow at least.

Could the discrepancy have been caused by create/update events that were put on the topic during normal inventory operation that didn't get read by the connector. If a duplicate host was updated right after cyndi shut down, and the update was added to the event topic, and then we ran the dupe deleting script, producing a delete event for the same host and deleting it from the inventory DB then:
- Would cyndi consider there to be a discrepancy even though, if processed chronologically it should see the current state of the host as deleted?
- Will this script correctly handle that case? 

This script looks to run chronologically, so it may do extra work recreating the deleted host and then deleting it again, but I think it would get it. I don't know how cyndi decides if the topic is in sync with the DB though so I can't speak to that.

Beyond that, if we know they are out of sync, but don't know why, how can we know that syncing them like this won't cause other problems with the data. We're just assuming that every delete on the topic is correct, and that every host missing in the database is correct. I'd like to get to the bottom of the cause, or at least better understand the possibilities before we run this.

kruai:
> Okay, very rambly but I'm leaving this as a note to investigate tomorrow at least.
> 
I'll respond inline!
> Could the discrepancy have been caused by create/update events that were put on the topic during normal inventory operation that didn't get read by the connector. If a duplicate host was updated right after cyndi shut down, and the update was added to the event topic, and then we ran the dupe deleting script, producing a delete event for the same host and deleting it from the inventory DB then:
> 
I'm trying to follow, but I'm not fully sure I understand the scenario you're describing. Let me know if this is in line with what you mean:

1. Cyndi shuts down.
2. HBI consumes an `add_host` message from MQ, updates the host in the DB, and emits the "updated" event.
3. We run `host_delete_duplicates.py`, which removes the host from the DB, and emits a "delete" event.
4. We run `rebuild_events_topic.py`.

> * Would cyndi consider there to be a discrepancy even though, if processed chronologically it should see the current state of the host as deleted?
AFAIK, if all of those events were correctly added to `platform.inventory.events` and none were lost, there shouldn't be a discrepancy.
> * Will this script correctly handle that case?
If the host was removed from the DB and the "delete" event was produced, the record is considered to be in-sync. So, the script wouldn't do anything for that host.
> 
> This script looks to run chronologically, so it may do extra work recreating the deleted host and then deleting it again, but I think it would get it. I don't know how cyndi decides if the topic is in sync with the DB though so I can't speak to that.
I don't know Cyndi's logic either, beyond the basics. This script won't create new hosts or produce `created`/`updated` events, though; it just deletes hosts from the DB or adds a delete event to the MQ if a record is out of sync.
> 
> Beyond that, if we know they are out of sync, but don't know why, how can we know that syncing them like this won't cause other problems with the data. We're just assuming that every delete on the topic is correct, and that every host missing in the database is correct. I'd like to get to the bottom of the cause, or at least better understand the possibilities before we run this.
That's a good point. Let's consider the assumptions we're making:

- Every `delete` event on `platform.inventory.events` is correct
  - HBI is the only app that emits messages to this topic, and we only do it when we delete a host. In what scenario would there be a `delete` event on that topic that isn't correct?
- Every host missing from the database is correct
  - In other words, there's a `created`/`updated` event on the events topic, and that host isn't in the DB anymore. In that situation, I would have to assume that the record was deleted, but the `delete` event was lost somehow, or that the producer failed to emit it. In that case, I would want this script to create a delete event.

Does that make sense? Please let me know if you've got specific situations in mind; I do want to cover any potential gaps

kruai:
Based on some feedback from @fijshion, we should treat the HBI DB as the source of truth. So, I removed the part that deletes hosts from the DB based on `delete` events. I have some more changes in progress as well, so you might wanna hold off on reviewing more for now

deployment.yaml
```diff
@@ -545,6 +545,54 @@ objects:
...
+          - name: INVENTORY_DB_SSL_MODE
+            value: ${INVENTORY_DB_SSL_MODE}
+          - name: INVENTORY_DB_SSL_CERT
+            value: ${INVENTORY_DB_SSL_CERT}
+          - name: KAFKA_BOOTSTRAP_SERVERS
+            value: ${KAFKA_BOOTSTRAP_HOST}:${KAFKA_BOOTSTRAP_PORT}
+          - name: PAYLOAD_TRACKER_KAFKA_TOPIC
+            value: platform.payload-status
+          - name: PAYLOAD_TRACKER_SERVICE_NAME
+            value: inventory-mq-service
```
ryandillinfelton, line 568: We might want to change this, but I don't think anything happening here would be tracked  by the payload tracker anyways

ryandillinfelton (dismissed):
LGTM

thearifismail (approved):
LGTM

PR #1005, opened 2021-09-23
[WIP] Removed NodeJS from from the image; not used by host-inventory

thearifismail:
# Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/ESSNTL-1140).

This PR removes NPM and its is dependencies including NodeJS, which are not used by host-inventory but their presence in the base image provides potential pathway for security vulnerabilities.  Though this JIRA card is about using "ubi8-minimal" image but using this image blocks using pipenv by non-root accounts which is how HBI services are run.  That's why for now removing unused packages is being tried.

This is not a permanent solution because it is very hard to maintain due to security vulnerabilities are always being reported and responding to every single vulnerability must be handled in the base images upstream.  For now this approach is being tried to see if it works if there is no other way.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,9 @@ USER root
 RUN dnf install -y postgresql
 RUN dnf upgrade -y --security
 
+# remove packages not used by host-inventory to avoid security vulnerabilityes
+RUN dnf remove -y npm
+
 USER 1001
 
 WORKDIR /opt/app-root/src


akofink (approved):
You may also want to run `dnf clean all -y` to reduce the image size.

thearifismail:
> You may also want to run `dnf clean all -y` to reduce the image size.

Done!  Thanks

akofink (approved):

kruai (dismissed):

thearifismail:
@fstavela can you please execute your tests using this branch, particularly the Dockerfile before the code is merged, IF we choose to adopt this approach.  Manually adding PRM should be the last ditch approach.

Dockerfile
```diff
@@ -2,11 +2,17 @@ FROM registry.access.redhat.com/ubi8/python-38
...
+RUN dnf remove -y npm
+
+# upgrade security patches and cleanup any clutter left.
 RUN dnf upgrade -y --security
+RUN dnf clean all -y
+
+# The following RPM should be removed when a patched postgresql package becomes available in yum repos
+# RUN dnf install -y postgresql
+COPY postgresql-13.3-1.module+el8.4.0+11254+85259292.x86_64.rpm ./
+RUN dnf install -y postgresql-13.3-1.module+el8.4.0+11254+85259292.x86_64.rpm
```
kruai, line 15: Storing the postgresql rpm in our git repo feels very hacky. Also, doing it this way would mean we'd have to manually replace that RPM and the references in our Dockerfile every time there's a new vulnerability in postgresql.
What was the issue with using dnf install for postgresql 13?

thearifismail: > Storing the postgresql rpm in our git repo feels very hacky. Also, doing it this way would mean we'd have to manually replace that RPM and the references in our Dockerfile every time there's a new vulnerability in postgresql.
> What was the issue with using dnf install for postgresql 13?

These are the [vulnerabilities](https://quay.io/repository/thearifismail/insights-inventory/manifest/sha256:21a8456446ab7a7ff8a9772a103754ee347ab5eeef9a98f198b08f3da42b41fb?tab=vulnerabilities) that psql v13 has fixes for.

I, personally, do not like including RPMs but we may have to when there are hight impact vulnerabilities like these ones.  Since inventory installs `postgresql`, it may be okay to include the rpm.  May the rpm should be placed in the `lib` directory to make it look a bit better than the project root.

kruai: Sorry, I'm just confused - we're already using `dnf install` to install postgresql. What's stopping us from doing that with postgresql 13?

thearifismail: > Sorry, I'm just confused - we're already using `dnf install` to install postgresql. What's stopping us from doing that with postgresql 13?

`dnf install postgres13` works if any of the yum repos has "postgresql13".  The problem is none of the yum repos have "postgresql13". That's why the .rpm package has been added here.

kruai: OK. Is there a way we can request that they add that package to the registry? And as a shorter-term measure, is there another postgres version (like 12) we could use in the meantime?

thearifismail: > OK. Is there a way we can request that they add that package to the registry? And as a shorter-term measure, is there another postgres version (like 12) we could use in the meantime?

Unfortunately no.  That's what I was working on; rather looking for a postgresql13 installation package.

kruai: OK. I understand having to go this way if this is the only remaining option, I just want to make sure there isn't another way first. Like, [this image](https://catalog.redhat.com/software/containers/rhel8/postgresql-13/5ffdbdef73a65398111b8362) has that same postgres module installed, which makes it really confusing if that package doesn't exist in the repo. I did also notice that that image has pgsql v13, but no NPM/Node... but I'm trying to not be tempted by it lol

Anway, @ryandillinfelton @RonnyPfannschmidt @fstavela @fpjrh: Do any of you know if we can request a package be added to the RH package repository?

RonnyPfannschmidt: do we actually need a full postgresql? i was under the impression we only needed the includes for the client lib?

maybe we could get away by ensuring a existing psycopg2-binary or a manylinux2014 wheel for it?

fpjrh: @kruai let's take a look at what it will take to get the package added to the RH repository; I agree that it's not good practice for us to take ownership of managing an RPM in our tree. My preferred option is to convince the owner(s) of the container images to keep them healthy and secure per latest CVEs and notify users of the image of any updates. 

Let's discuss this further with ProdSec to see if we can put together a better mechanism.

PR #1024, opened 2021-11-02
Implement generic object filter

kruai:
# Overview

This PR is being created to address [ESSNTL-1572](https://issues.redhat.com/browse/ESSNTL-1572).
The original goal of this PR was to allow filtering specifically on the `ansible` System Profile object. I've since refactored things so that it will construct a generic filter for xjoin for all "object" types that don't require their own custom filter (such as operating_system).

I considered adding a check to make sure the child fields being filtered exist on the parent object, but... Since xjoin should complain correctly if we pass an unsupported filter, that felt redundant. Let me know what you think.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


api/filtering/filtering.py
```diff
@@ -46,13 +46,28 @@ def _wildcard_string_filter(field_name, field_value):
...
+def _build_object_filter(field_name, input_object, field_filter):
+    # field name is unused but here because the generic filter builders need it,
+    # and this has to have the same interface
+    object_filter = {}
+
+    for name in input_object:
+        field_value = _get_field_value(input_object[name], "wildcard")
+        object_filter.update(
+            _generic_filter_builder(BUILDER_FUNCTIONS.wildcard.value, name, str(field_value), "wildcard", True)[0]
+        )
```
kruai, line 58: I just realized this "wildcard" stuff was leftover from when this was just the Ansible filter. I'm making this more generic right now

api/filtering/filtering.py
```diff
@@ -46,13 +46,30 @@ def _wildcard_string_filter(field_name, field_value):
...
 
 
+def _build_object_filter(field_name, input_object, field_filter):
+    # field name is unused but here because the generic filter builders need it,
+    # and this has to have the same interface
+    object_filter = {}
+    if not isinstance(input_object, dict):
+        raise ValidationException("Invalid filter value")
+
+    for name in input_object:
```
ryandillinfelton, line 56: We could easily make this a follow up. As is this will only work for single level deep objects. It should really be recursive (or equivalently iterative but that seems trickier)  if we'd like to make it more robust.

kruai:
The unit tests are passing, so I'm unclear as to why the PR check is failing atm. Is there an easy way to get it to run again without pushing another change? I seem to recall that re-running the job works, but doesn't show on the PR.

ryandillinfelton:
> The unit tests are passing, so I'm unclear as to why the PR check is failing atm. Is there an easy way to get it to run again without pushing another change? I seem to recall that re-running the job works, but doesn't show on the PR.

Usually re-running does work.

kruai:
> > The unit tests are passing, so I'm unclear as to why the PR check is failing atm. Is there an easy way to get it to run again without pushing another change? I seem to recall that re-running the job works, but doesn't show on the PR.
> 
> Usually re-running does work.

Oh, you're right. I must have used "Build Now" or "Rebuild Last" earlier. "Retry" appears to work, though the check is still failing. Investigating now

kruai:
Appears to be failing in a bonfire step where it's attempting to find a front-end deployment file. I'll post in Slack about it

ryandillinfelton (approved):

thearifismail (dismissed):
LGTM

api/filtering/filtering.py
```diff
@@ -46,17 +46,36 @@ def _wildcard_string_filter(field_name, field_value):
...
+        _check_field_in_spec(system_profile_spec()[field_name]["children"], name)
+        child_filter = system_profile_spec()[field_name]["children"][name]["filter"]
+        field_value = _get_field_value(input_object[name], child_filter)
```
ryandillinfelton, line 59: Whoops. This only works for singly nested stuff rn.

ryandillinfelton (changes requested):
Woops

tests/test_xjoin.py
```diff
@@ -2026,6 +2026,67 @@ def test_query_hosts_filter_spf_operating_system_exception_handling(
...
+        "filter[system_profile][ansible][controller_version]=7.1",
+        "filter[system_profile][ansible][hub_version]=8.0.*",
+        "filter[system_profile][ansible][catalog_worker_version]=nil",
+        "filter[system_profile][ansible][sso_version]=0.44.963",
+        "filter[system_profile][ansible][controller_version]=7.1&filter[system_profile][ansible][hub_version]=not_nil",
+        "filter[system_profile][ansible][catalog_worker_version]=not_nil",
+    )
+
+    graphql_queries = (
+        {"OR": [{"spf_ansible": {"controller_version": {"matches": "7.1"}}}]},
```
thearifismail, line 2041: Shouldn't `AND` be included also?

thearifismail (commented):
The `test_query_hosts_filter_spf_ansible` should include AND

tests/test_xjoin.py
kruai: No, I don't see how that would work. Where are you suggesting it would go?

api/filtering/filtering.py
kruai: I overhauled my implementation, and it should work for fully nested stuff now.

ryandillinfelton (approved):
Te functionality and implementation looks good to me, as does the test coverage. We've discussed the potential issues this may cause with future support of multiple values for object fields, but the conclusion is that the trade of is acceptable.

tests/test_xjoin.py
ryandillinfelton: > Shouldn't `AND` be included also?

It's not possible to request which boolean operation you want to use for the query in the REST API.  That said I think we should make this AND by default since we are not supporting multiple values per field here. I believe that's usually how our parameters work.

different fields are "anded" together, same field with multiple values is "ored"

ryandillinfelton (changes requested):
Whoops. I looked into Arifs comment more and I believe the boolean operation is wrong. Let me know your thoughts. It's not a hard and fast rule, it's mostly up to our judgement. The rule of thump i use is `AND` for combining different parameters, `OR` for combining multiple of the same type.

thearifismail (commented):
Minor comments

app/__init__.py
```diff
@@ -114,20 +114,24 @@ def _get_field_filter(field_name, props):
     return field_type
 
 
-def process_system_profile_spec():
-    with open(SYSTEM_PROFILE_SPECIFICATION_FILE) as fp:
-        # TODO: add some handling here for if loading fails for some reason
-        system_profile_spec = yaml.safe_load(fp)["$defs"]["SystemProfile"]["properties"]
-        system_profile_spec_processed = {}
+def process_spec(spec, process_unindexed=False):
```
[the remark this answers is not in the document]
kruai: `process_system_profile_spec` is not the only calling function. This function is now also called recursively by itself, so that it can generate queries based on deep objects.

kruai:
> Whoops. I looked into Arifs comment more and I believe the boolean operation is wrong. Let me know your thoughts. It's not a hard and fast rule, it's mostly up to our judgement. The rule of thump i use is `AND` for combining different parameters, `OR` for combining multiple of the same type.

OK. Because there's only one clause inside, this doesn't affect the query logic at all. However, if that's our rule of thumb, then I'll push the change.

app/__init__.py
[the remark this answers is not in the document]
thearifismail: oh I meant system_profile was the only context.  okay recursion is the reason.

thearifismail (approved):

ryandillinfelton (approved):
I'll  approve to unblock. I would like to also see test expansion in the follow up though.

With deep object queries we allow users to specify operation
`filter[system_profile][ansible][controller_version][eq]=7.1`
`filter[system_profile][ansible][controller_version][matches]=7*`

the way it's set up now is a sensible default and something we allow though.

PR #1034, opened 2021-11-15
Fix hostname_or_id filter when display_name and FQDN are different

thearifismail:
# Overview

This PR is being created to address [ESSNTL-1385](https://issues.redhat.com/browse/ESSNTL-1385).
Fixed the issue by removing  the assumption that display_name and FQDN have the same value.  This parameter used is "hostname_or_id" and the filter should not use `display_name`, which itself can be used for filtering.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


ryandillinfelton:
I don't think we can take out the host_name portion of the query. Even if it would make sense from an API design perspective it would be a breaking change for consumers.

thearifismail:
> I don't think we can take out the host_name portion of the query. Even if it would make sense from an API design perspective it would be a breaking change for consumers.

host-inventory should do the right thing.  With Ian not around, who can provide the provenance of this implementation? i.e. where has this requirement come from?

kruai:
> > I don't think we can take out the host_name portion of the query. Even if it would make sense from an API design perspective it would be a breaking change for consumers.
> 
> host-inventory should do the right thing. With Ian not around, who can provide the provenance of this implementation? i.e. where has this requirement come from?

The requirement comes from us having a search bar in our UI. It makes sense that if a user searches for a value in that box, that it should check against multiple things; the display_name, the fqdn, and the ID. We already had it implemented that way; if you're saying we should change it, I would like to know your reasoning.

kruai (changes requested):
As per yesterday's discussion, we should not be using `eq` for either of those comparisons.

thearifismail:
I am closing this PR because the fixed has been delivered in [xjoin-search PR #126](https://github.com/RedHatInsights/xjoin-search/pull/126)

PR #1036, opened 2021-11-18
Temporarily pin the ubi8 version

kruai:
# Overview

This PR is being created to address the issue discussed in [this Slack thread](https://coreos.slack.com/archives/CCRND57FW/p1637087808474000). They don't know how to fix the issue on their end yet, so they're recommending we pin our UBI8 version until they get it resolved.

After this gets merged, I'll create a follow-up PR to revert this change, which will be placed on hold until the issue is fixed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/python-38
+FROM registry.access.redhat.com/ubi8/python-38:1-74
 
 USER root
 


thearifismail:
The problem is with running dnf upgrad , which needs the system registered with redhat.   The temporary solution may be to use rhel8/python-38* image NOT ubi8/python-38*

thearifismail (approved):
LGTM

kruai:
OK, I'm going to go ahead and merge this so that we can see if it fixes the builds on the other PRs. If not, we'll try version 1-73; if that fails, we'll just revert this whole mess and work with DevProd to figure out a different solution.

PR #1039, opened 2021-11-22
Update System Profile spec

kruai:
# Overview

This PR just updates the system profile spec to the latest version. The main change is that now, if a reporter sends data for `operating_system`, they must send all 3 fields (`name`, `major`, and `minor`).
It also adds "CentOS" to the `operating_system.name` enum, so that value will now be allowed.


ryandillinfelton:
Since https://github.com/RedHatInsights/inventory-schemas/pull/84 was merged I think we should go ahead and merge this as well

kruai:
> Since [RedHatInsights/inventory-schemas#84](https://github.com/RedHatInsights/inventory-schemas/pull/84) was merged I think we should go ahead and merge this as well

OK. Then can you and @thearifismail please approve it?

ryandillinfelton (approved):
Whoops thought I had approved :+1:

PR #1041, opened 2021-11-29
Remove redundant spec parsing

Glutexo:
# Overview

_TranslatingParser_ does parsing automatically on initialization. Calling it again manually is not only unnecessary, but also inefficient. Removed the call from the _create_app_ function.

Loading, parsing and validating the specification is one of the heavier steps in the app initialization process. Doing it once instead of twice should measurably improve its boot time.

Does not relate to any JIRA ticket. This is a part of my Hackathon work on the Prance library.

## PR Checklist

- [x] Keep PR title short, ideally under 72 characters
- [ ] Descriptive comments provided in complex code blocks
- [ ] Tests: validate optimal/expected output
- [ ] Tests: validate exceptions and failure scenarios
- [ ] Tests: edge cases
- [ ] Recovers or fails gracefully during potential resource outages (e.g. DB, Kafka)
- [ ] Documentation, if this PR changes the way other services interact with host inventory
- [ ] Links to related PRs

## Secure Coding Practices Documentation Reference

You can find documentation on this checklist [here](https://github.com/RedHatInsights/secure-coding-checklist).

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [x] General Coding Practices

PR #1044, opened 2021-12-03
Revert "Remove redundant spec parsing"

thearifismail:
Reverts RedHatInsights/insights-host-inventory#1041.

The last merge is causing pr_check failures for other services e.g compliance and advisor


RonnyPfannschmidt (approved):

RonnyPfannschmidt:
@thearifismail do we have links to some issue/tracebacks seen so we can understand what happens there?
@Glutexo it seems we missed a hidden side-effect, i hope we can reveal it soon

ryandillinfelton (approved):

PR #1045, opened 2021-12-08
Allow filtering operating_system using "eq" operation

kruai:
# Overview

This PR is being created to address [ESSNTL-135](https://issues.redhat.com/browse/ESSNTL-135).
Because of how the filter works in the UI, we need to allow the System Profile to be filtered on the `operating_system` field using not only gt/gte/lt/lte, but also eq. This PR adds this capability, and modifies an existing test to verify that it works correctly.
After making some additional changes, it now allows for multiple version filters to be provided for the same operating system. For instance, if you wanted to get all hosts that are either RHEL 8.0 or RHEL 9.0, you can now filter using:
`?filter[system_profile][operating_system][RHEL][version][eq][]=8.0&filter[system_profile][operating_system][RHEL][version][eq][]=9.0`

This PR should only be merged after [xjoin-search PR 130](https://github.com/RedHatInsights/xjoin-search/pull/130), since that one enables using "eq" on the xjoin side.

## PR Checklist

- [ ] Tests: edge cases
- [x] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


api/filtering/custom_filters.py
```diff
@@ -8,21 +8,14 @@
...
-            },
-            {"spf_operating_system": {"major": {major_operation: major}, "name": {"eq": name}}},
-        ]
-    }
+    # same applies to gte and gt, and doesn't break for eq.
+
+    os_filter = [{"spf_operating_system": {"major": {"eq": major}, "minor": {operation: minor}, "name": {"eq": name}}}]
+
+    if operation != "eq":
+        os_filter.append({"spf_operating_system": {"major": {operation[0:2]: major}, "name": {"eq": name}}})
```
ryandillinfelton, line 16: I'm still a little partial to declaring the major operation explicitly to make the substring look less magic. The comment above does help with context but it's far enough away from this line that it might seem irrelevant.

api/filtering/custom_filters.py
```diff
@@ -8,21 +8,14 @@
...
+
+    return {"OR": os_filter}
```
ryandillinfelton, line 18: A few small notes here but overall I like the solution.

**major_operation**: I'm still a little partial to declaring the major operation explicitly to make the substring look less magic. The comment above does help with context but it's far enough away from this line that it might seem irrelevant.

**Unnecessary OR**: If the operation is `eq` we don't need to wrap the os_filter with or. I'd suggest we wrap `os_filter` in the `operation != "eq"` conditional to avoid the `OR` when we don't need it.

```suggestion
    if operation != "eq":
        #The major operation should only ever be 'lt' or 'gt'
        major_operation = operation[0:2]
        os_filter.append({"spf_operating_system": {"major": {major_operation: major}, "name": {"eq": name}}})
        os_filter = {"OR": os_filter}

    return os_filter
```
Then maybe remove the big comment up top, but I'll leave that to your discretion.

ryandillinfelton (changes requested):
Everything else looks muy bueno. Just saw some improvements that could be made. You don't have to take them all, but I think it's worth not adding the `OR` when we don't have to so he queries are smaller and more succinct.

api/filtering/custom_filters.py
kruai: OK, I've implemented your solution with some slight modifications to achieve the expected output you described.

ryandillinfelton (dismissed):
LGTM!

kruai:
I'm putting this on hold because I'm going to have to make additional changes to support passing multiple values for the same key. Right now, passing multiple RHEL values doesn't work.

thearifismail (approved):

kruai:
OK, I've added the ability to pass multiple OS values for the same operating system name. In other words, you can filter such that it returns hosts that are RHEL 7.0, RHEL 8.1, or RHEL 9.0 (for example). I also added tests to verify this behavior.

tests/test_xjoin.py
```diff
@@ -2058,22 +2060,33 @@ def test_query_hosts_filter_spf_operating_system(
...
                 },
+                {"AND": [{"spf_operating_system": {"major": {"eq": 12}, "minor": {"eq": 6}, "name": {"eq": "RHEL"}}}]},
+            ]
+        },
+        {
+            "OR": [
                 {
                     "AND": [
                         {
                             "OR": [
```
ryandillinfelton, line 2071: More of a discussion than a change request. Should this be `OR`? We've talked about the limitations of this REST API format before and how it's shortcoming is not having a way for users to specify the boolean operation, but do you think `OR` would be the more common use case or assumed behavior?

I think the rule of thumb I've been using throughout the API is to `OR` filters on different fields to expand the query (since RHEL AND CENT would return nothing) and `AND` filters on the same field to narrow the query. My expectation is that a user specifying two ranges for `[operating_system][RHEL]` would want to get their intersection rather than their union. I would advocate for switching this to `AND` for consistencies sake, but this honestly may be the only place we really have this problem so I could be swayed if you don't agree.

let me know what you think.

kruai: Hmm... we definitely need to use `OR` for checking multiple `eq` comparisons on one OS, as using `AND` would result in no records being returned. You're right that if we needed to do multiple inequality comparisons on one OS, it would make more sense to use AND. 
Priority-wise, getting in the `eq` functionality is critical, and I'm not sure if/when we'll need to use compound inequalities. Should I make a Jira to implement that?

ryandillinfelton (changes requested):
small improvement request. Everything looks solid otherwise!

One other thing, would do it inline but it's too far from the changes. I would like a test-case added to 
`test_query_hosts_filter_spf_operating_system_exception_handling`. I realized while reviewing we don't test how sending a query like `filter[system_profile][operating_system][RHEL][version][eq][]=9.0.1` is handled. I would like it to be with an exception, but as is it looks like we might just drop everything after the second `.`. If the case doesn't pass I think it'd make more sense to file a bug than fix it, but also push back if you think this is too far out of scope. Just seemed like it could be an easy add on if it already works.

tests/test_xjoin.py
ryandillinfelton: Dang you're right. Either `AND` or `OR` will make one thing wonky. The lesser of two evils seems to me to keep the `OR` for now since it would only make multiple range queries weird, while `AND` makes multiple `eq` useless. I'll make a ticket. I expect we could add some logic to create an `OR` group for the `eq` operations and an `AND` group for the `range` operations? But that's getting ahead of things.

api/filtering/custom_filters.py
```diff
@@ -6,23 +6,36 @@
...
+
+
+def _build_filter_from_version_string(os_value, name, operation):
+    major_version, *minor_version_list = os_value.split(".")
+
+    major_version = int(major_version)
+    minor_version = 0
+
+    if minor_version_list != []:
+        minor_version = int(minor_version_list[0])
```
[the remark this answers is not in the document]
kruai: Yep, it was inherited. I had to change the `>=1` to `>1` but it works perfectly otherwise!

kruai:
@ryandillinfelton I added the test you suggested, and made it so it raises a ValidationException if they pass in an OS filter with more than a major and minor version (e.g. `9.0.1` as you suggested).

ryandillinfelton (approved):
LGTM

api/filtering/custom_filters.py
```diff
@@ -6,23 +6,35 @@
...
+        # The major operation should only ever be 'lt' or 'gt'
+        major_operation = operation[0:2]
+        os_filter = {
+            "OR": [os_filter, {"spf_operating_system": {"major": {major_operation: major}, "name": {"eq": name}}}]
+        }
+
+    return os_filter
+
+
+def _build_filter_from_version_string(os_value, name, operation):
```
thearifismail, line 21: For clarity and ease of reading,  include return type `e.g. -> dict:` to function definitions

api/filtering/custom_filters.py
```diff
@@ -6,23 +6,35 @@
...
-                "spf_operating_system": {
-                    "major": {"gte": major, "lte": major},  # eq
-                    "minor": {operation: minor},
-                    "name": {"eq": name},
-                }
-            },
-            {"spf_operating_system": {"major": {major_operation: major}, "name": {"eq": name}}},
-        ]
-    }
+    os_filter = {"spf_operating_system": {"major": {"eq": major}, "minor": {operation: minor}, "name": {"eq": name}}}
```
thearifismail, line 9: should name matching be case-insensitive?

kruai: no - OS name is passed in from an enum, and we only need to check equality

api/filtering/custom_filters.py
kruai: We don't do that for any of our functions. Should that really block this PR from being merged?

thearifismail: Nah, but we should start adding them one by one, don't have to update all in one swoop

thearifismail (approved):

PR #1047, opened 2021-12-11
Check kafka availability before deleting hosts

thearifismail:
# Overview

This PR is being created to address [ESSNTL-38](https://issues.redhat.com/browse/ESSNTL-38).
(A description of your PR's changes, along with why/context to the PR, goes here.)

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


lib/host_kafka.py
```diff
@@ -0,0 +1,48 @@
...
+from app.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+def _check_connection(kafka_socket, server):
+    conn_check = 999
+    try:
+        host, port = server.split(":")
+        nw_addr = (host, int(port))
```
RonnyPfannschmidt, line 14: i suggest to put that into get_bootstrap servers instead of splitting it here, that way the check can be simplified

lib/host_kafka.py
```diff
@@ -0,0 +1,48 @@
...
+def _get_bootstrap_servers(servers):
+    if not servers:
+        config = Config(RuntimeEnvironment.SERVICE)
+        servers = config.bootstrap_servers
+    return servers
+
+
+def kafka_available(servers=None):
+    # random bad value. Set to zero by successful connection attempt
+    kafka_check = 999
```
RonnyPfannschmidt, line 30: lets remove the need for that variable alltogether

lib/host_kafka.py
```diff
@@ -0,0 +1,48 @@
...
+
+logger = get_logger(__name__)
+
+
+def _check_connection(kafka_socket, server):
+    conn_check = 999
+    try:
+        host, port = server.split(":")
+        nw_addr = (host, int(port))
+        conn_check = kafka_socket.connect_ex(nw_addr)
```
RonnyPfannschmidt, line 15:
```suggestion
        maybe_error_code = kafka_socket.connect_ex(nw_addr)
        if maybe_error_code == 0:
           return True
```

makes things way more easy on the loop

lib/host_kafka.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def kafka_available(servers=None):
+    # random bad value. Set to zero by successful connection attempt
+    kafka_check = 999
+
+    kafkas = _get_bootstrap_servers(servers)
+    kafka_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+    if isinstance(kafkas, str):
```
RonnyPfannschmidt, line 35: always normalize to a list of servers

lib/host_kafka.py
```diff
@@ -0,0 +1,48 @@
...
+        kafka_check = _check_connection(kafka_socket, kafkas)
+    elif isinstance(kafkas, list):
+        for server in kafkas:
+            kafka_check = _check_connection(kafka_socket, server)
+            if kafka_check == 0:
```
RonnyPfannschmidt, line 40: instead of break, use a return, and use `contextlib.closing` on the socket to do cleanup

lib/host_kafka.py
thearifismail: Done!

lib/host_kafka.py
thearifismail: Gone!

lib/host_kafka.py
thearifismail: Done!

lib/host_kafka.py
thearifismail: Done!

lib/host_kafka.py
thearifismail: IMO, not needed.  The code is too simple to use the suggestion.

thearifismail: Done.  Changed mind and update the code to use `contextlib.closing`

lib/host_kafka.py
```diff
@@ -0,0 +1,33 @@
...
+from contextlib import closing
+
+from app.config import Config
+from app.environment import RuntimeEnvironment
+from app.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+def _get_bootstrap_servers(kafka_socket, servers):
```
RonnyPfannschmidt, line 11: nice compression - new the name 
```suggestion
def _any_bootstrap_server_connects(kafka_socket, servers) ->  bool:
```

seems like match

tests/test_api_hosts_delete.py
```diff
@@ -32,9 +32,12 @@ def test_delete_with_invalid_host_id(api_delete_host):
     assert_response_status(response_status, expected_status=400)
 
 
-def test_create_then_delete(event_datetime_mock, event_producer_mock, db_create_host, db_get_host, api_delete_host):
+def test_create_then_delete(
+    event_datetime_mock, event_producer_mock, db_create_host, db_get_host, api_delete_host, mocker
+):
     host = db_create_host()
 
+    mocker.patch("api.host.kafka_available", return_value=True)
```
kruai, line 40: This change is made in so many tests that it would make more sense to use pytest fixtures. I recommend the following:

1. Create a fixture for when kafka should be available (let's call it `patch_kafka_available` for explanation purposes). It patches "api.host.kafka_available" the way you're doing in the above code.
2. Update the `api_delete_host` fixture to use the new `patch_kafka_available` fixture, so kafka will automatically be available for any test that uses `api_delete_host`.
3. Then, you can remove `mocker` from these tests, as they will be patched automatically.

When you test the functionality for kafka NOT being available, you can create another fixture (named `patch_kafka_unavailable`, for instance) and use that.

lib/host_kafka.py
```diff
@@ -0,0 +1,33 @@
...
+        servers = [config.bootstrap_servers]
+
+    try:
+        for server in servers:
+            try:
+                host, port = server.split(":")
+                new_addr = (host, int(port))
+                may_be_errcode = kafka_socket.connect_ex(new_addr)
+                if may_be_errcode == 0:
+                    return True
```
kruai, line 23: can be simplified
```suggestion
                if not kafka_socket.connect_ex(new_addr):
                    return True
```

lib/host_kafka.py
```diff
@@ -0,0 +1,33 @@
...
+            except ValueError as ve:
+                logger.error(f"Invalid server address: {str(ve)}")
+    except ValueError as ve:
+        logger.error(f"No servers available: {str(ve)}")
+
+
+def kafka_available(servers=None):
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as kafka_socket:
+        kafka = _any_bootstrap_server_connects(kafka_socket, servers)
+    return kafka is True
```
kruai, line 33: can be simplified
```suggestion
    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as kafka_socket:
        return _any_bootstrap_server_connects(kafka_socket, servers)
```

tests/test_unit.py
```diff
@@ -1184,8 +1186,8 @@ def test_without_culling_fields(self, deserialize_canonical_facts, deserialize_f
         }
         for additional_data in ({"stale_timestamp": "2019-12-16T10:10:06.754201+00:00"}, {"reporter": "puptoo"}):
             with self.subTest(additional_data=additional_data):
-                for mock in (deserialize_canonical_facts, deserialize_facts, deserialize_tags):
-                    mock.reset_mock()
+                for thismock in (deserialize_canonical_facts, deserialize_facts, deserialize_tags):
+                    thismock.reset_mock()
```
kruai, line 1190: Why is this change here?

tests/test_unit.py
```diff
@@ -2084,5 +2086,77 @@ def test_custom_regex_escape(self):
...
+
+    def test_happy_path(self):
+        real = host_kafka
+        real._any_bootstrap_server_connects = MagicMock()
+        real._any_bootstrap_server_connects.return_value = True
+
+        available = real.kafka_available()
+
+        assert available is True
+        real._any_bootstrap_server_connects.assert_called_once_with(mock.ANY, None)
```
kruai, line 2102: This is more of an issue of readability... but why did you make a variable named `real` which is set to the whole `host_kafka` class? `available` also seems unnecessary, since it's only read once. It would be more concise and IMO easier to understand without the extra vars:

```suggestion
        host_kafka._any_bootstrap_server_connects = MagicMock()
        host_kafka._any_bootstrap_server_connects.return_value = True

        assert host_kafka.kafka_available()
        host_kafka._any_bootstrap_server_connects.assert_called_once_with(mock.ANY, None)
```

tests/test_unit.py
```diff
@@ -2084,5 +2086,77 @@ def test_custom_regex_escape(self):
...
+    def test_one_invalid_bootstrap_server(self):
+        kafka_servers = ["localhos.129092"]
+        real = host_kafka
+        real._any_bootstrap_server_connects = MagicMock()
+        real._any_bootstrap_server_connects.return_value = False
+
+        available = real.kafka_available(kafka_servers)
+
+        assert available is False
+        real._any_bootstrap_server_connects.assert_called_once_with(mock.ANY, kafka_servers)
```
kruai, line 2158: These are all effectively testing the same things, which are:

1. `kafka_available` will call `_any_bootstrap_server_connects` and pass it the list of servers you provide
2. `kafka_available` will return what you mocked for `_any_bootstrap_server_connects.return_value`

It isn't actually testing whether any of the servers are invalid or malformed. In order to do that, you'll need some tests that actually validate `_any_bootstrap_server_connects`. In those tests, I'd recommend patching `kafka_socket.connect_ex` to return a mock error code.

lib/host_kafka.py
```diff
@@ -0,0 +1,33 @@
...
+from app.environment import RuntimeEnvironment
+from app.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+def _any_bootstrap_server_connects(kafka_socket, servers) -> bool:
+    if not servers:
+        config = Config(RuntimeEnvironment.SERVICE)
+        servers = [config.bootstrap_servers]
```
kruai, line 14: they're already a list
```suggestion
        servers = config.bootstrap_servers
```

kruai (changes requested):
The implementation mostly looks good! I have some suggestions to simplify some things, and some notes on the unit testing.

lib/host_kafka.py
```diff
@@ -0,0 +1,33 @@
...
+from contextlib import closing
+
+from app.config import Config
+from app.environment import RuntimeEnvironment
+from app.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+def _any_bootstrap_server_connects(kafka_socket, servers) -> bool:
```
ryandillinfelton, line 11: The function will never return false and may cause an error if non of the servers can be connected to.

lib/host_kafka.py
ryandillinfelton: If you do that it may be worth adding a comment explaining what's going on. Or maybe just keep the variable cast even though we only use the value once for clarity.
I think that
```
if not maybe_errorcode:
``` 
is pretty self documenting. reads like "if there's not an error code". However

```
if not kafka_socket.connect_ex(new_addr):
```
is not very clear IMO. It reads a bit like "if not connected to new address" or something.

tests/test_unit.py
```diff
@@ -2084,5 +2086,77 @@ def test_custom_regex_escape(self):
                 assert result == output
 
 
+class KafkaAvailabilityTests(TestCase):
```
ryandillinfelton, line 2089: Could use explicit testing for exception handling.

ryandillinfelton (changes requested):
I think Asa covered most of the bases here.

I somewhat disagree about on suggested change on style grounds but that's up for debate and will work correctly either way. Over all looks pretty close. 

Let me know if I'm off base with the return comment, it looks the there is a test covering that case so if it passes maybe I'm misunderstanding something.

tests/test_api_hosts_delete.py
thearifismail: Done.  Added a fixture `patch_kafka_available()` and a supporting function `_kafka_patch()`

lib/host_kafka.py
thearifismail: Added comment "# connect_ex() returns zero when the socket is open and accessible"

lib/host_kafka.py
thearifismail: Fixed per suggestion. I have an old habit of "not returning from within a loop

tests/test_unit.py
thearifismail: `thismock` had to be used in place of the original `mock` variable but now it is not needed due to code changes.

lib/host_kafka.py
thearifismail: When there is a single Kafka server, a string is returned as `host:port`.  `[]` has been added to deal with such cases.

api/host.py
```diff
@@ -144,20 +145,27 @@ def delete_by_id(host_id_list):
...
-                tracker_message = "deleted host"
-            else:
-                log_host_delete_failed(logger, host_id, get_control_rule())
-                tracker_message = "not deleted host"
-
-            with PayloadTrackerProcessingContext(
-                payload_tracker, processing_status_message=tracker_message
-            ) as payload_tracker_processing_ctx:
-                payload_tracker_processing_ctx.inventory_id = host_id
+        deletion_count = 0
```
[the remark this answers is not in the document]
thearifismail: oh this was left over after the reversion of delete filtered hosts

lib/host_kafka.py
thearifismail: In case of failures, callers have:
`logger.error("Kafka server not available.")`
`flask.abort(status.HTTP_503_SERVICE_UNAVAILABLE)`

tests/test_unit.py
thearifismail: oh still needed.  The presence of `from unittest import mock` cause `flake` to fail with: 
```
tests/test_unit.py:1189:21: F402 import 'mock' from line 11 shadowed by loop variable
```

tests/test_unit.py
thearifismail:
```
def _any_bootstrap_server_connects(kafka_socket, servers) -> bool:
```
When Kafka is not available, `_any_bootstrap_server_connects` simply logs a message and does not throw any exception.  In the API code, when `kafka_available()` does not return `True`, does the following:
```
            logger.error("Kafka server not available.")
            flask.abort(status.HTTP_503_SERVICE_UNAVAILABLE)
```

tests/test_unit.py
thearifismail: Can not patch `socket.connect_ex()` for being `read-only`.  

The intent for these tests to test different permutations which can be used as Kafka server addresses.

lib/host_kafka.py
ryandillinfelton: That will make a double nested given a list though right? Should it conditionally cast to a list like
```
servers = [config.bootstrap_servers] if config.bootstrap_servers is string else config.bootstrap_servers
```
or something like that?

thearifismail: Updated the code add suggested for being elegant.  Reason behind the previous implementation was `[[...]]` evaluated as string.  Here is how it was checked:
```
>>> fruit = [["apples", "oranges", "apricots"]]
>>> type(fruit)
<class 'list'>
>>> 
```

fijshion:
I could be misreading the code, it appears to do a single check for Kafka availability before deleting rows in the DB. What happens if Kafka goes down after the initial check but before the Kafka is message is attempted to be sent? I know this isn't very likely but when it happens it'll be a headache. Is it possible to roll back the DB change (via a DB transaction) if the Kafka message fails to be produced instead?

thearifismail:
> I could be misreading the code, it appears to do a single check for Kafka availability before deleting rows in the DB. What happens if Kafka goes down after the initial check but before the Kafka is message is attempted to be sent? I know this isn't very likely but when it happens it'll be a headache. Is it possible to roll back the DB change (via a DB transaction) if the Kafka message fails to be produced instead?

Being a very unlikely scenario, IMO it should be addressed in future as a refinement to this implementation.  Committing this one will help dealing with other places where ever hosts are deleted.

A bit of googling revealed that there is a notion of flashing back committed DB transactions but not sure if SQLAlchemy can or not.  The simpler option is checking "Kafka availability" [here](https://github.com/thearifismail/insights-host-inventory/blob/6069cf16d83648321f3a37ebc282088a4d560bd1/lib/host_delete.py#L37) before deleting every host.  This can be done in a separate PR.

fijshion:
SQLAlchemy does have transactions: https://docs.sqlalchemy.org/en/14/orm/session_transaction.html. I don't see any reason not to start with the code in this PR, but I just want to make sure using transactions isn't deprioritized because of it. DB transactions are standard practice and a clean way to handle this type of scenario. It also looks like SQLAlchemy makes them pretty simple to implement.

kruai:
> SQLAlchemy does have transactions: https://docs.sqlalchemy.org/en/14/orm/session_transaction.html. I don't see any reason not to start with the code in this PR, but I just want to make sure using transactions isn't deprioritized because of it. DB transactions are standard practice and a clean way to handle this type of scenario. It also looks like SQLAlchemy makes them pretty simple to implement.

That sounds like a good plan. We had been holding off on doing _Kafka_ transactions because we'd need to migrate to a different client lib first, but I'd love if we could start using transactions more.

FYI, we use the functions `session_guard` and `multi_session_guard` to handle session transactions already; we could probably use them here as well.

tests/test_api_hosts_delete.py
```diff
@@ -12,32 +11,33 @@
...
 
 
-def test_delete_non_existent_host(api_delete_host):
+def test_delete_non_existent_host(mocker, api_delete_host):
     host_id = generate_uuid()
 
-    response_status, response_data = api_delete_host(host_id)
+    response_status, response_data = api_delete_host(mocker, host_id)
 
     assert_response_status(response_status, expected_status=404)
```
ryandillinfelton, line 24: If you set up your fixtures differently you won't need to pass mocker to api_delete_host in every test you call it in. This change would cut down a lot of the line changes in this PR.

I tested it out just to make sure I was right and ended up just making the changes so I've opened a PR with the changes here: https://github.com/thearifismail/insights-host-inventory/pull/9

lib/host_kafka.py
ryandillinfelton: I was wondering about how statements like `if kafka_available():` in `host/delete.py` would resolve since this function doesn't abort the process or return false when it can't connect to any servers. Turns out functions will return `None` in python if there is no explicit return statement and since None will evaluate to false branching based on the return will work. I would still prefer an explicit `return False` fall though though for clarity. But I'll admit that's nitpicky so up to you.

tests/test_api_hosts_delete.py
thearifismail: Thanks @ryandillinfelton pulled last [PR #10](https://github.com/thearifismail/insights-host-inventory/pull/10)

tests/test_api_hosts_delete.py
```diff
@@ -104,45 +102,6 @@ def test_create_then_delete_without_insights_id(
     assert_delete_event_is_valid(event_producer=event_producer_mock, host=host, timestamp=event_datetime_mock)
 
 
-def test_delete_hosts_using_filter(
```
ryandillinfelton, line 107: Why is this test being deleted?

tests/fixtures/api_fixtures.py
```diff
@@ -76,3 +77,18 @@ def user_identity_mock(flask_app):
...
+
+    return _kafka_patch
+
+
+@pytest.fixture(scope="function")
+def patch_kafka_available():
+    def _patch_kafka_available(method):
+        kafka_patch(method)
+
+    return _patch_kafka_available
```
kruai, line 94: At the moment, these appear to only be replacing inline calls to `mocker.patch()`. Is that correct?
Since we want `kafka_available` to return True by default in our tests, I think it would be better to use a session-scoped patch, and then patch it as _unavailable_ when you need to test that scenario.

lib/host_delete.py
```diff
@@ -1,41 +1,49 @@
...
 
 
-def _delete_host(session, host):
+def _delete_host(session, event_producer, host):
     delete_query = session.query(Host).filter(Host.id == host.id)
-    delete_query.delete(synchronize_session="fetch")
-    delete_query.session.commit()
+    if kafka_available():
+        delete_query.delete(synchronize_session="fetch")
+        host_deleted = _deleted_by_this_query(host)
```
kruai, line 35: `_deleted_by_this_query(host)` only works after the deletion has been committed, according to its comments... is this tested and working as expected?

lib/host_delete.py
```diff
@@ -1,41 +1,49 @@
...
+        if host_deleted:
+            delete_host_count.inc()
+            event = build_event(EventType.delete, host)
+            insights_id = host.canonical_facts.get("insights_id")
+            headers = message_headers(EventType.delete, insights_id)
+            event_producer.write_event(event, str(host.id), headers, wait=True)
+            delete_query.session.commit()
```
kruai, line 42: I'm unable to find either a call to `rollback` or usage of `session_guard`. Is this attempting to use transactions as @fijshion suggested? If so, can you point me to where the rollback occurs?

tests/test_unit.py
```diff
@@ -8,6 +8,7 @@
 from json import dumps
 from random import choice
 from unittest import main
+from unittest import mock
```
kruai, line 11: You don't need to import this. `ANY`, `MagicMock` and `Mock` are all already imported.

tests/test_unit.py
kruai: That failed because you imported `mock`, which you didn't need to do (see above comment)

kruai (changes requested):
I have a few remaining questions and concerns

ryandillinfelton (approved):
LGTM! :tada:

lib/host_delete.py
thearifismail: `_deleted_by_this_query(host)` is called before committing the change. The comments has been changed to match the current implementation.  Besides, `ObjectDeletedError` is not mentioned or handled anywhere else in the host-inventory code.

lib/host_delete.py
thearifismail: just added rollback.  session_guard not needed because the transaction is committed when successful and rolled back when fails.

lib/host_delete.py
```diff
@@ -1,41 +1,49 @@
...
+            return True
+    else:
+        logger.error(f"host with {host.id} because Kafka server not available.")
```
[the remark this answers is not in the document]
thearifismail: Fixed the error message.

lib/host_delete.py
```diff
@@ -1,41 +1,49 @@
...
+        flask.abort(status.HTTP_503_SERVICE_UNAVAILABLE)
```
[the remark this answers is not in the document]
thearifismail: flask.abort removed.

lib/host_kafka.py
thearifismail: Now the method return False instead to running `flask.abort()`

lib/host_delete.py
```diff
@@ -1,41 +1,49 @@
...
-
-            host_deleted = _deleted_by_this_query(host)
-            if host_deleted:
-                delete_host_count.inc()
-
-                event = build_event(EventType.delete, host)
-                insights_id = host.canonical_facts.get("insights_id")
-                headers = message_headers(EventType.delete, insights_id)
-                event_producer.write_event(event, str(host.id), headers, wait=True)
+                host_deleted = _delete_host(select_query.session, event_producer, host)
```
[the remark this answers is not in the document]
thearifismail: Fixed code to return False when a host is not deleted.

tests/test_unit.py
thearifismail: No.  Removing them complains `not found` or `undefined`

thearifismail:
@kruai Changed fixture scope to session`@pytest.fixture(scope="session")` .  Kafka unavailable checks are tested by unit tests. No new integration tests created because there are existing tests to deal with KafkaError

tests/test_unit.py
kruai: I'm saying that right now, this file already imports a number of things from `unittest.mock`, including `ANY`.
It appears that you only added `from unittest import mock` so that you could use `mock.ANY`, but that's already imported.
If you just change the lines you added to `ANY` instead of `mock.ANY`, you can remove this import statement

thearifismail: Ah.  Thanks.  Done

lib/host_delete.py
kruai: I'm concerned about if `write_event()` fails and raises an exception. When using `session_guard`, it catches exceptions and automatically rolls back; in your code, I only see a rollback happen if `_deleted_by_this_query(host)` returns false (which only happens if something else deletes the host first). 

Is it being rolled back somewhere else I'm not seeing?

tests/fixtures/api_fixtures.py
```diff
@@ -87,3 +87,18 @@ def user_identity_mock(flask_app):
...
+
+    return _kafka_patch
+
+
+@pytest.fixture(scope="session")
+def patch_kafka_available():
+    def _patch_kafka_available(method):
+        kafka_patch(method)
+
+    return _patch_kafka_available
```
kruai, line 104: `scope="session"` applies the fixture for the whole session, which works, but only if you actually use the fixture. I don't see this fixture being used anywhere

tests/fixtures/api_fixtures.py
```diff
@@ -87,3 +87,18 @@ def user_identity_mock(flask_app):
     flask_app.user_identity = MockUserIdentity()
     yield flask_app.user_identity
     # flask_app.user_identity = None
+
+
+def kafka_patch(mocker):
+    def _kafka_patch(method):
+        return mocker.patch(method, return_value=True)
+
+    return _kafka_patch
```
kruai, line 96: As far as I can tell, you're patching the same function every time. Do we need a dedicated function for this, when we can just patch `mocker.patch("lib.host_delete.kafka_available")` directly?

tests/test_api_hosts_delete.py
```diff
@@ -34,7 +34,11 @@ def test_delete_with_invalid_host_id(api_delete_host):
     assert_response_status(response_status, expected_status=400)
 
 
-def test_create_then_delete(event_datetime_mock, event_producer_mock, db_create_host, db_get_host, api_delete_host):
+def test_create_then_delete(
+    event_datetime_mock, event_producer_mock, db_create_host, db_get_host, api_delete_host, mocker
+):
+    mocker.patch("lib.host_delete.kafka_available")
```
kruai, line 40: Not gonna comment on all of these, but... you have the `patch_kafka_available` fixture now. These patches should no longer be necessary once the fixture is requested. It would probably make sense to be requested by `event_producer_mock` - what do you think @ryandillinfelton?

kruai (changes requested):
In addition to the kafka availability tests you added, we also need some tests to validate the output when kafka fails to produce the DELETE messages. Or, if we already have tests that validate this, they should be updated to demonstrate that the deletion has been rolled back after the kafka error.

(Also, please update this PR's description to summarize at least the basics of what it's addressing)

api/host.py
```diff
@@ -177,21 +178,25 @@ def delete_host_list(
             tags,
             filter,
         )
+
+        if not len(ids_list):
```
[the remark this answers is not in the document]
thearifismail: Fixed!

lib/host_delete.py
thearifismail: Now using `with session_guard(select_query.session)` before starting the deletions.

lib/host_delete.py
```diff
@@ -1,47 +1,53 @@
...
+        raise KafkaError("Kafka server not available.  Stopping host deletions.")
 
 
 def _deleted_by_this_query(host):
-    # This process of checking for an already deleted host relies
-    # on checking the session after it has been updated by the commit()
-    # function and marked the deleted hosts as expired.  It is after this
-    # change that the host is called by a new query and, if deleted by a
-    # different process, triggers the ObjectDeletedError and is not emited.
+    # Before committing the change, verify that the host has been marked 'expired'
```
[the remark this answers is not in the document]
thearifismail: `instance_state(host).expired` means the host has marked for deletion but its deletion has not been committed yet.

tests/test_api_hosts_delete.py
thearifismail: Removed the `patch_kafka_available()` fixture.  Since `kafka_availablility` check stretches across the whole session, I tried to create a fixture for use in place of `mocker.patch("lib.host_delete.kafka_available")`, but have not found a way yet.

thearifismail (commented):
@asa, please take a look.  Thanks.

lib/host_delete.py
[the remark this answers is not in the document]
kruai: My understanding from the SQLAlchemy docs is that an object gets `expired` if another session has deleted the object; I think here we're making sure not to attempt to delete a host that was deleted by another process (like Reaper, or even another DELETE request).
But anyway, the code was returning whether the host was _not_ expired, so I was just correcting the comment.

tests/test_api_hosts_delete.py
thearifismail: @kruai @ryandillinfelton,
yesterday while trying to create a clean and crisp fixture to make `kafka_available` but the only way I could make it work was adding `kafka_patch("lib.host_delete.kafka_available")` to the `api_delete_host`, `api_delete_filtered_hosts`, and `api_delete_all_hosts` fixture but doing so required passing in `mocker` from every test using `api_delete_*`, which would lead to a lot more changes than what is in now.  That's why I took that out.

kruai: No, I said the `event_producer_mock` fixture, not any of the API ones; you're not patching an API function anyway.
I pulled down your code and removed all of those `mocker.patch()` statements and instead put one in the definition for `event_producer_mock`, and it functions as expected.

tests/test_culling.py
```diff
@@ -350,13 +356,11 @@ def test_culled_edge_host_is_not_removed(event_producer_mock, db_create_host, db
         shutdown_handler=mock.Mock(**{"shut_down.return_value": False}),
     )
 
-    assert db_get_host(created_host.id)
```
kruai, line 353: Why is this assertion being removed?

kruai (changes requested):
A few notes, but... I still take issue with this `patch` approach, especially because developers will just have to know to manually add this patch statement every time they use the event producer. See my comment about adding it to the `event_producer_mock` fixture

tests/test_culling.py
thearifismail: with use of `session_guard(select_query.session)` in `delete_hosts()`, the session is closed already when `host_reaper_run()` is called.  So instead whether an event was created or not.

tests/test_culling.py
```diff
@@ -439,7 +446,6 @@ def test_unknown_host_is_not_removed(
...
```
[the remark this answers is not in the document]
thearifismail: with use of `session_guard(select_query.session)` in `delete_hosts()`, the session is closed already when `host_reaper_run()` is called.  So instead whether an event was created or not.

lib/host_delete.py
```diff
@@ -1,47 +1,55 @@
...
+        raise KafkaError("Kafka server not available.  Stopping host deletions.")
 
 
 def _deleted_by_this_query(host):
-    # This process of checking for an already deleted host relies
-    # on checking the session after it has been updated by the commit()
-    # function and marked the deleted hosts as expired.  It is after this
-    # change that the host is called by a new query and, if deleted by a
-    # different process, triggers the ObjectDeletedError and is not emited.
+    # Before committing the change, verify that the host has been marked for deletion.
```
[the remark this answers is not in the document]
thearifismail: Put the original comment back.

tests/test_culling.py
kruai: True that the session gets closed, but I don't see why that means we can't check that the host still exists. If you're getting an error saying that `created_host` isn't bound to a session, I'd recommend only storing the ID, i.e.:
```
created_host_id = db_create_host(host=host).id
```
Then, you should be able to make both assertions without issue:
```
assert db_get_host(created_host_id)
assert event_producer_mock.event is None
```

That being said, the purpose of this test isn't to verify that the DB and Kafka messages are in sync, so consider this change optional - it shouldn't block the merge

tests/test_api_hosts_delete.py
```diff
@@ -90,7 +94,7 @@ def test_create_then_delete_without_request_id(
 
 
 def test_create_then_delete_without_insights_id(
-    event_datetime_mock, event_producer_mock, db_create_host, api_delete_host
+    event_datetime_mock, event_producer_mock, db_create_host, api_delete_host, mocker
```
kruai, line 97: `mocker` is defined but not used

tests/test_api_hosts_delete.py
```diff
@@ -154,7 +156,7 @@ def test_delete_hosts_using_filter(
 
 
 def test_delete_all_hosts(
-    event_producer_mock, db_create_multiple_hosts, db_get_hosts, api_delete_all_hosts, patch_xjoin_post
+    event_producer_mock, db_create_multiple_hosts, db_get_hosts, api_delete_all_hosts, patch_xjoin_post, mocker
```
kruai, line 159: `mocker` is defined but not used

tests/test_api_hosts_delete.py
```diff
@@ -344,8 +349,9 @@ def test_delete_hosts_chunk_size(
     ((mock.Mock(), mock.Mock(**{"get.side_effect": KafkaError()})), (mock.Mock(), KafkaError("oops"))),
 )
 def test_delete_stops_after_kafka_producer_error(
-    send_side_effects, kafka_producer, event_producer, db_create_multiple_hosts, api_delete_host, db_get_hosts
+    send_side_effects, kafka_producer, event_producer, db_create_multiple_hosts, api_delete_host, db_get_hosts, mocker
```
kruai, line 352: `mocker` is defined but not used

tests/test_culling.py
```diff
@@ -390,7 +389,7 @@ def test_non_culled_host_is_not_removed(
 
 
 @pytest.mark.host_reaper
-def test_reaper_shutdown_handler(event_datetime_mock, db_create_host, db_get_hosts, inventory_config):
+def test_reaper_shutdown_handler(db_create_host, db_get_hosts, inventory_config, mocker, event_producer_mock):
```
kruai, line 392: `mocker` is defined but not used

tests/test_culling.py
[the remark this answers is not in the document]
kruai: OK. I'm not gonna copy&paste what I said in the other place you made this change, but this check can still be done by changing one of the vars. Optional though and won't block the merge

tests/test_culling.py
```diff
@@ -460,14 +459,10 @@ def assert_system_culling_data(response_host, expected_stale_timestamp, expected
...
-    kafka_producer,
-    event_producer,
-    event_datetime_mock,
-    db_create_multiple_hosts,
-    db_get_hosts,
-    inventory_config,
+    send_side_effects, event_producer, db_create_multiple_hosts, db_get_hosts, inventory_config, mocker
 ):
+    mocker.patch("lib.host_delete.kafka_available")
+
```
kruai, line 465: `mocker` should be removed from the params, and the `mocker.patch` statement should be removed. Both the `event_producer` and `kafka_producer` fixtures are entirely mocked objects that don't call `kafka_available`.

kruai (changes requested):
Needs some cleanup, but it's almost there

tests/test_api_hosts_update.py
```diff
@@ -450,6 +450,7 @@ def test_patch_host_with_RBAC_bypassed_as_system(api_patch, db_create_host, even
 
 def test_update_delete_race(event_producer, db_create_host, db_get_host, api_patch, api_delete_host, mocker):
     mocker.patch.object(event_producer, "write_event")
+    mocker.patch("lib.host_delete.kafka_available")
```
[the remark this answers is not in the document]
thearifismail: `EventProducer.write_event()` needs `Kafka` for `send._kafka_prodcuer.send(...)`.  When `Kafka` is not available `assert event_producer.write_event.call_args_list[-1][0][2]["event_type"] == "delete"` fails.  The fix is to either update the `event_producer` fixture or use `mocker.patch()` like I did.  Since there were only 3 places where `event_producer` needed patching, I chose to go with explicit patching.

tests/test_culling.py
```diff
@@ -43,7 +43,7 @@ def test_with_stale_timestamp(mq_create_or_update_host, api_get):
     assert_system_culling_data(response_data["results"][0], stale_timestamp, reporter)
 
 
-def test_dont_get_only_culled(mq_create_hosts_in_all_states, api_get):
+def test_dont_get_only_culled(mq_create_hosts_in_all_states, event_producer_mock, api_get):
```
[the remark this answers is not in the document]
thearifismail: Fixed.

tests/test_api_hosts_delete.py
thearifismail: Gone!

tests/test_api_hosts_delete.py
```diff
@@ -183,15 +185,17 @@ def test_delete_all_hosts(
...
     # delete all hosts using incomplete filter
     response_status, response_data = api_delete_all_hosts({})
 
     assert_response_status(response_status, expected_status=400)
     assert event_producer_mock.event is None
 
 
-def test_create_then_delete_check_metadata(event_datetime_mock, event_producer_mock, db_create_host, api_delete_host):
+def test_create_then_delete_check_metadata(
+    event_datetime_mock, event_producer_mock, db_create_host, api_delete_host, mocker
```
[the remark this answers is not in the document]
thearifismail: Gone!

tests/test_api_hosts_delete.py
thearifismail: Gone!

tests/test_api_hosts_delete.py
```diff
@@ -116,17 +120,15 @@ def test_delete_hosts_using_filter(
     patch_xjoin_post,
     field,
     value,
+    mocker,
```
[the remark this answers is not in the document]
thearifismail: Gone!

tests/test_api_hosts_delete.py
thearifismail: Gone!

tests/test_api_hosts_delete.py
```diff
@@ -76,7 +80,7 @@ def test_create_then_delete_with_request_id(event_datetime_mock, event_producer_
 
 
 def test_create_then_delete_without_request_id(
-    event_datetime_mock, event_producer_mock, db_create_host, api_delete_host
+    event_datetime_mock, event_producer_mock, db_create_host, api_delete_host, mocker
```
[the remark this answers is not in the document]
thearifismail: Gone!

tests/test_api_hosts_delete.py
```diff
@@ -60,7 +62,9 @@ def test_create_then_delete_with_branch_id(
     assert not db_get_host(host.id)
 
 
-def test_create_then_delete_with_request_id(event_datetime_mock, event_producer_mock, db_create_host, api_delete_host):
+def test_create_then_delete_with_request_id(
+    event_datetime_mock, event_producer_mock, db_create_host, api_delete_host, mocker
```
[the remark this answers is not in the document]
thearifismail: Gone!

tests/test_api_hosts_delete.py
```diff
@@ -47,7 +49,7 @@ def test_create_then_delete(event_datetime_mock, event_producer_mock, db_create_
 
 
 def test_create_then_delete_with_branch_id(
-    event_datetime_mock, event_producer_mock, db_create_host, db_get_host, api_delete_host
+    event_datetime_mock, event_producer_mock, db_create_host, db_get_host, api_delete_host, mocker
```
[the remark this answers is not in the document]
thearifismail: Gone!

tests/test_api_hosts_delete.py
```diff
@@ -34,7 +34,9 @@ def test_delete_with_invalid_host_id(api_delete_host):
     assert_response_status(response_status, expected_status=400)
 
 
-def test_create_then_delete(event_datetime_mock, event_producer_mock, db_create_host, db_get_host, api_delete_host):
+def test_create_then_delete(
+    event_datetime_mock, event_producer_mock, db_create_host, db_get_host, api_delete_host, mocker
```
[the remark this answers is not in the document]
thearifismail: Gone!

tests/test_culling.py
```diff
@@ -201,15 +201,15 @@ def test_put_facts_works_on_non_culled(mq_create_hosts_in_all_states, api_put):
     assert response_status == 200
 
 
-def test_delete_ignores_culled(mq_create_hosts_in_all_states, api_delete_host):
+def test_delete_ignores_culled(mq_create_hosts_in_all_states, api_delete_host, mocker):
```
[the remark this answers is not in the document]
thearifismail: Gone!

tests/test_culling.py
```diff
@@ -201,15 +201,15 @@ def test_put_facts_works_on_non_culled(mq_create_hosts_in_all_states, api_put):
...
     culled_host = mq_create_hosts_in_all_states["culled"]
 
     response_status, response_data = api_delete_host(culled_host.id)
 
     assert response_status == 404
 
 
-def test_delete_works_on_non_culled(mq_create_hosts_in_all_states, api_delete_host):
+def test_delete_works_on_non_culled(mq_create_hosts_in_all_states, api_delete_host, mocker):
```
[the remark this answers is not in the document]
thearifismail: Gone!

tests/test_culling.py
thearifismail: without mocker.patch(), the test fails.   As mentioned above:
EventProducer.write_event() needs Kafka for send._kafka_prodcuer.send(...). When Kafka is not available assert event_producer.write_event.call_args_list[-1][0][2]["event_type"] == "delete" fails. The fix is to either update the event_producer fixture or use mocker.patch() like I did. Since there were only 3 places where event_producer needed patching, I chose to go with explicit patching.

tests/test_culling.py
```diff
@@ -390,7 +391,7 @@ def test_non_culled_host_is_not_removed(
 
 
 @pytest.mark.host_reaper
-def test_reaper_shutdown_handler(event_datetime_mock, db_create_host, db_get_hosts, inventory_config):
+def test_reaper_shutdown_handler(db_create_host, db_get_hosts, inventory_config, event_producer_mock):
```
kruai, line 394: Just curious... why is the `event_producer_mock` fixture being added here?

tests/test_culling.py
kruai: @thearifismail You shouldn't create the host again after it's deleted... I meant that because only the ID is used, you could replace this (L340):
```
created_host = db_create_host(host=host)
```
with this:
```
created_host_id = db_create_host(host=host).id
```
That way, you can just use `created_host_id` for your checks, since it will persist after the host is deleted.

tests/test_culling.py
```diff
@@ -439,7 +441,8 @@ def test_unknown_host_is_not_removed(
         shutdown_handler=mock.Mock(**{"shut_down.return_value": False}),
     )
 
-    assert db_get_host(created_host.id)
+    created_host_id = db_create_host(host=created_host).id
```
kruai, line 444: It doesn't make sense to create and validate a new host here. See my above comment about `created_host_id`.

kruai (changes requested):
One comment and a couple requests.

tests/test_culling.py
thearifismail: Had to add back `event_producer_mock`, the fixture because it provides `kafka` when not available.

tests/test_culling.py
thearifismail: I did wonder checking for an updated event NOT create but then it slipped off.  Since it is not needed, now it checks if an event was generated or not which does not.

lib/host_delete.py
```diff
@@ -1,41 +1,53 @@
...
+            headers = message_headers(EventType.delete, insights_id)
+            event_producer.write_event(event, str(host.id), headers, wait=True)
+            delete_query.session.commit()
+            return host_deleted
+        else:
+            delete_query.session.rollback()
+            return host_deleted
+    else:
+        logger.error(f"host with {host.id} NOT deleted because Kafka server not available.")
+        raise KafkaError("Kafka server not available.  Stopping host deletions.")
```
ryandillinfelton, line 50: I don't think that the code currently will stop _attempting_ to delete hosts when this error is raised. There is no exception handling done in the `delete_hosts` loop, so it will continue running and raising these exceptions.

I think this exception handling and structure need work. There's more nesting than needed, making it the responsibility of callers to raise the correct error when kafka can't be reached is error prone, and the exception you're raising is not being caught or acted upon.

I think you should instead check if Kafka is available first and if it's not re-raise the exception and return early.

```suggestion
    try:
        kafka_available()
    except KafkaError:
        logger.error(f"host with {host.id} NOT deleted because Kafka server not available.")
        raise

    delete_query.delete(synchronize_session="fetch")
    host_deleted = _deleted_by_this_query(host)
    if host_deleted:
        delete_host_count.inc()
        event = build_event(EventType.delete, host)
        insights_id = host.canonical_facts.get("insights_id")
        headers = message_headers(EventType.delete, insights_id)
        event_producer.write_event(event, str(host.id), headers, wait=True)
        delete_query.session.commit()
        return host_deleted
    else:
        delete_query.session.rollback()
        return host_deleted
```

lib/host_kafka.py
```diff
@@ -0,0 +1,35 @@
...
+
+    try:
+        for server in servers:
+            try:
+                host, port = server.split(":")
+                new_addr = (host, int(port))
+
+                # connect_ex() returns zero when the socket is open and accessible
+                errcode = kafka_socket.connect_ex(new_addr)
+                if not errcode:
```
ryandillinfelton, line 24: For clarity and to avoid any unlikely type coercion issues. Maybe also make 0 a constant in this file for more clarity. Then you could remove the comment

```suggestion
                if errcode == SUCCESS_CODE:
```

lib/host_kafka.py
```diff
@@ -0,0 +1,35 @@
...
+
+            except ValueError as ve:
+                logger.error(f"Invalid server address: {str(ve)}")
+    except ValueError as ve:
+        logger.error(f"No servers available: {str(ve)}")
+
+
+def kafka_available(servers=None):
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as kafka_socket:
+        return _any_bootstrap_server_connects(kafka_socket, servers)
```
ryandillinfelton, line 35: This goes along with the other comments in this review about changing how exceptions are handled. I think it would make sense to have the KafkaError raised here instead of where this is called. Then callers can chose to handle the exception as they see fit.

```suggestion
def kafka_available(servers=None):
    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as kafka_socket:
        if not _any_bootstrap_server_connects(kafka_socket, servers):
            raise KafkaError("No Kafka server available")
```

lib/host_delete.py
```diff
@@ -1,41 +1,53 @@
...
-        for host in select_query.limit(chunk_size):
-            host_id = host.id
-            with delete_host_processing_time.time():
-                _delete_host(select_query.session, host)
+    with session_guard(select_query.session):
+        while select_query.count():
+            for host in select_query.limit(chunk_size):
+                host_id = host.id
+                with delete_host_processing_time.time():
+                    host_deleted = _delete_host(select_query.session, event_producer, host)
```
ryandillinfelton, line 24: The KafkaError should be handled here to stop deletion if Kafka stops being available during a bulk deletion 

```suggestion
               try:
                   host_deleted = _delete_host(select_query.session, event_producer, host)
               except KafkaError:
                   logger.error("KafkaError: Stopping bulk deletion")
                   raise
```

tests/test_unit.py
```diff
@@ -2084,5 +2085,59 @@ def test_custom_regex_escape(self):
...
+        assert host_kafka.kafka_available(kafka_servers)
+        host_kafka._any_bootstrap_server_connects.assert_called_once_with(ANY, kafka_servers)
+
+    def test_one_invalid_bootstrap_server(self):
+        kafka_servers = ["localhos.129092"]
+        host_kafka._any_bootstrap_server_connects = MagicMock()
+        host_kafka._any_bootstrap_server_connects.return_value = None
+
+        assert host_kafka.kafka_available(kafka_servers) is None
+        host_kafka._any_bootstrap_server_connects.assert_called_once_with(ANY, kafka_servers)
```
ryandillinfelton, line 2139: As is `kafka_available` just returns whatever `_any_bootstrap_server_connects` returns. Since `_any_bootstrap_server_connects` is mocked these tests aren't very useful.

lib/host_kafka.py
```diff
@@ -0,0 +1,35 @@
...
+
+                # connect_ex() returns zero when the socket is open and accessible
+                errcode = kafka_socket.connect_ex(new_addr)
+                if not errcode:
+                    return True
+
+            except ValueError as ve:
+                logger.error(f"Invalid server address: {str(ve)}")
+    except ValueError as ve:
+        logger.error(f"No servers available: {str(ve)}")
```
ryandillinfelton, line 30: Where would the value error come from?

This looks to me like this should just return false if the for loop ends without returning true. Am I missing something?

ryandillinfelton (changes requested):
I left some suggestions around reworking exception handing and some structure and clarity changes.

I also think this need a more tests.

The current unit tests don't give me much confidence since basically all of the new logic is mocked. I really don't think you should be mocking `_any_bootstrap_server_connects` in the unit tests at all since it is what should be being tested.

I would also like:

- API tests that make sure we get the correct response from the API when Kafka isn't available.
- Tests that make sure it fails gracefully if Kafka goes down during a bulk delete operation.
- Tests that verify that the hosts are not deleted, and no delete events are produced if Kafka is not available

tests/test_api_hosts_delete.py
```diff
@@ -356,8 +355,8 @@ def test_delete_stops_after_kafka_producer_error(
     assert_response_status(response_status, expected_status=500)
 
     remaining_hosts = db_get_hosts(host_id_list)
-    assert remaining_hosts.count() == 1
-    assert event_producer._kafka_producer.send.call_count == 2
+    assert remaining_hosts.count() == 3
+    assert event_producer._kafka_producer.send.call_count == 0
```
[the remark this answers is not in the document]
thearifismail: When there is a Kafka error, no hosts are deleted and no events are generated.

api/host.py
```diff
@@ -178,21 +179,25 @@ def delete_host_list(
...
+        json_data = {"hosts_found": len(ids_list), "hosts_deleted": delete_count}
+
+        return flask_json_response(json_data, status.HTTP_202_ACCEPTED)
+
     except ValueError as err:
         log_get_host_list_failed(logger)
         flask.abort(400, str(err))
     except ConnectionError:
         logger.error("xjoin-search not accessible")
         flask.abort(503)
```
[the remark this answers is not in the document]
thearifismail: Done!

lib/host_kafka.py
thearifismail: Now using `errcode == 0`

lib/host_delete.py
thearifismail: > I don't think that the code currently will stop _attempting_ to delete hosts when this error is raised. There is no exception handling done in the `delete_hosts` loop, so it will continue running and raising these exceptions.
> 
> I think this exception handling and structure need work. There's more nesting than needed, making it the responsibility of callers to raise the correct error when kafka can't be reached is error prone, and the exception you're raising is not being caught or acted upon.
> 
> I think you should instead check if Kafka is available first and if it's not re-raise the exception and return early.

The current state of implementation is the result of review comments from the team.  The first implementation was to check for Kafka availability before starting the deletion.  If Kafka was inaccessible, don't start deleting hosts.  Then Chris K pointed out what if Kakfa disappeared after some hosts got delete but not all, which was a valid point.

When the first KafkaError is raised for the lack of Kafka availability, exception is caught by the caller in "api.host.py::delete*"

lib/host_delete.py
thearifismail: The KafkaError is handled by the caller in `api.host`

tests/test_unit.py
thearifismail: This is a unit test.  Relying on QE to run integration tests against it.

lib/host_kafka.py
thearifismail: `ValueError` is returned when a bogus kakfa server is provided.

tests/test_unit.py
ryandillinfelton: I just don't see what logic is being tested here. You're mocking the return value of `_any_bootstrap_server_connects` then checking the results of a call to `kafka_available` which just returns whatever `_any_bootstrap_server_connects` returns. No matter what values are provided for `kafka_servers` the test still passes.

Since `_any_bootstrap_server_connects` is entirely mocked I can replace it's whole body with pass and the tests still show no issues, and since no tests cover it's functionality the whole test suite still passes.

```
def _any_bootstrap_server_connects(kafka_socket, servers) -> bool:
    pass

def kafka_available(servers=None):
    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as kafka_socket:
        return _any_bootstrap_server_connects(kafka_socket, servers)
```

Since the core logic we're concerned with verifying lives in `_any_bootstrap_server_connects` we should make sure that the tests cover it. I understand a unit test should be focusing on one unit, but in this case `kafka_available` is basically just a context wrapper so I think it makes sense to test it and it's helper function in the same set of tests. You don't even need to mock anything since kafka does run during the tests and these are just verifying the handling of good/bad server addresses.

These all worked for me. And now if I effect how server addresses are verified they fail.
```
class KafkaAvailabilityTests(TestCase):
    def setUp(self,):
        super().setUp()
        self.config = Config(RuntimeEnvironment.TEST)

    def test_happy_path(self):
        assert host_kafka.kafka_available()

    def test_valid_server(self):
        assert host_kafka.kafka_available([self.config.bootstrap_servers])

    def test_list_of_valid_servers(self):
        kafka_servers = ["127.0.0.1:29092", "localhost:29092"]
        assert host_kafka.kafka_available(kafka_servers)

    def test_list_with_first_bad_second_good_server(self):
        kafka_servers = ["localhost29092", "127.0.0.1:29092"]
        assert host_kafka.kafka_available(kafka_servers)

    # second bad with missing ':'.  Returns as soon as the first kafka server found.
    def test_list_with_first_good_second_bad_server(self):
        kafka_servers = ["127.0.0.1:29092", "localhost29092"]
        assert host_kafka.kafka_available(kafka_servers)

    def test_one_invalid_bootstrap_server(self):
        kafka_servers = ["localhost29092"]
        assert host_kafka.kafka_available(kafka_servers) is None
```

lib/host_kafka.py
ryandillinfelton: In my testing I wasn't able to get this log to show up. What conditions would cause it and how are they different than those that get caught by the invalid server address except?
```
except ValueError as ve:
                logger.error(f"Invalid server address: {str(ve)}")

File: lib/host_delete.py (hunk 51 above)
Comment (ryandillinfelton) [reply lib/host_delete.py:50]: Gotcha, the raise stops the loop the function is called in. I thought you had to re-raise the exception for some reason.

I do think we could make this less nested though.
 
```suggestion
    if not kafka_available():
        logger.error(f"host with {host.id} NOT deleted because Kafka server not available.")
        raise KafkaError("Kafka server not available.  Stopping host deletions.")
    
    delete_query.delete(synchronize_session="fetch")
    host_deleted = _deleted_by_this_query(host)
    
    if not host_deleted:
        delete_query.session.rollback()
        return host_deleted

    delete_host_count.inc()
    event = build_event(EventType.delete, host)
    insights_id = host.canonical_facts.get("insights_id")
    headers = message_headers(EventType.delete, insights_id)
    event_producer.write_event(event, str(host.id), headers, wait=True)
    delete_query.session.commit()
    return host_deleted
```

ryandillinfelton (changes requested):

tests/test_unit.py
thearifismail: > Mocking is needed because pr_check runs in a standalone container with no Kafka available.  At this time, only DB is available but not sure if DB runs in the same container or has access to one.  In the past, kafka_availibility was not checked so mocking needs was not felt.

Working on how to verify failures.

tests/test_api_hosts_delete.py
```diff
@@ -339,27 +335,6 @@ def test_delete_hosts_chunk_size(
     query_wraper.query.limit.assert_called_with(5)
 
 
-@pytest.mark.parametrize(
```
thearifismail, line 342: This test is not needed any more and fails due to changes in code.

tests/test_unit.py
thearifismail: When Kafka not available None is returned and an exception thrown this func caller and the raised exception is caught by the caller in API code.

thearifismail (commented):
@ryandillinfelton please look at the responses.

lib/host_kafka.py
thearifismail: if a bogus host without `:` is used to get ValueError.

thearifismail (commented):
@ryandillinfelton, now connect_ex wrapper is used for patching

lib/host_kafka.py
kruai: Correct, but that gets caught by the `except ValueError as ve:` on line 32. At that point, the error is handled, and does not get reraised in order to be caught on line 38. Is that what you mean @ryandillinfelton ?

tests/test_culling.py
```diff
@@ -403,25 +403,26 @@ def test_reaper_shutdown_handler(event_datetime_mock, db_create_host, db_get_hos
...
+    facke_event_producer = mock.Mock()
 
     threadctx.request_id = UNKNOWN_REQUEST_ID_VALUE
+
     host_reaper_run(
         inventory_config,
         mock.Mock(),
         db.session,
-        event_producer_mock,
+        facke_event_producer,
```
kruai, line 414:
```suggestion
        fake_event_producer,
```

tests/test_culling.py
```diff
@@ -403,25 +403,26 @@ def test_reaper_shutdown_handler(event_datetime_mock, db_create_host, db_get_hos
...
         shutdown_handler=mock.Mock(**{"shut_down.side_effect": (False, True)}),
     )
 
     remaining_hosts = db_get_hosts(created_host_ids)
     assert remaining_hosts.count() == 1
-    assert event_producer_mock.write_event.call_count == 2
+    assert facke_event_producer.write_event.call_count == 2
```
kruai, line 420:
```suggestion
    assert fake_event_producer.write_event.call_count == 2
```

tests/test_api_hosts_delete.py
kruai: Can you explain your reasoning? I think it's valuable to know that a delete operation is aborted properly when the producer errors out

kruai (changes requested):
Requesting a few more changes, mostly typos. But there are also still some threads that remain unresolved

tests/test_culling.py
```diff
@@ -403,25 +403,26 @@ def test_reaper_shutdown_handler(event_datetime_mock, db_create_host, db_get_hos
     created_hosts = db_get_hosts(created_host_ids)
     assert created_hosts.count() == host_count
 
-    event_producer_mock = mock.Mock()
+    facke_event_producer = mock.Mock()
```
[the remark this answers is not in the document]
thearifismail: Fixed.

tests/test_culling.py
thearifismail: Fixed.

tests/test_culling.py
thearifismail: fixed.

tests/test_culling.py
```diff
@@ -403,25 +403,26 @@ def test_reaper_shutdown_handler(event_datetime_mock, db_create_host, db_get_hos
...
     remaining_hosts = db_get_hosts(created_host_ids)
     assert remaining_hosts.count() == 1
-    assert event_producer_mock.write_event.call_count == 2
+    assert facke_event_producer.write_event.call_count == 2
 
 
 @pytest.mark.host_reaper
 def test_unknown_host_is_not_removed(
-    event_producer_mock, db_create_host_in_unknown_state, db_get_host, inventory_config
+    event_producer_mock, db_create_host_in_unknown_state, db_get_host, inventory_config, db_create_host
```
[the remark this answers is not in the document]
thearifismail: Left over; should've been removed.

tests/test_api_hosts_delete.py
thearifismail: I don't understand the comment.  Is the comment placed?  Irregardless of this comment, this test has not been modified by this PR.

thearifismail (commented):
responded to Asa's comments.

lib/host_delete.py
thearifismail: This suggestion has been implemented.

tests/test_unit.py
kruai: Based on our conversation earlier, this was resolved.

kruai (approved):

ryandillinfelton (approved):
LGTM! Thanks for being such a good sport about the review process! I know these ones that go on for a long time are exhausting.

api/host.py
```diff
@@ -178,17 +179,22 @@ def delete_host_list(
...
 
-    if not len(ids_list):
-        flask.abort(status.HTTP_404_NOT_FOUND, "No hosts found for deletion.")
+    if len(ids_list) == 0:
```
RonnyPfannschmidt, line 190: this broke a qe test, i think we should change it back

RonnyPfannschmidt: (removal of the message that is)

PR #1050, opened 2021-12-14
Fix CJIs in template

bsquizz:
There are two CJIs which fail to run correctly when this template is deployed

One fails due to a bad cpu request/limit ratio:
```
  Warning  FailedCreate  9m1s   job-controller  Error creating: pods "host-inventory-events-topic-rebuilder-3n4ep5b--1-7pmj8" is forbidden: memory max limit to request ratio per Pod is 2, but provided ratio is 8.000000
```
so I am changing the template to use the CPU_REQUEST/MEMORY_REQUEST parameter everywhere. The values seemed to be set to `200m` and `256Mi` everywhere, which should still be the case since that will match the default values of the parameters.

The other fails because no 'job' exists on the ClowdApp anymore:
```
  Conditions:
    Last Transition Time:  2021-12-14T18:22:50Z
    Message:               Some Jobs are still incomplete
    Reason:                No such job duplicate-hosts-remover
    Status:                False
    Type:                  JobInvocationComplete
```
so I've removed this CJI from the template

### DIFF
--- a/deployment.yaml
+++ b/deployment.yaml
@@ -80,8 +80,8 @@ objects:
             cpu: ${CPU_LIMIT}
             memory: ${MEMORY_LIMIT}
           requests:
-            cpu: 200m
-            memory: 256Mi
+            cpu: ${CPU_REQUEST}
+            memory: ${MEMORY_REQUEST}
         livenessProbe:
           failureThreshold: 3
           httpGet:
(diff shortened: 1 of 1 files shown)


thearifismail (approved):
LGTM

kruai (approved):

PR #1011, opened 2021-10-06
Add API endpoint for deleting filtered hosts

thearifismail:
# Overview

This PR is being created to address [this Jira](https://issues.redhat.com/browse/ESSNTL-1203).
(A description of your PR's changes, along with why/context to the PR, goes here.)

## PR Checklist

- [ ] Keep PR title short, ideally under 72 characters
- [ ] Descriptive comments provided in complex code blocks
- [ ] Tests: validate optimal/expected output
- [ ] Tests: validate exceptions and failure scenarios
- [ ] Tests: edge cases
- [ ] Documentation, if this PR changes the way other services interact with host inventory
- [ ] Links to related PRs

## Secure Coding Practices Documentation Reference

You can find documentation on this checklist [here](https://github.com/RedHatInsights/secure-coding-checklist).

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #1055, opened 2021-12-15
Revert "Add API endpoint for deleting filtered hosts"

thearifismail:
Reverts RedHatInsights/insights-host-inventory#1011.  Reverting it per standup decision for not having time for testing for prod deployment before the holidays.


kruai (approved):
LGTM

PR #1056, opened 2021-12-15
Revert operating_system required fields

kruai:
We currently require `system_profile.operating_system` to have either **all** or **none** of its properties populated. This PR temporarily removes this requirement, in order to reduce the number of things that will likely go wrong right before holiday PTO.


ryandillinfelton (approved):
LGTM. Just remember to update the schemas repo first.

thearifismail (approved):
LGTM

PR #1059, opened 2022-01-04
Unpin version and disable beta RPMs

kruai:
# Overview

This PR is being created to address [ESSNTL-1961](https://issues.redhat.com/browse/ESSNTL-1961), which progresses work on the issue discussed in this [Slack thread](https://coreos.slack.com/archives/CCRND57FW/p1637087808474000). It unpins the UBI8 version, and disables a couple of beta RPM repos. 

The issues we saw were caused by newer versions of the UBI8 image using Beta RPM repos, which subscription-manager did not like.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
-FROM registry.access.redhat.com/ubi8/python-38:1-74
+FROM registry.access.redhat.com/ubi8/python-38
 
 USER root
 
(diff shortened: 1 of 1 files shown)


kruai:
Reading through the logs, this appears to be a valid workaround for the subscription-manager issue we were seeing earlier. We can remove the line that disables beta repos if they ever fix that on the base UBI8 image, but we'll have to leave it until then.

ryandillinfelton (approved):
Looks alright to me

PR #1066, opened 2022-01-13
Fix dependency issue and add Snappy support

kruai:
# Overview

This PR is being created to address [ESSNTL-2022](https://issues.redhat.com/browse/ESSNTL-2022).
It fixes the dependency resolution error that was triggered by a conflict between connexion and v0.2 of openapi-schema-validator, which released recently. It also adds support for Snappy, so that when our Kafka brokers implement compression, we can decompress them.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ USER root
 
 RUN dnf config-manager --disable rhel-8-for-x86_64-baseos-beta-rpms rhel-8-for-x86_64-appstream-beta-rpms
 RUN dnf module install -y postgresql:13
+RUN dnf module install -y snappy-devel
 
 # remove packages not used by host-inventory to avoid security vulnerabilityes
 RUN dnf remove -y npm
--- a/Pipfile
+++ b/Pipfile
@@ -17,6 +17,7 @@ flask-migrate = "~=2.5.2"
 itsdangerous = "~=1.1.0"
 gunicorn = "~=19.9.0"
 connexion = {version = "~=2.7",extras = ["swagger-ui"]}
+openapi-schema-validator = "< 0.2"
 pyyaml = "~=5.4"
 prometheus-client = "~=0.7.1"
 logstash-formatter = "~=0.5.17"
@@ -33,6 +34,7 @@ app-common-python = "~=0.1.2"
 honcho = "*"
 sqlalchemy = "<1.4.0"
 urllib3 = ">=1.26.5"
+python-snappy = "*"
 
 [dev-packages]
 pytest = "~=5.0.1"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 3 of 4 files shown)


kruai:
The main change I'm not sure about is the Dockerfile. I used `dnf module install` to match our approach with `postgresql`, but I'm not sure if that will work. If it doesn't, hopefully the standard `dnf install` will.

kruai:
Well, it looks like neither works. I might have to reach out to AppSRE to see what's available, or the Strimzi team to see how they implemented it on their end

Dockerfile
```diff
@@ -4,6 +4,7 @@ USER root
 
 RUN dnf config-manager --disable rhel-8-for-x86_64-baseos-beta-rpms rhel-8-for-x86_64-appstream-beta-rpms
 RUN dnf module install -y postgresql:13
+RUN dnf install snappy-devel
```
thearifismail, line 7: Was any attempt made to build a local image?

kruai: I answered when you asked in Slack, but: no, at this point, I had not. I ran this command on my own machine to get the Snappy installation

kruai:
Turns out we only need to install Snappy on our local machines in order to use python-snappy, so I removed the installation command from the Dockerfile. I'll also add that requirement to our README.

ryandillinfelton (approved):
LGTM! Thanks for tracking that down

ryandillinfelton:
> Turns out we only need to install Snappy on our local machines in order to use python-snappy, so I removed the installation command from the Dockerfile. I'll also add that requirement to our README.

Is that so? I sitll see Snappy installs on both Dockerfiles. I would expect it's necessary for deploys unless it's part of the base image but it' still in the dev one too. Am I missing something?

kruai:
> > Turns out we only need to install Snappy on our local machines in order to use python-snappy, so I removed the installation command from the Dockerfile. I'll also add that requirement to our README.
> 
> Is that so? I sitll see Snappy installs on both Dockerfiles. I would expect it's necessary for deploys unless it's part of the base image but it' still in the dev one too. Am I missing something?

I was trying out different permutations on the ephemeral nodes. I had removed both installation lines briefly to see if it would complain, and they didn't, but that could have been because they were already installed. I re-added them and it seems to work fine, so... fingers crossed that AppSRE doesn't have to add a module again!

ryandillinfelton:
> I was trying out different permutations on the ephemeral nodes. I had removed both installation lines briefly to see if it would complain, and they didn't, but that could have been because they were already installed. I re-added them and it seems to work fine, so... fingers crossed that AppSRE doesn't have to add a module again!

Shouldn't be an issue to reinstall a package so I see no issue with merging this as is.

thearifismail (approved):

PR #1071, opened 2022-01-19
essntl-1993: Add staleness and registered_with as params for deleting filtered hosts

thearifismail:
# Overview

This PR is being created to address [ESSNTL-1993](https://issues.redhat.com/browse/ESSNTL-1993), which is about using `staleness` and `registered_with` as input parameters to filter hosts on.  Previously default values of `['fresh', 'stale', 'unknown'] were set in the `api.spec.yaml`, which prevented `stateleness` to be used as input parameter. 

This PR removes the `staleness` default values from `api.spec.yaml` and set them in code where needed.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


api/host.py
```diff
@@ -147,7 +147,10 @@ def delete_host_list(
     tags=None,
     filter=None,
 ):
-    if not any([display_name, fqdn, hostname_or_id, insights_id, provider_id, provider_type, tags, filter]):
+    # "unknown" in staleness means staleness has default values and is not being used as a filter parameter
```
kruai, line 150: I don't think that's true... if the request has `?staleness=unknown`, it's definitely being used as a filter parameter.
`unknown` is the default value for a Host record's staleness, not the filter.

api/host.py
```diff
@@ -147,7 +147,10 @@ def delete_host_list(
...
+    if "unknown" in staleness and not any(
+        [display_name, fqdn, hostname_or_id, insights_id, provider_id, provider_type, registered_with, tags, filter]
+    ):
```
kruai, line 153: I think you just need to account for whether `staleness` was provided at all, i.e.:
```suggestion
    if not any(
        [display_name, fqdn, hostname_or_id, insights_id, provider_id, provider_type, registered_with, staleness, tags, filter]
    ):
```

fstavela: In the first moment I also thought that, but the problem is that staleness will never be `None`, because the default value for the parameter is `["fresh", "stale", "unknown"]`.

I think we have 4 options here.

The easiest option would be to do `if staleness != ["fresh", "stale", "unknown"] and not any(...)`, but it wouldn't be correct, because if someone manually sets the staleness parameter to these values, we would reject his request too.

Another one would be to check if the parameter was actually provided, but I don't know how and if that can be done.

Another option would be to create separate staleness API parameter for the deletion endpoint, which wouldn't have any default values and then we can simply add `staleness` to `if not any(...)`. But we would then have to check `if staleness is None: staleness = default_values_for_staleness`.

The best option in my opinion is to finally add `stale_warning` to the default options in the staleness filter. That way we could check `if len(set(staleness) & {"fresh", "stale", "stale_warning"}) == 3` (I'm intentionally leaving `unknown` here because it doesn't do anything anyway and in my opinion it should be removed). That way we don't have to worry about someone manually setting staleness to `["fresh", "stale", "stale_warning"]`, because even if they do, they are not really filtering the hosts (as it would return all of them), so we can safely reject the request.
Also, not having `stale_warning` in the default options causes another problem and that is that if someone uses different parameter for filtering, it will not result in deleting all hosts satisfying that filter, because hosts that are in `stale_warning` state would not be deleted.

api/host.py
fstavela: I think the intended interpretation was that `unknown` is in the default values and no one would set it manually anyway.

api/host.py
kruai: I'm resolving this, since we hashed out the details on Thursday.

thearifismail:
> 

Done!

ryandillinfelton (changes requested):
I don't like spreading out the default like this. Would it make more sense to defined two versions of `stalenessParam`: `stalenessParam` and `stalenessParamWithDefaults` and use the default free version on the delete endpoint. 

Then we wouldn't need to add the default values in the code itself. 

I think it also makes sense to have the delete endpoint have no defaults if we don't want users to use it without specifying the filter explicitly. Since adding `?staleness=[fresh,stale,unknow]` explicitly is valid, it wouldn't make sense to add the equivalent as a default but not allow users to use the endpoint with just the default filter when it would do the same thing.

api/filtering/filtering.py
```diff
@@ -252,6 +252,8 @@ def query_filters(
 
     if tags:
         query_filters += build_tag_query_dict_tuple(tags)
+    if staleness is None:
+        staleness = ["fresh", "stale", "unknown"]
```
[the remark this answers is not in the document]
thearifismail: Setting the default values in `api.spec.yaml` makes it very difficult to differentiate whether the `staleness` value passed in as a filter parameter or by the default values.  To delete filtered hosts, user must provide at least one filter parameter but if `default` values coming from `api.specs`, it is not possible.  That's why the default values have been removed from the api.spec.  I am open to suggestions for a more elegant implementation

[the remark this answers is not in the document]
kruai: IIRC when we discussed this in our chit-chat meeting, we agreed to one of the following:

- Remove the default values from the `staleness` definition in the API spec, and handle the defaults internally. 
- Create an additional `staleness` definition with no default values, which would get used by this delete endpoint.

I think we landed on the former, but LMK if that doesn't match with your notes @thearifismail. If we're doing it that way, we definitely need to just have one place where those values are defined, rather than several.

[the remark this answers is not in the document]
thearifismail: @ryandillinfelton Setting default values for staleness in api.spec.yaml poses problems when staleness is used for filtering hosts because it always has a set of values from api.spec.  To delete filtered hosts, user must provide at least one filter parameter but if default values coming from api.specs, it is not possible. That's why the default values have been removed from the api.spec.

@kruai It was a two step process.  First remove `staleness defaults` from `api.spec`.  Second define additional staleness definition in a separate PR, if needed.  For staleness filter to work, `api.filtering.query_filters()` is the only place to set "staleness_defaults", but the way tests are written, it has to be set in `api.host_query_db.py` and `app.xjoin.py`.  To have one place to set default, a new module `api.utils.py` has been added.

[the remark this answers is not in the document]
kruai: @ryandillinfelton @thearifismail We also talked about treating that filter like it wasn't provided, if it contains both "fresh" and "stale". If we did that, we wouldn't have to remove the defaults from the API spec, and we wouldn't have to make a bunch of changes to support it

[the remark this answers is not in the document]
kruai: So if someone calls `DELETE /hosts?staleness=fresh,stale`, it would return 400 bc that's the same as not providing a staleness filter at all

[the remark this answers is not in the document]
thearifismail: > @ryandillinfelton @thearifismail We also talked about treating that filter like it wasn't provided, if it contains both "fresh" and "stale". If we did that, we wouldn't have to remove the defaults from the API spec, and we wouldn't have to make a bunch of changes to support it

To meet this req, `api/utils.py` has been added to provide default values as needed by a large number of tests, yet allowing staleness filter getting used as intended.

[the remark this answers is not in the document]
thearifismail: @kruai @ryandillinfelton Put staleness defaults back in api.spec because `delete_all` is going to have its own separate `endpoint`.

swagger/api.spec.yaml
```diff
@@ -597,6 +597,20 @@ components:
...
+      schema:
+        type: array
+        items:
+          type: string
+          enum:
+            - fresh
+            - stale
+            - stale_warning
+            - unknown
+      description: "stalenessNoDefaultsParam reference used for deleting hosts using staleness"
```
ryandillinfelton, line 613: This description is for users, not us. I think it could have the same description as stalenessParam, just without the comment about defaults.

api/host.py
```diff
@@ -147,7 +157,27 @@ def delete_host_list(
...
+            provider_id,
+            provider_type,
+            registered_with,
+            staleness,
+            tags,
+            filter,
+        ]
+    ):
         logger.error("bulk-delete operation needs at least one input property to filter on.")
         flask.abort(400, "bulk-delete operation needs at least one input property to filter on.")
```
ryandillinfelton, line 182: I'm curious why the addition of args_count is necessary when it seems like it checks the same thing as the `if not any` statement. 

If I'm wrong there and it is necessary to have both though I'd prefer to have them in an or to avoid repeating the error and abort calls. I also slightly prefer using `== 0` to `not`, it just feels a little more descriptive to me, or fitting of the check being done, but I won't block or insist on it.
```suggestion
    if args_count == 0 or if not any(
        [
            display_name,
            fqdn,
            hostname_or_id,
            insights_id,
            provider_id,
            provider_type,
            registered_with,
            staleness,
            tags,
            filter,
        ]
    ):
        logger.error("bulk-delete operation needs at least one input property to filter on.")
        flask.abort(400, "bulk-delete operation needs at least one input property to filter on.")
```

ryandillinfelton (changes requested):
I think it could use some minor tweaks but overall it's looking good

api/host.py
kruai: I agree. In addition, we've already got all the args in the variable `input_args` at this point; instead of listing them all out again, you can just do `if not any( [*input_args] )`.

api/host.py
```diff
@@ -147,7 +157,27 @@ def delete_host_list(
     tags=None,
     filter=None,
 ):
-    if not any([display_name, fqdn, hostname_or_id, insights_id, provider_id, provider_type, tags, filter]):
+    input_args = deepcopy(locals())
```
kruai, line 160: What's the deepcopy for?

tests/test_xjoin.py
```diff
@@ -1666,7 +1666,7 @@ def test_xjoin_search_query_using_hostfilter(
     api_delete_filtered_hosts({field: value})
 
     graphql_query_empty_response.assert_called_once_with(
-        HOST_IDS_QUERY, {"filter": ({field: {"eq": value}}, mocker.ANY)}, mocker.ANY
+        HOST_IDS_QUERY, {"filter": ({field: {"eq": value}},)}, mocker.ANY
```
kruai, line 1669: This goes for the changes below as well, but what prompted this change?

kruai (changes requested):

swagger/api.spec.yaml
thearifismail: Done!

tests/test_xjoin.py
thearifismail: staleness does not have default values anymore.

api/host.py
thearifismail: Using `input_args = deepcopy(locals())` produces: 
```
input_args = {'display_name': None, 'fqdn': None, 'hostname_or_id': None, 'insights_id': None, 'provider_id': None, 'provider_type': None, 'registered_with': None, 'staleness': None, 'tags': None, 'filter': {}}
```
Whereas `input_args = locals()` produces: 
```
input_args = {'display_name': None, 'fqdn': None, 'hostname_or_id': None, 'insights_id': None, 'provider_id': None, 'provider_type': None, 'registered_with': None, 'staleness': None, 'tags': None, 'filter': {}, 'input_args': {...}}
```
The latter has `'input_args': {...}` at the end i.e. in the object, which does not lead zero arguments even when no input argument is provided.

I guess the symbols returned by `locals()` has some metadata in the object which gets dumped when the deepcopy() is made producing an object yielding the expected structure.

api/host.py
thearifismail: > I'm curious why the addition of args_count is necessary when it seems like it checks the same thing as the if not any statement.
@ryandillinfelton `args_count()` checks if any parameter with non-None value has been passed in.  `if not any([...]) could not tell me.  i.e. counts the number of arguments passed in.

thearifismail:
@kruai, I did not know that `([*input_args])` returned a list of the dictionary keys, i.e. somewhat equivalent to `input_args.keys()`.

Since `_get_input_args_count(input_args)` counts the args with `non-None` values, `if not any([...])` is not needed anymore, so it is gone

api/host.py
thearifismail: @kruai checking of number arguments passed in is gone because `if not any([...])` does both jobs of checking for `None` values and no args passed.

api/host.py
```diff
@@ -147,7 +147,20 @@ def delete_host_list(
...
+            hostname_or_id,
+            insights_id,
+            provider_id,
+            provider_type,
+            registered_with,
+            staleness,
+            tags,
+            filter,
+        ]
+    ):
```
kruai, line 163: Why did we abandon the `locals()` approach? I verified locally that this does the same thing as:
```suggestion
    if not any(locals()):
```

tests/test_xjoin.py
kruai: OK, that makes sense!

kruai (approved):
The way you have it now is fine with me; I just had one suggestion/question

ryandillinfelton (approved):
LGTM! I do agree with Asa's comment about `not any [locals()]` or however it is being preferable. But I'm keen to get this closed So I vote merge now and do that as a follow up if it's really bugging anyone.

api/host.py
thearifismail: There are more than one reasons.

1. `locals()` was to count the args passed in because at the time `if not any([...])` did not work due to my IDE or project being in bad state.
2.  The object returned by `locals()` structure is more complex than the simple dictionary of input arguments that deepcopy returns.  e.g. `len(locals())` would return the number of arguments where `if not locals()` would succeed for having all `None` In other I didn't feel confident in using it.
3.  The most important reason, `in not any([...])` is very explicit and easy to read.

thearifismail:
> LGTM! I do agree with Asa's comment about `not any [locals()]` or however it is being preferable. But I'm keen to get this closed So I vote merge now and do that as a follow up if it's really bugging anyone.

Thanks but I still favor `if not any9[....])` for being explicit.

PR #1082, opened 2022-02-01
Delete all hosts using a separate API endpoint.

thearifismail:
# Overview

This PR is being created to address ESSNTL-1967.
It adds a new API endpoint `/hosts/delete_all_hosts` which requires two parameters `delete_all=true` and `confirm_delete_all=true`.  Requiring two input parameters to delete all hosts is an extra caution to avoid any unintended use. 
Host IDs are retrieved using a DB query bypassing "xjoin-search" which does not allow fetching more than 100 hosts at a time.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


tests/test_xjoin.py
```diff
@@ -1737,6 +1737,14 @@ def test_xjoin_search_query_using_hostfilter_provider(
     )
 
 
+def test_xjoin_search_query_to_delete_all_hosts(mocker, graphql_query_empty_response, api_delete_all_hosts):
+    query_params = {"delete_all": True, "confirm_delete_all": True}
+    api_delete_all_hosts(query_params)
+
+    # using mocker.ANY because no filter is used to fetch all hosts.
+    graphql_query_empty_response.assert_called_once_with(HOST_IDS_QUERY, {"filter": (mocker.ANY)}, mocker.ANY)
```
ryandillinfelton: Doesn't using any make the test ineffective? If anything could be in filter, then it could be adding filters that would prevent it from selecting all hosts.

tests/test_api_hosts_delete.py
```diff
@@ -153,6 +153,51 @@ def test_delete_hosts_using_filter(
...
+    # for querying for deletion using filters
+    patch_xjoin_post(response, status=200)
+
+    new_hosts = db_create_multiple_hosts()
+    new_ids = [str(host.id) for host in new_hosts]
+
+    # delete hosts using the IDs supposedly returned by the query_filter
+    response_status, response_data = api_delete_all_hosts({"delete_all": True, "confirm_delete_all": True})
+
+    assert '"type": "delete"' in event_producer_mock.event
```
kruai, line 178: It would also be good to verify that the event producer was used the correct number of times, since the above only tells you that at least one "delete" event was written. As we expect all of the hosts you created to be deleted, it would look something like this:
```suggestion
    assert '"type": "delete"' in event_producer_mock.event
    assert event_producer_mock.write_event.call_count == len(created_hosts)
```

tests/test_api_hosts_delete.py
```diff
@@ -153,6 +153,51 @@ def test_delete_hosts_using_filter(
...
+    assert len(host_ids) == response_data["hosts_deleted"]
+
+    # check db for the deleted hosts using their IDs
+    host_id_list = [str(host.id) for host in created_hosts]
+    deleted_hosts = db_get_hosts(host_id_list)
+    assert deleted_hosts.count() == 0
+
+    # now verify that the second set of hosts still available.
+    remaining_hosts = db_get_hosts(new_ids)
+    assert len(new_hosts) == remaining_hosts.count()
```
kruai, line 189: This is invalid - there should be no hosts remaining.

swagger/api.spec.yaml
```diff
@@ -68,6 +68,26 @@ paths:
           description: Request for deletion of filtered hosts has been accepted.
         '400':
           description: Invalid request.
+  /hosts/delete_all_hosts:
```
ryandillinfelton, line 71: I think `/hosts/all` would look a cleaner.

api/host.py
```diff
@@ -226,6 +226,38 @@ def _delete_filtered_hosts(host_id_list):
...
+@api_operation
+@rbac(Permission.WRITE)
+@metrics.api_request_time.time()
+def delete_all_hosts(delete_all=None, confirm_delete_all=None):
+    if not any([delete_all, confirm_delete_all]):
+        logger.error("To delete all hosts, provide delete_all=true and confirm_delete_all=true in the request.")
+        flask.abort(400, "To delete all hosts, provide delete_all=true and confirm_delete_all=true in the request.")
+    if (delete_all and not confirm_delete_all) or (confirm_delete_all and not delete_all):
+        logger.error("Deleting all hosts requires delete_all=true and confirm_delete_all=true.")
+        flask.abort(400, "Deleting all hosts requires delete_all=true and confirm_delete_all=true.")
```
ryandillinfelton, line 238: I don't think it's really necessary to have two separate errors for sending none and sending one.
```suggestion
    if not all([delete_all, confirm_delete_all]):
        logger.error("To delete all hosts, provide delete_all=true and confirm_delete_all=true in the request.")
        flask.abort(400, "To delete all hosts, provide delete_all=true and confirm_delete_all=true in the request.")
```

kruai (changes requested):
Looks mostly good. Just a couple points

api/host.py
```diff
@@ -226,6 +226,38 @@ def _delete_filtered_hosts(host_id_list):
...
+        flask.abort(503)
+
+    if not len(ids_list):
+        flask.abort(status.HTTP_404_NOT_FOUND, "No hosts found for deletion.")
+
+    delete_count = _delete_filtered_hosts(ids_list)
+
+    json_data = {"hosts_found": len(ids_list), "hosts_deleted": delete_count}
+
+    return flask_json_response(json_data, status.HTTP_202_ACCEPTED)
```
ryandillinfelton, line 258: Did we decide to keep going with the 202 transition?

ryandillinfelton (changes requested):

swagger/api.spec.yaml
thearifismail: IMO, `/hosts/all` hides the only intent of this endpoint.  That's why `delete_all_hosts` is better.

api/host.py
thearifismail: I fretted about it and forgo to check with you guys

tests/test_api_hosts_delete.py
```diff
@@ -153,6 +153,51 @@ def test_delete_hosts_using_filter(
...
+    for ind, id in enumerate(host_ids):
+        resp["hosts"]["data"][ind]["id"] = id
+    response = {"data": resp}
+
+    # Make the new hosts available in xjoin-search to make them available
+    # for querying for deletion using filters
+    patch_xjoin_post(response, status=200)
+
+    new_hosts = db_create_multiple_hosts()
+    new_ids = [str(host.id) for host in new_hosts]
```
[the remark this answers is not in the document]
thearifismail: Thanks to copy/paste; removed.  The source had it to check for unintended deletion.

tests/test_api_hosts_delete.py
thearifismail: the `event_producer_mock` does not have `call_count` for the write event but the `delete count` is available in response data; so using `assert response_data.get('hosts_deleted') == len(created_hosts)`

tests/test_api_hosts_delete.py
thearifismail: Gone!

tests/test_xjoin.py
thearifismail: Now the code uses an `empty filter` using `{"filter": ()}`

api/host.py
ryandillinfelton: I'm still a bit mixed on it, but It does seem like an outlier in terms of scale so that make sit feel less incongruous

api/host.py
ryandillinfelton: Yes. I'm suggesting we just return one error message if both or either is missing, and `if not all([delete_all, confirm_delete_all])` will cover all of those cases.

Also, do we really need the `delete_all_hosts=true` parameter if that's the sole purpose of the endpoint? I'm still all for requiring confirmation, but I think `confirm=true` is enough. Originally we only had `delete_all_hosts` to tell the api what we were trying to do since it was part of the filtered host delete endpoint but it feels superfluous here.

swagger/api.spec.yaml
ryandillinfelton: I think since it's sent with a DELETE verb it's clear.
In relation to other endpoints that support the DELETE verb it's
`DELETE hosts?{filters}`
`DELETE hosts/{host_id_list}`
`DELETE hosts/all`
I think, especially with context, it's clear that the endpoint deletes all hosts

thearifismail: It is not a vanity issue but a willful intent to make it hard to delete all hosts.  That's why I would like to leave it as is.

api/host.py
thearifismail: Yes, earlier missed `all`.  Changed it to use `if not all([delete_all, confirm_delete_all]):`

api/host.py
thearifismail: let's leave it to `202` because deleting filtered hosts also returns `202`.

thearifismail: If you insist, then I can change it using a separate PR with Asa's input.  It may require changing some tests also.

kruai: Pretty sure we decided to use 202 for update/delete endpoints going forward. We're not changing the existing ones because of the risk of breaking existing implementations

swagger/api.spec.yaml
kruai: I'm for `DELETE /hosts/all` as well, fwiw... I don't think `DELETE /hosts/all` hides any intent at all, and in fact is pretty explicit. It is pretty redundant to do `DELETE /hosts/delete_all_hosts?delete_all=true&confirm_delete_all=true`.

ryandillinfelton (changes requested):

tests/test_api_hosts_delete.py
kruai: Ah, you're right... you'd need to do `mocker.patch.object(event_producer_mock, "write_event")` before calling the endpoint, but then you wouldn't be able to check the value of `event`. In that case, I'm resolving this, and we can deal with `write_event` counts in the Kafka PR

api/host.py
```diff
@@ -226,6 +226,35 @@ def _delete_filtered_hosts(host_id_list):
...
+@metrics.api_request_time.time()
+def delete_all_hosts(delete_all=None, confirm_delete_all=None):
+    if not all([delete_all, confirm_delete_all]):
+        logger.error("To delete all hosts, provide delete_all=true and confirm_delete_all=true in the request.")
+        flask.abort(400, "To delete all hosts, provide delete_all=true and confirm_delete_all=true in the request.")
+
+    try:
+        # Send None for all filter params, except {} for the filter expected for checking system_profile presence
+        # Sending no filter, gets all hosts on the account
+        ids_list = get_host_ids_list_xjoin(None, None, None, None, None, None, None, None, None, {})
```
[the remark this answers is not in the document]
thearifismail: Since the issue has to be addressed on the `xjoin-search` side, I have removed the use of limit which I inserted for debugging.  So pagination is not an issue.  Reaper gets hosts using DB query and bypasses xjoin altogether. 

Questions:

1.  Should deleting all hosts bypass xjoin and get hosts list from the DB directly?
2.  Should inventory show the total number of hosts and the number of hosts deleted?
3.  Should the inventory code iterate over until all hosts are deleted?
4.  Should this PR be delivered with the current behavior?  The design and behavior short comings be discussed and addressed in a separate PR?

thearifismail (commented):
@kruai @ryandillinfelton the endpoint has been changed to `hosts/all`.  Please look at the response about limit use.

api/host_query_xjoin.py
```diff
@@ -159,7 +161,7 @@ def get_host_ids_list(
     if current_identity.identity_type == IdentityType.SYSTEM and current_identity.auth_type != AuthType.CLASSIC:
         all_filters += owner_id_filter()
 
-    variables = {"filter": all_filters}
+    variables = {"filter": all_filters, "limit": 100}  # max limit 100 allowed by xjoin-search
```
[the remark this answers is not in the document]
thearifismail: removed limit altogether.  I had added for investigation purpose.

api/host.py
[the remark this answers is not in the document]
kruai: My thoughts:

1. Good point... now that it's a separate endpoint, that would make more sense. 
2. It would make sense to return the total number of hosts deleted in the API response.
3. Not sure I understand, but since we have to write an event for each host we delete, we must iterate by necessity.
4. I think we can save the above for a follow-up PR so as not to block the Prod deploy. We'll likely need to prioritize pagination, though, as @ryandillinfelton brings up a good point... that would need to be implemented for both bulk-delete endpoints.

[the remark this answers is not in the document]
thearifismail: > 3\. Not sure I understand, but since we have to write an event for each host we delete, we must iterate by necessity.

The iteration IS done but is limited to the number of hosts returned.  Though `xjoin-search` query does return the `total` number of hosts, it is not used to keep track.  Inventory code needs an update to track of the total number of hosts available and how many have been deleted.  Then API caller function, like "delete_all_hosts()"  can keep track of total vs deleted and keep iterating until all available have been deleted.  That said, bypassing "xjoin-search" and query DB sounds simple.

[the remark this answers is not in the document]
kruai: I don't think returning the number of deleted hosts is required to merge this, but it wouldn't hurt either.

api/host.py
```diff
@@ -226,6 +227,34 @@ def _delete_filtered_hosts(host_id_list):
...
+        log_get_host_list_failed(logger)
+        flask.abort(400, str(err))
+    except ConnectionError:
+        logger.error("xjoin-search not accessible")
+        flask.abort(503)
+
+    if len(ids_list) == 0:
+        flask.abort(status.HTTP_404_NOT_FOUND, "No hosts found for deletion.")
+
+    delete_count = _delete_filtered_hosts(ids_list)
```
ryandillinfelton, line 251: Why make two calls to the database rather than just using a single delete expression and letting the database handle it?
We should be able to essentially send `DELETE FROM hosts WHERE account_number={account_number}` since we're just deleting every host on an account.

swagger/openapi.json
```diff
@@ -1087,6 +1118,24 @@
...
+      },
+      "deleteAll": {
+        "in": "query",
+        "name": "delete_all",
+        "schema": {
+          "type": "boolean"
+        },
+        "description": "The option to delete all hosts on an account. Combine with `confirm_delete_all=true` to confirm deletion of all hosts.",
+        "required": false
+      },
```
ryandillinfelton, line 1130: I mentioned this in another thread on this PR but I don't think I got a conclusive answer. I don't think we really need the delete_all_hosts=true parameter. It made sense when it was an option on the more general bulk delete endpoint, but now that deleting all is the sole purpose of the endpoint it feels redundant. I'm still all for requiring confirmation, but I think `confirm_delete_all=true` is enough.

ryandillinfelton (changes requested):

api/host.py
thearifismail: I myself realized that two DB calls but the first one is very quick and the second one `_delete_filtered_hosts(ids_list)` deletes one host at a time and generates events which lead to synching of host-inventory with `xjoin-search` index and synching with other databases.  Furthermore, `delete_all_hosts` is expected to get used seldom.  There making two calls should not be an issue.

ryandillinfelton: I see, I completely overlooked needing to publish events about those deletes. I think it's acceptable, 2 DB calls really isn't that expensive.

swagger/openapi.json
ryandillinfelton: @thearifismail @kruai thoughts on this? I'm ultimately okay with either but I'd prefer to simplify to just one.

kruai: I agree with this. When we originally planned on using 1 endpoint, we wanted 2 booleans:

- 1 to specify that we wanted to delete *all* hosts instead of providing a filter
- 1 to confirm that you want to do that

Now that we're splitting it into 2 endpoints, I think choosing to use the `DELETE /hosts/all` endpoint replaces the first boolean; using one confirmation boolean on a dedicated endpoint should suffice IMO.

api/host.py
kruai: Yeah, I agree that it's fine here.

swagger/openapi.json
thearifismail: `delete_all` gone!  Just pushed the update

api/host.py
```diff
@@ -226,6 +227,34 @@ def _delete_filtered_hosts(host_id_list):
     return deletion_count
 
 
+@api_operation
+@rbac(Permission.WRITE)
+@metrics.api_request_time.time()
+def delete_all_hosts(delete_all=None, confirm_delete_all=None):
```
ryandillinfelton, line 233:
```suggestion
def delete_all_hosts(confirm_delete_all=None):
```

ryandillinfelton (changes requested):
Just found one little isue

api/host.py
thearifismail: Thanks.  Fixed.

ryandillinfelton (approved):
LGTM

api/host_query_db.py
```diff
@@ -72,6 +72,15 @@ def get_host_list(
     return query_results.items, query_results.total, additional_fields
 
 
+def get_all_hosts():
+    query = _find_all_hosts()
+    query_results = query.all()
+    ids_list = [str(host.id) for host in query_results]
```
kruai, line 78: My only concern with this approach is that `query_results` will become much larger than it needs to be; optimally, we'd want to query just the IDs instead of the entire hosts. But, that needn't block this PR if people do not agree.

kruai (approved):

thearifismail:
Created https://issues.redhat.com/browse/ESSNTL-2197 for fetching just host IDs

PR #1084, opened 2022-02-01
Remove bulk query source / fix OS ordering

kruai:
# Overview

This PR was initially created to address [ESSNTL-1879](https://issues.redhat.com/browse/ESSNTL-1879), but by necessity has grown to enforce a xjoin-only setup ([ESSNTL-2277](https://issues.redhat.com/browse/ESSNTL-2277)). This means removal of the bulk-query-source configuration, as we will require xjoin going forward. Note that in the interest of not making this PR even larger, not all endpoints are changed here - the Tag endpoints will be handled by [ESSNTL-2282](https://issues.redhat.com/browse/ESSNTL-2282).

By request, this has been updated to also address [ESSNTL-1755](https://issues.redhat.com/browse/ESSNTL-1755).

## Changes

Major changes to codebase:
- Removed `BulkQuerySource`, including references and deployment param.
- Changed non-tag Host endpoints to use Xjoin instead of DB.
- Changed System Profile endpoint to **always** use xjoin, even if `fields` is not provided.
  - This necessitated a refactor in _api/sparse_host_list_system_profile.py_.
- Removed a number of now-unused DB functions from _api/host_query_db.py_.

Major changes to testing:
- Added `patch_xjoin_post` to the `flask_client` fixture, so all calls to `app.xjoin.post` return an empty hosts response by default. Note that this is overwritten any time a test mocks this method.
- Removed all `query_source_*` fixtures.
- Removed a lot of tests from _tests/test_api_hosts_get.py_. These specifically tested that the correct hosts were returned in DB-mode; the corresponding tests for xjoin-mode already exist in _tests/test_xjoin.py_.
- Removed a lot of tests from _tests/test_culling.py_. These specifically tested that the correct hosts were returned in DB-mode; the corresponding tests for xjoin-mode already exist in _tests/test_xjoin.py_.
- Added `test_query_all_hosts` to validate the graphql request when calling the host-list endpoint without providing any filters, which I don't think was originally covered.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


thearifismail (commented):
Not done with review yet but would like some input from team memebers.  

1. Now that `xjoin-search` is the only source for fetching hosts, `get_host_list` and similar functions should be consolidated into one module instead of two `api.host.py` and `api.host.xjoin.query.py`.  Putting the code in one module will make it cohesive, concise, and easy to understand.  Most of the clutter cause by the large number of parameters will be removed.  I am okay with postponing such refactoring to a separate PR to keep this one from bloating too much.
2. Another issue is the presence `host_query.py`, `host_query_db.py`, and `host_query_xjoin.py`.  These modules should be renamed, particularly `host_query.py` which does nothing related to queries.
3. GraphQL queries templates are in at least two different modules and more queries are to come.  How about moving all GraphQL query templates to one module?

kruai:
> Not done with review yet but would like some input from team memebers.
> 
> 1. Now that `xjoin-search` is the only source for fetching hosts, `get_host_list` and similar functions should be consolidated into one module instead of two `api.host.py` and `api.host.xjoin.query.py`.  Putting the code in one module will make it cohesive, concise, and easy to understand.  Most of the clutter cause by the large number of parameters will be removed.  I am okay with postponing such refactoring to a separate PR to keep this one from bloating too much.
> 2. Another issue is the presence `host_query.py`, `host_query_db.py`, and `host_query_xjoin.py`.  These modules should be renamed, particularly `host_query.py` which does nothing related to queries.

I agree with these two points - we should refactor and consolidate. Personally, since this PR is big enough already, and since I'm going to make another PR to convert the rest of the endpoints after this gets merged, I think it would be better to do this cleanup task after those are complete. I'm amenable to doing it in this PR, though, if y'all think it would be best. Thoughts @ryandillinfelton?

> 3. GraphQL queries templates are in at least two different modules and more queries are to come.  How about moving all GraphQL query templates to one module?

I'm torn on this one, since there are pros and cons for different methods of organization. As I typed out my reasons, I think I started favoring the way we do it now tbh.

Currently, constants are defined in the files they're most relevant to. For instance, `host_query_xjoin.py` contains the standard queries for hosts and for host IDs, `tag.py` contains the query for tags, etc. This is nice because these queries are generally used in the same files that define them, so you just have to scroll up to see what the query is. It's also easy to see where each query is being used. The negative in this case is that they take up a fair amount of space, which could be seen as clutter in the file.

One could make the case that having all the queries in one file would make those other files cleaner. One negative for this is that you have one huge file that contains every single xjoin query we need to work with, which could be overwhelming, and could be hard to find what you're looking for due to its length. It would also be harder to tell which queries are being used and which are not, which may present an issue as things change.

Regardless, I'm pretty sure reorganizing our query definitions is out of scope for this PR.

swagger/api.spec.yaml
```diff
@@ -138,6 +138,7 @@ paths:
         - $ref: '#/components/parameters/pageParam'
         - $ref: '#/components/parameters/orderByParam'
         - $ref: '#/components/parameters/orderHowParam'
+        - $ref: '#/components/parameters/fields_param'
```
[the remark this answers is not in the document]
kruai: It's to address [ESSNTL-1755](https://issues.redhat.com/browse/ESSNTL-1755), which @fstavela requested I add to facilitate testing this PR.

tests/helpers/graphql_utils.py
```diff
@@ -99,6 +103,23 @@
     }
 }
 
+XJOIN_INVALID_SYSTEM_PROFILE = {
```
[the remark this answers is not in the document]
kruai: Is that necessary when the value is literally the word `invalid`? :smile:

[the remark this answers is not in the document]
thearifismail: Just for clarity.  At first glance all fields looked good; then I guess `random` key appeared as the suspect.  Since the question popped up, it doesn't hurt to add a comment.

tests/test_api_hosts_delete.py
```diff
@@ -397,6 +397,6 @@ def __init__(self, mocker):
         self.mocker = mocker
 
     def mock_get_host_list_by_id_list(self, host_id_list):
-        self.query = _get_host_list_by_id_list(host_id_list)
+        self.query = _get_host_list_by_id_list_db(host_id_list)
```
thearifismail, line 400: The function name is confusing; how about `_get_host_list_by_id_from_db`?

kruai: It's being retrieved `by_id_list`, so I don't want to lose that clarity. But, I can change it to `_get_host_list_by_id_list_from_db` if that helps

tests/helpers/graphql_utils.py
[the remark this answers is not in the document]
kruai: OK, added a comment noting the invalid field

tests/test_culling.py
```diff
@@ -18,29 +17,6 @@
 from tests.helpers.db_utils import minimal_db_host
 from tests.helpers.mq_utils import assert_delete_event_is_valid
 from tests.helpers.test_utils import get_staleness_timestamps
-from tests.helpers.test_utils import minimal_host
-from tests.helpers.test_utils import now
-
-
-def test_with_stale_timestamp(mq_create_or_update_host, api_get):
```
thearifismail, line 25: Why these tests has been removed?  If the reason is that it `def api_get()` used to get them form DB and now xjoin provides, then shouldn't the remaining `api_get` be removed? OR shouldn't the `def api_get` fixture to moved to `graph-ql` fixture.

kruai: This is something we discussed in the meeting for this PR, but basically yes; these are only testing the DB implementation. We already have tests that validate this same functionality on the xjoin side here: https://github.com/RedHatInsights/insights-host-inventory/blob/master/tests/test_xjoin.py#L536-L643

You're right that the other DB-specific tests for the updated endpoints need to be removed! I'll get on that now.

kruai: OK, I've removed the remaining tests in this file that validated culling against `/hosts` and `/system_profile` endpoints.

thearifismail (dismissed):
LGTM

RonnyPfannschmidt:
/retest

triggering to recreate the expired images

api/filtering/filtering.py
```diff
@@ -282,6 +282,10 @@ def build_tag_query_dict_tuple(tags):
     return query_tag_tuple
 
 
+def host_id_list_query_filter(host_id_list):
+    return ({"OR": [{"id": {"eq": host_id}} for host_id in host_id_list]},)
```
RonnyPfannschmidt, line 286: @kruai @ryandillinfelton would there be any way to map this to something like https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-ids-query.html

kruai:
@RonnyPfannschmidt I resolved the merge conflicts and retried the build, but the IQE tests failed again. Assuming that's why you were doing the `/retest` earlier

RonnyPfannschmidt:
@kruai retest was for getting the image rebuild, they timeout after 3 days

Im still not done completing the fixes

api/filtering/filtering.py
RonnyPfannschmidt: i meant something like `id IN [host_id_1, host_id_2, host_id_3]`

in todays standup ryan pointed out that we currently have something like this, @fijshion may be able to elaborate if we can use something like it in the next iteration

maybe its a low prio backlog item

kruai: @RonnyPfannschmidt I chatted with @ryandillinfelton and it seems that functionality does not exist yet. I'm going to resolve this, but if this is planned to be implemented on the xjoin side and you think it's worth tracking this in the meantime, please feel free to make a card for it.

ryandillinfelton (approved):
LGTM 🚀

RonnyPfannschmidt:
/retest for recreating the images

@kruai given the current details collected from the other qe's and the hardships in testing out the details, its unlikelty to fully complete - so i strongly recommend to defer at least to end of april, if not mid may

pr_check.sh
```diff
@@ -11,6 +11,7 @@ export APP_ROOT=$(pwd)
 export WORKSPACE=${WORKSPACE:-$APP_ROOT}  # if running in jenkins, use the build's workspace
 export IMAGE_TAG=$(git rev-parse --short=7 HEAD)
 export GIT_COMMIT=$(git rev-parse HEAD)
+export QUAY_EXPIRE_TIME="40d"
```
RonnyPfannschmidt, line 14: reminder comment so that we take this out before merge

api/filtering/filtering.py
```diff
@@ -312,6 +312,10 @@ def build_tag_query_dict_tuple(tags):
     return query_tag_tuple
 
 
+def host_id_list_query_filter(host_id_list):
```
RonnyPfannschmidt, line 315: @kruai we need to ensure hiding culled hosts here

InsightsDroid:
Test results are available in [Ibutsu](https://url.corp.redhat.com/ibutsu-runs). The test run IDs are:
1b6e6874-fb49-4053-9a4a-0a262f336f84
ff03af87-66a7-49c0-ac1b-0fbbdbb51ba6

InsightsDroid:
Test results are available in [Ibutsu](https://url.corp.redhat.com/ibutsu-runs). The test run IDs are:
80310bd1-d8cc-4ab9-8945-c1e811bfa6ab
e8a9cd67-2282-4b07-a8e4-9f704d20828c

RonnyPfannschmidt:
/retest

after release of the test changes

roliveri (approved):

thearifismail (approved):
approving it based on successful local runs.  Due to problems with the available linux system, could not verify by creating an image from this code and running the newly built image

PR #1089, opened 2022-02-04
essntl-1897: set stale_timestamp of edge systems.

thearifismail:
# Overview

This PR is being created to address [ESSNTL-1897](https://issues.redhat.com/browse/ESSNTL-1897).
It sets `stale_timestamp` of edge systems to 999999 days in future, which will be sometime in Year 4760.

An MR https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/32922 has been submitted for getting the number edge hosts in `host-inventory-prod` to see if a manual job should be run to extend edge hosts' stale_timestamp.
## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/queue/queue.py
```diff
@@ -228,7 +231,12 @@ def add_host(host_data, platform_metadata):
             if identity.identity_type == IdentityType.SYSTEM:
                 host_data = _set_owner(host_data, identity)
 
+            if host_data.get("system_profile").get("host_type") == "edge":
+                # edge system should not be deleted.  To avoid deletion, set the stale_timestamp to 4760-01-01
+                host_data["stale_timestamp"] = (datetime.now(timezone.utc) + timedelta(days=999999)).isoformat()
```
kruai, line 236: This works, though this might not be the best place to put this Edge override. If we only do it here, the logic will not be applied in the other locations `Timestamps.from_config()` is used.
I wonder if we should instead add a "host_type" parameter to `from_config()`, so that we do the staleness timestamp calculation in the culling logic. What do you think, @ryandillinfelton ?

thearifismail: > This works, though this might not be the best place to put this Edge override. If we only do it here, the logic will not be applied in the other locations `Timestamps.from_config()` is used. I wonder if we should instead add a "host_type" parameter to `from_config()`, so that we do the staleness timestamp calculation in the culling logic. What do you think, @ryandillinfelton ?

Though the code does not reflect it, during a standup call I provided the rationale for this approach.  Using this approach rest of the code remains untouched because once the `stale_timestamp` has been set, nothing else have to be done.  I also ran a query against "host-inventory-prod" and there were less than 20 hosts of type "edge".  Since hosts are added primarily using "MQ", there aren't many places to get affected by it.  Furthermore, "edge" is the `ONLY` known host_type.  If more host-types are added in future, then it should be added to `app/config`.  Not just that, `host_type` should be a column in "hosts" table instead of being buried in "system_prorfile".

As far as the current "edge" hosts are confirmed, they can be updated using a one time job, i.e. ClowdInvocationJob.

kruai: It's true that Hosts are only created in the MQ right now; I'm just saying that `queue.py` might not be the most logical location to put the code that overrides the stale timestamp. 
I was wrong about using `from_config()` earlier, since that's just calculating the offsets based off of the stale timestamp. Since this has to do with the host's data itself, it might make more sense in `host.py`; we even have a function that modifies the timestamp [right here](https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/models.py#L293). Thoughts?

thearifismail: IMO, this is the best place for this one time adjustment for the following reasons:

1. Updates to incoming host payloads are made in this function.  The `if` block immediately before this one updates owner_id and the line following it validates the payload as a part of deserialization call.
2. MQ triggers the code in `app` module whereas `host.py` is in `api` module.   It does not feel right to trigger API  code for just this one independent call
3. Moving it to `app.models.py` will unnecessarily complicate the code by pushing this change deeper into the code.

fstavela (commented):
I tested this and it seems that now I can't get any edge hosts with /hosts endpoint. It doesn't matter what `?staleness` parameter I use in the request, I'm always getting 0 results.
I'm not sure why though. Maybe xjoin doesn't support such high values for datetimes?

kruai:
> I tested this and it seems that now I can't get any edge hosts with /hosts endpoint. It doesn't matter what `?staleness` parameter I use in the request, I'm always getting 0 results. I'm not sure why though. Maybe xjoin doesn't support such high values for datetimes?

Just to be clear, are you filtering on `[system_profile][host_type]=edge`?

app/queue/queue.py
ryandillinfelton: I think Asa may have misspoken? He linked to the timestamp updating funciton in `models.py` not anything in `host.py`.
I do think putting it in `models.py` would be more appropriate if only to consolidate the timestamp effecting logic. Looking into it more I think the reason the owner is modified here is because of it's reliance on the identity header, although it's debatable if that's enough of a justification. Point is, it's the exception not the rule.

I think we should be modifying the staleness timestamp in `models.py` using function Asa linked to, since that's where updates to that value are currently handled. It looks like some changes will also need to be made to the `init` function to set the staleness timestamp correctly on first creation as well if I understand correctly.

ryandillinfelton (changes requested):
I agree with Asa that the logic should be moved from queue. I would also like to see another test that checks that updates work as expected.

fstavela:
> > I tested this and it seems that now I can't get any edge hosts with /hosts endpoint. It doesn't matter what `?staleness` parameter I use in the request, I'm always getting 0 results. I'm not sure why though. Maybe xjoin doesn't support such high values for datetimes?
> 
> Just to be clear, are you filtering on `[system_profile][host_type]=edge`?

Nope, I was doing `GET /hosts` request without `filter` parameter and I got 0 results, but that's because I didn't have any other (non-edge) hosts in the database.
I tried again with more hosts in the database, and getting non edge hosts behaves as expected, but I'm getting 0 edge hosts. I tried using all available values for `staleness=` parameter and I also tried getting the hosts without providing `staleness` parameter, and it all results in correct number of non-edge hosts, but 0 edge hosts.

I don't know why this is happening, it doesn't make a lot of sense why the code changes should cause this and there also is a test for adding the host and having it in the database, so the only thing that comes to my mind is that maybe xjoin can't index hosts with so high datetime values for some reason.

thearifismail:
> 

@fstavela I will try to reproduce the problem today and will let you know my findings

thearifismail:
@fstavela I have been able to reproduce the problem.  I created 20 hosts, 10 "edge" and 10 "non-edge".  Strange fact is that `.../api/inventory/v1/hosts` returns only 10.  However, using the ID of a edge host can retrieve the host

```http://localhost:8000/api/inventory/v1/hosts//77348304-1088-4c2d-ae2d-03dee292b2e5
{
    "total": 1,
    "count": 1,
    "page": 1,
    "per_page": 50,
    "results": [
        {
            "bios_uuid": "5343ece4-2bee-4810-bb7a-40dc21342d1d",
            "id": "77348304-1088-4c2d-ae2d-03dee292b2e5",
            "account": "test",
            "display_name": "edge.foo.redhat.com",
            "facts": [],
            "reporter": "puptoo",
            "per_reporter_staleness": {
                "puptoo": {
                    "last_check_in": "2022-02-24T00:44:27.116234+00:00",
                    "stale_timestamp": "2022-02-25T00:44:26.359488+00:00",
                    "check_in_succeeded": true
                }
            },
            "stale_timestamp": "4760-01-21T00:16:16.418365+00:00",
            "stale_warning_timestamp": "4760-01-28T00:16:16.418365+00:00",
            "culled_timestamp": "4760-02-04T00:16:16.418365+00:00",
            "created": "2022-02-24T00:44:27.121445+00:00",
            "updated": "2022-02-24T00:44:27.121452+00:00"
        }
    ]
}


http://localhost:8000/api/inventory/v1/hosts//77348304-1088-4c2d-ae2d-03dee292b2e5/system_profile
{
    "total": 1,
    "count": 1,
    "page": 1,
    "per_page": 50,
    "results": [
        {
            "id": "77348304-1088-4c2d-ae2d-03dee292b2e5",
            "system_profile": {
                "arch": "x86-64",
                "owner_id": "1b36b20f-7fa0-4454-a6d2-008294e06378",
                "cpu_flags": [
                    "flag1",
                    "flag2"
                ],
                "cpu_model": "Intel(R) Xeon(R) CPU E5-2690 0 @ 2.90GHz",
                "host_type": "edge",
                "yum_repos": [
                    {
```

thearifismail:
It appears that edge hosts are not getting indexed.   Here is the first one, non-edge:
```
Query:
{
  hosts (
    filter: {
      id: {eq: "16a5b2aa-c19c-4c76-b4be-bc0c78f6b2d8"}
    }
    limit: 10,
    offset: 0
  ) {
    meta { count, total }
    data {
      id
      account
      display_name
    }
  }
}

output:
{
  "data": {
    "hosts": {
      "meta": {
        "count": 1,
        "total": 1
      },
      "data": [
        {
          "id": "16a5b2aa-c19c-4c76-b4be-bc0c78f6b2d8",
          "account": "test",
          "display_name": "test.foo.redhat.com"
        }
      ]
    }
  }
}
```

Here is query for an "edge" host:
```
Query:
{
  hosts (
    filter: {
      id: {eq: "77348304-1088-4c2d-ae2d-03dee292b2e5"}
    }
    limit: 10,
    offset: 0
  ) {
    meta { count, total }
    data {
      id
      account
      display_name
    }
  }
}


output:
{
  "data": {
    "hosts": {
      "meta": {
        "count": 0,
        "total": 0
      },
      "data": []
    }
  }
}
```

thearifismail:
@ryandillinfelton It appears that edge hosts are not getting indexed to elasticsearch. Above are two hosts, one edge and one non-edge.  They are from "ephemeral-l97532".  Can you please take a look?

thearifismail:
> @ryandillinfelton It appears that edge hosts are not getting indexed to elasticsearch. Above are two hosts, one edge and one non-edge. They are from "ephemeral-l97532". Can you please take a look? What are other choices aside from `date_nanos`?

The issue IS related to "date_nanos".  The farthest a date can be set to is " Year 2262".   Here is the error from "env-ephemeral-l97532-4b0892bc-connect-79f974d5d9-xplqk" pod.
```
2022-02-24 15:43:46,123 WARN Encountered an illegal document error when executing batch 117 of 1 records. Ignoring and will not index record. Error was [{"type":"mapper_parsing_exception","reason":"failed to parse field [stale_timestamp] of type [date_nanos] in document with id 'c0cbdcb3-8197-4676-a7a1-8aae277a08c0'. Preview of field's value: '4760-01-21T00:16:16.418365Z'","caused_by":{"type":"illegal_argument_exception","reason":"date[4760-01-21T00:16:16.418365Z] is after 2262-04-11T23:47:16.854775807 and cannot be stored in nanosecond resolution"}}] (io.confluent.connect.elasticsearch.bulk.BulkProcessor) [BulkProcessor@192107613-1]
2022-02-24 15:44:04,873 INFO WorkerSourceTask{id=xjoin.inventory.db.1645705331435752925-0} flushing 0 outstanding messages for offset commit (org.apache.kafka.connect.runtime.WorkerSourceTask) [SourceTaskOffsetCommitter-1]
2022-02-24 15:45:04,874 INFO WorkerSourceTask{id=xjoin.inventory.db.1645705331435752925-0} flushing 0 outstanding messages for offset commit (org.apache.kafka.connect.runtime.WorkerSourceTask) [SourceTaskOffsetCommitter-1]
```

thearifismail:
@fstavela the problem should be fixed now.

app/models.py
```diff
@@ -48,6 +49,9 @@
 SPECIFICATION_DIR = "./swagger/"
 SYSTEM_PROFILE_SPECIFICATION_FILE = "system_profile.spec.yaml"
 
+# set edge host stale_timestamp way out in future to Year 2222
+EDGE_HOST_STALE_TIMESTAMP = datetime.now(timezone.utc) + timedelta(days=73000)
```
ryandillinfelton, line 53: If there is a limit to the timestamp of Year 2262 shouldn't we make this not relative to today's date? Seems like if we start it close to the limit now then it's a time bomb. Could we just hardcore the date to just under the maximum?

app/models.py
```diff
@@ -244,8 +252,12 @@ def update(self, input_host, update_system_profile=False):
...
             self.update_system_profile(input_host.system_profile_facts)
 
-        self._update_stale_timestamp(input_host.stale_timestamp, input_host.reporter)
-        self._update_per_reporter_staleness(input_host.stale_timestamp, input_host.reporter)
+        if self.system_profile_facts and self.system_profile_facts.get("host_type") == "edge":
+            self._update_stale_timestamp(EDGE_HOST_STALE_TIMESTAMP, input_host.reporter)
+            self._update_per_reporter_staleness(EDGE_HOST_STALE_TIMESTAMP, input_host.reporter)
+        else:
+            self._update_stale_timestamp(input_host.stale_timestamp, input_host.reporter)
+            self._update_per_reporter_staleness(input_host.stale_timestamp, input_host.reporter)
```
ryandillinfelton, line 260: Since it's the same two calls either way could we just conditionally assign the timestamp to avoid repeating the calls? It will make it easier if we need to add other timestamp dependent calls later, or if the interface for either function changes.

Also switched to a optional chaining pattern to simplify/shorten the condition. LMK what you think of it. I like that it keeps you from having to check each part of the chain individually, but the syntax isn't as nice as say JS's `?.`. I don't personally think it hurts readability, especially if we start using it in more places, but that doesn't mean it doesn't.

It will just return None if any part of the chain fails, since an empty dict will never have whatever member you're asking for.

```suggestion
        stale_timestamp = input_host.stale_timestamp
        
        if self.get(system_profile, {}).get("host_type") == "edge":
            stale_timestamp = EDGE_HOST_STALE_TIMESTAMP
   
        self._update_stale_timestamp(stale_timestamp, input_host.reporter)
        self._update_per_reporter_staleness(stale_timestamp, input_host.reporter)
```

app/models.py
```diff
@@ -220,7 +224,11 @@ def __init__(
...
         super().__init__(canonical_facts, display_name, ansible_host, account, facts, tags, system_profile_facts)
-        self.stale_timestamp = stale_timestamp
+
+        if system_profile_facts and system_profile_facts.get("host_type") == "edge":
+            self.stale_timestamp = EDGE_HOST_STALE_TIMESTAMP
+        else:
+            self.stale_timestamp = stale_timestamp
         self.reporter = reporter
         self.per_reporter_staleness = per_reporter_staleness or {}
         if not per_reporter_staleness:
```
ryandillinfelton, line 234: I think you have a bug here. Currently I believe a new edge host will use the payload stale_timestamp in per_reporter_staleness rather than the edge timestamp. I mention in the main review comment that it would be safer to modify the behavior of the staleness update functions themselves so they decide which timestamp to use. I think that approach would prevent this bug and others like it.

ryandillinfelton (changes requested):
Functionally it looks good other than the moving constant timestamp worry I mentioned.

I do think this could be improved though. Instead of having the edge check in update and init you could modify `_update_stale_timestamp` and `_update_per_reporter_staleness` to determine which timestamp to use themselves. I think it would be more maintainable since everything would in one spot, and safer since we don't have to check that we're passing the right timestamp in every spot we call those functions.

The existing test looks good, but I think it need another test for updating an existing edge host still, since they call different code.

app/models.py
```diff
@@ -220,7 +224,11 @@ def __init__(
             raise InventoryException(title="Invalid request", detail="The tags field cannot be null.")
 
         super().__init__(canonical_facts, display_name, ansible_host, account, facts, tags, system_profile_facts)
-        self.stale_timestamp = stale_timestamp
+
+        if system_profile_facts and system_profile_facts.get("host_type") == "edge":
```
ryandillinfelton, line 228: I ran into this concept recently and it seemed neat, so I thought i'd suggest it. You can do JS style optional chaining in python pretty easily and it simplifies checks like this nicely. This one is only two checks so it's far from worst case but if you want to try the alternative I've suggested the changes below. The behavior is identical either way.

I realized after writing that that in this case we don't know if the root object is None or not, so it doesn't come out looking quite as nice as when you know it exists. Like with `self` which looks like this.

```
self.get(system_profile_facts, {}).get("host_type") == "edge"
```

anyways, take it or leave it, just thought it was interesting, and I always hate writing checks like this.

```suggestion
        if (system_profile_facts or {}).get("host_type") == "edge":
```

app/models.py
thearifismail: Trying to understand the comment... it is not a starting point rather keep the host "fresh" until this date is reached, which is 200 years away.  For clarity and better readability, it can be set explicitly rather relative to `now` as `datetime.datetime(2260, 1, 1, tzinfo=datetime.timezone.utc)` to January 1, 2260.

thearifismail:
> Functionally it looks good other than the moving constant timestamp worry I mentioned.
> 
> I do think this could be improved though. Instead of having the edge check in update and init you could modify `_update_stale_timestamp` and `_update_per_reporter_staleness` to determine which timestamp to use themselves. I think it would be more maintainable since everything would in one spot, and safer since we don't have to check that we're passing the right timestamp in every spot we call those functions.
> 
> The existing test looks good, but I think it need another test for updating an existing edge host still, since they call different code.

IMO, this is the right place for two reasons:

1.  Since `host_type` is specified in `system_profile`, checking for `edge` makes more sense where system_profile is handled.  Checking for `edge` in `_update...` functions is a bit more abstract because `self` has to be used to get to it.
2.  Setting in `update...` function will missing setting an edge host stale_timestamp to far-out-future altogether, if the reporter provides `per_reporter_staleness` with staleness to now are 14 days.

So it make more sense to set it when a new host is created or an existing host update.

app/models.py
thearifismail: Fixed.  Now stale_timestamp is set in update functions.

app/models.py
thearifismail: Plan to leave it this time.

app/models.py
kruai: Yes, I think it only makes sense to do this as an absolute date. If we kept it relative, we'd see this start breaking in a year or two.

app/models.py
kruai: I'm fine with either implementation. I do prefer the more concise one and didn't realize you could do that in Python, so thanks for the info!

kruai (dismissed):
LGTM

ryandillinfelton (changes requested):
On line 223 in `app/models.py` you set the stale_timestamp to just the incoming stale_timestamp.  I think that means that new edge systems will start with the wrong timestamp. I think you need to switch that to use the stale_timestamp update function

thearifismail:
> On line 223 in `app/models.py` you set the stale_timestamp to just the incoming stale_timestamp. I think that means that new edge systems will start with the wrong timestamp. I think you need to switch that to use the stale_timestamp update function

The stale_timestamp is adjusted for `host_type: edge` in [_update_per_reporter_staleness](https://github.com/thearifismail/insights-host-inventory/blob/95a4e5fcb47ee486a7ab9e2151208ad22603f8d5/app/models.py#L230)

ryandillinfelton:
> > On line 223 in `app/models.py` you set the stale_timestamp to just the incoming stale_timestamp. I think that means that new edge systems will start with the wrong timestamp. I think you need to switch that to use the stale_timestamp update function
> 
> 
> 
> The stale_timestamp is adjusted for `host_type: edge` in [_update_per_reporter_staleness](https://github.com/thearifismail/insights-host-inventory/blob/95a4e5fcb47ee486a7ab9e2151208ad22603f8d5/app/models.py#L230)

I don't think you're invoking that helper function when you create a new host.

app/models.py
```diff
@@ -291,10 +294,19 @@ def update_facts(self, facts_dict):
...
+            self.stale_timestamp = stale_timestamp
         self.reporter = reporter
 
     def _update_per_reporter_staleness(self, stale_timestamp, reporter):
+
+        if self.system_profile_facts and self.system_profile_facts.get("host_type") == "edge":
+            self.stale_timestamp = EDGE_HOST_STALE_TIMESTAMP
+        else:
+            self.stale_timestamp = stale_timestamp
+
```
ryandillinfelton, line 309: Why are we also updating the `stale_timestamp` here? I appreciate that it means the bug I found isn't happening, but it's not good delineation of responsibility to have this also effect the `stale_timestamp`. It also means we have the same exact update in two spots which is not very DRY.

ryandillinfelton (changes requested):
I think this is a bit off the mark still but headed in the right direction.

Mainly I'd like to see the separation of concerns preserved with the update helper functions.
Right now `_update_per_reporter_staleness` also updates `stale_timestamp` but they should usually be called in sequence which would make it redundant.

For example on line 247 in `models.py`
```
        self._update_stale_timestamp(input_host.stale_timestamp, input_host.reporter)
        self._update_per_reporter_staleness(input_host.stale_timestamp, input_host.reporter)
```
 
 They aren't called sequentially when a new host is created. On line 223 instead of manually setting the stale_timestamp to the incoming we should use the `_update_stale_timestamp` helper function to make sure that the same behavior is used for edge hosts.
 
 So 
 ```
         super().__init__(canonical_facts, display_name, ansible_host, account, facts, tags, system_profile_facts)
        self.reporter = reporter
        self.per_reporter_staleness = per_reporter_staleness or {}
        if not per_reporter_staleness:
            self._update_per_reporter_staleness(stale_timestamp, reporter)
 ```
 Should instead be something like this. Not sure if we need to iniitalize the stale timestamp the same way we do the per reporter staleness since we know for sure that reporters send it.
 ```
         super().__init__(canonical_facts, display_name, ansible_host, account, facts, tags, system_profile_facts)
        self._update_stale_timestamp()
        self.reporter = reporter
        self.per_reporter_staleness = per_reporter_staleness or {}
        if not per_reporter_staleness:
            self._update_per_reporter_staleness(stale_timestamp, reporter)
 ```
 
 
 Also, in `_update_per_reporter_staleness` the actually timestamp used in the per_reporter_stalness object doesn't get adjusted for edge systems.

thearifismail:
> I think this is a bit off the mark still but headed in the right direction.
> 
> Mainly I'd like to see the separation of concerns preserved with the update helper functions. Right now `_update_per_reporter_staleness` also updates `stale_timestamp` but they should usually be called in sequence which would make it redundant.
> 
> For example on line 247 in `models.py`
> 
> ```
>         self._update_stale_timestamp(input_host.stale_timestamp, input_host.reporter)
>         self._update_per_reporter_staleness(input_host.stale_timestamp, input_host.reporter)
> ```
> 
> They aren't called sequentially when a new host is created. On line 223 instead of manually setting the stale_timestamp to the incoming we should use the `_update_stale_timestamp` helper function to make sure that the same behavior is used for edge hosts.
> 
> So
> 
> ```
>         super().__init__(canonical_facts, display_name, ansible_host, account, facts, tags, system_profile_facts)
>        self.stale_timestamp = stale_timestamp
>        self.reporter = reporter
>        self.per_reporter_staleness = per_reporter_staleness or {}
>        if not per_reporter_staleness:
>            self._update_per_reporter_staleness(stale_timestamp, reporter)
> ```
> 
> Should instead be something like this. Not sure if we need to iniitalize the stale timestamp the same way we do the per reporter staleness since we know for sure that reporters send it.
> 
> ```
>         super().__init__(canonical_facts, display_name, ansible_host, account, facts, tags, system_profile_facts)
>        self.stale_timestamp = stale_timestamp
>        self._update_stale_timestamp()
>        self.reporter = reporter
>        self.per_reporter_staleness = per_reporter_staleness or {}
>        if not per_reporter_staleness:
>            self._update_per_reporter_staleness(stale_timestamp, reporter)
> ```
> 
> Also, in `_update_per_reporter_staleness` the actually timestamp used in the per_reporter_stalness object doesn't get adjusted for edge systems.

Fixed all issues.  Now timestamp is set in `update_stale_timestamp`.  Now `per_reporter_staleness` sets `stale_timestamp` to `stale_timestamp=self.stale_timestamp.isoformat()`

ryandillinfelton (approved):

PR #1096, opened 2022-02-16
ESSNTL-2138: Fix problem with os version range operation

thearifismail:
# Overview

This PR is being created to address [ESSNTL-2138](https://issues.redhat.com/browse/ESSNTL-2138).  The fix is to use `minor= None` instead of defaulting to `zero` when minor version is not provided.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


api/filtering/custom_filters.py
```diff
@@ -5,11 +5,32 @@
...
+            os_filter = _build_no_minor_filter(operation, major, name)
+        else:
+            os_filter = _build_major_equality_filter(major, minor, name, operation)
+
+    if operation in ["lt", "gt", "lte", "gte"]:
+        if minor is None:
+            os_filter = _build_no_minor_filter(operation, major, name)
+        else:
+            # handle 'eq' here and 'gt' 'lt' in !='eq'
+            os_filter = _build_major_equality_filter(major, minor, name, operation)
```
ryandillinfelton, line 29: The branching on operation type is unnecessary, both branches do the same thing. We can just short circuit if the minor version is missing and leave the rest of the logic the same. I chose to inline the function calls since they are only used in one place.
```suggestion
    if minor is None:
        return {"spf_operating_system": {"major": {operation: major}, "name": {"eq": name}}}

    os_filter = {"spf_operating_system": {"major": {"eq": major}, "minor": {operation: minor}, "name": {"eq": name}}}
```

api/filtering/custom_filters.py
```diff
@@ -5,11 +5,32 @@
...
 
     if operation != "eq":
-        # The major operation should only ever be 'lt' or 'gt'
+        # The major operation should only ever be 'lt' or 'gt'.
+        # For 'gte' and 'lte', the 'eq' check handled in the previous block.
```
ryandillinfelton, line 33: If you commit the suggestion in my other comment this should be removed since that block would be gone.

ryandillinfelton (changes requested):
The functionality is technically correct, but there is a bug in the query generation causing it to create an over complicated query. I've suggested a way to fix the bug and reduce the complexity of the generation function.

api/filtering/custom_filters.py
thearifismail: I knew this would be questioned immediately but left it in for clarity because it explicitly shows how each operator is getting filtered; makes the code easy to read for someone not too familiar.  That said, the code is update per the recommendation here.  The code compact but hides the filter building complexity.  IMO, the explicit code is easier to follow, but the change is here if you don't buy the explicit reasoning.

kruai: Personally, I think the longer and more explicit version is harder to read and understand, because it creates a logical split where both branches do the same thing

api/filtering/custom_filters.py
```diff
@@ -5,11 +5,14 @@
 logger = get_logger(__name__)
 
 
+# build filter based on the operation and OS name and version
 def _build_operating_system_version_filter(major, minor, name, operation):
+    if minor is None:
+        return {"spf_operating_system": {"major": {operation: major}, "name": {"eq": name}}}
+
```
kruai, line 12: If this filter is used:
```
filter[system_profile][operating_system][RHEL][version][gte]=7
```
then this function would return:
```
{"spf_operating_system": {"major": {"gte": 7}, "name": {"eq": "RHEL"}}}
```
We used to only allow `eq`, `lt`, and `gt` on the major version. Have you checked whether these are allowed on the xjoin side?

api/filtering/custom_filters.py
```diff
@@ -5,11 +5,14 @@
...
     os_filter = {"spf_operating_system": {"major": {"eq": major}, "minor": {operation: minor}, "name": {"eq": name}}}
 
     if operation != "eq":
-        # The major operation should only ever be 'lt' or 'gt'
```
kruai, line 12: This restriction still applies, doesn't it?

api/filtering/custom_filters.py
thearifismail: Yes `xjoin-search` has `gte` and `lte` implemented.  Thanks to major and minor version handled separately, this filter has to be split into `eq` and `lt`/`gt`

api/filtering/custom_filters.py
thearifismail: Yes it does.

tests/test_xjoin.py
```diff
@@ -2018,13 +2018,7 @@ def test_query_hosts_filter_spf_operating_system(
...
                             "OR": [
-                                {
-                                    "spf_operating_system": {
-                                        "major": {"eq": 7},
-                                        "minor": {"gt": 0},
-                                        "name": {"eq": "RHEL"},
-                                    }
-                                },
+                                {"spf_operating_system": {"major": {"gt": 7}, "name": {"eq": "RHEL"}}},
                                 {"spf_operating_system": {"major": {"gt": 7}, "name": {"eq": "RHEL"}}},
```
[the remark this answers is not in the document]
thearifismail: Duplicate entry is not needed anymore due to the following code and has been removed:
```
    if minor is None:
        return {"spf_operating_system": {"major": {operation: major}, "name": {"eq": name}}}
```

api/filtering/custom_filters.py
thearifismail: Ok

api/filtering/custom_filters.py
kruai: OK - why is the comment being removed, then?

api/filtering/custom_filters.py
kruai: OK. In that case, I have no issue with this

api/filtering/custom_filters.py
thearifismail: Since this code block gets used for `lte` and `gte`, I found this comment confusing.  It took me a while to realize that the prior `eq` were being used `lte` and `gte` also.  That's why I took it out.

kruai: Ah, maybe it needs an update to clarify it, then. It's just saying that the major operation should only ever be `lt` or `gt` when using an inequality on a major+minor version. So, it converts `lte` and `gte` by removing the third character (if any).

ryandillinfelton: Yeah I agree with Asa here. I added a comment originally because it's really weird to just remove the last letter of the string for seemingly no reason.
maybe this comment would be more clear
```
# This portion of the query is filtering only on major version to select all hosts with a major version greater
# or lesser than the target. If GTE or LTE were used in this portion it would return render the previous portion 
# of the query useless. To prevent this and get the desired behavior we use a sub-string of the first two characters
# of the supplied operation so that the major operation is always "gt" or "lt".
```

api/filtering/custom_filters.py
```diff
@@ -5,11 +5,14 @@
 logger = get_logger(__name__)
 
 
+# build filter based on the operation and OS name and version
```
ryandillinfelton, line 8: I'm not against function header comments if they clarify it's use or workings, but I think all the information in this comment is already covered by the functions name and parameters.

ryandillinfelton (changes requested):
Mainly blocking on the return of a comment explaining the major_version substring since I think that behavior is very confusing if not explained explicitly. I can take or leave the header comment, but I don't think it adds clarity so I would prefer to leave it out.

api/filtering/custom_filters.py
thearifismail: Set this comment.

api/filtering/custom_filters.py
thearifismail: the header comment gone.

ryandillinfelton (approved):
LGTM

kruai (approved):

PR #1106, opened 2022-03-11
new endpoint for listing unique OS versions per account

ryandillinfelton:
# Overview

This PR is being created to address [ESSNTL-564](https://issues.redhat.com/browse/ESSNTL-564). (Kind of)

I list 564 as the Jira since it was what was planned, but the goals have changed drastically so I may swap it in the future with a different more specific card. Essentially with xjoin-search being replaced soon we thought it unwise to add full generic unique value listing as a feature. This will unblock the main issue that was proposed to solve, listing OS versions on an account for use with filtering.

The companion xjoin-search PR was:
https://github.com/RedHatInsights/xjoin-search/pull/146 

But now replaced by
https://github.com/RedHatInsights/xjoin-search/pull/152

Here is an example query and response. Please note I am using local test data that sidesteps our usual ingestion validation. All fields on operating_system are currently required, and worst case, `name` has always been required. So the first result should never appear in production. That said it's good to know it will not 500 if it does.
```
/api/inventory/v1/system_profile/operating_system
```
```
{
    "total": 4,
    "count": 4,
    "page": 1,
    "per_page": 50,
    "results": [
        {
            "value": {
                "name": "",
                "major": null,
                "minor": null
            },
            "count": 4
        },
        {
            "value": {
                "name": "FED",
                "major": 1,
                "minor": 1
            },
            "count": 1
        },
        {
            "value": {
                "name": "RHEL",
                "major": 0,
                "minor": 0
            },
            "count": 1
        },
        {
            "value": {
                "name": "RHEL",
                "major": 2,
                "minor": 2
            },
            "count": 1
        }
    ]
}
```

TODO (This is still a WIP):
- [ ]  Add tests

## PR Checklist

- [ ] Tests: edge cases
- [x] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [x] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/instrumentation.py
```diff
@@ -98,6 +98,15 @@ def log_get_sap_sids_failed(logger):
     logger.debug("sap_sids not found", extra={"access_rule": get_control_rule()})
 
 
+# get operating_system
+def log_get_operating_system_succeeded(logger, data):
```
thearifismail, line 102: This logger should be generalized for all properties in `system_profile`.

swagger/api.spec.yaml
```diff
@@ -441,6 +441,36 @@ paths:
         '404':
           $ref: '#/components/responses/PageOutOfBounds'
 
+  /system_profile/operating_system:
```
thearifismail, line 444: Same comment as earlier, can this endpoint be generalized over `system_profile`?

swagger/api.spec.yaml
```diff
@@ -705,6 +735,32 @@ components:
       required: false
 
   schemas:
+    SystemProfileOperatingSystemOut:
+      properties:
+        total:
+          $ref: '#/components/schemas/Total'
```
thearifismail, line 741: Is `Total` really needed when count is available?

tests/test_xjoin.py
```diff
@@ -3305,7 +3305,7 @@ def test_generic_filtering_wildcard_invalid_values(subtests, query_source_xjoin,
     for url_builder in endpoint_url_builders:
         for prefix in prefixes:
             for suffix in suffixes:
-                with subtests.test(prefix=prefix, suffix=suffix):
+                with subtests.test(url_builder=url_builder, prefix=prefix, suffix=suffix):
                     url = url_builder(query=prefix + suffix)
```
thearifismail, line 3309: Personally, I prefer `+` for ease of understanding and other languages concatenate strings this way but fashion and modernity police likes f'{prefix}{suffix}'.

ryandillinfelton: I'm not sure what you mean, I'm not concatenating any strings here. I'm just adding the url_builder to the subtest so that when the test fails we may have a better indication of which endpoint the test is failing on. When I made this change I accidentally made a change that broke the filters on the /sap_system endpoint but none of the other. I was confused for a while because I was only testing that it worked on the /hosts endpont. I think noting it will help others in the same situation in the future.

api/system_profile.py
```diff
@@ -202,6 +236,68 @@ def get_sap_sids(search=None, tags=None, page=None, per_page=None, staleness=Non
...
+    if hostfilter_and_variables != ():
+        variables["hostFilter"]["AND"] = hostfilter_and_variables
+
+    response = graphql_query(OPERATING_SYSTEM_QUERY, variables, log_get_operating_system_failed)
+
+    data = response["hostSystemProfile"]
+
+    check_pagination(offset, data["operating_system"]["meta"]["total"])
+
+    log_get_operating_system_succeeded(logger, data)
```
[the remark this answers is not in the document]
ryandillinfelton: This is how the other /system_profile endpoints handle failure. And considering this whole file will need to be refactored whenever we actually do a generic unique values endpoint I'm inclined to kick any refactory changes down the road.

api/system_profile.py
```diff
@@ -202,6 +236,68 @@ def get_sap_sids(search=None, tags=None, page=None, per_page=None, staleness=Non
...
+):
+    if not xjoin_enabled():
+        logger.error("xjoin-search not enabled")
+        flask.abort(503)
+
+    limit, offset = pagination_params(page, per_page)
+
+    variables = {
+        "hostFilter": {
+            # we're not indexing null timestamps in ES
```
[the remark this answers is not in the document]
ryandillinfelton: Honestly IDK. Copy paste casualty. Was probably relevant when we first added the SAP endpoints.

api/system_profile.py
```diff
@@ -86,6 +90,36 @@
...
+    ) {
+        hostSystemProfile (
+            hostFilter: $hostFilter
+        ) {
+            operating_system (
+                limit: $limit
+                offset: $offset
+            ) {
+                meta {
+                    total
```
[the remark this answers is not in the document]
ryandillinfelton: total is present on the other system_profile endpoints and is not equivalent to count. Both are necessary.

api/system_profile.py
```diff
@@ -86,6 +90,36 @@
     }
 """
 
+OPERATING_SYSTEM_QUERY = """
```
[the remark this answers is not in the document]
ryandillinfelton: This is a stop-gap solution to unblock teams waiting on getting unique `operating_system` values. You're absolutely right that this will not scale, but it is the existing pattern we will have to deal with when we actually can implement the generic unique values endpoint. Until then i don't want to spend extra cycles implementing any changes we *think* will make the transition to generic easier.

swagger/api.spec.yaml
ryandillinfelton: We're not generalizing the endpoint at this time. But when we do, if we follow the design I proposed, the generalized endpoints will follow this pattern. `/system_profile/{field_name}/{subfield_name}` etc. From a user perspective the interface will remain the same. As for now there is also the issue of `sap_sids` having a unqiue sap_sids_filter that would be meaningless to other system_profile endpoints which would make generalizing fairly complex, not to mention trying to keep an eye out for any other fields that may need special treatment. Basically it's just out of scope for this PR.

api/system_profile.py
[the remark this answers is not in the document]
thearifismail: Yes, please refactor the code when generic unique values endpoint is done.  Right now it does not look good.

swagger/api.spec.yaml
thearifismail: What about this question?

tests/test_xjoin.py
thearifismail: Isn't this `prefix + suffix` string concatenation?  The concatenation of prefix+suffix is assigned to the query?

api/system_profile.py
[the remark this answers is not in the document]
kruai: Can you clarify the issue or link to an example of the way you'd prefer it be implemented @thearifismail ? The way he's implementing it is in line with how we've implemented the API endpoints for both the Host and System Profile. If `graphql_query` fails, it logs the failure and aborts, so the success is never called

swagger/api.spec.yaml
```diff
@@ -385,8 +385,8 @@ paths:
   /system_profile/sap_system:
     get:
       tags:
-        - sap_system
-      summary: get sap system values
+        - system_profile
+      summary: get all sap system values and counts on an account
```
kruai, line 389: The tags got swapped with the `/system_profile/sap_sids` endpoint

swagger/api.spec.yaml
```diff
@@ -414,8 +414,8 @@ paths:
   /system_profile/sap_sids:
     get:
       tags:
-        - sap_system
-      summary: get sap system values
+        - system_profile
+      summary: get all sap sids values and counts on an account
```
kruai, line 418: The tags got swapped with the `/system_profile/sap_system` endpoint

kruai (changes requested):
Mostly got some questions

swagger/api.spec.yaml
ryandillinfelton: This is Intentional. I think having them all categorized under system profile makes more sense, especially as we add more system_profile resources like operating_system.

kruai: My bad, I misread; thought they were being unintentionally swapped. I'm fine with consolidation

api/system_profile.py
```diff
@@ -202,6 +236,68 @@ def get_sap_sids(search=None, tags=None, page=None, per_page=None, staleness=Non
...
+@api_operation
+@rbac(Permission.READ)
+@metrics.api_request_time.time()
+def get_operating_system(
+    tags=None,
+    page: Optional[int] = None,
+    per_page: Optional[int] = None,
+    staleness: Optional[str] = None,
+    registered_with: Optional[str] = None,
+    filter=None,
```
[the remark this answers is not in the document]
ryandillinfelton: It's got the same parameters as the other system profile endpoints (aside from the sap_sids filter). I think it would be good to assess expanding the filtering on them to match `/hosts` but since I don't know how involved that would be, and I'd like all of the system profile endpoints to support the same host filters, I'd prefer to make that a standalone task rather than fold it into this.

api/system_profile.py
```diff
@@ -202,6 +236,68 @@ def get_sap_sids(search=None, tags=None, page=None, per_page=None, staleness=Non
...
+):
+    if not xjoin_enabled():
+        logger.error("xjoin-search not enabled")
+        flask.abort(503)
```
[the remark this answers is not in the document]
ryandillinfelton: Removed :+1:

api/system_profile.py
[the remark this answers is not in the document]
kruai: Fair point, but considering the use case, it might make more sense for it to match the `/hosts` implementation. My understanding system profile endpoints are like that so we can fetch data for a specific page of hosts; for the unique values, we want to get the full list of unique operating systems for all hosts (on all pages) in the current search. Does that make sense, or am I off base?

thearifismail:
@ryandillinfelton shouldn't this be closed now because this changed is already in https://github.com/RedHatInsights/insights-host-inventory/pull/1109/files?

ryandillinfelton:
> @ryandillinfelton shouldn't this be closed now because this changed is already in https://github.com/RedHatInsights/insights-host-inventory/pull/1109/files?

This is about returning the operating system versions represented on a set of hosts, nothing to do with per_reporter_staleness.

thearifismail (approved):
LGTM

kruai (approved):
LGTM

PR #1112, opened 2022-03-29
Migration to add org_id column, and its associated index to hosts/

roliveri:
# Overview

Add migration to add org_id column, and its associated index to the hosts table.

This PR is being created to address [ESSNTL-2522](https://issues.redhat.com/browse/ESSNTL-2522).

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/migrations/versions/268debe6db17_add_org_id_column_and_index.py
+++ b/migrations/versions/268debe6db17_add_org_id_column_and_index.py
@@ -5,19 +5,19 @@ Revises: 3d06b0a8828f
 Create Date: 2022-03-28 13:46:34.982026
 
 """
-from alembic import op
 import sqlalchemy as sa
+from alembic import op
 
 
 # revision identifiers, used by Alembic.
-revision = '268debe6db17'
-down_revision = '3d06b0a8828f'
+revision = "268debe6db17"
+down_revision = "3d06b0a8828f"
 branch_labels = None
 depends_on = None
 
 
 def upgrade():
-    op.add_column("hosts", sa.Column("org_id", sa.String(length=255), nullable=True))
+    op.add_column("hosts", sa.Column("org_id", sa.String(length=36), nullable=True))
     op.create_index("idxorgid", "hosts", ["org_id"])
 
 
--- a/migrations/versions/268debe6db17_add_org_id_column_and_index.py
+++ b/migrations/versions/268debe6db17_add_org_id_column_and_index.py
@@ -1,0 +1,26 @@
+"""Add org_id column and index
+
+Revision ID: 268debe6db17
+Revises: 3d06b0a8828f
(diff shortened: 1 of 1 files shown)


roliveri:
Not sure what the length of the column should be. The account column is currently 10, but I'm not sure if the org_id has a well defined format, to I made it 255 to be safe.

kruai (approved):

thearifismail (approved):
LGTM

kruai (changes requested):
I actually just learned that the [official Postgres guidance](https://wiki.postgresql.org/wiki/Don't_Do_This#Text_storage) is to not use char(n)/strings of fixed length. They actually recommend Text, but we can use a CHECK constraint to verify that they're not longer than 36 chars. IDK how to do that in SQLAlchemy off the top of my head though

thearifismail:
I guess 
```
op.add_column("hosts", sa.Column("org_id", sa.String(length=36), nullable=True))
``` 
should be replaced by 
```
op.add_column("hosts", sa.Column("org_id", UUID, nullable=True))
```
where `UUID` provided by `sqlalchemy.dialects.postgresql`

roliveri:
@thearifismail Is it a UUID? UUID implies a format in addition to length. Validation would fail if the org_id isn't in "UUID" format. Unless we know that the org_id will always be in a recognized UUID format, we probably should just restrict the length.

kruai:
> I guess
> 
> ```
> op.add_column("hosts", sa.Column("org_id", sa.String(length=36), nullable=True))
> ```
> 
> should be replaced by
> 
> ```
> op.add_column("hosts", sa.Column("org_id", UUID, nullable=True))
> ```
> 
> where `UUID` provided by `sqlalchemy.dialects.postgresql`

We don't know that it's going to be UUID format yet; we just know that it will not be longer than a UUID. It could end up being no longer than our current account numbers. We should use a Text type and validate the length using a CHECK constraint

roliveri:
@kruai But we're not using `char(n)` strings. It seems these columns are defined as `varchar(n)` strings in postgres. The document you reference seems to have less of a problem with `varchar(n)` than it does with `char(n)`. In our usage, it seems to me that the larger fields might be more of a problem: `display_name(200)`, `ansibel_host(255)` and `reporter(255)`. However, I think these are ok, as I believe using `text` would be overkill in these instances.

If we do determine that our string definitions should change, should they all be addressed in a followup PR?

PR #1114, opened 2022-03-30
Unleash integration; hide edge hosts by default if feature flag enabled

kruai:
# Overview

This PR is being created to address [ESSNTL-2469](https://issues.redhat.com/browse/ESSNTL-2469) and [ESSNTL-2425](https://issues.redhat.com/browse/ESSNTL-2425).
This connects HBI to our Unleash feature flag service, and implements filtering out Edge hosts by default when the feature flag is enabled. The feature flag's name is `hbi.api.hide-edge-by-default`, and is located [here in Stage Unleash](https://insights-stage.unleash.devshift.net/#/features/strategies/hbi.api.hide-edge-by-default).

I also added and modified some xjoin tests to validate the functionality with and without the flag enabled. With the flag disabled, the current functionality should not change; with it enabled, Edge hosts get filtered out unless the request is specifically filtering on `system_profile.host_type`.

## PR Checklist

- [x] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -341,10 +341,11 @@ def query_filters(
         query_filters += ({"provider_id": {"eq": provider_id.casefold()}},)
 
     # FEATURE FLAG: Edge hosts should be hidden by default.
-    # If the feature flag "hbi.api-hide-edge-default" is enabled, filter here on the API side.
+    # If the feature flag is enabled, filter here on the API side.
     sp_filter = deepcopy(filter)
     if UNLEASH.client.is_enabled(FLAG_HIDE_EDGE_BY_DEFAULT):
-        sp_filter.setdefault("system_profile", []).append({"host_type": "edge"})
+        if sp_filter.get("system_profile", {}).get("host_type") is None:
+            sp_filter.setdefault("system_profile", {}).update({"host_type": "nil"})
 
     for key in sp_filter:
         if key == "system_profile":
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from enum import Enum
 from functools import partial
 from uuid import UUID
@@ -14,6 +15,8 @@ from app.utils import Tag
 from app.validators import is_custom_date as is_timestamp
 from app.xjoin import staleness_filter
 from app.xjoin import string_contains_lc
+from lib.feature_flags import FLAG_HIDE_EDGE_BY_DEFAULT
+from lib.feature_flags import UNLEASH
 
 logger = get_logger(__name__)
 
@@ -337,9 +340,15 @@ def query_filters(
     if provider_id:
         query_filters += ({"provider_id": {"eq": provider_id.casefold()}},)
 
-    for key in filter:
+    # FEATURE FLAG: Edge hosts should be hidden by default.
+    # If the feature flag "hbi.api-hide-edge-default" is enabled, filter here on the API side.
+    sp_filter = deepcopy(filter)
+    if UNLEASH.client.is_enabled(FLAG_HIDE_EDGE_BY_DEFAULT):
+        sp_filter.setdefault("system_profile", []).append({"host_type": "edge"})
+
+    for key in sp_filter:
         if key == "system_profile":
-            query_filters += build_system_profile_filter(filter["system_profile"])
+            query_filters += build_system_profile_filter(sp_filter["system_profile"])
         else:
             raise ValidationException("filter key is invalid")
 
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -416,9 +416,10 @@ class ConfigTestCase(TestCase):
 
 
 @patch("app.db.get_engine")
+@patch("app.init_unleash_app")
 @patch("app.Config", **{"return_value.mgmt_url_path_prefix": "/"})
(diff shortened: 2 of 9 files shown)


api/filtering/filtering.py
```diff
@@ -337,9 +340,16 @@ def query_filters(
     if provider_id:
         query_filters += ({"provider_id": {"eq": provider_id.casefold()}},)
 
-    for key in filter:
+    # FEATURE FLAG: Edge hosts should be hidden by default.
+    # If the feature flag is enabled, filter here on the API side.
+    sp_filter = deepcopy(filter)
+    if UNLEASH.client.is_enabled(FLAG_HIDE_EDGE_BY_DEFAULT):
```
RonnyPfannschmidt, line 346: we might want to move this particular check/change into build_system_profile_filter

kruai: We can... I was just concerned it would increase the logical complexity. I'll give it a shot, though!

kruai: OK, after giving it a couple attempts, I don't think it makes sense to put this in `build_system_profile_filter()`. That function is called for each `filter[system_profile]` provided in the request params, and its purpose is to process each of those into a graphql condition for the xjoin filter.

By contrast, if the feature flag is set, we need to provide this graphql condition whether or not there's a `filter[system_profile]` in the request. Does that make sense @RonnyPfannschmidt ?

RonnyPfannschmidt: the controll fow for the loop is a little strange, i propose we change things around a bit, will need a bit to account for the fact that the only valid key is system profile

and bascially now we always have to build the filter list, which is allowed to be empty!

RonnyPfannschmidt: @kruai if `build_system_profile_filter` will take a None as input, this can be done nicely

i recommend adding a rough type annotation if feasible

kruai: @RonnyPfannschmidt `build_system_profile_filter` accepting None is fine, but doesn't simplify the host_type filtering logic, as we need to check specifically whether that system_profile field was filtered on.

I shifted the code around a bit to move the system-profile-related filter logic into its own function, which executes even if the filter list is empty; it handles the feature flag and then iterates through the SP filters as usual. I also added some type hinting. Is this closer to how you were envisioning it?

thearifismail (commented):
The code LGTM but how is"unleash" integration related to whether hide or show "edge" systems?

tests/test_unit.py
```diff
@@ -429,19 +429,21 @@ def test_kafka_producer_defaults(self):
 
 
 @patch("app.db.get_engine")
+@patch("app.init_unleash_app")
 @patch("app.Config", **{"return_value.mgmt_url_path_prefix": "/"})
 class CreateAppConfigTestCase(TestCase):
-    def test_config_is_assigned(self, config, get_engine):
+    def test_config_is_assigned(self, config, get_engine, init_unleash_app):
```
thearifismail, line 435: What's the benefit of patching `init_unleash_app` when no verification is being done?

tests/test_unit.py
```diff
@@ -429,19 +429,21 @@ def test_kafka_producer_defaults(self):
...
         self.assertEqual(config.return_value, app.config["INVENTORY_CONFIG"])
 
 
 @patch("app.connexion.App")
 @patch("app.db.get_engine")
+@patch("app.init_unleash_app")
 class CreateAppConnexionAppInitTestCase(TestCase):
     @patch("app.TranslatingParser")
-    def test_specification_is_provided(self, translating_parser, get_engine, app):
+    def test_specification_is_provided(self, translating_parser, get_engine, init_unleash_app, app):
```
thearifismail, line 446: Same as above

tests/test_unit.py
kruai: Calling the real `init_unleash_app()` would cause an error, since `app.Config` is patched above. The real function requires certain values to be set, and initializes a connection to the configured Unleash server (which we don't want to do for these config tests).

kruai:
> The code LGTM but how is"unleash" integration related to whether hide or show "edge" systems?

In a couple of our stand-up meetings, we collectively determined that this functionality would be a good place to start using feature flags (thus why this PR is also addressing [ESSNTL-2469](https://issues.redhat.com/browse/ESSNTL-2469)). "Unleash" is the tool that people are standardizing on.

tests/test_unit.py
kruai: Ah, maybe I wasn't clear... this `init_unleashed_app` parameter is patched (above, on line 432). I had to patch this because it's called as part of the base `create_app()` function (in `app/__init__.py`), and if the _real_ function is called using the mock config used in these tests, it will raise an error and break `create_app()`.

So yeah, the tests in this file aren't testing Unleash - the tests in `tests/test_xjoin.py` cover that.

thearifismail: Ah got it.

thearifismail (dismissed):

kruai:
Closing because they no longer want this.

PR #1134, opened 2022-06-01
Check response for None before indexing

roliveri:
HBI returns 500 on /hosts request, through xjoin, if there are no hosts in the database. This is because xjoin returns None in this situation and we try to index that return value.

# Overview

This PR is being created to address [ESSNTL-1158](https://issues.redhat.com/browse/ESSNTL-1158).
In get_host_list_using_filters(), check response for None before indexing.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/host_query_xjoin.py
+++ b/api/host_query_xjoin.py
@@ -1,3 +1,5 @@
+import flask
+
 from api.filtering.filtering import host_id_list_query_filter
 from api.filtering.filtering import query_filters
 from app.auth import get_current_identity
@@ -94,8 +96,10 @@ def get_host_list_using_filters(all_filters, page, per_page, param_order_by, par
         "fields": system_profile_fields,
     }
     response = graphql_query(QUERY, variables, log_get_host_list_failed)
-    if response is None:
-        return iter([]), 0, additional_fields
+    if response is None or "hosts" not in response:
+        # Log an error implicating xjoin, then abort with status 503
+        logger.error("xjoin-search responded with invalid format")
+        flask.abort(503)
 
     response = response["hosts"]
 
--- a/api/host_query_xjoin.py
+++ b/api/host_query_xjoin.py
@@ -93,7 +93,11 @@ def get_host_list_using_filters(all_filters, page, per_page, param_order_by, par
         "filter": all_filters,
         "fields": system_profile_fields,
     }
-    response = graphql_query(QUERY, variables, log_get_host_list_failed)["hosts"]
+    response = graphql_query(QUERY, variables, log_get_host_list_failed)
+    if response is None:
+        return iter([]), 0, additional_fields
+
+    response = response["hosts"]
 
     total = response["meta"]["total"]
     check_pagination(offset, total)
--- a/tests/test_xjoin.py
+++ b/tests/test_xjoin.py
@@ -62,6 +62,28 @@ def test_host_request_xjoin_status_403(patch_xjoin_post, api_get):
     assert response_status == 500
 
 
+def test_host_request_xjoin_invalid_response_none(patch_xjoin_post, api_get):
+    patch_xjoin_post(response={"data": None}, status=200)
+    request_id = generate_uuid()
+
+    response_status, response_data = api_get(
+        HOST_URL, extra_headers={"x-rh-insights-request-id": request_id, "foo": "bar"}
+    )
+
+    assert response_status == 503
+
+
+def test_host_request_xjoin_invalid_response_empty(patch_xjoin_post, api_get):
+    patch_xjoin_post(response={"data": {}}, status=200)
+    request_id = generate_uuid()
+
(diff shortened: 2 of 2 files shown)


api/host_query_xjoin.py
```diff
@@ -93,7 +93,11 @@ def get_host_list_using_filters(all_filters, page, per_page, param_order_by, par
         "filter": all_filters,
         "fields": system_profile_fields,
     }
-    response = graphql_query(QUERY, variables, log_get_host_list_failed)["hosts"]
+    response = graphql_query(QUERY, variables, log_get_host_list_failed)
+    if response is None:
```
kruai, line 97: After setting up xjoin-search and creating the index but _not_ populating it with hosts, this is the response I get from the host query:
```
{
  "data": {
    "hosts": {
      "meta": {
        "count": 0,
        "total": 0
      },
      "data": []
    }
  }
}
```
The `graphql_query` function returns the value of the (inner) `data` field, which I'd think would be an empty array, not None. Correct me if I'm wrong, but should this not be checking against an empty array instead of None?

kruai (changes requested):
I left a comment, but can you please also create one test that validates this fix? It should patch the xjoin response (likely with `patch_xjoin_post`), using the response that xjoin-search actually returns when there are 0 hosts. My comment in this PR has the response data, if that helps!

api/host_query_xjoin.py
kruai: IMO we should expect that the index exists; otherwise, something is wrong with xjoin, and our usual exception handling should suffice. To simulate not having any hosts, I just ran seed.js after removing the code that actually adds/indexes the hosts, which happens here: https://github.com/RedHatInsights/xjoin-search/blob/master/test/seed.js#L84

roliveri: @kruai as we discussed, when the index exists and there are no hosts, the code works as expected.

We also discussed the possibility of propagating the error from the `xjoin-search` layer to the `insights-host-inventory` API return. However, it appears `xjoin-search` does not return an error in this case. The  `insights-host-inventory` code is indeed checking for non-200 return codes from `xjoin-search`.

It looks like `elasticsearch` is returning a 404 status to `xjoin-search`, as it appears in the `xjoin-search` log, but `xjoin-search` ultimately returns a 200 status to `insights-host-inventory`.

Given the fix for this would require changes to `xjoin-search`, it would be a high-risk change for a rare (or non-existent) edge case.

Should we just check for None on our end, and possibly log the unexpected return value from `xjoin-search`?

kruai: @roliveri LMK your thoughts - I'm open to other ideas if you'd prefer a different approach

fstavela (approved):

api/host_query_xjoin.py
roliveri: @kruai Should the returned status be 503 or 500?
It seems to me 500 would be more appropriate in this case.

kruai: I try to avoid 500 since that's the generic catch-all. Generally if our app returns status 500 it means that we failed to handle an exception

tests/test_xjoin.py
```diff
@@ -62,6 +62,28 @@ def test_host_request_xjoin_status_403(patch_xjoin_post, api_get):
...
+    patch_xjoin_post(response={"data": {}}, status=200)
+    request_id = generate_uuid()
+
+    response_status, response_data = api_get(
+        HOST_URL, extra_headers={"x-rh-insights-request-id": request_id, "foo": "bar"}
+    )
+
+    assert response_status == 503
+
+
```
kruai, line 86: Looks good, but this can also be parameterized to reduce duplication:
```suggestion
@pytest.mark.parametrize("response_data", (None, {}))
def test_host_request_xjoin_invalid_response_none(patch_xjoin_post, api_get, response_data):
    patch_xjoin_post(response={"data": response_data}, status=200)
    request_id = generate_uuid()

    response_status, response_data = api_get(
        HOST_URL, extra_headers={"x-rh-insights-request-id": request_id, "foo": "bar"}
    )

    assert response_status == 503


```

kruai (changes requested):
Just one more request!

tests/test_xjoin.py
roliveri: Done.

thearifismail (approved):
LGTM

kruai (approved):
LGTM!

PR #1138, opened 2022-06-07
Update returned message when no host is found for deletion 

FabriciaDinizRH:
# Overview

This PR is being created to address [ESSNTL-2359](https://issues.redhat.com/browse/ESSNTL-2359).

Changes `api/host.py` so that when a user tries to delete a host that does not exist, an appropriate message is returned.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


InsightsDroid:
Test results are available in [Ibutsu](https://url.corp.redhat.com/ibutsu-runs). The test run IDs are:
1fbbe5f4-7245-4a1d-b4c1-9237e4ecac5a
5901316b-288a-43e8-ab78-1b2797e65af6

kruai (approved):
LGTM!

thearifismail (approved):
LGTM

api/host.py
```diff
@@ -165,11 +165,8 @@ def delete_hosts_by_filter(
         logger.error("xjoin-search not accessible")
         flask.abort(503)
 
-    if len(ids_list) == 0:
-        flask.abort(status.HTTP_404_NOT_FOUND, "No hosts found for deletion.")
-
     try:
-        delete_count = _delete_host_list(ids_list)
+        delete_count = _delete_host_list(ids_list) if len(ids_list) else 0
```
RonnyPfannschmidt, line 169:
```suggestion
        delete_count = _delete_host_list(ids_list) if ids_list else 0
```

api/host.py
```diff
@@ -188,25 +185,23 @@ def _delete_host_list(host_id_list):
...
-                deletion_count += 1
-            else:
-                log_host_delete_failed(logger, host_id, get_control_rule())
-                tracker_message = "not deleted host"
-
-            with PayloadTrackerProcessingContext(
-                payload_tracker, processing_status_message=tracker_message
-            ) as payload_tracker_processing_ctx:
-                payload_tracker_processing_ctx.inventory_id = host_id
+        if query.count():
```
RonnyPfannschmidt, line 189: i recall that count queries can be expensive for the database, given that the actual deletion should be empty anyway, wouldnt a black iteration by some niceness?

FabriciaDinizRH: @RonnyPfannschmidt what's a black iteration?

api/host.py
```diff
@@ -188,25 +185,23 @@ def _delete_host_list(host_id_list):
...
-                deletion_count += 1
-            else:
-                log_host_delete_failed(logger, host_id, get_control_rule())
-                tracker_message = "not deleted host"
-
-            with PayloadTrackerProcessingContext(
-                payload_tracker, processing_status_message=tracker_message
-            ) as payload_tracker_processing_ctx:
-                payload_tracker_processing_ctx.inventory_id = host_id
+        if query:
```
kruai, line 189: I don't think this works as expected; afaict the most performant way to verify that 1 or more records are returned is:
```suggestion
        if query.first()
```
Correct me if I'm wrong though @RonnyPfannschmidt

FabriciaDinizRH: @RonnyPfannschmidt I was taking a deeper look into this and saw that the query is a SQLAlchemy BaseQuery object, so even if no hosts have been found, query is True and the code goes into the for loop to try to delete the hosts. According to the good people of the internet Asa's suggestion is the best one to check for results.

RonnyPfannschmidt: the point is - if we run the deletion, and there where no results , its jsut a empty loop

if there are results, we have to run the loop anyway, and thus have a extra count query

additionally count queries on postgresql can be expensive (for correctness constraints)

so why not jsut run the deletion function regardess of the count

FabriciaDinizRH: Ok, let me see if were on the same page. The actual deletion occurs on line 190, so if there are no results from `_get_host_list_by_id_list_from_db()`,  `deleted_count` stays as 0 and the whole deletion part is not executed. That way the deletion doesn't happen on a empty list. That was the previous behavior and what I tried to maintain without having to send a 404 to the user.
So the way I see it the deletion loop only happens if there are results from the db and I'm having a hard time understanding your point because of that.

kruai: Ohhhhh I see what you mean now @RonnyPfannschmidt. Yes, since `delete_hosts()` will return no results in that scenario, we shouldn't need to check `query.count()` or `query.first()`. I'm in agreement that the conditional can just be removed.

@FabriciaDinizRH Since we already do `while select_query.count()` in [delete_hosts on L20](https://github.com/FabriciaDinizRH/insights-host-inventory/blob/ESSNTL-2359/lib/host_delete.py#L20), checking to see if any records are returned right before that is redundant. If no hosts are returned by that query, there are no records to iterate through, and the whole loop is effectively skipped.

FabriciaDinizRH: Oh, I see. I guess i was not looking at the right place.
Fixed it.

kruai:
The IQE issue appears to be resolved, so it's worth investigating what's causing the issue now. I don't recognize the error at first glance, so it might still be an issue with Clowder or something.
Just in case it's a transient issue, I'm going to trigger a retry on the build and see if it resolves itself.

PR #1140, opened 2022-06-17
ESSNTL-2863: Inventory doesn't send update event messages after updating facts

FabriciaDinizRH:
# Overview

This PR is being created to address [ESSNTL-2863](https://issues.redhat.com/browse/ESSNTL-2863).

Adds a event emitter when host facts are updated by namespace.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -365,10 +365,10 @@ def update_facts_by_namespace(operation, host_id_list, namespace, fact_dict):
         else:
             host.merge_facts_in_namespace(namespace, fact_dict)
 
-    if db.session.is_modified(host):
-        db.session.commit()
-        serialized_host = serialize_host(host, staleness_timestamps(), EGRESS_HOST_FIELDS)
-        _emit_patch_event(serialized_host, host.id, host.canonical_facts.get("insights_id"))
+        if db.session.is_modified(host):
+            db.session.commit()
+            serialized_host = serialize_host(host, staleness_timestamps(), EGRESS_HOST_FIELDS)
+            _emit_patch_event(serialized_host, host.id, host.canonical_facts.get("insights_id"))
 
     logger.debug("hosts_to_update:%s", hosts_to_update)
 
--- a/api/host.py
+++ b/api/host.py
@@ -366,7 +366,7 @@ def update_facts_by_namespace(operation, host_id_list, namespace, fact_dict):
         else:
             host.merge_facts_in_namespace(namespace, fact_dict)
 
-        db.session.commit()
+    db.session.commit()
 
     logger.debug("hosts_to_update:%s", hosts_to_update)
 
--- a/api/host.py
+++ b/api/host.py
@@ -366,7 +366,10 @@ def update_facts_by_namespace(operation, host_id_list, namespace, fact_dict):
         else:
             host.merge_facts_in_namespace(namespace, fact_dict)
 
-    db.session.commit()
+        if db.session.is_modified(host):
+            db.session.commit()
+            serialized_host = serialize_host(host, staleness_timestamps(), EGRESS_HOST_FIELDS)
+            _emit_patch_event(serialized_host, host.id, host.canonical_facts.get("insights_id"))
 
     logger.debug("hosts_to_update:%s", hosts_to_update)
 
--- a/api/host.py
+++ b/api/host.py
@@ -366,10 +366,7 @@ def update_facts_by_namespace(operation, host_id_list, namespace, fact_dict):
         else:
             host.merge_facts_in_namespace(namespace, fact_dict)
 
-        if db.session.is_modified(host):
-            db.session.commit()
-            serialized_host = serialize_host(host, staleness_timestamps(), EGRESS_HOST_FIELDS)
-            _emit_patch_event(serialized_host, host.id, host.canonical_facts.get("insights_id"))
+        db.session.commit()
 
     logger.debug("hosts_to_update:%s", hosts_to_update)
 
--- a/api/host.py
+++ b/api/host.py
@@ -366,7 +366,10 @@ def update_facts_by_namespace(operation, host_id_list, namespace, fact_dict):
         else:
             host.merge_facts_in_namespace(namespace, fact_dict)
 
-    db.session.commit()
+    if db.session.is_modified(host):
+        db.session.commit()
+        serialized_host = serialize_host(host, staleness_timestamps(), EGRESS_HOST_FIELDS)
+        _emit_patch_event(serialized_host, host.id, host.canonical_facts.get("insights_id"))
 
     logger.debug("hosts_to_update:%s", hosts_to_update)
 
--- a/tests/test_api_hosts_update.py
+++ b/tests/test_api_hosts_update.py
@@ -303,7 +303,7 @@ def test_add_facts_without_fact_dict(api_patch, db_create_host):
     assert_error_response(response_data, expected_status=400, expected_detail="Request body is not valid JSON")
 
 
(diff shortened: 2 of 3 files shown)


InsightsDroid:
Test results are available in Ibutsu:
https://url.corp.redhat.com/ibutsu-runs-f00a647b-b384-4ed4-9a09-a36b8559a6f9

fstavela:
I tested this and it works and sends update event message after the facts have been updated via REST API.

However, I found out that this message, or any other event message, doesn't include `facts`. I don't know if this was decided in the past and it is intentional, or if it is a bug, so I created another Jira card for it and I'm okay with fixing it in another PR, so I think this is good to go.

Mentioned Jira: https://issues.redhat.com/browse/ESSNTL-3012

fstavela (dismissed):

kruai:
> I tested this and it works and sends update event message after the facts have been updated via REST API.
> 
> However, I found out that this message, or any other event message, doesn't include `facts`. I don't know if this was decided in the past and it is intentional, or if it is a bug, so I created another Jira card for it and I'm okay with fixing it in another PR, so I think this is good to go.
> 
> Mentioned Jira: https://issues.redhat.com/browse/ESSNTL-3012

I don't think I was around when this was decided, but I also don't know which services need each host's facts to be sent on every update. It seems like this was an intentional choice based on the documentation, since the `facts` field is defined in the expected message format, but is not in the created/updated event interface. We leave a number of other fields out of the events we produce, like `subscription_manager_id`, `satellite_id`, and `bios_uuid`. I'm guessing we only included the fields we absolutely needed to provide in order to keep the message size down, which is an issue we're hitting today.

kruai:
Actually, which services need to keep track of the `facts` field @fstavela? It only makes sense to send an update event when `facts` gets updated if `facts` is also included in the created/updated event, and it only makes sense to add it to those events if an existing app is consuming our messages and needs to synchronize that field

fstavela:
> 

I see all of those UUID fields in the created/updated events. Basically, I see everything there (including the entire system_profile) except of the `facts`. You can see the example received event in the description of the created Jira card.

fstavela:
> Actually, which services need to keep track of the `facts` field @fstavela? It only makes sense to send an update event when `facts` gets updated if `facts` is also included in the created/updated event, and it only makes sense to add it to those events if an existing app is consuming our messages and needs to synchronize that field

Hmm, you are right. I don't know about any such application, but I would guess that there is some, since the field is actively being populated by the reporters.

However, I would also like to know what exactly is that field used for, for me it looks like a lot of times the data there is duplicated from `system_profile`. Isn't `facts` just some historic predecessor of `system_profile`? Do we even still need it?

kruai:
> > > I tested this and it works and sends update event message after the facts have been updated via REST API.
> > > However, I found out that this message, or any other event message, doesn't include `facts`. I don't know if this was decided in the past and it is intentional, or if it is a bug, so I created another Jira card for it and I'm okay with fixing it in another PR, so I think this is good to go.
> > > Mentioned Jira: https://issues.redhat.com/browse/ESSNTL-3012
> > 
> > 
> > I don't think I was around when this was decided, but I also don't know which services need each host's facts to be sent on every update. It seems like this was an intentional choice based on the documentation, since the `facts` field is defined in the expected message format, but is not in the created/updated event interface. We leave a number of other fields out of the events we produce, like `subscription_manager_id`, `satellite_id`, and `bios_uuid`. I'm guessing we only included the fields we absolutely needed to provide in order to keep the message size down, which is an issue we're hitting today.
> 
> I see all of those UUID fields in the created/updated events. Basically, I see everything there (including the entire system_profile) except of the facts. You can see the example received event in the description of the created Jira card.

Oh, you're right. In that case, we need to update our documentation to reflect that...


> > Actually, which services need to keep track of the `facts` field @fstavela? It only makes sense to send an update event when `facts` gets updated if `facts` is also included in the created/updated event, and it only makes sense to add it to those events if an existing app is consuming our messages and needs to synchronize that field
> 
> Hmm, you are right. I don't know about any such application, but I would guess that there is some, since the field is actively being populated by the reporters.
> 
> However, I would also like to know what exactly is that field used for, for me it looks like a lot of times the data there is duplicated from `system_profile`. Isn't `facts` just some historic predecessor of `system_profile`? Do we even still need it?

I don't know! It would be great if we could just remove it, but we need to check with our dependent services first. I _do_ see that it's not a syndicated field in Cyndi, but we need to make sure nobody is querying it from us directly

api/host.py
```diff
@@ -365,7 +365,10 @@ def update_facts_by_namespace(operation, host_id_list, namespace, fact_dict):
         else:
             host.merge_facts_in_namespace(namespace, fact_dict)
 
-    db.session.commit()
+    if db.session.is_modified(host):
+        db.session.commit()
+        serialized_host = serialize_host(host, staleness_timestamps(), EGRESS_HOST_FIELDS)
+        _emit_patch_event(serialized_host, host.id, host.canonical_facts.get("insights_id"))
```
kruai, line 371: This needs to be inside the `for host` loop in order to produce an update for each host:
```suggestion
        if db.session.is_modified(host):
            db.session.commit()
            serialized_host = serialize_host(host, staleness_timestamps(), EGRESS_HOST_FIELDS)
            _emit_patch_event(serialized_host, host.id, host.canonical_facts.get("insights_id"))
```

kruai (changes requested):
In addition to my suggested change, we should also cover the event message production in our testing. It should at least verify that a message was produced for each host that was updated

tests/test_api_hosts_update.py
```diff
@@ -271,6 +271,25 @@ def test_patch_produces_update_event_no_insights_id(
     )
 
 
+def test_patch_by_namespace_produces_update_event(
+    event_producer_mock, event_datetime_mock, db_create_host, db_get_host, api_patch
```
kruai, line 275: `db_get_host` is not used and should be removed.

tests/test_api_hosts_update.py
```diff
@@ -271,6 +271,25 @@ def test_patch_produces_update_event_no_insights_id(
...
+):
+    host = db_host()
+    created_host = db_create_host(host=host, extra_data={"facts": DB_FACTS})
```
kruai, line 278: This is an optional change, but since `host` is only used once, I personally wouldn't create a variable for it. Up to you, though! If you think it makes it less readable, I'm fine with leaving it the way it is.
```suggestion
    created_host = db_create_host(host=db_host(), extra_data={"facts": DB_FACTS})
```

tests/test_api_hosts_update.py
```diff
@@ -271,6 +271,25 @@ def test_patch_produces_update_event_no_insights_id(
...
+
+    facts_url = build_facts_url(host_list_or_id=created_host.id, namespace=DB_FACTS_NAMESPACE)
+    response_status, response_data = api_patch(facts_url, DB_NEW_FACTS)
```
kruai, line 281: Kind of a nitpick, but generally when a function returns a value that won't be used, you just use `_`.
```suggestion
    response_status, _ = api_patch(facts_url, DB_NEW_FACTS)
```

FabriciaDinizRH: I like to learn those things. I'll use _ from now on

kruai (approved):
LGTM!

thearifismail (approved):
LGTM

PR #1164, opened 2022-07-19
Make it possible to use other than platform.inventory.events topic

jhutar:
# Overview

Do not fail with KeyError when we configure other than default value


thearifismail (approved):

kruai (approved):

jhutar:
Thank you again!

PR #1165, opened 2022-07-19
Avoid KeyError if we do not have platform.inventory.host-ingress-p1 and app/config.py is used

jhutar:
# Overview

If we have configured other than default `platform.inventory.host-ingress-p1` in ClowdApp and the code runs through app/config.py, there is:

```
        self.additional_validation_topic = topic(
            os.environ.get("KAFKA_ADDITIONAL_VALIDATION_TOPIC", "platform.inventory.host-ingress-p1")
        )
```

So we need a way how to specify this variable for every app so this cod is happy even if we have this in the ClowdApp:

```
    kafkaTopics:
      - topicName: platform.inventory.system-profile
        partitions: 1
      - topicName: platform.payload-status
        partitions: 1
      - topicName: platform.inventory.events-devel
        partitions: 1
      - topicName: platform.inventory.host-ingress-devel
        partitions: 1
      - topicName: platform.inventory.host-ingress-p1-devel
        partitions: 1
```

### DIFF
--- a/deployment.yaml
+++ b/deployment.yaml
@@ -70,6 +70,8 @@ objects:
           value: ${KAFKA_SECURITY_PROTOCOL}
         - name: KAFKA_SASL_MECHANISM
           value: ${KAFKA_SASL_MECHANISM}
+        - name: KAFKA_ADDITIONAL_VALIDATION_TOPIC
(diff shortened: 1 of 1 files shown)


thearifismail (approved):
LGTM

jhutar:
And thank you again :)

PR #1173, opened 2022-07-27
Replace kafka-python with confluent-kafka.

thearifismail:
# Overview

This is DFRAFT PR implements [ESSNTL-1878](https://issues.redhat.com/browse/ESSNTL-1878) for using `confluent-kafka` in place of `kafka-python`.

Using confluent-kafa in place of kafka-python had to deal with two changes:
1. Calling Consumer.poll() returned one message instead of a list of messages returned by kafka-python.  This required changing some tests.
2. Producing an event did not return anything and the status was provided back as a message or error to the callback which was called automatically when Producer.poll() was explicitly called.

Changes are primarily refactoring the code and modifying tests to respond to changes in the way produce event and callbacks are used.

NOTE: The message_processed and message_not_processed from the instrumentation module are not applicable anymore to the test-api-host-updates. Another integration test is needed but not sure where?

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/queue/event_producer.py
```diff
@@ -1,37 +1,57 @@
...
 
 from app.instrumentation import message_not_produced
 from app.instrumentation import message_produced
 from app.logging import get_logger
 
+# from confluent_kafka.error import KafkaError
+
 logger = get_logger(__name__)
 
+__event__ = None
```
RonnyPfannschmidt, line 12: those globals have to go, all the data should be available on the message object

app/queue/queue.py
```diff
@@ -290,23 +290,23 @@ def handle_message(message, event_producer, message_operation=add_host):
...
-                        authentication failure occurs or the DB is not accessible.
-                        Exit the process to restart the pod
-                        """
-                        logger.error(f"Could not access DB {str(oe)}")
-                        sys.exit(3)
-                    except Exception:
-                        metrics.ingress_message_handler_failure.inc()
-                        logger.exception("Unable to process message", extra={"incoming_message": message.value})
+            message = consumer.poll(timeout=CONSUMER_POLL_TIMEOUT_SECONDS)
+            if message is None or message.error() is not None:
```
RonnyPfannschmidt, line 294: the error should be raised

thearifismail: Now an error is logged without exiting/stopping the service.

app/config.py
```diff
@@ -180,7 +180,7 @@ def __init__(self, runtime_environment):
             **self.kafka_ssl_configs,
         }
 
-        self.payload_tracker_kafka_producer = {"bootstrap_servers": self.bootstrap_servers, **self.kafka_ssl_configs}
+        self.payload_tracker_kafka_producer = {"bootstrap.servers": self.bootstrap_servers}
```
CodeHeeler, line 183: How is confluent kafka handling SSL now?

CodeHeeler (approved):
LGTM, though I'm not well-versed in kafka.  Had a comment on the SSL handling.  Thanks for taking care of this library swap!

kruai (changes requested):
I haven't reviewed the tests yet; I think there are some things in the implementation that need to be ironed out first

thearifismail (commented):
@asa Fixed SSL config and producer and consumer.  With code update, your original comments related to SSL and consumer config have disappeared.

inv_mq_service.py
```diff
@@ -25,13 +25,13 @@ def main():
...
-        group_id=config.host_ingress_consumer_group,
-        bootstrap_servers=config.bootstrap_servers,
-        api_version=(0, 10, 1),
-        value_deserializer=lambda m: m.decode(),
-        **config.kafka_consumer,
+        {
+            "group.id": config.host_ingress_consumer_group,
+            "bootstrap.servers": config.bootstrap_servers,
+            "auto.offset.reset": "earliest",
+        }
```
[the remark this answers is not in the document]
thearifismail: Added back in but had to use the config for confluent-kafka.  The link provided as a code comment.

app/queue/queue.py
```diff
@@ -290,23 +290,27 @@ def handle_message(message, event_producer, message_operation=add_host):
...
-                        sys.exit(3)
-                    except Exception:
-                        metrics.ingress_message_handler_failure.inc()
-                        logger.exception("Unable to process message", extra={"incoming_message": message.value})
+            message = consumer.poll(timeout=CONSUMER_POLL_TIMEOUT_SECONDS)
+            # if message is None or message.error() is not None:
+            if message is None:
+                continue
+            elif message.error():
+                logger.error(f"Message received but has an error, which is {str(message.error())}")
```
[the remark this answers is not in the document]
kruai: What error was it? We should probably have a bit more handling for it, like incrementing a failure metric at the very least.

app/queue/queue.py
```diff
@@ -290,23 +290,26 @@ def handle_message(message, event_producer, message_operation=add_host):
...
-                        metrics.ingress_message_handler_failure.inc()
-                        logger.exception("Unable to process message", extra={"incoming_message": message.value})
+            message = consumer.poll(timeout=CONSUMER_POLL_TIMEOUT_SECONDS)
+            if message is None:
+                continue
+            elif message.error():
+                logger.error(f"Message received but has an error, which is {str(message.error())}")
+            else:
+                try:
+                    logger.info("A message received.")
```
kruai, line 300: Doesn't make sense; `A message was received` would work, though I think it should probably just stay as `Message received`.

tests/test_api_hosts_delete.py
```diff
@@ -350,13 +350,20 @@ def test_delete_hosts_chunk_size(
...
+    event_producer_mock,
+    event_producer,
+    db_create_multiple_hosts,
+    api_delete_host,
+    db_get_hosts,
+    mocker,
 ):
-    event_producer._kafka_producer.send.side_effect = send_side_effects
+    mocker.patch("lib.host_delete.kafka_available", return_value=False)
+    event_producer._kafka_producer.produce.side_effect = produce_side_effects
```
kruai, line 366: This changes what is being tested. This test wasn't validating that the producer works as expected when kafka isn't available; it was testing that the deletion stops if an exception is raised. As it is, the producer never gets to produce a message, and the side effects don't come into play.

tests/test_api_hosts_delete.py
```diff
@@ -366,8 +373,9 @@ def test_delete_stops_after_kafka_producer_error(
     assert_response_status(response_status, expected_status=500)
 
     remaining_hosts = db_get_hosts(host_id_list)
-    assert remaining_hosts.count() == 2
-    assert event_producer._kafka_producer.send.call_count == 2
+    assert remaining_hosts.count() == 3
+    # since kafka was not available none of the hosts were deleted and no event was generated.
+    assert event_producer._kafka_producer.produce.call_count == 0
```
kruai, line 378: Here you've changed the results of the test so that it passes even though the functionality has changed. The first message does not raise an error, so one host should have been deleted.

tests/test_culling.py
```diff
@@ -337,5 +337,6 @@ def test_reaper_stops_after_kafka_producer_error(
         )
 
     remaining_hosts = db_get_hosts(created_host_ids)
-    assert remaining_hosts.count() == 2
-    assert event_producer._kafka_producer.send.call_count == 2
+    assert remaining_hosts.count() == 3
+    # since kafka was not available none of the hosts were reaped and no event was generated.
+    assert event_producer._kafka_producer.produce.call_count == 0
```
kruai, line 342: You've done the same thing here as above, which invalidates the test. Please fix.

tests/test_mq_service.py
```diff
@@ -36,13 +37,13 @@
 OWNER_ID = SYSTEM_IDENTITY["system"]["cn"]
 
 
-def test_event_loop_exception_handling(mocker, flask_app):
+def test_event_loop(mocker, flask_app):
```
kruai, line 40: No, this still needs to test that it handles exceptions. Please do not keep changing the purpose of unit tests just so that they pass.

tests/test_mq_service.py
```diff
@@ -153,7 +155,7 @@ def test_shutdown_handler(mocker, flask_app):
     )
     fake_consumer.poll.assert_called_once()
 
-    assert handle_message_mock.call_count == 2
+    assert handle_message_mock.call_count == 1
```
kruai, line 158: Same as above. The expected results have been changed to hide that the functionality has changed.

tests/test_unit.py
```diff
@@ -1883,35 +1882,71 @@ def test_happy_path(self):
...
                 self.event_producer.write_event(event, host_id, headers)
 
-                send.assert_called_once_with(
-                    self.topic_name,
-                    key=host_id.encode("utf-8"),
-                    value=event.encode("utf-8"),
-                    headers=expected_encoded_headers(event_type, threadctx.request_id, host_id),
-                )
-                send.reset_mock()
+                produce.assert_called_once_with(self.topic_name, event.encode("utf-8"), callback=ANY)
```
kruai, line 1885: This used to test that they `key` was correct, but that has been removed. The key is a very important part of the message and needs to be tested here.

tests/test_unit.py
```diff
@@ -1883,35 +1882,71 @@ def test_happy_path(self):
...
+                poll.assert_called_once_with()  # calls on callback
```
kruai, line 1886: If you're not testing any arguments, use `poll.assert_called_once()`.

tests/test_unit.py
```diff
@@ -1883,35 +1882,71 @@ def test_happy_path(self):
...
             self.event_producer.write_event(event, key, headers)
 
-        message_not_produced_mock.assert_called_once_with(
-            event_producer_logger,
-            self.config.event_topic,
-            event,
-            key,
-            headers,
-            self.event_producer._kafka_producer.send.side_effect,
-        )
```
kruai, line 1914: This assertion should not be removed.

kruai (changes requested):
Requesting some more changes. Please don't change the output of tests just to make them pass; make sure they still validate the same functionality they were originally built to test.
Also, please make sure you remove classes, fixtures, and such that are no longer used (`MockFuture`, for instance).

app/instrumentation.py
```diff
@@ -10,38 +10,38 @@
...
-    info_message = "Message %s offset=%d timestamp=%d topic=%s, key=%s"
-    logger.info(info_message, status, offset, timestamp, topic, key, extra=info_extra)
+    info_message = f"Message status={status}, offset={offset} timestamp={timestamp} topic=%{topic}, key={key}"
+    logger.info(f"{info_message}, extra={info_extra}")
 
-    debug_message = "Message offset=%d timestamp=%d topic=%s key=%s value=%s"
+    debug_message = f"Message offset={offset} timestamp={timestamp}] topic={topic} key={key} value={value}"
     debug_extra = {**extra, "value": value}
-    logger.debug(debug_message, offset, timestamp, topic, key, value, extra=debug_extra)
+    logger.debug(debug_message, extra=debug_extra)
```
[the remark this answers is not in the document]
thearifismail: No logs don't change.   I used to use formatted strings but Kyle Lape had adamantly insisted to use `f-string` for being easier than anything else available.

tests/test_mq_service.py
kruai: This still hasn't been updated. I think it's best to keep testing what was originally being tested, which is that multiple messages are processed before the shutdown signal is received. `call_count` should remain at 2, but the interrupt should have one more `False` added to its side effects

tests/test_unit.py
kruai: In the old test, the assertion validated that the message was being produced with the correct key (`host_id`). The new assertion does not test that, but it needs to. Does that help clarify?

app/queue/event_producer.py
```diff
@@ -1,37 +1,65 @@
...
+
+    def __init__(self, topic, event, headers, key):
+        self.event = event
+        self.headers = headers
+        self.key = key
+        self.topic = topic
+
+    def send(self, producer):
+        producer.produce(self.topic, self.event, callback=self.on_delivered)
+        producer.poll()
```
kruai, line 26: I didn't realize this before, but why is this function in MessageDetails instead of the producer? Having it here makes it so you have to call `MessageDetails.send(producer)`, which makes it seem like you're telling the message to send the producer. `Producer.send(messageDetails)` makes more sense to me...

tests/test_unit.py
kruai: This assertion is still missing.

tests/test_api_hosts_delete.py
```diff
@@ -350,24 +350,31 @@ def test_delete_hosts_chunk_size(
 
 
 @pytest.mark.parametrize(
-    "send_side_effects",
-    ((mock.Mock(), mock.Mock(**{"get.side_effect": KafkaError()})), (mock.Mock(), KafkaError("oops"))),
+    "produce_side_effects",
+    (
+        (mock.Mock(), KafkaException()),
+        (mock.Mock(), KafkaException("oops")),
+    ),
```
kruai, line 357: This used to validate that the deletion would stop after either of these scenarios:
1. The `future` returned by `send()` raises an exception
2. The `send()` itself raises an exception

Now, this function only tests scenario 2 twice. In order to test both, you'll have to also test the scenario where `on_delivered()` receives an error.

tests/test_culling.py
```diff
@@ -304,15 +304,13 @@ def assert_system_culling_data(response_host, expected_stale_timestamp, expected
 
 @pytest.mark.host_reaper
 @pytest.mark.parametrize(
-    "send_side_effects",
-    ((mock.Mock(), mock.Mock(**{"get.side_effect": KafkaError()})), (mock.Mock(), KafkaError("oops"))),
+    "produce_side_effects",
+    ((mock.Mock(), KafkaException()), (mock.Mock(), KafkaException("oops"))),
```
kruai, line 308: Same issue as what I commented in `test_delete_stops_after_kafka_producer_error`

kruai (changes requested):

app/queue/event_producer.py
```diff
@@ -1,37 +1,65 @@
...
 from app.instrumentation import message_produced
 from app.logging import get_logger
 
+
 logger = get_logger(__name__)
 
 
+class MessageDetails:
+    event: str
+    headers: str
```
kruai, line 14: Are you sure this is a `str`? In `write_event()`, it's an array of tuples.

tests/test_mq_service.py
thearifismail: Fixed.

tests/test_unit.py
thearifismail: A host ID IS passed in to the `producer.produce()` with the host info in the event.  In `kafka-python`, a `host_id` was sent as an argument to `producer.send()` despite being present in the event.  

Here is the event json with "host.id" that is passed in as an argument to `produce.assert_called_once_with(<topic>, <event>, <callback>)`
```
{
  "type": "created",
  "host": {
    "reporter": "test_reporter",
    "stale_timestamp": "2022-08-25T12:19:11.169401+00:00",
    "account": "test",
    "org_id": "test",
    "fqdn": "fqdn",
    "id": "fe9a8aaf-dfc0-4ecd-bb55-de31f489d182"
  },
  "metadata": {
    "request_id": "e8123e45-80ae-4763-9150-e2e8b63535fa"
  },
  "timestamp": "2022-08-25T12:19:11.260830+00:00",
  "platform_metadata": null
}
```

thearifismail (commented):
@asa, this PR is ready review.

kruai (changes requested):

PR #1184, opened 2022-08-30
Standardize missing request_id values

kruai:
# Overview

This PR is being created to address [ESSNTL-566](https://issues.redhat.com/browse/ESSNTL-566).
This consolidates all uses of missing / unknown request_id. There is now only one constant variable that can be referenced, and the value has been standardized across the board to be None. I [verified with the payload-tracker team](https://coreos.slack.com/archives/C022YV4E0NA/p1661867070902129) that this change is OK and should not break anything on their end.

I also checked cyndi-operator, xjoin-operator, and xjoin-search to make sure they would not be affected by this change. From what I can tell, they don't care about the request ID at all, so there should be no side effects.

I've also created an MR to update our docs with this info: https://gitlab.cee.redhat.com/consoledot/consoledot.pages.redhat.com/-/merge_requests/78

## PR Checklist

- [ ] Tests: edge cases
- [x] Documentation, if this PR changes the way other services interact with host inventory
- [x] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/payload_tracker/__init__.py
```diff
@@ -11,7 +11,7 @@
 
 _CFG = None
 _PRODUCER = None
-_UNKNOWN_REQUEST_ID = "-1"
+UNKNOWN_REQUEST_ID_VALUE = None
```
CodeHeeler, line 14: The standardization is a good step, but can we take it a step further at this stage and assume that this is not needed and we can just state "None" in the cases below that would be calling this value?   I generally avoid declaring these unless it's for a value we expect to change, like a Kafka topic, so that we only will need to change it in one place.

CodeHeeler (approved):
LGTM, left a non-blocking comment discussing going even further, but this is ready to go.

app/payload_tracker/__init__.py
kruai: Honestly, that's a good call. I think that will simplify it further. I'll update a few things

kruai: Done! Let me know if I missed any @CodeHeeler

CodeHeeler (approved):
Compared to grepping locally, pretty sure you got all the instances.  Thanks!

CodeHeeler (approved):
LGTM, thanks!

thearifismail (approved):
LGTM

PR #1186, opened 2022-09-02
[ESSNTL-629] Throw 400 sparse field not in schema

CodeHeeler:
# Overview

This PR is being created to address [ESSNTL-629](https://issues.redhat.com/browse/ESSNTL-629).
This PR adds the ability to throw a 400 if a request for a system profile sparse field is a field that is not in our schema for the system profile.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/sparse_host_list_system_profile.py
+++ b/api/sparse_host_list_system_profile.py
@@ -2,7 +2,7 @@ import flask
 from flask_api import status
 
 from api.host_query_xjoin import owner_id_filter
-from app.__init__ import process_system_profile_spec
+from api.validate_fields import validate_fields_in_schema
 from app.auth import get_current_identity
 from app.auth.identity import IdentityType
 from app.instrumentation import log_get_sparse_system_profile_failed
@@ -97,11 +97,8 @@ def get_sparse_system_profile(host_id_list, page, per_page, order_by, order_how,
     }
 
     if fields.get("system_profile"):
+        validate_fields_in_schema(fields)
         variables["fields"] = list(fields.get("system_profile").keys())
-        system_profile_schema = process_system_profile_spec()
-        for field in variables["fields"]:
-            if field not in system_profile_schema.keys():
-                flask.abort(400, "Requested system_profile field is not present in the schema.")
         sp_query = SYSTEM_PROFILE_SPARSE_QUERY
 
     response = graphql_query(sp_query, variables, log_get_sparse_system_profile_failed)
--- a/api/sparse_host_list_system_profile.py
+++ b/api/sparse_host_list_system_profile.py
@@ -2,6 +2,7 @@ import flask
 from flask_api import status
 
 from api.host_query_xjoin import owner_id_filter
+from app.__init__ import process_system_profile_spec
 from app.auth import get_current_identity
 from app.auth.identity import IdentityType
 from app.instrumentation import log_get_sparse_system_profile_failed
@@ -97,6 +98,10 @@ def get_sparse_system_profile(host_id_list, page, per_page, order_by, order_how,
 
     if fields.get("system_profile"):
         variables["fields"] = list(fields.get("system_profile").keys())
+        system_profile_schema = process_system_profile_spec()
+        for field in variables["fields"]:
+            if field not in system_profile_schema.keys():
+                flask.abort(400, "Requested system_profile field is not present in the schema.")
         sp_query = SYSTEM_PROFILE_SPARSE_QUERY
 
     response = graphql_query(sp_query, variables, log_get_sparse_system_profile_failed)
--- a/api/validate_fields.py
+++ b/api/validate_fields.py
@@ -1,12 +1,12 @@
 import flask
 
-from app.__init__ import process_system_profile_spec
+from app import system_profile_spec
 
 
 def validate_fields_in_schema(fields):
     if fields.get("system_profile"):
         query_fields = list(fields.get("system_profile").keys())
-        system_profile_schema = process_system_profile_spec()
+        system_profile_schema = system_profile_spec()
         for field in query_fields:
             if field not in system_profile_schema.keys():
                 flask.abort(400, f"Requested field '{field}' is not present in the system_profile schema.")
--- a/api/validate_fields.py
+++ b/api/validate_fields.py
@@ -1,0 +1,12 @@
+import flask
+
+from app.__init__ import process_system_profile_spec
+
+
+def validate_fields_in_schema(fields):
+    if fields.get("system_profile"):
+        query_fields = list(fields.get("system_profile").keys())
+        system_profile_schema = process_system_profile_spec()
+        for field in query_fields:
+            if field not in system_profile_schema.keys():
+                flask.abort(400, f"Requested field '{field}' is not present in the system_profile schema.")
--- a/api/host.py
+++ b/api/host.py
@@ -18,6 +18,7 @@ from api.host_query_xjoin import get_host_list as get_host_list_xjoin
 from api.host_query_xjoin import get_host_list_by_id_list
 from api.host_query_xjoin import get_host_tags_list_by_id_list
 from api.sparse_host_list_system_profile import get_sparse_system_profile
+from api.validate_fields import validate_fields_in_schema
 from app import db
 from app import inventory_config
 from app import Permission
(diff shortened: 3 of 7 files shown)


CodeHeeler:
Wanted to get this opened as a draft PR.  I'm having some tests complain about "ValueError: I/O operation on closed file" that may be related to the method for processing the schema in order to obtain the keys for the fields to compare against, but I'm not sure how to address this.

In the meanwhile, I'll look through our tests to see if I need to add coverage to test incorrect fields like these.

kruai (commented):
This looks like a nice, clean way to do this; I wonder what's causing that IO error. But yes, it would also be great to have a test or two for this.

CodeHeeler:
Looks like the I/O Error is part of the ongoing errors we see after running all tests, see this comment in our Jira: https://issues.redhat.com/browse/ESSNTL-1915?focusedCommentId=20896632&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20896632

and the relevant pytest issue:
https://github.com/pytest-dev/pytest/issues/5502

CodeHeeler:
Added tests, ready for review.

api/sparse_host_list_system_profile.py
```diff
@@ -97,6 +98,10 @@ def get_sparse_system_profile(host_id_list, page, per_page, order_by, order_how,
 
     if fields.get("system_profile"):
         variables["fields"] = list(fields.get("system_profile").keys())
+        system_profile_schema = process_system_profile_spec()
+        for field in variables["fields"]:
+            if field not in system_profile_schema.keys():
+                flask.abort(400, "Requested system_profile field is not present in the schema.")
```
kruai, line 104: A bit more detail could be useful here. Something like this:
```suggestion
                flask.abort(400, f"Requested field '{field}' is not present in the system_profile schema.")
```

kruai (changes requested):
This looks good, but I'm realizing that the scope of the issue may be larger than just `get_sparse_system_profile`. There are other endpoints that allow sparse fields, such as `GET /hosts` and `GET /hosts/{host_ids}`. You may need to refactor the validation logic into a function that can be used in each of those endpoint

CodeHeeler:
> This looks good, but I'm realizing that the scope of the issue may be larger than just `get_sparse_system_profile`. There are other endpoints that allow sparse fields, such as `GET /hosts` and `GET /hosts/{host_ids}`. You may need to refactor the validation logic into a function that can be used in each of those endpoint

Ok, I'll work on a refactor.  Thanks for the suggestion!

CodeHeeler:
Refactored, and updated other tests and utils to include real fields like "arch" and "operating_system" instead of fake placeholders that were there like "test_data" and "random".

There is one test still failing which I have yet to figure out how to resolve.
`_______________________________________ test_system_profile_includes_owner_id ________________________________________

api_get = <function api_get.<locals>._api_get at 0x7f62ce017310>
patch_xjoin_post = <function patch_xjoin_post.<locals>._patch_xjoin_post at 0x7f62ce2045e0>

    def test_system_profile_includes_owner_id(api_get, patch_xjoin_post):
        host_id = generate_uuid()
        response = xjoin_host_response("2021-02-10T08:07:03Z")
        response["hosts"]["data"][0]["id"] = host_id
        patch_xjoin_post({"data": response})
    
        url = build_system_profile_url(host_list_or_id=host_id)
    
        response_status, response_data = api_get(url)
    
>       assert "owner_id" in response_data["results"][0]["system_profile"]
E       KeyError: 'results'

tests/test_system_profile.py:44: KeyError
`
Note, I did not later the order or number of fields anywhere, but for some reason the key error is saying the value is "results", and I'm unsure how this could happen.  Will dig into it fresh tomorrow or see if someone else sees something I'm missing.

CodeHeeler:
FabriciaDinezRH helped me narrow down where to look and I figured out there was other schema validation going on where the fields I'd chosen to be real in the schema didn't have those data types that we were looking to test; adjusted to another field, tests pass.

Ready for review.

kruai (approved):
LGTM! 
There's one thing I'm curious about, and that's whether we could/should run `validate_fields_in_schema` from the `api_operation` decorator, or somewhere like that... like if a `fields` argument is used in the function, it runs the validation check on the value. This shouldn't block this being merged, though, especially since it requires more investigation into decorators and such :) 

Edit: We do have a [custom parser](https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/parsing.py#L32) for the `fields` param as well. Wonder if that would be a decent place to validate. I'll poke around a bit more

api/validate_fields.py
```diff
@@ -0,0 +1,12 @@
+import flask
+
+from app.__init__ import process_system_profile_spec
+
+
+def validate_fields_in_schema(fields):
+    if fields.get("system_profile"):
+        query_fields = list(fields.get("system_profile").keys())
+        system_profile_schema = process_system_profile_spec()
```
kruai, line 9: Oh, I just noticed that this runs the whole processing function, even though it's already done in the app setup. This should instead be:
```suggestion
from app import system_profile_spec


def validate_fields_in_schema(fields):
    if fields.get("system_profile"):
        query_fields = list(fields.get("system_profile").keys())
        system_profile_schema = system_profile_spec()
```

kruai (changes requested):
oops, I actually noticed one thing while I was digging around.

api/sparse_host_list_system_profile.py
```diff
@@ -2,6 +2,7 @@
 from flask_api import status
 
 from api.host_query_xjoin import owner_id_filter
+from app.__init__ import process_system_profile_spec
```
thearifismail: importing `__init__`, shouldn't it be just `from app import process_system_profile_spec`?

thearifismail (commented):
One comment about importing __init__

CodeHeeler:
> LGTM! There's one thing I'm curious about, and that's whether we could/should run `validate_fields_in_schema` from the `api_operation` decorator, or somewhere like that... like if a `fields` argument is used in the function, it runs the validation check on the value. This shouldn't block this being merged, though, especially since it requires more investigation into decorators and such :)
> 
> Edit: We do have a [custom parser](https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/parsing.py#L32) for the `fields` param as well. Wonder if that would be a decent place to validate. I'll poke around a bit more

Agreed, I'm not super familiar with decorators, but if we could automatically check fields when present, that would likely be a cleaner solution in the long term.

CodeHeeler:
Adjusted import and went with system_profile_spec instead of process_system_profile_spec, as suggested.

@kruai I opened another ticket to track looking into custom parser or decorator options: https://issues.redhat.com/browse/ESSNTL-3388

thearifismail:
> > LGTM! There's one thing I'm curious about, and that's whether we could/should run `validate_fields_in_schema` from the `api_operation` decorator, or somewhere like that... like if a `fields` argument is used in the function, it runs the validation check on the value. This shouldn't block this being merged, though, especially since it requires more investigation into decorators and such :)
> > Edit: We do have a [custom parser](https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/parsing.py#L32) for the `fields` param as well. Wonder if that would be a decent place to validate. I'll poke around a bit more
> 
> Agreed, I'm not super familiar with decorators, but if we could automatically check fields when present, that would likely be a cleaner solution in the long term.

I did not see this comment from Asa ... I like the current implementation for being clear.  It would make sense to use decorator for fields validation if other parameters for getters were being validated using decorators.

thearifismail:
Talking about clarity, created [ESSNTL-3389](https://issues.redhat.com/browse/ESSNTL-3389)

kruai (approved):
Good points! I might grab that investigation ticket just because I've gone down a curiosity rabbit hole with it. I'm cool with this in its current state though

thearifismail (approved):

PR #1193, opened 2022-09-09
ESSNTL-1878: Replace kafka-python with confluent-kafka

thearifismail:
# Overview

This PR is being created to address [ESSNTL-1878](https://issues.redhat.com/browse/ESSNTL-1878).

This PR is continuation of [PR #1173](https://github.com/RedHatInsights/insights-host-inventory/pull/1173), which has been closed.  For background, please look at that PR first.  This PR has been created to resolve conflicts cause by multiple merges since the creation of PR #1173.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


lib/host_delete.py
```diff
@@ -39,15 +39,15 @@ def _delete_host(session, event_producer, host):
             event = build_event(EventType.delete, host)
             insights_id = host.canonical_facts.get("insights_id")
             headers = message_headers(EventType.delete, insights_id)
-            event_producer.write_event(event, str(host.id), headers, wait=True)
```
CodeHeeler, line 42: If we're not waiting for these, is there are retry later for events that get rolled back?

lib/host_synchronize.py
```diff
@@ -36,8 +36,8 @@ def synchronize_hosts(select_query, event_producer, chunk_size, config, interrup
         try:
             # pace the events production speed as flush completes sending all buffered records.
             event_producer._kafka_producer.flush(300)
-        except KafkaTimeoutError:
-            raise KafkaTimeoutError(f"KafkaTimeoutError: failure to flush {chunk_size} records within 300 seconds")
+        except ProduceError:
+            raise ProduceError(f"ProduceError: failure to flush {chunk_size} records within 300 seconds")
```
CodeHeeler, line 40: Should we mention Kafka in the text of this message? "ProduceError" isn't as clear on the source as "KafkaTimeoutError" was.

CodeHeeler: Also, I didn't see any expected messages for errors in tests updated; is MagicMock handling this or did we not have any tests checking the exact text of the error for comparison?

CodeHeeler (dismissed):
Overall LGTM though you should probably have someone more versed in Kafka give it a review as well.  Had some questions, but nothing I saw as a blocker.  Thanks for making these changes, especially all the fiddly details like changing underscore to periods for everything.

app/config.py
```diff
@@ -125,73 +125,67 @@ def __init__(self, runtime_environment):
...
+            "enable.auto.commit": False,
+            "max.poll.interval.ms": int(os.environ.get("KAFKA_CONSUMER_MAX_POLL_INTERVAL_MS", "300000")),
+            "session.timeout.ms": int(os.environ.get("KAFKA_CONSUMER_SESSION_TIMEOUT_MS", "10000")),
+            "heartbeat.interval.ms": int(os.environ.get("KAFKA_CONSUMER_HEARTBEAT_INTERVAL_MS", "3000")),
             **self.kafka_ssl_configs,
         }
 
         self.events_kafka_consumer = {
             "group_id": "inventory-events-rebuild",
-            "auto_offset_reset": "earliest",
```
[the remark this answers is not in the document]
thearifismail: Put back in, `"auto.offset.reset": "earliest",`

lib/host_delete.py
thearifismail: This implementation is different and wait option is not available but `timeout` is used when consuming messages.  Timeout of one second is used when polling for messages.

lib/host_synchronize.py
thearifismail: Now the produceError prints `raise ProduceError(f"ProduceError: Kafka failure to flush {chunk_size} records within 300 seconds")`.

Confluent-kafka provides only an Error code but no further details.  That's why `event_producer` module contains `MessageDetails`, which provides details about the host which encountered problems.

app/instrumentation.py
```diff
@@ -12,44 +12,45 @@
...
+    topic = message.topic()
+
+    timestamp = message_dict["timestamp"]
+
     extra = {"status": status, "offset": offset, "timestamp": timestamp, "topic": topic, "key": key}
 
     info_extra = {**extra, "headers": headers}
-    info_message = "Message %s offset=%d timestamp=%d topic=%s, key=%s"
-    logger.info(info_message, status, offset, timestamp, topic, key, extra=info_extra)
+    info_message = f"Message status={status}, offset={offset} timestamp={timestamp} topic=%{topic}, key={key}"
```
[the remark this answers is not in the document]
CodeHeeler: Good eye!

lib/host_delete.py
kruai: I may be misunderstanding, but isn't that what `Producer.flush()` does? [The docs](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Producer.flush) say that it makes it wait until the messages in the queue are delivered; would that not allow us to implement a sync producer in cases where we should wait (bulk deletions, etc.)?

tests/helpers/mq_utils.py
```diff
@@ -25,12 +26,30 @@ def __init__(self):
...
         self.event = event
         self.key = key
         self.headers = headers
-        self.wait = wait
+
+        self._message_details.topic = "platform.inventory.events"
+        self._message_details.event = event
+        self._message_details.key = key
+        self._message_details.headers = headers
+        self._message_details.send = MagicMock()
```
kruai, line 40: `MessageDetails.send()` is no longer a function, and should not be mocked

app/queue/event_producer.py
```diff
@@ -14,30 +16,46 @@ def _encode_headers(headers):
...
+    def on_delivered(self, error, message, msgdet):
+        if error:
+            message_not_produced(logger, error, self.topic, self.event, msgdet.key, msgdet.headers)
+        else:
+            message_produced(logger, message, msgdet.key, msgdet.headers)
+
+
 class EventProducer:
-    def __init__(self, config, topic):
+    def __init__(self, config, topic=None):
```
kruai, line 34: What's the reasoning behind making `topic` optional?

tests/helpers/mq_utils.py
```diff
@@ -25,12 +26,30 @@ def __init__(self):
         self.wait = None
         self._kafka_producer = Mock()
         self._kafka_producer.flush = Mock(return_value=True)
+        self._message_details = Mock()
```
kruai, line 29: This seems strange to mock, just because the real EventProducer doesn't have a local MessageDetails field. Is this ever used?

tests/helpers/mq_utils.py
```diff
@@ -25,12 +26,30 @@ def __init__(self):
...
+    def write_event(self, event, key, headers):
         self.event = event
         self.key = key
         self.headers = headers
-        self.wait = wait
+
+        self._message_details.topic = "platform.inventory.events"
+        self._message_details.event = event
+        self._message_details.key = key
+        self._message_details.headers = headers
```
kruai, line 39: Are these ever used? I don't see any references to `_message_details`

tests/test_api_hosts_delete.py
```diff
@@ -352,25 +354,49 @@ def test_delete_hosts_chunk_size(
...
+    msgdet = MessageDetails(topic=None, event=message, headers=headers, key=host.id)
+    event_producer._kafka_producer.produce.side_effects = msgdet.on_delivered(error, message, msgdet)
+
+    response_status, response_data = api_delete_host(",".join([str(host.id)]))
+
+    assert_response_status(response_status, expected_status=200)
+
+    remaining_hosts = db_get_hosts([host.id])
+
+    assert remaining_hosts.count() == 0
```
kruai, line 397: I said this on #1173, but... If this test is meant to cover the scenario where the deletion stops after the callback returns an error, then it needs to test that. Similar to `test_delete_stops_after_kafka_exception`, it needs to have some hosts remaining after deletion. Otherwise you're not testing that the deletion stopped

tests/test_mq_service.py
```diff
@@ -159,9 +179,9 @@ def test_shutdown_handler(mocker, flask_app):
         handler=handle_message_mock,
         interrupt=mocker.Mock(side_effect=(False, True)),
     )
-    fake_consumer.poll.assert_called_once()
+    fake_consumer.consume.assert_called_once()
 
-    assert handle_message_mock.call_count == 2
+    assert handle_message_mock.call_count == 1
```
kruai, line 184: We should keep testing what was originally being tested, which is that multiple messages are processed before the shutdown signal is received. This can be done by either:
- Adding one more `False` to the interrupt's side effects and leaving the consumer mock as-is
- Changing the consumer's side effect to be one array of 2 messages (like in the original), and leaving the interrupt as-is

tests/test_unit.py
```diff
@@ -1901,34 +1901,35 @@ def test_happy_path(self):
...
 
-                send.assert_called_once_with(
-                    self.topic_name,
-                    key=host_id.encode("utf-8"),
-                    value=event.encode("utf-8"),
-                    headers=expected_encoded_headers(event_type, threadctx.request_id, host_id),
-                )
-                send.reset_mock()
+                produce.assert_called_once_with(self.topic_name, event.encode("utf-8"), callback=ANY)
+                poll.assert_called_once()  # calls on callback
```
kruai, line 1905: Just a note that `poll()` is called immediately, not on callback. It does trigger callbacks for any newly-completed messages though

kruai (changes requested):

lib/host_delete.py
thearifismail: Thanks for pointing out `flush()`.  Replaced `poll()` with `flush()` because flush communicates the intent more appropriately than poll().  Also flush() make the call synchronous

thearifismail (commented):
All comments addressed, except:
https://github.com/RedHatInsights/insights-host-inventory/pull/1193/files#r969775333
https://github.com/RedHatInsights/insights-host-inventory/pull/1193/files#r969785237

tests/test_api_hosts_delete.py
thearifismail: IMO, event production should NOT stop producing an event fails for whatever reason.  It is important to note `event_producer.write_event()` is called after verifying that Kafka is available.  An event_production may fail due to a KafkaError caused by any one of these [event errors](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#kafkaerror).

Say 10 hosts are being deleted.  Before deleting a host, `kafka_available()` succeeds, then the host is deleted. Say the deletion of first two hosts succeeds but for the 3rd one Kafka is available but `EventProducer.writ_event()` for some reason.  IMO, attempts should be made to delete all subsequent hosts and hosts deletion should NOT stop upon encountering the first failure.

CodeHeeler: That list of possible event errors is quite wide and I would think different ones would be handled differently, like retry or key serializer errors vs queue full or time outs.  What is the standard practice for bulk delete operations?  Try all in the list and return which ones failed or error out and rollback any from this batch that had "completed" already or something in the middle?

That being said, I don't think the review comment was on how we implemented our bulk delete so much as which scenarios this test is meant to check.  Maybe add another case where there are hosts left to delete and then assert that case completes as you expect it to.

thearifismail: For this implementation, specific errors after successful `kafka_available()` checks will be beyond HBI control.  HBI stops immediately when Kafka is not available.  Question is should it stop, when an error occurs when delivering a message.  I am voting for continuing because the error with this specific event/host and not all.

FabriciaDinizRH: In my head it makes sense to continue with the deletion of the remaining hosts, but perhaps this is a matter of creating a new task to tackle the issue if the team decides that it makes sense as well.

kruai: I think we had originally made it stop upon Kafka delivery errors, because it implied that there was an issue communicating with the brokers, and each failure makes Cyndi get more and more out of sync with HBI.
While `kafka_available()` does verify that there is no network issue connecting to the bootstrap server (it checks that `socket.connect_ex` doesn't error), I'm concerned about other kinds of Kafka outages or configuration issues. For instance, I don't think `connect_ex` catches the NoBrokersAvailable error we see pretty frequently, because that error only happens in response to a message being produced (it would only show in the callback); I also don't think it would catch any of the other errors that would be raised from broker-side issues, since those are only sent in the callbacks.

I personally think we should stop after kafka errors the way we currently do. But if we do decide to make it keep trying to delete hosts after a callback error, we need to make it so it behaves the same way when it hits a KafkaException (tested by test_delete_stops_after_kafka_exception).

Apologies for the long post, but lmk what you think @CodeHeeler @thearifismail @FabriciaDinizRH

tests/test_unit.py
```diff
@@ -1899,36 +1899,59 @@ def test_happy_path(self):
                 event = build_event(event_type, host)
                 headers = message_headers(event_type, host_id)
 
-                self.event_producer.write_event(event, host_id, headers)
+                self.event_producer.write_event(event, host_id, headers, wait=True)
```
kruai, line 1902: This is testing single-host operations that do not use `wait=True`
```suggestion
                self.event_producer.write_event(event, host_id, headers)
```

tests/test_unit.py
```diff
@@ -1899,36 +1899,59 @@ def test_happy_path(self):
...
+        for (event_type, host) in (
+            (EventType.created, self.basic_host),
+            (EventType.updated, self.basic_host),
+            (EventType.delete, deserialize_host(self.basic_host)),
+        ):
+            with self.subTest(event_type=event_type):
+                event = build_event(event_type, host)
+                headers = message_headers(event_type, host_id)
+
+                self.event_producer.write_event(event, host_id, headers, wait=False)
```
kruai, line 1924: `wait=False` is the default
```suggestion
                self.event_producer.write_event(event, host_id, headers)
```

kruai (changes requested):

tests/test_api_hosts_delete.py
CodeHeeler: I think at this point, I'd like to stick with the current way of stopping deletion upon error so there's no change in expected behaviour for customers, and like Fabricia mentioned create a new Jira story to capture this discussion, decide, and implement on a separate PR.  We could get input from architects or others on preferred methodology for bulk delete behaviours there.

CodeHeeler (dismissed):
LGTM, thanks for making the suggested changes. Hopefully iqe will be fixed and build won't be unstable soon.

tests/test_unit.py
kruai: This suggestion still stands. `wait=True` is used in bulk operations (like our bulk deletion endpoint), but this test is validating single-host operations.

kruai (changes requested):
There are still suggestions/comments from my previous review that have gone unaddressed. I've commented on those threads again to make them more visible

tests/test_api_hosts_delete.py
thearifismail: For the execution to stop upon seeing first `message_not_produced(...)` error, add `raise Exception(error)` after the `message_not_produced` call in `event_producer.MessageDetails.on_delivered()`

app/queue/event_producer.py
```diff
@@ -23,15 +21,15 @@ def __init__(self, topic: str, event: str, headers: list(tuple()), key: str):
...
-            message_produced(logger, message, msgdet.key, msgdet.headers)
+            message_produced(logger, message, self.headers)
 
 
 class EventProducer:
-    def __init__(self, config, topic=None):
+    def __init__(self, config, topic):
         logger.info("Starting EventProducer()")
         self._kafka_producer = KafkaProducer({"bootstrap.servers": config.bootstrap_servers, **config.kafka_producer})
         self.egress_topic = topic if topic else config.event_topic
```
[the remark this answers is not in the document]
kruai: @thearifismail this still hasn't been updated

kruai (dismissed):
All of my concerns have been addressed, except for that one test. However, after drowning in mocking hell for a day, I'm not sure how best to test that. My main concern now is the IQE failures. @fstavela, what's your opinion on those?

fstavela:
> All of my concerns have been addressed, except for that one test. However, after drowning in mocking hell for a day, I'm not sure how best to test that. My main concern now is the IQE failures. @fstavela, what's your opinion on those?

As I mentioned [here](https://ansible.slack.com/archives/CQFKM031T/p1664310956860469), it looks like all API DELETE requests are resulting in error 500. Let me take a better look.

fstavela:
> > All of my concerns have been addressed, except for that one test. However, after drowning in mocking hell for a day, I'm not sure how best to test that. My main concern now is the IQE failures. @fstavela, what's your opinion on those?
> 
> As I mentioned [here](https://ansible.slack.com/archives/CQFKM031T/p1664310956860469), it looks like all API DELETE requests are resulting in error 500. Let me take a better look.

Yes, it really does that:
```
$ curl -H "$IDENTITY" -X "DELETE" "http://host-inventory-service.ephemeral-uswgsa.svc:8000/api/inventory/v1/hosts/fe2f3141-6e73-4bec-8962-e241cb336d3b" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   251  100   251    0     0  13210      0 --:--:-- --:--:-- --:--:-- 13210
{
  "detail": "The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.",
  "status": 500,
  "title": "Internal Server Error",
  "type": "about:blank"
}
```
Here are Inventory logs:
```
[2022-10-03 07:32:51,544] [8] [139961934350080] [inventory.app.queue.event_producer] [ERROR] Message status=NOT PRODUCED, topic=platform.inventory.events, key=b'fe2f3141-6e73-4bec-8962-e241cb336d3b', headers=[('event_type', b'delete'), ('request_id', b''), ('producer', b'host-inventory-service-7d6b48c9dc-b27zc'), ('insights_id', b'4fd5aefb-1aef-456e-b4b3-f655ad40e9f5')], error='host', event=b'{"timestamp": "2022-10-03T07:32:51.543239+00:00", "org_id": "3340851", "insights_id": "4fd5aefb-1aef-456e-b4b3-f655ad40e9f5", "request_id": null, "account": "0369233", "id": "fe2f3141-6e73-4bec-8962-e241cb336d3b", "metadata": {"request_id": null}, "type": "delete"}'
[2022-10-03 07:32:51,545] [8] [139961934350080] [inventory] [ERROR] Exception on /api/inventory/v1/hosts/fe2f3141-6e73-4bec-8962-e241cb336d3b [DELETE]
File "/opt/app-root/lib64/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
response = self.full_dispatch_request()
File "/opt/app-root/lib64/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/opt/app-root/lib64/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
reraise(exc_type, exc_value, tb)
File "/opt/app-root/lib64/python3.8/site-packages/flask/_compat.py", line 39, in reraise
raise value
File "/opt/app-root/lib64/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
rv = self.dispatch_request()
File "/opt/app-root/lib64/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
return self.view_functions[rule.endpoint](**req.view_args)
File "/opt/app-root/lib64/python3.8/site-packages/connexion/decorators/decorator.py", line 68, in wrapper
response = function(request)
File "/opt/app-root/lib64/python3.8/site-packages/connexion/security/security_handler_factory.py", line 388, in wrapper
return function(request)
File "/opt/app-root/lib64/python3.8/site-packages/connexion/decorators/uri_parsing.py", line 149, in wrapper
response = function(request)
File "/opt/app-root/lib64/python3.8/site-packages/connexion/decorators/validation.py", line 399, in wrapper
return function(request)
File "/opt/app-root/lib64/python3.8/site-packages/connexion/decorators/response.py", line 112, in wrapper
response = function(request)
File "/opt/app-root/lib64/python3.8/site-packages/connexion/decorators/parameter.py", line 120, in wrapper
return function(**kwargs)
File "/opt/app-root/src/api/__init__.py", line 37, in new_func
results = old_func(*args, **kwargs)
File "/opt/app-root/src/lib/middleware.py", line 72, in modified_func
return func(*args, **kwargs)
File "<decorator-gen-10>", line 2, in delete_by_id
File "/opt/app-root/lib64/python3.8/site-packages/prometheus_client/context_managers.py", line 66, in wrapped
return func(*args, **kwargs)
File "/opt/app-root/src/api/host.py", line 244, in delete_by_id
delete_count = _delete_host_list(host_id_list)
File "/opt/app-root/src/api/host.py", line 192, in _delete_host_list
for host_id, deleted in delete_hosts(
File "/opt/app-root/src/lib/host_delete.py", line 24, in delete_hosts
host_deleted = _delete_host(select_query.session, event_producer, host)
File "/opt/app-root/src/lib/host_delete.py", line 42, in _delete_host
event_producer.write_event(event, str(host.id), headers, wait=True)
File "/opt/app-root/src/app/queue/event_producer.py", line 57, in write_event
raise error
File "/opt/app-root/src/app/queue/event_producer.py", line 49, in write_event
self._kafka_producer.flush()
File "/opt/app-root/src/app/queue/event_producer.py", line 28, in on_delivered
message_produced(logger, message, self.headers)
File "/opt/app-root/src/app/instrumentation.py", line 18, in message_produced
key = message_dict["host"]["id"]
KeyError: 'host'
```
I hope this helps

thearifismail:
@fstavela The problem is fixed now, which was related to delete host event which had "id" from the API request but not the host that is available for create and update events

fstavela:
@thearifismail thanks, yes, the issue with deleting hosts has been fixed. There is one more issue remaining and that is that produced Kafka messages, both events and notifications are missing "headers"

thearifismail:
> @thearifismail thanks, yes, the issue with deleting hosts has been fixed. There is one more issue remaining and that is that produced Kafka messages, both events and notifications are missing "headers"

@fstavela I added `headers` as an optional parameter to `Producer.produce()` call and see if it adds the missing headers.

fstavela:
> > @thearifismail thanks, yes, the issue with deleting hosts has been fixed. There is one more issue remaining and that is that produced Kafka messages, both events and notifications are missing "headers"
> 
> @fstavela I added `headers` as an optional parameter to `Producer.produce()` call and see if it adds the missing headers.

It helped, it looks like that problem is fixed.
This is a big change, so I'm now running full IQE test suite on this. Please don't merge until I finish testing.

thearifismail:
@fstavela The events should have key now.  Are there any tests which have not been run yet?

fstavela:
From the functionality point of view, everything works as expected. But I'm concerned about the performance. Running all of our tests on Stage image takes about 1:40h, but on this PR it took about 3:15h, so the time is almost doubled. I don't have any more specific information yet, but it looks like the processing and production of Kafka messages are significantly slower with this PR.

@RonnyPfannschmidt You have more experience with confluent-kafka than me, would you please take a look at this PR and try to find what can cause this performance drop?

thearifismail:
hmmm, confluence-kafka is supposed to have better performance than Python
Kafka as shown by
https://towardsdatascience.com/3-libraries-you-should-know-to-master-apache-kafka-in-python-c95fdf8700f2
.

I wonder if something else is going on.
- Arif

tests/test_unit.py
```diff
@@ -1923,7 +1929,14 @@ def test_producer_poll(self):
 
                 self.event_producer.write_event(event, host_id, headers)
 
-                produce.assert_called_once_with(self.topic_name, event.encode("utf-8"), callback=ANY)
+                produce.assert_called_once_with(
+                    self.topic_name,
+                    event.encode("utf-8"),
+                    host_id.encode("utf-8"),
+                    callback=ANY,
+                    # timestamp=ANY,
```
kruai, line 1937: This should be deleted, not commented out

kruai (approved):

tests/test_unit.py
thearifismail: Gone!

thearifismail:
Work Done, pending the merging of PR #1229 and PR #1232

fstavela (approved):
LGTM
The 2 failures in "All tests" are not related to this change, so we can merge this.
I've updated iqe-hbi, the next build should pass.

fstavela (approved):

PR #1201, opened 2022-09-28
[ESSNTL-629] Account for unindexed fields in validation

CodeHeeler:
# Overview

This PR is being created to address [ESSNTL-629](https://issues.redhat.com/browse/ESSNTL-629).

In addition to the original changes for this issue in #1186 , it was found that fields not indexed by xjoin were also throwing 400 even though they are in the schema.  As we cannot filter these fields, this is correct behaviour, but we now check for them and throw a custom error message if a user attempts to include them in their query.

Currently, these fields include:
- bios_release_date
- running_processes
- yum_repos
- installed_packages_delta

We do not hard code them but rather check for `x-indexed: False` in the properties of fields from the system profile schema.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


fstavela:
@CodeHeeler It looks like some of the unit tests are failing

CodeHeeler:
> @CodeHeeler It looks like some of the unit tests are failing

Investigating

app/__init__.py
```diff
@@ -140,7 +141,10 @@ def process_spec(spec, process_unindexed=False):
             if field_filter == "object":
                 system_profile_spec_processed[field]["children"] = process_spec(props["properties"], True)
 
-    return system_profile_spec_processed
+        if props.get("x-indexed", False):
```
kruai, line 144: Unless I'm misunderstanding, shouldn't it be like this?
```suggestion
        if not props.get("x-indexed", True):
```
We want the default value to be True, and we only want to add the field to the list if the value is explicitly False

kruai (changes requested):
I think the logic might not be correct - adding a test or two would help validate whether or not it works.

app/__init__.py
CodeHeeler: Good catch, made the change, thanks!

CodeHeeler:
Moving back to draft PR to add some new tests and investigate the current test failures in xjoin which I believe are all due to changes to the expected values being sent now that validation map builder has two values (unindexed_fields as well as the system_profile_schema_spec)

fstavela (approved):
LGTM

app/__init__.py
```diff
@@ -138,9 +139,12 @@ def process_spec(spec, process_unindexed=False):
             }
 
             if field_filter == "object":
-                system_profile_spec_processed[field]["children"] = process_spec(props["properties"], True)
+                system_profile_spec_processed[field]["children"], _ = process_spec(props["properties"], True)
```
FabriciaDinizRH, line 142: Why was the `, _` necessary in this case?

FabriciaDinizRH (changes requested):
I feel like at least one test for the unindexed fields should be added.

app/__init__.py
CodeHeeler: @FabriciaDinizRH Because the process_spec return statement now returns a tuple instead of a single value, from line 147 below:
`return system_profile_spec_processed, unindexed_fields`

FabriciaDinizRH: got it. Thanks.

CodeHeeler:
> I feel like at least one test for the unindexed fields should be added.

@FabriciaDinizRH  Added, thank you for the suggestion.  We also already have "test_invalid_fields" in tests/test_api_hosts_get.py, but you're right that we should test in xjoin as well.

kruai (approved):
Looks good!

FabriciaDinizRH (approved):
Nice work, LGTM!

PR #1210, opened 2022-10-14
Increase memory limit for host synchronizer job

kruai:
# Overview

This PR increases the memory request and memory limit for the host-synchronizer job. This is needed because the latest job I ran yesterday was OOMKilled in Stage.


FabriciaDinizRH (approved):

CodeHeeler (approved):
LGTM

PR #1211, opened 2022-10-14
Decrease script chunk size, add env var

kruai:
# Overview

This PR is being created to address the OOMKill failures that the host-sync job is running into. I'm hoping that decreasing the number of hosts that get loaded into memory at once will make it take up less memory. I think there may still be a memory leak hiding in the code somewhere, but I haven't found it yet.


CodeHeeler (approved):
LGTM, thanks!

PR #1215, opened 2022-10-19
Increase sync script memory limit

CodeHeeler:
# Overview

This PR is being created to continue troubleshooting the OOMKilled jobs in stage.  Calculations were made by @fijshion estimating the buffer per partitions and number of partitions to arrive at 2Gi, settting to 3Gi for comfortable extra.


roliveri (approved):
lgtm

PR #1208, opened 2022-10-13
Allow encoded commas when querying tags

kruai:
# Overview

This PR is being created to address [ESSNTL-1675](https://issues.redhat.com/browse/ESSNTL-1675).
It changes the spec for Tags to be pipe-delimited instead of comma-delimited. We also have the option to make it space-delimited; it just depends on what character we don't want to allow in the tag.

We also had some tests that validated the csv-style of supplying multiple values to this query param; I've modified these to send the `tags` param multiple times, since that's the way the front-ends supply that query.

## PR Checklist

- [x] Keep PR title short, ideally under 72 characters
- [ ] Descriptive comments provided in complex code blocks
- [x] Tests: validate optimal/expected output
- [x] Tests: validate exceptions and failure scenarios
- [ ] Tests: edge cases
- [ ] Recovers or fails gracefully during potential resource outages (e.g. DB, Kafka)
- [ ] Uses [type hinting](https://docs.python.org/3/library/typing.html), if convenient
- [ ] Documentation, if this PR changes the way other services interact with host inventory
- [ ] Links to related PRs

## Secure Coding Practices Documentation Reference

You can find documentation on this checklist [here](https://github.com/RedHatInsights/secure-coding-checklist).

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #1228, opened 2022-10-27
Allow commas in tag query params

kruai:
# Overview

This PR is being created to address [ESSNTL-1675](https://issues.redhat.com/browse/ESSNTL-1675).
Overrides the URIParser's query param resolver in order to distinguish between an encoded comma and a non-encoded one. Like with #1208, this also removes support for comma-delimited query params, so I've updated some tests.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/tests/test_xjoin.py
+++ b/tests/test_xjoin.py
@@ -997,15 +997,15 @@ def test_query_variables_tags_collection_multi(mocker, assert_tag_query_host_fil
 
 def test_query_variables_tags_collection_encoded_commas(mocker, assert_tag_query_host_filter_single_call):
     assert_tag_query_host_filter_single_call(
-        build_tags_url(query="?tags=Sat/env=prod%2Cstage&tags=insights-client/os=fedora%2Cubuntu"),
+        build_tags_url(query="?tags=Sat/env=prod%2Cstage&tags=insights-client/Location=Santa%20Fe%2C%20NM"),
         host_filter={
             "AND": (
                 {"tag": {"namespace": {"eq": "Sat"}, "key": {"eq": "env"}, "value": {"eq": "prod,stage"}}},
                 {
                     "tag": {
                         "namespace": {"eq": "insights-client"},
-                        "key": {"eq": "os"},
-                        "value": {"eq": "fedora,ubuntu"},
+                        "key": {"eq": "Location"},
+                        "value": {"eq": "Santa Fe, NM"},
                     }
                 },
             ),
--- a/tests/test_xjoin.py
+++ b/tests/test_xjoin.py
@@ -367,24 +367,24 @@ def test_query_variables_invalid(query, mocker, graphql_query_empty_response, ap
 @pytest.mark.parametrize(
     "tags,query_param",
     (
-        (({"namespace": {"eq": "a"}, "key": {"eq": "b"}, "value": {"eq": "c"}},), "a/b=c"),
-        (({"namespace": {"eq": "a"}, "key": {"eq": "b"}, "value": {"eq": None}},), "a/b"),
+        (({"namespace": {"eq": "a"}, "key": {"eq": "b"}, "value": {"eq": "c"}},), f"?tags={quote('a/b=c')}"),
+        (({"namespace": {"eq": "a"}, "key": {"eq": "b"}, "value": {"eq": None}},), f"?tags={quote('a/b')}"),
         (
             (
                 {"namespace": {"eq": "a"}, "key": {"eq": "b"}, "value": {"eq": "c"}},
                 {"namespace": {"eq": "d"}, "key": {"eq": "e"}, "value": {"eq": "f"}},
             ),
-            "a/b=c,d/e=f",
+            f"?tags={quote('a/b=c')}&tags={quote('d/e=f')}",
         ),
         (
             ({"namespace": {"eq": "a/a=a"}, "key": {"eq": "b/b=b"}, "value": {"eq": "c/c=c"}},),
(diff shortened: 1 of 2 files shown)


tests/test_xjoin.py
```diff
@@ -367,24 +367,24 @@ def test_query_variables_invalid(query, mocker, graphql_query_empty_response, ap
...
                 {"namespace": {"eq": "a"}, "key": {"eq": "b"}, "value": {"eq": "c"}},
                 {"namespace": {"eq": "d"}, "key": {"eq": "e"}, "value": {"eq": "f"}},
             ),
-            "a/b=c,d/e=f",
+            f"?tags={quote('a/b=c')}&tags={quote('d/e=f')}",
         ),
         (
             ({"namespace": {"eq": "a/a=a"}, "key": {"eq": "b/b=b"}, "value": {"eq": "c/c=c"}},),
-            quote("a/a=a") + "/" + quote("b/b=b") + "=" + quote("c/c=c"),
+            "?tags=" + quote(quote("a/a=a") + "/" + quote("b/b=b") + "=" + quote("c/c=c")),
```
CodeHeeler, line 381: Minor, but here you use double quotes within the quote() call when above on line 377 you use single quotes--shouldn't these be kept consistent?

tests/test_xjoin.py
```diff
@@ -998,13 +995,19 @@ def test_query_variables_tags_collection_multi(mocker, assert_tag_query_host_fil
...
         host_filter={
             "AND": (
-                {"tag": {"namespace": {"eq": "Sat"}, "key": {"eq": "env"}, "value": {"eq": "prod"}}},
-                {"tag": {"namespace": {"eq": "insights-client"}, "key": {"eq": "os"}, "value": {"eq": "fedora"}}},
+                {"tag": {"namespace": {"eq": "Sat"}, "key": {"eq": "env"}, "value": {"eq": "prod,stage"}}},
+                {
+                    "tag": {
+                        "namespace": {"eq": "insights-client"},
+                        "key": {"eq": "os"},
+                        "value": {"eq": "fedora,ubuntu"},
```
CodeHeeler, line 1008: are these meant to be equivalent to the `("Location": ["Santa Fe, NM"])` where they will not be split?  I was just confused by the case where something would be both fedora and ubuntu and wanted to doublecheck the intent of this test.

tests/test_xjoin.py
```diff
@@ -1331,26 +1334,26 @@ def test_system_profile_sap_sids_endpoint(mocker, graphql_system_profile_sap_sid
...
         ),
-        (({"namespace": {"eq": "ɑ"}, "key": {"eq": "β"}, "value": {"eq": "ɣ"}},), "ɑ/β=ɣ"),
+        (({"namespace": {"eq": "ɑ"}, "key": {"eq": "β"}, "value": {"eq": "ɣ"}},), f"?tags={quote('ɑ/β=ɣ')}"),
     ),
 )
 def test_system_profile_sap_sids_endpoint_tags(
     tags, query_param, mocker, graphql_system_profile_sap_sids_query_empty_response, api_get
 ):
-    url = build_system_profile_sap_sids_url(query=f"?tags={quote(query_param)}")
+    url = build_system_profile_sap_sids_url(query=query_param)
```
CodeHeeler, line 1356: if someone is providing tags.yaml to insights-client for an upload, we should be fine, but if someone is updating tags directly via our api, this would need a docs update to explain the expected query format.  Do we allow direct post/patch for tags on a host?

CodeHeeler (commented):
Overall looks good, but I have a few questions so I understand more clearly.

tests/test_xjoin.py
kruai: The single-quotes are used there because they're inside a double-quoted string. I could alternatively use an escaped double-quote, but I thought this approach looked cleaner. Preference?

tests/test_xjoin.py
kruai: The intent is just to make sure commas work in the query and aren't interpreted as separate strings. I can change it to a location if you prefer.

tests/test_xjoin.py
kruai: You can't directly update host tags from the UI, no. And our documentation already explains the query format here: https://consoledot.pages.redhat.com/docs/dev/services/inventory.html#_query_syntax

tests/test_xjoin.py
CodeHeeler: you're right, your current way looks cleaner.

tests/test_xjoin.py
CodeHeeler: Thanks!  I think it's helpful for our test examples to be real world clear when possible so that it's easier to understand the code when reading through it and you need to make some new addition or change.

tests/test_xjoin.py
CodeHeeler: Ok cool, then we're good.

CodeHeeler (approved):
Thanks for addressing my questions.  LGTM now.

FabriciaDinizRH (approved):
LGTM

fstavela (approved):
LGTM

PR #1232, opened 2022-11-02
Fix datasource ID for production

kruai:
# Overview

This PR is being created to address [ESSNTL-3603](https://issues.redhat.com/browse/ESSNTL-3603). It turns out the data source ID we were using works in Stage, but not Prod. Since Prod is more important, I've updated the data source ID to the one that works in Prod. This will result in a blank panel for Stage, but a working one for Prod.

### DIFF
--- a/dashboards/grafana-dashboard-insights-inventory-general.configmap.yaml
+++ b/dashboards/grafana-dashboard-insights-inventory-general.configmap.yaml
@@ -2541,7 +2541,7 @@ data:
         {
           "datasource": {
             "type": "cloudwatch",
-            "uid": "C_IyqZfnz"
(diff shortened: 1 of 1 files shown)


CodeHeeler:
Is there a way to take this value from a config environment variable so that both stage and prod will work?  If not, I agree with your decision to go with prod.

kruai:
> Is there a way to take this value from a config environment variable so that both stage and prod will work? If not, I agree with your decision to go with prod.

Not sure... That would be a better way to do it, though. I'll look through app-interface and see if I can find examples of people doing that.

fstavela (approved):

CodeHeeler (approved):

PR #1248, opened 2022-12-08
Remove excess host-synchronizer logs

kruai:
# Overview

This PR is being opened because the host-synchronizer produces a ton of logs in Prod, which can cause issues for Kibana.


CodeHeeler (approved):
LGTM

PR #1259, opened 2023-01-24
feat(ESSNTL-3754): Add support for Unleash feature flags

kruai:
# Overview

This PR is being created to address [ESSNTL-3754](https://issues.redhat.com/browse/ESSNTL-3754).
It does the following:
- Adds support for Unleash feature flags, which are automatically configured using the `UNLEASH_URL` and `UNLEASH_SECRET_NAME` parameters provided by our app-interface environment configurations.
- Configures a helper method to get the value of a feature flag toggle, and returns a fallback value (configured per-flag) if it is unable to connect to the configured Unleash server. The helper method returns the value of the toggle as well as metadata informing the caller whether it's returning the actual value or the default fallback value.
- Adds a a temporary GET `/unleash/test_toggle` endpoint in order to make it easier for QE and manual testing to validate whether the feature flag is enabled and working.
- Removes an old import that was no longer needed, but wasn't getting caught due to a #noqa comment

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


CodeHeeler (approved):
LGTM, thanks for implementing feature flags!

FabriciaDinizRH (dismissed):

app/__init__.py
```diff
@@ -200,6 +200,19 @@ def create_app(runtime_environment):
     flask_app.config["SYSTEM_PROFILE_SPEC"] = sp_spec
     flask_app.config["UNINDEXED_FIELDS"] = unindexed_fields
 
+    # Configure Unleash (feature flags)
+    if app_config.unleash_token and not app_config.bypass_unleash:
```
vkrizan, line 204: Is the bypass option needed when an empty token can practically bypass/disable it?

app/config.py
```diff
@@ -216,9 +216,16 @@ def __init__(self, runtime_environment):
             self.pendo_timeout = int(os.environ.get("PENDO_TIMEOUT", "240"))
             self.pendo_request_size = int(os.environ.get("PENDO_REQUEST_SIZE", "500"))
 
+        self.unleash_url = os.environ.get("UNLEASH_URL", "http://unleash:4242/api")
+        self.unleash_token = os.environ.get("UNLEASH_TOKEN", "")
+        self.bypass_unleash = (
+            os.environ.get("BYPASS_UNLEASH", "false").lower() == "true" if self.unleash_token else True
```
vkrizan, line 222: There is a combination of a string and boolean values.

Isn't there an existing function that would convert a string value to boolean?

lib/feature_flags.py
```diff
@@ -0,0 +1,29 @@
...
+
+# Gets a feature flag's value from Unleash, if available.
+# Accepts a feature flag in the format:
+#   (<feature_flag_name>, <fallback_value>)
+# Returns a tuple containing the flag's value and whether or not the fallback value was used.
+def get_flag_value(feature_flag: Tuple[str, bool]) -> Tuple[bool, bool]:
+    try:
+        # Attempt to get the feature flag via Unleash
+        return UNLEASH.client.is_enabled(feature_flag[0], fallback_function=custom_fallback), False
+    except AttributeError:
```
vkrizan, line 27: Is this the correct error to watch for? Can it be done with a regular check?

swagger/api.spec.yaml
```diff
@@ -708,6 +708,17 @@ paths:
         '403':
             description: Forbidden
 
+  '/unleash/test_toggle':
```
vkrizan, line 711: Is this a right path for a concrete feature flag?

vkrizan (commented):
I've left some comments, see below. I think this might be a `feat` instead of `chore`.

@kruai would you mind rebasing the PR on top of master. There are couple merge commits from master.

Thank you.

swagger/api.spec.yaml
kruai: Ah - I guess it would make more sense to use the name of the flag, since it's only checking that flag. I'll update it

lib/feature_flags.py
kruai: `AttributeError` was raised on me whenever the `UNLEASH` client didn't get initialized. Is there a reason you think it should be a different error?
Also, can you please explain what you mean by "regular check"?

deploy/clowdapp.yml
```diff
@@ -80,6 +80,14 @@ objects:
           value: ${BYPASS_TENANT_TRANSLATION}
         - name: CLOWDER_ENABLED
           value: "true"
+        - name: UNLEASH_URL
```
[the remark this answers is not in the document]
kruai: Ooh OK, will do. Thanks for finding that

app/__init__.py
kruai: Probably not _needed_; I was just trying to stay in line with how we bypass the other services (RBAC, tenant translator, etc). I'll update this to go along with the `featureFlags: true` changes Joao mentioned

app/config.py
kruai: Both conditions result in a boolean; a string is never returned. There is a method that converts strings to booleans, as long as they're capitalized correctly, so I could change this to:
```
from ast import literal_eval
...
self.bypass_unleash = (
  literal_eval(os.environ.get("BYPASS_UNLEASH", "False").capitalize()) if self.unleash_token else True
)
```
Do you prefer it that way instead?

kruai:
> I've left some comments, see below. I think this might be a `feat` instead of `chore`.
> 
> @kruai would you mind rebasing the PR on top of master. There are couple merge commits from master.
> 
> Thank you.

I can rename it to `feat` for sure. I've rebased; you should also have permission to rebase if you want it rebased before you review.

app/config.py
vkrizan: Sorry, the forced value for tests a few lines down have confused me.
```
if self._runtime_environment == RuntimeEnvironment.TEST:
   self.bypass_unleash = "true"
```

strider:
Some timeouts on reaper tests.

app/__init__.py
kruai: Removed it, and made all the logic rely on the presence of the token instead

app/config.py
kruai: You're right! Looks like all the bypasses were doing the same thing, so I changed them to use a boolean instead.

swagger/api.spec.yaml
kruai: Changed to GET `/unleash/inventory_groups_toggle`.

lib/feature_flags.py
kruai: Unfortunately, the documentation [is pretty sparse](https://docs.getunleash.io/unleash-client-python/unleashclient.html), but it doesn't seem like it raises exceptions. Based on the docs and what I've experienced testing it so far, if there's an error connecting or authenticating, or it can't find the flag, it will call `fallback_function` (or return False, if that's not defined).

So yes, basically the AttributeError is just there to catch the scenario where the client was not correctly initialized.

deploy/clowdapp.yml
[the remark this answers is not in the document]
kruai: I've added the new config using your PR and those docs as a reference. I wonder if the docs got updated sometime after I made my original feature flags PR (a long while back)...

strider:
LGTM! Good work Asa!

deploy/clowdapp.yml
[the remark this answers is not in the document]
jpramos123: Thanks!

jpramos123 (approved):
LGTM

Thanks for this PR

lib/feature_flags.py
kruai: OK, I can do that. I'm never sure which way's better, since people tend to say that optimistically catching exceptions might be more pythonic. I don't really mind either way though

FabriciaDinizRH (approved):

lib/feature_flags.py
kruai: OK. I still think it makes sense to raise an error in the fallback, but I've switched it to pessimistic handling while checking for `UNLEASH.client`, eliminating the `AttributeError`. I've switched the error type raised by the fallback method to `ConnectionError`, which seems more appropriate to me.

lib/feature_flags.py
```diff
@@ -0,0 +1,42 @@
...
+    flag_value = feature_flag[1]
+    using_fallback = True
+
+    # Attempt to get the feature flag via Unleash
+    try:
+        if hasattr(UNLEASH, "client"):
+            flag_value = UNLEASH.client.is_enabled(feature_flag[0], fallback_function=custom_fallback)
+            using_fallback = False
+    except ConnectionError as ce:
+        logger.warning(ce)
```
vkrizan, line 38: Nit: I'd maybe more info to the log message about the feature flag name and it's fallback value that is going to be used.

lib/feature_flags.py
```diff
@@ -0,0 +1,42 @@
...
+    raise ConnectionError(f"Could not contact Unleash server, or feature toggle {feature_name} not found!")
+
+
+# Gets a feature flag's value from Unleash, if available.
+# Accepts a feature flag in the format:
+#   (<feature_flag_name>, <fallback_value>)
+# Returns a tuple containing the flag's value and whether or not the fallback value was used.
+def get_flag_value(feature_flag: Tuple[str, bool]) -> Tuple[bool, bool]:
+    # Set defaults
+    flag_value = feature_flag[1]
```
vkrizan, line 29: Nit: readablity
```
flag_name, flag_value = feature_flag
```

lib/feature_flags.py
vkrizan: Looks a bit better. The `hasattr` check is insufficient, as the `client` instance var is created within constructor.

lib/feature_flags.py
```diff
@@ -0,0 +1,42 @@
...
+# Without this fallback function, is_enabled just returns False without error.
+def custom_fallback(feature_name: str, context: dict) -> bool:
+    raise ConnectionError(f"Could not contact Unleash server, or feature toggle {feature_name} not found!")
+
+
+# Gets a feature flag's value from Unleash, if available.
+# Accepts a feature flag in the format:
+#   (<feature_flag_name>, <fallback_value>)
+# Returns a tuple containing the flag's value and whether or not the fallback value was used.
+def get_flag_value(feature_flag: Tuple[str, bool]) -> Tuple[bool, bool]:
```
vkrizan, line 27: Does a feature flag's fallback value need to be always carried over to request a value from Unleash, hence requiring a tuple? Would each caller need to have different fallback values for the same FF?

swagger/api.spec.yaml
```diff
@@ -708,6 +708,17 @@ paths:
         '403':
             description: Forbidden
 
+  '/unleash/inventory_groups_toggle':
+    get:
+      summary: Check whether the Groups feature is enabled
+      description: Returns the value of the hbi.api.inventory-groups toggle, if able to be retrieved.
+      operationId: api.unleash.get_inventory_groups_toggle
+      security:
+        - ApiKeyAuth: []
```
vkrizan, line 717: Do we want to have this endpoint open in production?

kruai: The endpoint will just tell the user whether groups are enabled - this is mostly just to help with QE testing. Do you think it's a security risk? (CC @strider)

lib/feature_flags.py
kruai: I figured this setup would make it easy on us when checking the value; we just import `FLAG_INVENTORY_GROUPS` and use that in the function. But it would be just as easy to use an enum or something; I'll make another pass at it and see if I can make it cleaner.

vkrizan: Wouldn't separating flag name and fallback values be a better design (logical)?

swagger/api.spec.yaml
vkrizan: Ah, I've missed the intention. "toggle" part have confused me there. :slightly_smiling_face: 

Maybe the API would not be needed, as QE might lookup the value from Unleash. We need to bear in mind that the switch can flip to fallback value on any error. So actually testing with value from Unleash might be more precise.

kruai: Ah yeah, they call them "toggles" in Unleash, which can get a bit confusing. And you're correct that it would be more precise to get the values directly, but I think that implementation isn't finished yet, so this was meant to help QE in the meantime. Is that accurate @computercamplove?

lib/feature_flags.py
kruai: Probably! It works either way in my head, but I can certainly see the appeal of separating them, and have pushed those changes.

lib/feature_flags.py
kruai: Done

lib/feature_flags.py
kruai: Done

thearifismail:
The code `LGTM` but added a comment to the [JIRA card](https://issues.redhat.com/browse/ESSNTL-3754) regarding how unleash feature is related to host-inventory.

app/config.py
```diff
@@ -63,6 +63,21 @@ def topic(t):
...
 
+        unleash_configs = dict(
+            UNLEASH_URL=os.getenv("UNLEASH_URL"),
+            UNLEASH_TOKEN=os.getenv("UNLEASH_TOKEN"),
+        )
+        unleash = cfg.featureFlags
+        if unleash:
+            UNLEASH_URL = f"{unleash.hostname}:{unleash.port}/api"
+            if unleash.port == 443:
+                UNLEASH_URL = f"https://{UNLEASH_URL}"
```
vkrizan, line 74: Lint: upper case names are usually reserved for constants.

swagger/api.spec.yaml
computercamplove: @vkrizan @kruai yes, it's not finished yet

app/config.py
kruai: good call, fixed

app/config.py
```diff
@@ -63,6 +63,21 @@ def topic(t):
...
             self.kafka_security_protocol = "PLAINTEXT"
 
+        unleash_configs = dict(
+            UNLEASH_URL=os.getenv("UNLEASH_URL"),
+            UNLEASH_TOKEN=os.getenv("UNLEASH_TOKEN"),
+        )
+        unleash = cfg.featureFlags
+        if unleash:
+            UNLEASH_URL = f"{unleash.hostname}:{unleash.port}/api"
+            if unleash.port == 443:
```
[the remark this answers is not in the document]
kruai: updated it

app/config.py
```diff
@@ -63,6 +63,21 @@ def topic(t):
...
+        )
+        unleash = cfg.featureFlags
+        if unleash:
+            unleash_url = f"{unleash.hostname}:{unleash.port}/api"
+            if unleash.port in (80, 8080):
+                unleash_url = f"http://{unleash_url}"
+            else:
+                unleash_url = f"https://{unleash_url}"
+
+            self.unleash_url = unleash_configs["UNLEASH_URL"] or unleash_url
```
vkrizan, line 78: If clowder does not provide the config, then these config vars would not be set from env vars.

kruai: I've updated the logic so that the config vars are set whether or not it's using clowder.

kruai:
FYI, PR #1265 seems to have caused issues for local `pipenv install`, and it turns out that you need to have `pg_config` on your system before it will work. I dealt with that while merging to this branch, so I added a small section with instructions to the README.

vkrizan (approved):
LGTM. Thank you!

PR #1265, opened 2023-01-31
Build psycopg2 from source

bsquizz:
host-inventory fails to connect to the DB in a FIPS-enabled environment:
```
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) SSL error: disabled for FIPS
```

After doing some research, this is due to the `psycopg2-binary` using an SSL library that was compiled without FIPS support.

Switching away from the binary and building `psycopg2` from source resolves the problem.

This change breaks the unit tests since they were running directly on the jenkins agent VM which does not have the 'devel' packages present on it that would allow `pipenv install` to work when using `psycopg2` instead of `psycopg2-binary`. I have changed them to run using containers. The tests run inside the host-inventory image after it is built during the pr_check. This is a similar approach to how advisor runs their unit tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 
+ARG TEST_IMAGE=false
+
 USER root
 
 # install postgresql from centos if not building on RHSM system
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -11,18 +13,22 @@ RUN FULL_RHEL=$(microdnf repolist --enabled | grep rhel-8) ; \
     fi
 
 RUN microdnf module enable postgresql:13 python38:3.8 && \
-    microdnf install --setopt=tsflags=nodocs -y postgresql python38 && \
+    microdnf install --setopt=tsflags=nodocs -y postgresql python38 libpq-devel python38-devel gcc && \
```
vkrizan, line 16: I'd suggest to remove the devel packages once it is built, or, build the binary in other container and copy it over to production one.

bsquizz: It's a good point -- I'm familiar with how that looks using the go-toolset and I'll see what examples I can find for doing this with a python environment.

bsquizz: I decided to just remove the devel packages and all their dependencies. It ends up removing 16 packages at the end of the build. I think this is the faster way to remove the 'extras' for now -- if we want to move to a multi-stage build we could do that in a follow-up.

pr_check.sh
```diff
@@ -29,7 +29,8 @@ IQE_CJI_TIMEOUT="30m"
 # ---------------------------
 
 # Get bonfire helper scripts
-CICD_URL=https://raw.githubusercontent.com/RedHatInsights/bonfire/master/cicd
+CICD_URL=https://raw.githubusercontent.com/RedHatInsights/bonfire/test_image_arg/cicd
+export BONFIRE_REPO_BRANCH=test_image_arg
```
kruai, line 33: Is `test_image_arg` an intentional permanent change?

kruai (commented):
Mostly looks good; I just had one question

pr_check.sh
bsquizz: Oops, good catch, no it is not, fixing.

kruai (approved):
LGTM... I'll post it in the channel for others to review as well. Thanks!

vkrizan (approved):
LGTM

PR #1268, opened 2023-02-03
Revert "feat(ESSNTL-3754): Add support for Unleash feature flags"

kruai:
This caused Stage to crashloop, so we need to revert. I've copied a couple of logs that will hopefully be of use while investigating the issue.


CodeHeeler (approved):
LGTM, thanks for getting to this revert quickly.

PR #1269, opened 2023-02-03
feat(ESSNTL-3754): Add Unleash support, with fix

kruai:
# Overview

**NOTE: The only changes from the original PR are from [this commit](https://github.com/RedHatInsights/insights-host-inventory/pull/1269/commits/c3bf5d6fdf555d9810ba0176e0cb03af7e215c29) onwards.**

This PR is being created to address [ESSNTL-3754](https://issues.redhat.com/browse/ESSNTL-3754).
This re-implements #1259, which was reverted by #1268. In addition to undoing that reversion, it configures a cache directory for Unleash in Clowder environments. It uses the `/tmp/.unleashcache` folder by default; if we decide we want to configure this path later, we can add `UNLEASH_CACHE_DIR` to our clowdapp config.

For non-clowder environments, it leaves `UNLEASH_CACHE_DIRECTORY` unset; this makes it use fcache to decide, which is generally recommended (when there aren't permission issues). I got this config info from the flask-unleash docs [here](https://docs.getunleash.io/Flask-Unleash/#configuration).

## PR Checklist

- [ ] Tests: edge cases
- [x] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,28 @@ The `python-snappy` package requires the core Snappy library to be installed on 
 sudo dnf install snappy
 ```
 
+### pg_config
+
+Local development also requires the `pg_config` file, which is installed with the postgres developer library. To install this, use the command appropriate to your system:
+
+#### Fedora/Centos
+
+```bash
+sudo dnf install libpq-devel
+```
+
+#### Debian/Ubuntu
+
+```bash
+sudo apt-get install libpq-dev
+```
+
+#### MacOS (using Homebrew)
+
+```bash
+brew install postgresql
+```
+
 ### Install dependencies
 
 This project uses pipenv to manage the development and deployment environments.
--- a/lib/feature_flags.py
+++ b/lib/feature_flags.py
@@ -1,0 +1,47 @@
+from typing import Tuple
+
+from flask_unleash import Unleash
+
+from app.logging import get_logger
+
+UNLEASH = Unleash()
+logger = get_logger(__name__)
+
+FLAG_INVENTORY_GROUPS = "hbi.api.inventory-groups"
+
+FLAG_FALLBACK_VALUES = {FLAG_INVENTORY_GROUPS: True}
+
+
+def init_unleash_app(app):
+    UNLEASH.init_app(app)
+
+
+# Raise an error if the toggle is not found on the configured Unleash server.
+# Without this fallback function, is_enabled just returns False without error.
+def custom_fallback(feature_name: str, context: dict) -> bool:
+    raise ConnectionError(f"Could not contact Unleash server, or feature toggle {feature_name} not found.")
+
+
+# Gets a feature flag's value from Unleash, if available.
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -14,6 +14,28 @@ The `python-snappy` package requires the core Snappy library to be installed on
 sudo dnf install snappy
 ```
 
+### pg_config
```
thearifismail, line 17: Why are these needed now but not before?
Couldn't these be added to the Pipfile?

lib/feature_flags.py
```diff
@@ -0,0 +1,47 @@
...
+FLAG_FALLBACK_VALUES = {FLAG_INVENTORY_GROUPS: True}
+
+
+def init_unleash_app(app):
+    UNLEASH.init_app(app)
+
+
+# Raise an error if the toggle is not found on the configured Unleash server.
+# Without this fallback function, is_enabled just returns False without error.
+def custom_fallback(feature_name: str, context: dict) -> bool:
```
thearifismail, line 21: shouldn't this function be called default_fallback?

thearifismail (approved):
LGTM.  A couple of minor comments though

README.md
kruai: I explained this in the [original PR that we merged](https://github.com/RedHatInsights/insights-host-inventory/pull/1259#issuecomment-1414225703). But no, it can't be added to the pipfile because it's not python-related

lib/feature_flags.py
kruai: `default_fallback` is kinda redundant imo, since we could refer to that value as either a default value or a fallback value. I called it `custom_fallback` because that's the name they used in the official documentation, and because it's a custom function that's providing a fallback value

swagger/api.spec.yaml
```diff
@@ -708,6 +708,17 @@ paths:
         '403':
             description: Forbidden
 
+  '/unleash/inventory_groups_toggle':
+    get:
+      summary: Check whether the Groups feature is enabled
+      description: Returns the value of the hbi.api.inventory-groups toggle, if able to be retrieved.
```
fstavela, line 714: `tags` is missing in this endpoint, look at https://github.com/RedHatInsights/insights-host-inventory/blob/master/swagger/api.spec.yaml#L644 for example. Here it should be probably set to `unleash`.

computercamplove:
@kruai currently i get `None` for response while testing in ephemeral, since unleash is not working for ephemeral could we get `True` instead?

thearifismail (approved):
LGTM

thearifismail (approved):
LGTM

PR #1273, opened 2023-02-08
feat(ESSNTL-3711): DELETE /groups/<group_ids> endpoint

kruai:
# Overview

This PR is being created to address [ESSNTL-3711](https://issues.redhat.com/browse/ESSNTL-3711).
Adds the `DELETE /groups/<ids>` endpoint, plus some tests. Some caveats:

- I designed the delete function to take a list of group IDs for futureproofing reasons, and to keep the design consistent with what we're using for our host deletion endpoints. I'd actually vote to make the delete endpoint take a list of IDs, like how our `DELETE /hosts/<id_list>` endpoint works, so that the UI can delete multiple at once, if needed later. **Edit: We've decided to make it so you can delete a list of groups at once.**
- This sets up the integration with the DB only. Kafka integration has not been set up yet, because we don't yet know the specifics of how the Kafka messages are going to be consumed. I think we'll need to make a second pass after the endpoints are set up to produce the correct messages to the correct place.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


FabriciaDinizRH (commented):
Overall, very good, I just made a few suggestions.
Regarding the deletion endpoint receiving an ID of a list of IDs, I vote for a list of IDs as well. If more people agree, we would also need a test for that scenario.

lib/group_repository.py
```diff
@@ -0,0 +1,56 @@
...
+        delete_query.session.rollback()
+
+    return group_deleted
+
+
+def delete_group_list(group_id_list, chunk_size):
+    deletion_count = 0
+    select_query = Group.query.filter((Group.org_id == get_current_identity().org_id) & Group.id.in_(group_id_list))
+
+    # Do this for one group at a time, because once we add Kafka integration,
```
CodeHeeler, line 43: Are we expecting users to not want to delete large groups often?  I'm wondering if choosing one at a time as opposed to bulk delete will be too slow down the road. Could be over engineering at this stage, but as we use bulk delete as well as a delete all for our hosts, wanted to bring it up.

api/group.py
```diff
@@ -24,8 +37,19 @@ def update_group_details(group_id, group_daty):
     pass
 
 
+@api_operation
+@rbac(Permission.WRITE)
+@metrics.api_request_time.time()
 def delete_group(group_id):
```
thearifismail, line 43: Shouldn't there be some control on who can delete a group?  Does it mean that any body can delete a group with no hosts in it?

thearifismail (commented):
Please see the comment.  
Question is should this PR wait until `DELETE group/<group_id_list>` has been added?

api/group.py
kruai: There's RBAC restricting access, and also users are only able to delete groups that have the same org_id as them.

lib/group_repository.py
kruai: It might be slower, but it feels safer sync-wise, since we have to emit an event after each change to a host or group; if we delete from the DB and then Kafka has a connectivity issue, we can roll back the delete and abort. WDYT?

kruai:
> > ```
> > * I designed the delete function to take a list of group IDs for futureproofing reasons, and to keep the design consistent with what we're using for our host deletion endpoints. I'd actually vote to make the delete endpoint take a list of IDs, like how our `DELETE /hosts/<id_list>` endpoint works, so that the UI can delete multiple at once, if needed later. **What do you all think?**
> > ```
> 
> Sounds good to me; let's keep it consistent and futureproof by asking for lists from the beginning.
> 
> > ```
> > * This sets up the integration with the DB only. Kafka integration has not been set up yet, because we don't yet know the specifics of how the Kafka messages are going to be consumed. I think we'll need to make a second pass after the endpoints are set up to produce the correct messages to the correct place.
> > ```
> 
> Do we have a jira issue tracking this later kafka step so we don't let it slip our minds?

https://issues.redhat.com/browse/ESSNTL-3714

swagger/api.spec.yaml
```diff
@@ -835,6 +835,17 @@ components:
...
       description: Ordering field name
+    groupOrderByParam:
+      name: order_by
+      in: query
+      required: false
+      schema:
+        type: string
+        enum:
+          - name
+          - updated
```
gkarat, line 846: updated_at

swagger/openapi.json
```diff
@@ -1373,6 +1373,20 @@
...
       },
+      "groupOrderByParam": {
+        "name": "order_by",
+        "in": "query",
+        "required": false,
+        "schema": {
+          "type": "string",
+          "enum": [
+            "name",
+            "updated",
```
gkarat, line 1384: updated_at

lib/group_repository.py
CodeHeeler: Makes sense.  Data integrity is the most important, so having the rollback is great, and though we could roll back a bulk action as well, I'm fine with moving forward as is.

CodeHeeler (approved):
Thanks for addressing the questions/concerns.  LGTM now.

strider:
LGTM. Nice Work, Asa!

strider (approved):

thearifismail (approved):
LGTM

computercamplove:
> /retest

@strider this should resolve problem in PR checks https://gitlab.cee.redhat.com/insights-qe/iqe-core/-/merge_requests/1447

PR #1275, opened 2023-02-13
DRAFT [ESSNTL-3701] Add resource_types endpoints

CodeHeeler:
# Overview

This PR is being created to address [ESSNTL-3701](https://issues.redhat.com/browse/ESSNTL-3701).
It creates two new GET endpoints, /api/inventory/v1/resource-types/ and /api/inventory/v1/resource-types/inventory-groups/ and updates logging, schemas, and testing accordingly.  The endpoints are only accessible by users with admin privileges.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/resource_type.py
+++ b/api/resource_type.py
@@ -1,0 +1,33 @@
+import flask
+from flask import current_app
+from flask_api import status
+
+from api import api_operation
+from api import flask_json_response
+from api import metrics
+from api.resource_query import build_paginated_resource_list_response
+from app import Permission
+from lib.feature_flags import FLAG_INVENTORY_GROUPS
+from lib.feature_flags import get_flag_value
+from lib.middleware import rbac
+
+
+@api_operation
+@rbac(Permission.ADMIN)
+@metrics.api_request_time.time()
+def get_resource_type_list(
+    page=1,
+    per_page=100,
+    order_by=None,
+    order_how=None,
+    filter=None,
+):
+    try:
+        resource_list = get_resource_type_list()
+    except ValueError as e:
+        log_get_resource_type_list_failed(logger)
+        flask.abort(400, str(e))
+
+    return flask_json_response(
+        build_paginated_resource_list_response(total, page, per_page, resource_list, additional_fields)
+    )
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -542,6 +542,31 @@ paths:
         '400':
           description: Invalid request.
 
+  /resource_types:
+    get:
+    operationId: api.resource_type.get_resource_type_list
+    tags:
+      - resource_types
+    summary: Read the entire list of resource types
+    description: >-
+      Read the entire list of all resource types available to the account.
+      <br /><br />
+      Required permissions: inventory:*:*
+    security:
+      - ApiKeyAuth: []
+    parameters:
+      - $ref: '#/components/parameters/perPageParam'
+      - $ref: '#/components/parameters/pageParam'
+      - $ref: '#/components/parameters/orderByParam'
+      - $ref: '#/components/parameters/orderHowParam'
+    responses:
+      '200':
+        description: Successfully read the resource types list.
+        content:
+          application/json:
+            schema:
+              #$ref: '#/components/schemas/ResourceTypeQueryOutput'
+
   /tags:
     get:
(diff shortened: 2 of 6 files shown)


api/resource_type.py
```diff
@@ -0,0 +1,33 @@
...
+from api import flask_json_response
+from api import metrics
+from api.resource_query import build_paginated_resource_list_response
+from app import Permission
+from lib.feature_flags import FLAG_INVENTORY_GROUPS
+from lib.feature_flags import get_flag_value
+from lib.middleware import rbac
+
+
+@api_operation
```
CodeHeeler, line 15: So, I'm used to how Django handles routes/urls files, but looking at how Flask does it, I was expecting:
`flask routes:
@ app.route('/')
def home():

@ app.route('/about')
def about():
`
Are we just auto parsing the def for the name and the route is being created in the background instead of explicitly here?  I only actually see the routes being listed in our swagger/openapi spec schema files and in our /tests/helpers/api_utils.py

api/resource_type.py
```diff
@@ -0,0 +1,33 @@
...
+@rbac(Permission.ADMIN)
```
CodeHeeler, line 16: My understanding is that in order to restrict this to the admin only, we need them to have inventory:*:* perms and we have implemented read and * at the inventory level, so this should work, but the tests for failing and passing rbac on this endpoint should double check.

api/resource_type.py
```diff
@@ -0,0 +1,33 @@
...
+    filter=None,
+):
+    try:
+        resource_list = get_resource_type_list()
+    except ValueError as e:
+        log_get_resource_type_list_failed(logger)
+        flask.abort(400, str(e))
+
+    return flask_json_response(
+        build_paginated_resource_list_response(total, page, per_page, resource_list, additional_fields)
```
CodeHeeler, line 32: My biggest question is what exactly are the resource_types?  As I understand it from the ADR, resource_types are only groups in our context, but we are implementing these endpoints to work together with an existing wizard that allows for more granular custom rbac control of them.  So it doesn't sound like we need any additional models or fields for handling this, only GET endpoints for listing groups in a paginated way, which should be essentially a clone of the endpoint for listing groups, correct (which in itself is duplicating that of hosts, calling an equivalent to build_paginated_host_list_response, or generalizing that method for our new expanded purposes)?

swagger/api.spec.yaml
```diff
@@ -542,6 +542,31 @@ paths:
...
+      - $ref: '#/components/parameters/orderByParam'
+      - $ref: '#/components/parameters/orderHowParam'
+    responses:
+      '200':
+        description: Successfully read the resource types list.
+        content:
+          application/json:
+            schema:
+              #$ref: '#/components/schemas/ResourceTypeQueryOutput'
+
```
CodeHeeler, line 569: TODO: adding the second nested endpoint at api/inventory/v1/resource_types/inventory_groups

api/resource_type.py
```diff
@@ -0,0 +1,33 @@
...
+def get_resource_type_list(
+    page=1,
+    per_page=100,
+    order_by=None,
+    order_how=None,
+    filter=None,
+):
+    try:
+        resource_list = get_resource_type_list()
+    except ValueError as e:
```
kruai, line 27: Does this function call itself? Also, in what situation would a ValueError occur?

api/resource_type.py
kruai: I think so... based on the comments in the ADR, I'm pretty sure the `/resource-types` endpoint will just return "inventory-groups", but in the format specified in that doc. And then the `/resource-types/inventory-groups` endpoint will just reuse most of the code from the usual GET method in order to get the actual list of groups, but with simplified params (no complex filtering or anything)

CodeHeeler:
Closed in favor of this PR instead https://github.com/RedHatInsights/insights-host-inventory/pull/1342/

PR #1276, opened 2023-02-16
feat[ESSNTL 3704]: initial POST groups/ structure

FabriciaDinizRH:
# Overview

This PR is being created to address [ESSNTL-3704](https://issues.redhat.com/browse/ESSNTL-3704).
This PR implements the functionality for creating groups.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


lib/group_repository.py
```diff
@@ -7,13 +8,47 @@
...
+        if created_group:
+            select_query.session.commit()
+            create_group_count.inc()
+            logger.debug("Created group:%s", created_group)
+
+            # this needs to be improved
+            # also have to check if the hosts don't already belong to a group
+            # try/catch to notify customers that the hosts couldn't be added if something goes wrong?
+            for host_id in host_ids:
+                select_query.session.query(HostGroupAssoc).add(group_id=created_group.id, host_id=host_id)
```
FabriciaDinizRH, line 45: I'm not sure how to do all the checking and logging necessary here in an optimal way.

api/group.py
```diff
@@ -21,8 +22,20 @@ def get_group_list(
     pass
 
 
+@api_operation
+@rbac(Permission.WRITE)
+@metrics.api_request_time.time()
 def create_group(group_data):
-    pass
+    # the [0] has to stay while Asa's recent changes aren't merged
+    if not get_flag_value(FLAG_INVENTORY_GROUPS)[0]:
```
kruai, line 30: It should be fine to remove that now!

app/instrumentation.py
```diff
@@ -89,6 +89,26 @@ def log_get_host_list_failed(logger):
     logger.debug("hosts not found", extra={"access_rule": get_control_rule()})
 
 
+# create group
+def log_add_group_succeeded(logger, group_id, control_rule):
+    logger.info(f"Created group: {group_id}", extra={"access_rule": control_rule})
```
vkrizan, line 94: Nit: Loggers have they own formatting, that formats the string only if the log is actually being used/printed.

lib/group_repository.py
```diff
@@ -18,6 +29,82 @@
...
 
+def _get_group_from_id(group_id, org_id):
+    return Group.query.filter((Group.org_id == org_id) & (Group.id == group_id)).one_or_none()
+
+
+def _get_group_from_name(group_name, org_id):
+    return Group.query.filter((Group.org_id == org_id) & (Group.name == group_name)).one_or_none()
+
+
+def _is_valid_uuid(host_id):
```
vkrizan, line 40: Can you please elaborate on this validation function? Is there an existing one (from library) that can be used?

lib/group_repository.py
```diff
@@ -18,6 +29,82 @@
...
+                new_hostgroupassoc = HostGroupAssoc(group_id=group_query_result.id, host_id=host_id)
+
+                assoc_query.session.add(new_hostgroupassoc)
+                assoc_query.session.commit()
+
+                log_host_group_add_succeeded(logger, host_id, group_query_result.id, get_control_rule())
+            else:
+                log_host_group_add_failed(logger, host_id, group_query_result.id, get_control_rule())
+
+        return serialize_group(_object_as_dict(group_query_result), host_ids)
```
vkrizan, line 70: Is there a better way of handling the `group_query_result` that does not require custom `_object_as_dict()`?

lib/group_repository.py
```diff
@@ -18,6 +29,82 @@
...
+            # i.e. isn't already associated with a group
+            if _is_valid_uuid(host_id) and host_id not in assoc_query.all():
+                new_hostgroupassoc = HostGroupAssoc(group_id=group_query_result.id, host_id=host_id)
+
+                assoc_query.session.add(new_hostgroupassoc)
+                assoc_query.session.commit()
+
+                log_host_group_add_succeeded(logger, host_id, group_query_result.id, get_control_rule())
+            else:
+                log_host_group_add_failed(logger, host_id, group_query_result.id, get_control_rule())
```
vkrizan, line 68: Should a non valid UUID return a failure?

lib/group_repository.py
```diff
@@ -18,6 +29,82 @@
...
+        assoc_query = HostGroupAssoc.query.filter(HostGroupAssoc.host_id.in_(host_ids))
+
+        for host_id in host_ids:
+            # check if the host doesn't already exist in the HostGroupAssoc table
+            # i.e. isn't already associated with a group
+            if _is_valid_uuid(host_id) and host_id not in assoc_query.all():
+                new_hostgroupassoc = HostGroupAssoc(group_id=group_query_result.id, host_id=host_id)
+
+                assoc_query.session.add(new_hostgroupassoc)
+                assoc_query.session.commit()
```
vkrizan, line 64: Why do we commit this early?

lib/group_repository.py
```diff
@@ -18,6 +29,82 @@
...
+    new_group_org_id = get_current_identity().org_id
+    new_group_account = get_current_identity().account_number
+
+    if not _validate_group_name(new_group_name, new_group_org_id):
+        return 0
+
+    created_group = Group(name=new_group_name, org_id=new_group_org_id, account=new_group_account)
+
+    session.add(created_group)
+    session.commit()
```
vkrizan, line 89: Why do we commit this early?

fstavela:
@FabriciaDinizRH 
It looks like the response doesn't contain correct GroupOut object. All I get is this:
```
$ curl -H "$IDENTITY" -H "Content-Type: application/json" -H "Accept: application/json" -d '{"name": "fero-test5", "host_ids": []}' -X POST "http://host-inventory-service.ephemeral-nphnpv.svc:8000/api/inventory/v1/groups"
idaccountorg_idnamecreated_onmodified_onhost_id
```

lib/group_repository.py
FabriciaDinizRH: Maybe @kruai or @thearifismail would know. I tried a lot of ways and this is what I found to be the cleanest. `group_query_result` has a built-in `__dict__` but it returns an extra property that is ORM-related and not a part of the actual object that would have to be removed.

lib/group_repository.py
FabriciaDinizRH: yes, but the error here is raised by the API itself, it doesn't accept invalid uuids. You can see [here](https://github.com/RedHatInsights/insights-host-inventory/blob/master/swagger/api.spec.yaml#L1532) that a predefined schema for `host_ids` is being used and that schema checks for the formatting (or at least that what I think based on the fact that I kept getting 400 responses when submitting invalid host_ids on the tests).

lib/group_repository.py
FabriciaDinizRH: that's true, I can put that after. I'll fix that

FabriciaDinizRH:
> @FabriciaDinizRH It looks like the response doesn't contain correct GroupOut object. All I get is this:
> 
> ```
> $ curl -H "$IDENTITY" -H "Content-Type: application/json" -H "Accept: application/json" -d '{"name": "fero-test5", "host_ids": []}' -X POST "http://host-inventory-service.ephemeral-nphnpv.svc:8000/api/inventory/v1/groups"
> idaccountorg_idnamecreated_onmodified_onhost_id
> ```

weird, I'll take a look at that

lib/group_repository.py
vkrizan: Commit should be handled by the context manager (`session_guard`) that these queries are wrapped with.

lib/group_repository.py
vkrizan: Thanks for explanation. In that case the check here is unnecessary, and custom `_is_valid_uuid` might be dropped. If this function would be called outside of the REST API, then it will hit an error from database side (where the UUID would be validated).

api/group.py
```diff
@@ -22,8 +24,21 @@ def get_group_list(
...
+        return flask.Response(None, status.HTTP_501_NOT_IMPLEMENTED)
+
+    created_group_id = add_group(body)
+
+    if not created_group_id:
+        flask.abort(status.HTTP_400_BAD_REQUEST, "Group could not be created.")
+
+    updated_group = add_hosts_to_group(created_group_id, body.get("host_ids"))
+
+    return flask.Response(updated_group, status.HTTP_201_CREATED)
```
vkrizan, line 41: This is the line that returns incorrect data. @FabriciaDinizRH  @fstavela 

The response is suppose to be an iterable, but it gets `dict`. https://flask.palletsprojects.com/en/2.0.x/api/#flask.Response

This gets back to serialization, from https://github.com/RedHatInsights/insights-host-inventory/pull/1276/files#r1114644955. I assume that we want a JSON here. Check the libraries and middleware.

tests/test_api_groups_create.py
```diff
@@ -0,0 +1,39 @@
+from tests.helpers.api_utils import assert_response_status
+
+
+def test_create_group_without_hosts(api_create_group, db_get_hosts_for_group):
+    group_data = {"name": "my_awesome_group", "host_ids": []}
+
+    response_status, _ = api_create_group(group_data)
+
+    assert_response_status(response_status, expected_status=201)
```
kruai, line 9: The response body should also be validated

lib/group_repository.py
kruai: I'm not sure; I made a serialization function in my PR as well, but it was explicitly structured, similar to the host serialization function. I do like how clean this is, though...

kruai (changes requested):

app/instrumentation.py
FabriciaDinizRH: I had to search what nit means, but I'm not sure I get the comment. Is it because of the `f` there?

lib/group_repository.py
FabriciaDinizRH: Apparently there's no method for validating in the library (https://docs.python.org/3/library/uuid.html#uuid.UUID). We have one in the project but it returns the UUID object instead of a string. I might create a PostGroupSchema to to that validation instead of using this function

lib/group_repository.py
```diff
@@ -18,6 +29,82 @@
...
+
+def add_hosts_to_group(group_id, host_ids):
+    with session_guard(db.session):
+        group_query_result = _get_group_from_id(group_id, get_current_identity().org_id)
+        assoc_query = HostGroupAssoc.query.filter(HostGroupAssoc.host_id.in_(host_ids))
+
+        for host_id in host_ids:
+            # check if the host doesn't already exist in the HostGroupAssoc table
+            # i.e. isn't already associated with a group
+            if _is_valid_uuid(host_id) and host_id not in assoc_query.all():
```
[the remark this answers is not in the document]
FabriciaDinizRH: got it! I will do that.

tests/test_api_groups_create.py
```diff
@@ -0,0 +1,39 @@
+from tests.helpers.api_utils import assert_response_status
+
+
+def test_create_group_without_hosts(api_create_group, db_get_hosts_for_group):
```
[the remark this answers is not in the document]
FabriciaDinizRH: @kruai yeah, I was actually going to talk to you about that. I wanted to check the hosts for the groups on the tests, but I haven't figured out a way to get the `group_id` after creating the group through the API fixture.

lib/group_repository.py
vkrizan: @kruai I've check the existing codebase for serialization. I see that we do it in two passes. First pass converts a SQLAlchemy instance of a model into `dict`, and second pass serializes it into JSON string. Couldn't this be smth that SQLAlchemy lib(s) and some Flask middleware will help with? Ideally SQLAlchemy models can hold serialization rules, and API requests would return whole models that would be serialized by a middleware.

Anyhow, if we'd want to continue with the current design of manual conversions, then I'd recommend to avoid use of "meta" func `_object_as_dict()`. `serialize_group()` should be able to handle SQLAlchemy models.

fstavela:
I've shared all of the bugs I found during testing here: https://issues.redhat.com/browse/ESSNTL-4390

PR #1277, opened 2023-02-16
feat(ESSNTL-3712): Create endpoint to remove hosts from group

kruai:
# Overview

This PR is being created to address [ESSNTL-3712](https://issues.redhat.com/browse/ESSNTL-3712).
This creates the endpoint `DELETE /groups/{group_id}/{host_id1},{host_id2},{etc}`. It allows the user to remove host-group associations so that hosts no longer appear to be in a group. It also adds a new fixture and logging functions, and updates the `DELETE /groups/{group_ids}` endpoint to return 204 instead of 200.

Update: I've also corrected the `GroupIn` component in the swagger spec to include Org ID and account number. Additionally, I added the component `NonStrictUUID` because it was really confusing having a GroupID use the HostID component (and description) for its schema. Finally, I've removed the endpoint that GETs a single group, since it's redundant and clashes with our current design.

The main thing I'm mulling over right now is: **should we return the same 404 whether it's the group or the host that isn't found? How descriptive should we be?**

## PR Checklist

- [x] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/group.py
+++ b/api/group.py
@@ -67,6 +67,6 @@ def delete_hosts_from_group(group_id, host_id_list):
     delete_count = remove_hosts_from_group(group_id, host_id_list)
 
     if delete_count == 0:
-        flask.abort(status.HTTP_404_NOT_FOUND, "Hosts not found in group.")
+        flask.abort(status.HTTP_404_NOT_FOUND, "Group or hosts not found.")
 
     return flask.Response(None, status.HTTP_204_NO_CONTENT)
--- a/api/group.py
+++ b/api/group.py
@@ -42,7 +42,7 @@ def update_group_details(group_id, group_data):
 @rbac(Permission.WRITE)
 @metrics.api_request_time.time()
 def delete_groups(group_id_list):
-    if not get_flag_value(FLAG_INVENTORY_GROUPS)[0]:
+    if not get_flag_value(FLAG_INVENTORY_GROUPS):
         return flask.Response(None, status.HTTP_501_NOT_IMPLEMENTED)
 
     delete_count = delete_group_list(group_id_list, inventory_config().host_delete_chunk_size)
@@ -61,7 +61,7 @@ def get_groups_by_id(group_id_list):
 @rbac(Permission.WRITE)
 @metrics.api_request_time.time()
 def delete_hosts_from_group(group_id, host_id_list):
-    if not get_flag_value(FLAG_INVENTORY_GROUPS)[0]:
+    if not get_flag_value(FLAG_INVENTORY_GROUPS):
         return flask.Response(None, status.HTTP_501_NOT_IMPLEMENTED)
 
     delete_count = remove_hosts_from_group(group_id, host_id_list)
--- a/api/group.py
+++ b/api/group.py
@@ -26,10 +26,6 @@ def create_group(group_data):
     pass
 
 
-def get_group(group_id):
-    pass
-
-
 def patch_group_by_id(group_id, group_data):
     pass
 
--- a/api/group.py
+++ b/api/group.py
@@ -8,6 +8,7 @@ from app import Permission
 from lib.feature_flags import FLAG_INVENTORY_GROUPS
 from lib.feature_flags import get_flag_value
 from lib.group_repository import delete_group_list
+from lib.group_repository import remove_hosts_from_group
 from lib.middleware import rbac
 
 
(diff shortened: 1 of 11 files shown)


lib/group_repository.py
```diff
@@ -19,6 +21,22 @@ def _remove_all_hosts_from_group(session, group):
...
 
+def _delete_host_group_assoc(session, assoc):
+    # TODO: Produce Kafka message upon deletion; don't delete if Kafka is not available
+    delete_query = session.query(HostGroupAssoc).filter(
+        HostGroupAssoc.group_id == assoc.group_id, HostGroupAssoc.host_id == assoc.host_id
+    )
+    delete_query.delete(synchronize_session="fetch")
+    assoc_deleted = _deleted_by_this_query(assoc)
+    if assoc_deleted:
+        delete_group_count.inc()
```
FabriciaDinizRH, line 32: should this metric be the same as for the deletion of a group? It seems misleading from my point of view (unless `delete_group_count` is related to all deletions regarding groups, not only group deletions)

kruai: That's a good point and an oversight on my part. I can make a new metric for that

api/group.py
```diff
@@ -49,12 +50,23 @@ def delete_groups(group_id_list):
...
 def get_groups_by_id(group_id_list):
     pass
 
 
+@api_operation
+@rbac(Permission.WRITE)
+@metrics.api_request_time.time()
 def delete_hosts_from_group(group_id, host_id_list):
-    pass
+    if not get_flag_value(FLAG_INVENTORY_GROUPS)[0]:
```
vkrizan, line 64: Caution that this use of the `get_flag_value` would be error prone. I'm getting back to my comment from https://github.com/RedHatInsights/insights-host-inventory/pull/1259#discussion_r1093001277

For a quick (but separate) fix, I'd suggest to rename the function and have one that would return just the flag value (single value, excl. is_fallback.

lib/group_repository.py
```diff
@@ -54,3 +72,24 @@ def delete_group_list(group_id_list, chunk_size):
...
+
+def remove_hosts_from_group(group_id, host_id_list):
+    deletion_count = 0
+    group_query = Group.query.filter(Group.org_id == get_current_identity().org_id, Group.id == group_id)
+
+    # First, find the group to make sure the org_id matches
+    found_group = group_query.one_or_none()
+
+    host_group_query = HostGroupAssoc.filter(
+        HostGroupAssoc.group_id == found_group.id, HostGroupAssoc.host_id.in_(host_id_list)
```
vkrizan, line 87: `None` case of `found_group` not covered.

kruai: Do you mean by tests? Because that's covered by `test_remove_hosts_from_nonexistent_group` and `test_remove_hosts_from_someone_elses_group`

lib/group_repository.py
kruai: OK, added some new metrics specific to host-group associations

api/group.py
kruai: Ah. Yeah, I separated the two logically so that you're passing in a string and it uses a lookup to find the fallback value (rather than a tuple), but `get_flag_value()` still returns a tuple. For my next PR, I'll make a separate function that returns the tuple and make `get_flag_value()` return just a boolean

kruai: It was really quick, so I just added it to this PR

lib/group_repository.py
vkrizan:
```
found_group = group_query.one_or_none()
```
this one might return `None`, wouldn't it?

lib/feature_flags.py
```diff
@@ -45,3 +44,25 @@ def get_flag_value(flag_name: str) -> Tuple[bool, bool]:
         )
     finally:
         return flag_value, using_fallback
+
+
+# Gets a feature flag's value from Unleash, if available.
+# Accepts a string with the name of the feature flag.
+# Returns the value of the feature flag, whether it's the fallback or real value.
+def get_flag_value(flag_name: str) -> bool:
```
vkrizan, line 52: DRY

lib/group_repository.py
kruai: Oh, interesting... this comment appears to be on an older version of the code, but isn't marked as outdated. After the `one_or_none()` call, there's this:
```
if not found_group:
        return 0
```
GitHub bug? Not sure what's going on there. But check the "Files changed" tab to see the current version of the code

lib/feature_flags.py
kruai: Oh right... I can just re-use the other function here :facepalm: 
I've pushed the refactor.

vkrizan (approved):
Thx

FabriciaDinizRH (approved):
LGTM!

PR #1285, opened 2023-03-07
Add groupName param to endpoints; remove PUT /groups

kruai:
# Overview

Does the following:
- Adds the groupName param to relevant Host and Tag endpoints
- Removes PUT /groups as per our previous discussions (it's replaced entirely by PATCH)

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


CodeHeeler (approved):
LGTM

swagger/api.spec.yaml
```diff
@@ -431,37 +433,6 @@ paths:
           description: Invalid request.
         '404':
           description: Group not found.
-    put:
```
thearifismail, line 434: why has this method being taken out?

thearifismail (commented):
Please see comments.

swagger/api.spec.yaml
kruai: Because the group decision ([on Slack](https://redhat-internal.slack.com/archives/CQFKM031T/p1677615447233499)) was to just use PATCH instead. The ADR was updated to match this as well.

api/host.py
```diff
@@ -122,6 +123,7 @@ def delete_hosts_by_filter(
     provider_type=None,
     updated_start=None,
     updated_end=None,
+    group_name=None,  # To be implemented
```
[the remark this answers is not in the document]
kruai: Not yet, since `group_name` isn't actually filtering the list of hosts yet.
If it was included in that statement, and the user made a request to this endpoint passing only `group_name`, it would simply delete the full unfiltered list of hosts (which the `if not any()` block was designed to catch).

thearifismail (dismissed):

swagger/api.spec.yaml
```diff
@@ -26,6 +26,7 @@ paths:
         - $ref: '#/components/parameters/providerType'
         - $ref: '#/components/parameters/updatedStart'
         - $ref: '#/components/parameters/updatedEnd'
+        - $ref: '#/components/parameters/groupName'
```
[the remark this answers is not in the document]
kruai: Good call. I've made a separate parameter labeled `groupNameExplicit`, which uses the name `group_name` to make it explicit that it's for the group name and not a host name.

api/host.py
```diff
@@ -69,6 +69,7 @@ def get_host_list(
...
```
FabriciaDinizRH, line 72: `group_name` will be added to `get_host_list_xjoin()` only when we have the xjoin side of things set up?

FabriciaDinizRH (approved):

api/host.py
kruai: Correct. It will exist here to keep consistency in the schema, but it will not be filterable until we have groups working in xjoin.

strider (approved):
LGTM, let's launch this.

PR #1287, opened 2023-03-08
feat(ESSNTL-3704, ESSNTL-4390): Add POST /groups endpoint

FabriciaDinizRH:
# Overview

This PR is being created to address [ESSNTL-3704](https://issues.redhat.com/browse/ESSNTL-3704) and [ESSNTL-4390](https://issues.redhat.com/browse/ESSNTL-4390).

Add the POST /groups endpoint. The payload is as follows:
```
{
     "name": "test-host",
     "host_ids": []
}
```
`name` is mandatory but `host_ids` can be an empty list.

I have improved the error messages in comparison to the first PR, but I noticed that API is the one that checks if the host_ids are valid based on the spec. In order for the API to accept invalid host_ids (and a group to be created even if the customer sends invalid ids), the spec would have to change.  More details about this issue [here](https://issues.redhat.com/browse/ESSNTL-4390) and [here](https://redhat-internal.slack.com/archives/CQFKM031T/p1676505465420629). I want to make sure that this change makes sense to everybody before going on with it.

**Other changes introduced:
 - `PatchGroupSchema` was renamed to `InputGroupSchema` to allow it to be used for validation of all group input operations
 - `created_at` and `updated_at` were renamed to `created` and `updated` in the API spec to keep the spec consistent with the host's
 - fixture `db_get_group` was renamed to `db_get_group_by_id` since `db_get_group_by_name` has been added

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/group.py
+++ b/api/group.py
@@ -17,6 +17,7 @@ from app.logging import get_logger
 from app.models import PatchGroupSchema
 from lib.feature_flags import FLAG_INVENTORY_GROUPS
 from lib.feature_flags import get_flag_value
+from lib.group_repository import add_group
 from lib.group_repository import delete_group_list
 from lib.group_repository import get_group_by_id_from_db
 from lib.group_repository import remove_hosts_from_group
@@ -37,8 +38,19 @@ def get_group_list(
     pass
 
 
+@api_operation
+@rbac(Permission.WRITE)
+@metrics.api_request_time.time()
 def create_group(group_data):
-    pass
+    if not get_flag_value(FLAG_INVENTORY_GROUPS)[0]:
+        return flask.Response(None, status.HTTP_501_NOT_IMPLEMENTED)
+
+    created_group = add_group(group_data)
+
+    if not created_group:
+        flask.abort(status.HTTP_400_BAD_REQUEST, "Group could not be created.")
+
+    return flask.Response(created_group, status.HTTP_201_CREATED)
 
 
 @api_operation
--- a/api/group.py
+++ b/api/group.py
@@ -14,16 +14,18 @@ from app.exceptions import InventoryException
 from app.instrumentation import log_patch_group_failed
 from app.instrumentation import log_patch_group_success
 from app.logging import get_logger
+from app.models import CreateGroupSchema
 from app.models import PatchGroupSchema
 from lib.feature_flags import FLAG_INVENTORY_GROUPS
 from lib.feature_flags import get_flag_value
 from lib.group_repository import add_group
-from lib.group_repository import add_hosts_to_group
 from lib.group_repository import delete_group_list
 from lib.group_repository import get_group_by_id_from_db
 from lib.group_repository import remove_hosts_from_group
 from lib.group_repository import replace_host_list_for_group
 from lib.middleware import rbac
+
+logger = get_logger(__name__)
 
 
 logger = get_logger(__name__)
@@ -46,14 +48,20 @@ def create_group(body):
     if not get_flag_value(FLAG_INVENTORY_GROUPS):
         return flask.Response(None, status.HTTP_501_NOT_IMPLEMENTED)
 
-    created_group_id = add_group(body)
+    # from Asa's PR
+    try:
+        validated_create_group_data = CreateGroupSchema().load(body)
+    except ValidationError as e:
+        logger.exception(f"Input validation error while creating group: {body}")
+        return flask.Response({"detail": str(e.messages)}, status.HTTP_400_BAD_REQUEST)
 
-    if not created_group_id:
+    created_group_id = add_group(validated_create_group_data)
+    created_group = get_group_by_id_from_db(created_group_id)
+
+    if not created_group:
         flask.abort(status.HTTP_400_BAD_REQUEST, "Group could not be created.")
 
-    updated_group = add_hosts_to_group(created_group_id, body.get("host_ids"))
-
-    return flask.Response(updated_group, status.HTTP_201_CREATED)
+    return flask.Response(build_group_response(created_group), status.HTTP_201_CREATED)
 
 
 @api_operation
--- a/api/group.py
+++ b/api/group.py
@@ -14,8 +14,7 @@ from app.exceptions import InventoryException
 from app.instrumentation import log_patch_group_failed
 from app.instrumentation import log_patch_group_success
 from app.logging import get_logger
-from app.models import CreateGroupSchema
-from app.models import PatchGroupSchema
+from app.models import GroupSchema
 from lib.feature_flags import FLAG_INVENTORY_GROUPS
 from lib.feature_flags import get_flag_value
 from lib.group_repository import add_group
@@ -50,22 +49,40 @@ def create_group(body):
         return flask.Response(None, status.HTTP_501_NOT_IMPLEMENTED)
 
     try:
-        validated_create_group_data = CreateGroupSchema().load(body)
+        validated_create_group_data = GroupSchema().load(body)
     except ValidationError as e:
         logger.exception(f"Input validation error while creating group: {body}")
-        return flask.Response({"detail": str(e.messages)}, status.HTTP_400_BAD_REQUEST)
+        return flask.Response(
+            {
+                "status": status.HTTP_406_NOT_ACCEPTABLE,
+                "title": ValidationError,
+                "detail": str(e.messages),
+                "type": "unknown",
+            },
+            status.HTTP_406_NOT_ACCEPTABLE,
+        )
 
-    created_group_id = add_group(validated_create_group_data)
+    try:
+        created_group_id = add_group(validated_create_group_data)
+    except IntegrityError:
+        logger.exception(f"A group with details {validated_create_group_data} already exists.")
+        return (
+            {
+                "status": status.HTTP_400_BAD_REQUEST,
+                "title": "Integrity error",
+                "detail": f"A group with details {validated_create_group_data} already exists.",
+                "type": "unknown",
+            },
+            status.HTTP_400_BAD_REQUEST,
+        )
+
     created_group = get_group_by_id_from_db(created_group_id)
     host_id_list = validated_create_group_data.get("host_ids")
 
     if host_id_list:
         add_hosts_to_group(created_group, host_id_list)
 
-    if not created_group:
-        flask.abort(status.HTTP_400_BAD_REQUEST, "Group could not be created.")
-
-    return flask.Response(build_group_response(created_group), status.HTTP_201_CREATED)
+    return flask_json_response(build_group_response(created_group), status.HTTP_201_CREATED)
 
 
 @api_operation
@@ -73,7 +90,7 @@ def create_group(body):
 @metrics.api_request_time.time()
 def patch_group_by_id(group_id, body):
     try:
-        validated_patch_group_data = PatchGroupSchema().load(body)
+        validated_patch_group_data = GroupSchema().load(body)
     except ValidationError as e:
         logger.exception(f"Input validation error while patching group: {group_id} - {body}")
         return ({"status": 400, "title": "Bad Request", "detail": str(e.messages), "type": "unknown"}, 400)
--- a/api/group.py
+++ b/api/group.py
@@ -52,15 +52,7 @@ def create_group(body):
         validated_create_group_data = InputGroupSchema().load(body)
     except ValidationError as e:
         logger.exception(f"Input validation error while creating group: {body}")
-        return (
-            {
-                "status": status.HTTP_400_BAD_REQUEST,
-                "title": "Validation Error",
-                "detail": str(e.messages),
-                "type": "unknown",
-            },
-            status.HTTP_400_BAD_REQUEST,
-        )
+        return error_json_response("Validation Error", str(e.messages))
 
     try:
         # Create group with validated data
@@ -79,27 +71,11 @@ def create_group(body):
 
         log_create_group_failed(logger, group_name)
         logger.exception(error_message)
+        return error_json_response("Integrity error", error_message)
 
-        return (
-            {
-                "status": status.HTTP_400_BAD_REQUEST,
-                "title": "Integrity error",
-                "detail": error_message,
-                "type": "unknown",
-            },
-            status.HTTP_400_BAD_REQUEST,
-        )
     except InventoryException as inve:
         logger.exception(inve.detail)
-        return (
-            {
-                "status": status.HTTP_400_BAD_REQUEST,
-                "title": inve.title,
-                "detail": inve.detail,
-                "type": "unknown",
-            },
-            status.HTTP_400_BAD_REQUEST,
-        )
+        return error_json_response(inve.title, inve.detail)
 
     return flask_json_response(build_group_response(created_group), status.HTTP_201_CREATED)
 
@@ -182,3 +158,7 @@ def delete_hosts_from_group(group_id, host_id_list):
         flask.abort(status.HTTP_404_NOT_FOUND, "Group or hosts not found.")
 
     return flask.Response(None, status.HTTP_204_NO_CONTENT)
+
+
+def error_json_response(title, detail, status=status.HTTP_400_BAD_REQUEST):
+    return flask_json_response({"title": title, "detail": detail}, status)
--- a/api/group.py
+++ b/api/group.py
@@ -11,17 +11,15 @@ from app import db
 from app import inventory_config
 from app import Permission
 from app.exceptions import InventoryException
-from app.instrumentation import get_control_rule
-from app.instrumentation import log_add_group_failed
-from app.instrumentation import log_add_group_succeeded
+from app.instrumentation import log_create_group_failed
+from app.instrumentation import log_create_group_succeeded
 from app.instrumentation import log_patch_group_failed
 from app.instrumentation import log_patch_group_success
 from app.logging import get_logger
-from app.models import GroupSchema
+from app.models import InputGroupSchema
 from lib.feature_flags import FLAG_INVENTORY_GROUPS
 from lib.feature_flags import get_flag_value
 from lib.group_repository import add_group
-from lib.group_repository import add_hosts_to_group
 from lib.group_repository import delete_group_list
 from lib.group_repository import get_group_by_id_from_db
 from lib.group_repository import remove_hosts_from_group
@@ -49,8 +47,9 @@ def create_group(body):
     if not get_flag_value(FLAG_INVENTORY_GROUPS):
         return flask.Response(None, status.HTTP_501_NOT_IMPLEMENTED)
 
+    # Validate group input data
     try:
-        validated_create_group_data = GroupSchema().load(body)
+        validated_create_group_data = InputGroupSchema().load(body)
     except ValidationError as e:
         logger.exception(f"Input validation error while creating group: {body}")
         return (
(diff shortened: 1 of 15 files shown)


tests/test_api_groups_create.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+
+from tests.helpers.api_utils import assert_response_status
+
+
+def test_create_group_without_hosts(api_create_group, db_get_group_by_name):
+    group_data = {"name": "my_awesome_group"}
```
CodeHeeler, line 7: "my_awesome_group"
^ the best group :)

api/group.py
```diff
@@ -35,8 +41,48 @@ def get_group_list(
...
+    except ValidationError as e:
+        logger.exception(f"Input validation error while creating group: {body}")
+        return flask.Response(
+            {
+                "status": status.HTTP_406_NOT_ACCEPTABLE,
+                "title": ValidationError,
+                "detail": str(e.messages),
+                "type": "unknown",
+            },
+            status.HTTP_406_NOT_ACCEPTABLE,
```
CodeHeeler, line 62: I'm not sure we should use a 406 here.  Usually we use a blanket 400, but if we wanted to be more granular than that since technically 400 is for bad syntax but here the syntax may be correct but the specific content is incorrect when compared to the schema, from what I'm seeing online, some use 409, some use 422, but 406 is being listed as a rarely used error that doesn't seem to match a schema validation error use case.

CodeHeeler (dismissed):
Had one question/suggestion on the specific error code used, but otherwise looks great.  Very clean code, well thought out, addresses the problems raised in the jira discussion, and well tested.  Thank you!

lib/group_repository.py
```diff
@@ -23,7 +28,43 @@
 logger = get_logger(__name__)
 
 
-def _remove_all_hosts_from_group(session: scoped_session, group: Group):
+def add_hosts_to_group(group: Group, host_id_list: List[str]) -> List[HostGroupAssoc]:
+    for host_id in host_id_list:
+        try:
+            db_create_host_group_assoc(group_id=group.id, host_id=host_id)
```
vkrizan, line 34: This call adds hosts one by one with each immediately committed to db. Can you think of a better way to handle this association? Imagine million hosts being associated for a group.

lib/group_repository.py
```diff
@@ -23,7 +28,43 @@
...
+
+def _add_group(session: scoped_session, group_name: str) -> Group:
+    current_org_id = get_current_identity().org_id
+    current_account = get_current_identity().account_number
+    new_group = Group(name=group_name, org_id=current_org_id, account=current_account)
+
+    session.add(new_group)
+    create_group_count.inc()
+
+    created_group = Group.query.filter(Group.name == group_name, Group.org_id == current_org_id).one_or_none()
```
vkrizan, line 48: Is the group committed to db before the select here is done? Is the select even needed (can we trust the `new_group` model instance when it is written to db)?

api/group.py
```diff
@@ -12,15 +13,20 @@
...
 from lib.feature_flags import get_flag_value
+from lib.group_repository import add_group
+from lib.group_repository import add_hosts_to_group
 from lib.group_repository import delete_group_list
 from lib.group_repository import get_group_by_id_from_db
 from lib.group_repository import remove_hosts_from_group
 from lib.group_repository import replace_host_list_for_group
 from lib.middleware import rbac
 
+logger = get_logger(__name__)
```
vkrizan, line 28: Duplicated on line 31.

lib/group_repository.py
kruai: The main reason we'd want to do it this way is so we can send updates to the inventory.events Kafka topic, and stop+rollback upon MQ failure. That hasn't been finalized yet, though; it might actually be best to send the list of hosts to add/remove as a list in one kafka message. What do y'all think?

lib/group_repository.py
FabriciaDinizRH: since it was within the same session, the select will return the group even if it was not committed yet (which I now see can be a problem if something's wrong and the session rollbacks). I have improved this part. 
The select is needed to get the `group_id` to associate the hosts with the group.

api/group.py
```diff
@@ -35,8 +42,62 @@ def get_group_list(
...
+        logger.exception(f"Error adding group with details {validated_create_group_data}")
+        return (
+            {
+                "status": status.HTTP_400_BAD_REQUEST,
+                "title": f"Error adding group with details {validated_create_group_data}",
+                "detail": str(e.messages),
+                "type": "unknown",
+            },
+            status.HTTP_400_BAD_REQUEST,
+        )
```
kruai, line 93: Feels like bad practice to catch any exception that could occur and then tell the user it's their fault :)
Also, if it could be any type of Exception, we don't know that it'll have the field `e.messages`.

lib/group_repository.py
```diff
@@ -23,7 +25,36 @@
...
+    for host_id in host_id_list:
+        try:
+            db_create_host_group_assoc(group_id=group.id, host_id=host_id)
+            log_host_group_add_succeeded(logger, host_id, group.id, get_control_rule())
+        except Exception:
+            log_host_group_add_failed(logger, host_id, group.id, get_control_rule())
+
+
+def _add_group(session: scoped_session, group_name: str, org_id: str) -> Group:
+    current_account = get_current_identity().account_number
```
kruai, line 38: This should probably be defined in the same place as the `org_id`. Personally, I prefer having them both defined in `add_group` and passed into `_add_group`, so that you're passing in all the fields required to make a Group.

vkrizan:
image

@FabriciaDinizRH I highly recommend to utilize Git fixup (and force-push) to clean up PRs before they get merged. https://itnext.io/git-fixup-your-commit-e1047f2fb966

tests/test_api_groups_create.py
kruai, line 7: Is this valid? It's my understanding that you're supposed to pass the full object to a POST, which would lead me to expect something like this:
```suggestion
    group_data = {"name": "my_awesome_group", "host_ids": []}
```

tests/test_api_groups_create.py
```diff
@@ -0,0 +1,49 @@
...
+
+    response_status, _ = api_create_group(group_data)
+
+    assert_response_status(response_status, expected_status=201)
```
kruai, line 11: The response data should be validated as well, to make sure that the structure is correct and that the group is being properly serialized.

tests/test_api_groups_create.py
```diff
@@ -0,0 +1,49 @@
...
+def test_create_group_with_hosts(db_create_host, api_create_group, db_get_group_by_name):
+    host1 = db_create_host()
+    host2 = db_create_host()
+    host_id_list = [str(host1.id), str(host2.id)]
+
+    group_data = {"name": "my_awesome_group", "host_ids": host_id_list}
+
+    response_status, _ = api_create_group(group_data)
+
+    assert_response_status(response_status, expected_status=201)
```
kruai, line 26: This response data should be validated as well

kruai (changes requested):

lib/group_repository.py
vkrizan: I'd recommend here to do a bach update(s). Can you elaborate on the Kafka events (through Slack or other forum)?

FabriciaDinizRH: I'll ask @kruai to weight in on this, because from what I can tell from the patch endpoint (and from all the hosts endpoints), in the future we are going to emit a kafka event for each of the assoc created, so I think we have to do that one by one.

vkrizan: @kruai Do we have requirements for these Kafka events? I think we would need to reassess their design. Sending it one by one would create a storm :cloud_with_lightning: .

vkrizan: @FabriciaDinizRH Changing how we insert it to db doesn't have an impact to how Kafka events would be produced. That would likely be a post-process after a successful assignment (unless the MQ failure is critical for this to be reverted).

api/group.py
```diff
@@ -9,19 +10,25 @@
 from app import db
 from app import inventory_config
 from app import Permission
+from app.instrumentation import get_control_rule
+from app.instrumentation import log_add_group_failed
+from app.instrumentation import log_add_group_succeeded
 from app.instrumentation import log_patch_group_failed
 from app.instrumentation import log_patch_group_success
 from app.logging import get_logger
+from app.models import CreateGroupSchema
```
thearifismail, line 19: Should rename to `GroupSchema` because `CreateGroupSchema` sounds like a method and not a class.  Furthermore there is not class or function named `GroupSchema`

app/instrumentation.py
```diff
@@ -89,6 +89,31 @@ def log_get_host_list_failed(logger):
     logger.debug("hosts not found", extra={"access_rule": get_control_rule()})
 
 
+# create group
+def log_add_group_succeeded(logger, group_id, control_rule):
```
thearifismail, line 93: Should choose either `add` or `create` group and that should be used consistently across multiple modules.  For logging "log_add_group_*" is being used but the group.py has `create_group`

lib/group_repository.py
thearifismail: I like the idea of creating hosts in batches but playing devil's advocate here.  One of the challenges inventory has is hosts in DB getting out of sync with Elasticsearch index.  When using batch additions/updates, say 100 hosts;  before all 100 hosts are added, a problem occurs.  Now accounting for what host got added and what host was not could be tricky issue to deal with.

lib/group_repository.py
```diff
@@ -23,7 +25,36 @@
...
+    new_group = Group(name=group_name, org_id=org_id, account=current_account)
```
thearifismail, line 39: `org_id` should be retrieved from the identity the same way was account to avoid problems caused by difference between the passed in org_id and the one in the identity providing account number

tests/test_api_groups_create.py
```diff
@@ -0,0 +1,49 @@
+import pytest
```
thearifismail, line 1: Need tests creating hosts by providing a group id and not providing one.
Create groups by providing non-existent-hosts

tests/test_api_groups_delete.py
```diff
@@ -18,15 +18,15 @@ def test_delete_with_invalid_group_id(api_delete_groups):
...
+def test_delete_group_ids(db_create_group, db_get_group_by_id, api_delete_groups):
     group_list = [db_create_group(f"test_group{g_index}") for g_index in range(3)]
 
     response_status, response_data = api_delete_groups([group.id for group in group_list])
 
     assert_response_status(response_status, expected_status=204)
 
     for group in group_list:
-        assert not db_get_group(group.id)
+        assert not db_get_group_by_id(group.id)
```
thearifismail, line 30: Need a test to delete a group with hosts in it.

lib/group_repository.py
vkrizan: That out-of sync problem would occur regardless of the insertion method. AFAIK Elastic taps into binary log of a database. Indexing would still need to be done for each host association.

api/group.py
FabriciaDinizRH: Yeah, I think I went to the other extreme there in trying to capture exceptions.

api/group.py
kruai: I think it's being named `CreateGroupSchema` to differentiate from `PatchGroupSchema`

tests/test_api_groups_delete.py
kruai: That's not related to the POST /groups endpoint, though

tests/test_api_groups_create.py
kruai: Since you can't specify the group_id when creating a new group, putting group_id on the POST body would be the same as adding `"foo": "bar"` to the payload. Is the goal to test whether extra fields are rejected?

api/group.py
FabriciaDinizRH: Which just made me realize we don't need both haha

api/group.py
```diff
@@ -35,16 +41,70 @@ def get_group_list(
...
+                "title": "Validation Error",
+                "detail": str(e.messages),
+                "type": "unknown",
+            },
+            status.HTTP_400_BAD_REQUEST,
+        )
+
+    try:
+        created_group = add_group(validated_create_group_data)
+        log_create_group_succeeded(logger, created_group.id, get_control_rule())
```
vkrizan, line 67: `get_control_rule()` call is repeated for each log message.

kruai (changes requested):
I just have a couple more comments!

fstavela:
I shared my findings in a Jira comment: https://issues.redhat.com/browse/ESSNTL-4390?focusedId=21931326&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21931326

vkrizan (commented):
Looks good Fabricia. The exception handling might need to be looked at before merging.
Thank you.

api/group.py
```diff
@@ -35,16 +41,71 @@ def get_group_list(
...
+        logger.exception(f"Error adding group with details {validated_create_group_data}")
+        return (
+            {
+                "status": status.HTTP_400_BAD_REQUEST,
+                "title": f"Error adding group with details {validated_create_group_data}",
+                "detail": str(e.messages),
+                "type": "unknown",
+            },
+            status.HTTP_400_BAD_REQUEST,
+        )
```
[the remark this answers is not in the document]
FabriciaDinizRH: That must've come back in some of the conflicts, removing again.

lib/group_repository.py
```diff
@@ -25,6 +28,50 @@
 logger = get_logger(__name__)
 
 
+def add_hosts_to_group(group_id: str, host_id_list: List[str]) -> List[HostGroupAssoc]:
+    current_account = get_current_identity().account_number
+
+    for host_id in host_id_list:
+        # Check if the host belongs to the current account
+        host = Host.query.filter(Host.account == current_account, Host.id == host_id).one_or_none()
```
fstavela, line 36: I think we should check for org_id here instead of account number because an account number is not required anymore so it is possible that a host's account number is `None`.

lib/group_repository.py
```diff
@@ -25,6 +28,50 @@
...
+def add_group(group_data) -> Group:
+    logger.debug("Creating a new group: %s", group_data)
+    group_name = group_data.get("name")
+    org_id = get_current_identity().org_id
+    account = get_current_identity().account_number
+
+    with session_guard(db.session):
+        new_group = Group(name=group_name, org_id=org_id, account=account)
+        db.session.add(new_group)
+        db.session.commit()
```
fstavela, line 62: Should this `commit` be here? Would it be better to commit after we add hosts to group, or even let the `session_guard` commit for us?

fstavela (commented):
@FabriciaDinizRH  Thank you for addressing all of the issues! I left just a few last comments.

lib/group_repository.py
```diff
@@ -25,6 +28,60 @@
...
+        )
+
+    # Check if any hosts in the list don't belong to the current org_id
+    host_query = Host.query.filter((Host.org_id != current_org_id) & Host.id.in_(host_id_list)).all()
+    if host_query:
+        nonexistent_hosts = [host.id for host in host_query]
+        log_host_group_add_failed(logger, nonexistent_hosts, group_id)
+        raise InventoryException(
+            title="Invalid request", detail=f"Could not find an existing host(s) with ID {host_query}."
+        )
```
fstavela, line 51:
```suggestion
    # Check if the hosts exist in Inventory and correct account
    host_query = Host.query.filter((Host.org_id == current_org_id) & Host.id.in_(host_id_list)).all()
    found_ids_set = {str(host.id) for host in host_query}
    if found_ids_set != set(host_id_list):
        nonexistent_hosts = set(host_id_list) - found_ids_set
        log_host_group_add_failed(logger, nonexistent_hosts, group_id)
        raise InventoryException(
            title="Invalid request", detail=f"Could not find an existing host(s) with ID {nonexistent_hosts}."
        )
```
What do you say about something like this? The original check query would catch hosts in different account, but it wouldn't catch non-existing hosts, which then return error "A group with host list ['290fec21-f2b1-49df-92af-3194c667cac2'] already exists.", because it will pass this and fail on DB addition.

fstavela: Also, can we move this check above the check for already assigned hosts? That way we can return this error message even if we try to add a host from different account which is already assigned to a group.

lib/group_repository.py
```diff
@@ -25,6 +28,60 @@
...
 
 
+def add_hosts_to_group(group_id: str, host_id_list: List[str]):
+    current_org_id = get_current_identity().org_id
+
+    # Check if the hosts are already associated with another group
+    assoc_query = HostGroupAssoc.query.filter(HostGroupAssoc.host_id.in_(host_id_list)).all()
+    if assoc_query:
+        taken_hosts = [assoc.host_id for assoc in assoc_query]
+        log_host_group_add_failed(logger, taken_hosts, group_id)
```
fstavela, line 38:
```suggestion
        log_host_group_add_failed(logger, host_id_list, group_id)
```
I'm not sure about this, so feel free to ignore this suggestion if you think we should only log the bad hosts.

lib/group_repository.py
FabriciaDinizRH: hahaha, that looks a lot like what I had written before changing to this version I committed. I'll change back to what you have suggested

api/group.py
```diff
@@ -37,16 +40,76 @@ def get_group_list(
...
+
+        log_create_group_failed(logger, group_name)
+        logger.exception(error_message)
+
+        return (
+            {
+                "status": status.HTTP_400_BAD_REQUEST,
+                "title": "Integrity error",
+                "detail": error_message,
+                "type": "unknown",
```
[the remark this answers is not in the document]
FabriciaDinizRH: I'm not sure. I copied the response used in `patch_group_by_id`.

lib/group_repository.py
```diff
@@ -25,6 +28,60 @@
 logger = get_logger(__name__)
 
 
+def add_hosts_to_group(group_id: str, host_id_list: List[str]):
+    current_org_id = get_current_identity().org_id
+
+    # Check if the hosts are already associated with another group
+    assoc_query = HostGroupAssoc.query.filter(HostGroupAssoc.host_id.in_(host_id_list)).all()
+    if assoc_query:
+        taken_hosts = [assoc.host_id for assoc in assoc_query]
```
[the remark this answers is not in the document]
FabriciaDinizRH: corrected that

api/group.py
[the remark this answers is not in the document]
FabriciaDinizRH: I started a discussion on slack regarding this, no one know why it exists and nothing breaks when I take it out, so I took it out.

api/group.py
```diff
@@ -37,16 +40,76 @@ def get_group_list(
...
+        if group_name in str(inte.params):
+            error_message = f"A group with name {group_name} already exists."
+        else:
+            error_message = f"A group with host list {host_id_list} already exists."
+
+        log_create_group_failed(logger, group_name)
+        logger.exception(error_message)
+
+        return (
+            {
```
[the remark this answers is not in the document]
FabriciaDinizRH: I put the responses now as flask responses and regarding the errors.
The Integrity error is already covered in the model tests and I'm not sure how to test the Inventory exception since it would be basically repeating the tests that I already have but checking for something else. The other tests for the groups endpoints also check for the response error status rather than for specific exceptions, so I followed the general trend, testing all these scenarios but checking only the final response.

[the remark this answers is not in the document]
FabriciaDinizRH: The way that we have been testing exceptions in Inventory is by catching exceptions that haven't been dealt with, so that's why I can't catch the exception in the existing tests, they test the whole API functionality and the functionality deals with its  expected exceptions.

kruai (approved):
LGTM!

CodeHeeler (approved):
LGTM. Great job responding to all the feedback. Good clean work. Thanks!

PR #1289, opened 2023-03-09
[ESSNTL-4421] Cleanup serialization

CodeHeeler:
# Overview

This PR is being created to address [ESSNTL-4421](https://issues.redhat.com/browse/ESSNTL-4421).
Refactor to reduce manual serialization for cleaner code.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/serialization.py
+++ b/app/serialization.py
@@ -101,38 +101,31 @@ def serialize_host(host, staleness_timestamps, fields=DEFAULT_FIELDS):
 
     serialized_host = {**serialize_canonical_facts(host.canonical_facts)}
 
-    if "id" in fields:
-        serialized_host["id"] = _serialize_uuid(host.id)
-    if "account" in fields:
-        serialized_host["account"] = host.account
-    if "org_id" in fields:
-        serialized_host["org_id"] = host.org_id
-    if "display_name" in fields:
-        serialized_host["display_name"] = host.display_name
-    if "ansible_host" in fields:
-        serialized_host["ansible_host"] = host.ansible_host
-    if "facts" in fields:
-        serialized_host["facts"] = serialize_facts(host.facts)
-    if "reporter" in fields:
-        serialized_host["reporter"] = host.reporter
-    if "per_reporter_staleness" in fields:
-        serialized_host["per_reporter_staleness"] = host.per_reporter_staleness
-    if "stale_timestamp" in fields:
-        serialized_host["stale_timestamp"] = stale_timestamp and _serialize_datetime(stale_timestamp)
-    if "stale_warning_timestamp" in fields:
-        serialized_host["stale_warning_timestamp"] = stale_timestamp and _serialize_datetime(stale_warning_timestamp)
-    if "culled_timestamp" in fields:
-        serialized_host["culled_timestamp"] = stale_timestamp and _serialize_datetime(culled_timestamp)
-        # without astimezone(timezone.utc) the isoformat() method does not include timezone offset even though iso-8601
-        # requires it
-    if "created" in fields:
-        serialized_host["created"] = _serialize_datetime(host.created_on)
-    if "updated" in fields:
-        serialized_host["updated"] = _serialize_datetime(host.modified_on)
-    if "tags" in fields:
-        serialized_host["tags"] = _serialize_tags(host.tags)
-    if "system_profile" in fields:
-        serialized_host["system_profile"] = host.system_profile_facts or {}
+    for field in fields:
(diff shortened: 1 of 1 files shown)


app/serialization.py
```diff
@@ -101,38 +101,29 @@ def serialize_host(host, staleness_timestamps, fields=DEFAULT_FIELDS):
...
-        serialized_host["updated"] = _serialize_datetime(host.modified_on)
-    if "tags" in fields:
-        serialized_host["tags"] = _serialize_tags(host.tags)
-    if "system_profile" in fields:
-        serialized_host["system_profile"] = host.system_profile_facts or {}
+    for field in fields:
+        if field == "id":
+            serialized_host["id"] = _serialize_uuid(host.id)
+        elif isinstance(field, str):
+            serialized_host[field] = host.field
```
vkrizan, line 108: Does `host.field` return value from the selected field?

app/serialization.py
```diff
@@ -101,38 +101,29 @@ def serialize_host(host, staleness_timestamps, fields=DEFAULT_FIELDS):
...
+        elif field == "created":
+            serialized_host["created"] = _serialize_datetime(host.created_on)
+        elif field == "updated":
+            serialized_host["updated"] = _serialize_datetime(host.modified_on)
```
vkrizan, line 112: Created and updated timestamps have the same processing.

CodeHeeler: they do, but I can't combine these two and create the serialization call from field automatically due to the names changing (updated->modified_on), that's why I reverted this to what it was in master.

app/serialization.py
```diff
@@ -101,38 +101,31 @@ def serialize_host(host, staleness_timestamps, fields=DEFAULT_FIELDS):
...
-        serialized_host["updated"] = _serialize_datetime(host.modified_on)
-    if "tags" in fields:
-        serialized_host["tags"] = _serialize_tags(host.tags)
-    if "system_profile" in fields:
-        serialized_host["system_profile"] = host.system_profile_facts or {}
+    for field in fields:
+        if field == "id":
+            serialized_host["id"] = _serialize_uuid(host.id)
+        elif isinstance(field, str):
+            serialized_host[field] = getattr(host, field)
```
vkrizan, line 108: Are list of fields coming from untrusted input? Are those sanitized? The reason I'm asking is to prevent exceptions when a nonexistent field is requested or limit where it can be reached.

app/serialization.py
```diff
@@ -101,38 +101,31 @@ def serialize_host(host, staleness_timestamps, fields=DEFAULT_FIELDS):
...
-    if "updated" in fields:
-        serialized_host["updated"] = _serialize_datetime(host.modified_on)
-    if "tags" in fields:
-        serialized_host["tags"] = _serialize_tags(host.tags)
-    if "system_profile" in fields:
-        serialized_host["system_profile"] = host.system_profile_facts or {}
+    for field in fields:
+        if field == "id":
+            serialized_host["id"] = _serialize_uuid(host.id)
+        elif isinstance(field, str):
```
vkrizan, line 107: Is this check correct? What type is a `field` in `fields`?

app/serialization.py
CodeHeeler: I would think the best way to handle preventing exceptions like these would be to leave the serialization the way it was originally where each is handled manually.  Is there another alternative?

app/serialization.py
CodeHeeler: Yes, fields by default is assigned fields=DEFAULT_FIELDS, DEFAULT_FIELDS on line 29 is a tuple of strings, so field is a string.

app/serialization.py
vkrizan: Can you think of any mechanism that would check what it can actually return? Bear in mind that `fields` can be trusted input (isn't it?).

app/serialization.py
CodeHeeler: ah, I see what you're saying.  Yes, this checks the key, I will fix it to compare to the value. Thanks for catching this!

CodeHeeler: updated

PR #1302, opened 2023-04-14
feat(ESSNTL-3698): Implement group_name host filter

kruai:
# Overview

This PR is being created to address [ESSNTL-3698](https://issues.redhat.com/browse/ESSNTL-3698).
Implements the `group_name` filter for /hosts endpoints. The format of the filter it creates in the graphql query is shown in the xjoin-search PR that implements the filter itself (https://github.com/RedHatInsights/xjoin-search/pull/191).

## PR Checklist

- [ ] Keep PR title short, ideally under 72 characters
- [ ] Descriptive comments provided in complex code blocks
- [ ] Tests: validate optimal/expected output
- [ ] Tests: validate exceptions and failure scenarios
- [ ] Tests: edge cases
- [ ] Recovers or fails gracefully during potential resource outages (e.g. DB, Kafka)
- [ ] Uses [type hinting](https://docs.python.org/3/library/typing.html), if convenient
- [ ] Documentation, if this PR changes the way other services interact with host inventory
- [x] Links to related PRs

## Secure Coding Practices Documentation Reference

You can find documentation on this checklist [here](https://github.com/RedHatInsights/secure-coding-checklist).

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #1304, opened 2023-04-17
feat(ESSNTL-3707, ESSNTL-3708): Implement groups GET endpoints

kruai:
# Overview

This PR is being created to address [ESSNTL-3707](https://issues.redhat.com/browse/ESSNTL-3707) and [ESSNTL-3708](https://issues.redhat.com/browse/ESSNTL-3708). Does the following:

- Adds filters for group name & IDs
- Implements `GET /groups` endpoint
- Implements `GET /groups/<id_list>` endpoint
- Swaps out the host_ids on Group responses for host_count (as discussed in [this slack thread](https://redhat-internal.slack.com/archives/CQFKM031T/p1681481939158569))
- Adds fixtures, utils, and tests

Because it relies on the `group_name` filter introduced in #1302, this PR is a superset of those changes. This will remain in Draft status until that PR is merged so you don't end up reviewing some of the same changes twice.

## PR Checklist

- [ ] Tests: edge cases
- [x] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


CodeHeeler (dismissed):
LGTM

FabriciaDinizRH (commented):
I left a couple of comments.
Once the tests are fixed I'll approve the PR

api/group_query.py
```diff
@@ -1,7 +1,89 @@
...
+from app.xjoin import check_pagination
+from app.xjoin import graphql_query
+from app.xjoin import groups_params_to_order
+from app.xjoin import pagination_params
+
+
+logger = get_logger(__name__)
+
+QUERY = """query Query (
+    $hostFilter: [HostFilter!],
```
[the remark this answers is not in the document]
kruai: It's `hostFilter` because it's filtering host data. What we want is group data, but the records in xjoin are all host records. So, we filter those using a hostFilter and get the group data from their `groups` field

tests/helpers/api_utils.py
```diff
@@ -386,6 +386,10 @@ def build_host_id_list_for_url(host_list_or_id):
     return str(host_list_or_id)
 
 
+def build_groups_url(group_id=None, query=None):
+    return _build_url(base_url=GROUP_URL, host_list_or_id=group_id, query=query)
```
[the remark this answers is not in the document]
kruai: I agree with you. Might be a bit out of scope, but it's bugging me too, so I might as well change it here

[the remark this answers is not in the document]
kruai: OK, I've refactored things to have clearer division between host-specific functions and more generic ones

app/xjoin.py
```diff
@@ -31,8 +31,16 @@
...
 
+GROUPS_ORDER_BY_MAPPING = {
+    None: "group",
+    "name": "group",
+    "host_ids": "count",
+}
+
 ORDER_HOW_MAPPING = {"modified_on": "DESC", "display_name": "ASC", "operating_system": "DESC"}
 
+GROUPS_ORDER_HOW_MAPPING = {"group": "DESC", "count": "DESC"}
```
fstavela, line 42: I know that this will be reworked to get groups from DB directly, but just as a reminder, should the default `group` (or `name`) ordering be ASC? Similar to `display_name` in hosts?

kruai: That makes sense. I'll change it to ASC when ordering by `group`.

tests/test_api_groups_create.py
thearifismail, line 1: Needs a test creating a new group with no hosts in it

api/filtering/filtering.py
```diff
@@ -422,12 +438,15 @@ def query_filters(
...
-    for key in filter:
-        if key == "system_profile":
-            query_filters += build_system_profile_filter(filter["system_profile"])
-        else:
-            raise ValidationException("filter key is invalid")
+    if group_ids:
+        query_filters += group_id_list_query_filter(group_ids)
+
+    if filter:
+        for key in filter:
```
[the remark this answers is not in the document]
kruai: My only change was the addition of the `if filter:` conditional, which makes it so this code does not error out if a filter is not provided.
But to answer your question, it's possible for `filter` to have more than one value. We only support "system_profile" currently, but we designed it to be flexible.

tests/test_api_groups_create.py
kruai: @thearifismail This PR is to implement GET endpoints, not host creation (which has already been done). Furthermore, the very first test in `test_api_groups_create.py` tests that exact scenario https://github.com/RedHatInsights/insights-host-inventory/blob/master/tests/test_api_groups_create.py#L11

fstavela:
@kruai With this PR, if I try to get groups by IDs, but none of them is found, I get 404.

I don't have a problem with that, but I get an empty list if I try to do the same thing on the `/hosts/<host_id>` endpoint. I think there was a team conversation about this somewhere and we decided to return an empty list instead of 404 for hosts.

It makes sense to also return an empty list of groups in this case, just to keep it consistent, but as I said, I'm OK with both approaches. Just to know how to test this, will you change it to an empty list, or keep erroring 404?

kruai:
> @kruai With this PR, if I try to get groups by IDs, but none of them is found, I get 404.
> 
> I don't have a problem with that, but I get an empty list if I try to do the same thing on the `/hosts/<host_id>` endpoint. I think there was a team conversation about this somewhere and we decided to return an empty list instead of 404 for hosts.
> 
> It makes sense to also return an empty list of groups in this case, just to keep it consistent, but as I said, I'm OK with both approaches. Just to know how to test this, will you change it to an empty list, or keep erroring 404?

@fstavela I can't recall precisely where we landed on that discussion. From what I remember... if no hosts are found for the filtered `GET /hosts` endpoint, we should return an empty list since it's querying all hosts; for `GET /hosts/<host_id>` we should return 404 since you're requesting a specific resource that doesn't exist. Does that sound correct?

fstavela:
> > @kruai With this PR, if I try to get groups by IDs, but none of them is found, I get 404.
> > I don't have a problem with that, but I get an empty list if I try to do the same thing on the `/hosts/<host_id>` endpoint. I think there was a team conversation about this somewhere and we decided to return an empty list instead of 404 for hosts.
> > It makes sense to also return an empty list of groups in this case, just to keep it consistent, but as I said, I'm OK with both approaches. Just to know how to test this, will you change it to an empty list, or keep erroring 404?
> 
> @fstavela I can't recall precisely where we landed on that discussion. From what I remember... if no hosts are found for the filtered `GET /hosts` endpoint, we should return an empty list since it's querying all hosts; for `GET /hosts/<host_id>` we should return 404 since you're requesting a specific resource that doesn't exist. Does that sound correct?

It sounds good to me, but it's not what we currently do :D Currently I get an empty list on `GET /hosts/<host_id>` if hosts are not found.

kruai:
> > > @kruai With this PR, if I try to get groups by IDs, but none of them is found, I get 404.
> > > I don't have a problem with that, but I get an empty list if I try to do the same thing on the `/hosts/<host_id>` endpoint. I think there was a team conversation about this somewhere and we decided to return an empty list instead of 404 for hosts.
> > > It makes sense to also return an empty list of groups in this case, just to keep it consistent, but as I said, I'm OK with both approaches. Just to know how to test this, will you change it to an empty list, or keep erroring 404?
> > 
> > 
> > @fstavela I can't recall precisely where we landed on that discussion. From what I remember... if no hosts are found for the filtered `GET /hosts` endpoint, we should return an empty list since it's querying all hosts; for `GET /hosts/<host_id>` we should return 404 since you're requesting a specific resource that doesn't exist. Does that sound correct?
> 
> It sounds good to me, but it's not what we currently do :D Currently I get an empty list on `GET /hosts/<host_id>` if hosts are not found.

OK, I've changed it so they return an empty list now. We can update the GET endpoints that fetch by ID to use 404 when we eventually version the API.

kruai:
Not sure why the required check is failing here - tests all pass locally. I'll rebase and re-run once our Jenkins build queue is a little less overloaded

CodeHeeler (approved):
LGTM, including the new db portions.

FabriciaDinizRH (dismissed):
LGTM!

tests/test_api_groups_get.py
```diff
@@ -30,13 +30,93 @@ def test_query_variables_group_name(db_create_group, api_get):
     assert response_data["results"][0]["id"] == str(group_id)
 
 
+@pytest.mark.parametrize(
+    "order_how_query,reverse_list",
+    [
+        (
+            "&order_how=ASC",
```
thearifismail, line 37: Shouldn't `order_how=ASC, True be tested?

tests/test_api_groups_get.py
```diff
@@ -30,13 +30,93 @@ def test_query_variables_group_name(db_create_group, api_get):
...
+            False,
+        ),
+        (
+            "&order_how=DESC",
```
thearifismail, line 41: Shouldn't `order_how=DESC, False be tested?

tests/test_api_groups_get.py
```diff
@@ -30,13 +30,93 @@ def test_query_variables_group_name(db_create_group, api_get):
...
+            True,
+        ),
+        (
+            "",
```
thearifismail, line 45: Does this return random order or nothing done to the list returned by query?

tests/test_api_groups_get.py
```diff
@@ -30,13 +30,93 @@ def test_query_variables_group_name(db_create_group, api_get):
...
+            False,
+        ),
+    ],
+)
+def test_sort_by_name(db_create_group, api_get, order_how_query, reverse_list):
```
thearifismail, line 50: Are there sorting tests really needed when the sort order is not verified in the records returned?

thearifismail (commented):
minor comments

tests/test_api_groups_get.py
kruai: Nope. When `order_how` is ASC, the expected host list should not be reversed (so `reverse_list=False`).

tests/test_api_groups_get.py
kruai: Nope. When `order_how` is DESC, the expected host list should be reversed (so `reverse_list=True`).

tests/test_api_groups_get.py
kruai: This means no `order_how` is specified in the query, so the behavior should default to `order_how=ASC` (and thus `reverse_list=False`)

tests/test_api_groups_get.py
kruai: What do you mean? The sort order is verified on line 68 for this test.

thearifismail: Do you mean this `assert response_data["results"][idx]["id"] == group_id_list[idx]`?  If yes, I think it compares the items in the list but not their exact order

kruai: It checks, in order, that each result in the response has the same ID as the corresponding host in `group_id_list`, which is already sorted in the expected order. Can you explain what's missing in more detail?

PR #1352, opened 2023-05-16
fix: Change RBAC enums

fellipeh:
# Overview

This PR is being created to address [ESSNTL-4505](https://issues.redhat.com/browse/ESSNTL-4505).

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/__init__.py
```diff
@@ -50,10 +50,14 @@
 
 
 class Permission(Enum):
-    READ = "inventory:hosts:read"
-    WRITE = "inventory:hosts:write"
-    ADMIN = "inventory:*:*"
+    HOSTS_READ = "inventory:hosts:read"
+    HOSTS_WRITE = "inventory:hosts:write"
+    HOSTS_ADMIN = "inventory:hosts:*:*"
```
fstavela, line 55: Is this a valid permissions? I've never seen a permission like this. And what would be the difference between this and `HOSTS_ALL`?

fellipeh: Thinking here make no sense to have `HOSTS_ADMIN` and `GROUPS_ADMIN` since we already have ALL for both cases.. makes more sense to have only ADMIN.. I'll change here.

thearifismail (commented):
Since I never worked on `rbac`, I never noticed the existence of `rbac-mock-data` in `utility` folder.  Shouldn't `rback-mock-data` be located in `tests/helpers` folder?

lib/middleware.py
```diff
@@ -88,7 +88,7 @@ def modified_func(*args, **kwargs):
 
             for rbac_permission in rbac_data:
                 if (
-                    rbac_permission["permission"] == Permission.ADMIN.value  # inventory:*:*
+                    rbac_permission["permission"] == Permission.HOSTS_ADMIN.value  # inventory:*:*
```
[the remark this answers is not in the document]
fellipeh: This was changed on my last commit 😃

api/host_group.py
```diff
@@ -22,7 +22,7 @@
 
 
 @api_operation
-@rbac(Permission.WRITE)
+@rbac(Permission.HOSTS_WRITE)
```
fstavela, line 25: I think `GROUPS_WRITE` is more appropriate here, since this is a Groups endpoint.

api/unleash.py
```diff
@@ -8,7 +8,7 @@
 
 
 @api_operation
-@rbac(Permission.READ)
+@rbac(Permission.HOSTS_READ)
```
fstavela, line 11: This probably doesn't matter, but `GROUPS_READ` would probably be better here.

Maybe not even that, maybe there are some RBAC roles for feature flags that we should use here instead. I don't know.

kruai: I actually have no idea what kind of permission would make the most sense here. Should it be a secret to know whether we have a FF enabled or not?

thearifismail:
> Since I never worked on `rbac`, I never noticed the existence of `rbac-mock-data` in `utility` folder. Shouldn't `rback-mock-data` be located in `tests/helpers` folder?

To check it, I suggest moving the folder to `tests/helpers` folder and run tests.  You may have to update the path in test files.  The only reason I can think of putting rback-mocks in the utility folder could issues with python paths resolution, which I hope is not the case.

api/unleash.py
fstavela: That's a good point. So maybe we don't need any RBAC restrictions on that endpoint?

kruai:
This had all PR checks pass, so I'm going to merge it once it's done rebasing.

api/unleash.py
kruai: I think removing the restriction makes the most sense, yeah.

kruai: @fellipeh This conversation was marked as resolved, but the issue was not addressed

thearifismail (dismissed):
LGTM

kruai (approved):

kruai:
The PR checks passed earlier, so I'm going to go ahead and merge. The PR check queue is really long rn

PR #1373, opened 2023-05-25
fix(ESSNTL-3701): Fixes for resource-types endpoints

kruai:
# Overview

This PR is being created as a follow-up to #1342 to address some bugs discovered in [ESSNTL-3701](https://issues.redhat.com/browse/ESSNTL-3701):

- Updates the resource-types groups base path
- Adds `per_page` to the links in the resource-types endpoints
- Defaults `per_page` to 10 for resource-types endpoints
- Removes ordering params from resource-types endpoints

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


api/resource_query.py
```diff
@@ -1,6 +1,6 @@
 from math import ceil
 
-RESOURCES_TYPES_GROUPS_PATH = "/inventory/v1/resource-types/inventory-groups/"
+RESOURCES_TYPES_GROUPS_PATH = "/inventory/v1/resource-types/inventory-groups"
 
 
 def get_resources_types():
```
fstavela, line 6: Should the returned `count` be 1? I'm not sure about this, but since the `count` field is inside each resource type, to me it looks like it should indicate how many resources of that type are there. In other words, how many inventory groups are available. The number of different resource types is then indicated in the `meta.count` section of the `/resource-types` endpoint response.

Otherwise, the `count` in the data section would be redundant and it would be the same in each resource type, if we had more than 1 (something else then Groups).

swagger/api.spec.yaml
```diff
@@ -568,7 +568,7 @@ paths:
       security:
         - ApiKeyAuth: []
       parameters:
-        - $ref: '#/components/parameters/perPageParam'
+        - $ref: '#/components/parameters/resourceTypesPerPageParam'
         - $ref: '#/components/parameters/pageParam'
         - $ref: '#/components/parameters/groupOrderByParam'
```
fstavela, line 573: These should be removed

fstavela:
The response type of `/resource-types/inventory-groups` data array objects shouldn't be `GroupOut` on line 1697 in `api.spec.yaml`. The expected output looks different in ADR (just `{"value": "<group-name>"}`). We should probably introduce something like `ResourceTypesGroupsOut`.

fstavela (approved):
LGTM

fellipeh (approved):
lgtm

PR #1377, opened 2023-05-31
[ESSNTL 4873]: Add assignment_rules table and ORM

FabriciaDinizRH:
# Overview

This PR is being created to address [ESSNTL-4873](https://issues.redhat.com/browse/ESSNTL-4873).
This PR creates the `assignment_rules` table and ORM as well as add a few tests.

As of today, examples for the filter structure have not been provided and help with that is appreciated if folks feel like it is necessary at this point. 

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


kruai:
Note: We had to merge a new migration to fix a data issue, so references to the down-revision (`b48ad9e4c8c5`) will need to be changed to `8277117fa057`.

tests/test_models.py
```diff
@@ -646,3 +646,26 @@ def test_add_delete_host_group_happy(
...
         InputGroupSchema().load(data)
+
+
+def test_create_assignment_rule(db_create_group, db_create_assignment_rule, db_get_assignment_rule):
+    group_name = "MainGroup"
+    group = db_create_group(group_name)
+
+    # I'm leaving the filter as this dummy because I couldn't quite understand how to format it, help is appreciated
+    # the intended format is in the jira card
+    ar = db_create_assignment_rule("default assignment", group.id, {"facts": {"fqdn": "54321"}})
```
kruai, line 657: I elaborated a bit about my idea for the structure in the Jira, but it shouldn't make a difference for these tests, as long as they're in JSON format.

kruai: Side note: I think "dummy" is a word that may be considered insensitive, and is recommended that we use `placeholder` or `sample` instead

app/models.py
```diff
@@ -478,6 +478,58 @@ def __init__(
     group_id = db.Column(UUID(as_uuid=True), ForeignKey("groups.id"), primary_key=True)
 
 
+class AssignmentRule(db.Model):
+    __tablename__ = "assignment_rules"
+    __table_args__ = (Index("idxassrulesorgid", "org_id"),)
```
kruai, line 483: hehe (I think this is the correct way to abbreviate it though, no need to modify)

migrations/versions/1689cda89252_add_assignment_rules_table.py
```diff
@@ -0,0 +1,43 @@
+"""Add assignment_rules table
+
+Revision ID: 1689cda89252
+Revises: b48ad9e4c8c5
```
kruai, line 4:
```suggestion
Revises: 8277117fa057
```

migrations/versions/1689cda89252_add_assignment_rules_table.py
```diff
@@ -0,0 +1,43 @@
...
+Create Date: 2023-05-30 18:06:39.983078
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "1689cda89252"
+down_revision = "b48ad9e4c8c5"
```
kruai, line 14:
```suggestion
down_revision = "8277117fa057"
```

kruai (changes requested):
Looking good - just a few suggestions

app/models.py
```diff
@@ -478,6 +478,58 @@ def __init__(
...
+    id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    org_id = db.Column(db.String(36), nullable=False)
+    account = db.Column(db.String(10))
+    name = db.Column(db.String(255), nullable=False)
+    description = db.Column(db.String(255))
+    group_id = db.Column(UUID(as_uuid=True), ForeignKey("groups.id"))
+    filter = db.Column(JSONB, nullable=False)
+    enabled = db.Column(db.Boolean(), default=True)
+    created_on = db.Column(db.DateTime(timezone=True), default=_time_now)
+    updated_on = db.Column(db.DateTime(timezone=True), default=_time_now, onupdate=_time_now)
```
fstavela, line 530: Just to keep it consistent with other DB tables, I would rename this to `modified_on`.

Although, in the API we use `updated`, so there already is some level of inconsistency. @kruai @FabriciaDinizRH What do you think, should we use `updated_on` or `modified_on`?

fstavela (changes requested):
Just a small detail, otherwise looks good.

migrations/versions/1689cda89252_add_assignment_rules_table.py
kruai: Never mind - we undid that change, so I'm resolving this

migrations/versions/1689cda89252_add_assignment_rules_table.py
kruai: Never mind - we undid that change, so I'm resolving this

app/models.py
kruai: Good call. I don't think we should do `updated_on`, since that's a third variation we'll have to keep track of. I think `modified_on` makes the most sense for consistency across DB tables. I'd definitely hear an argument for `updated` as well though if y'all prefer that

app/models.py
```diff
@@ -478,6 +478,58 @@ def __init__(
...
+        if not filter:
+            raise ValidationException("Assignment rule filter cannot be null.")
+
+        self.org_id = org_id
+        self.account = account
+        self.name = name
+        self.group_id = group_id
+        self.filter = filter
+
+    def update_modified_on(self):
```
thearifismail, line 508: Adding introducing a new attribute after the object has been initialized does not sound right.  The `__init__()` should have `self.modified_on = None` and `self.created_on = None`.

thearifismail (commented):
over looks good.  some minor comments.

tests/helpers/db_utils.py
```diff
@@ -97,6 +98,21 @@ def db_group_dict(**values):
...
 
 
+def db_assignment_rule(**values):
+    data = db_assignment_rule_dict(**values)
+    return AssignmentRule(**data)
+
+
+def db_assignment_rule_dict(**values):
+    data = {**values}
+    if "org_id" in values:
```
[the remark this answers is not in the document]
kruai: It's the same as what we do in other creation functions, for the sake of simplicity when writing tests. `org_id` is required to exist on the object, but we don't want to be required to supply it every time we create one; you can pass in an `org_id` to override the value if you need, but otherwise it just defaults to the one in USER_IDENTITY

app/models.py
```diff
@@ -478,6 +478,58 @@ def __init__(
...
+        self.modified_on = _time_now()
+
+    def update(self, input_ar):
```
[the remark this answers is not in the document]
kruai: I think it's named `input_ar` to be explicit that it's the assignment rule that's being input, rather than the assignment rule that's being updated. Its naming is also consistent with every other `update` function in `models.py`.

app/models.py
kruai: That's fair. The only reason we have it for Group is because there were certain situations where we needed the `modified_on` date to updated even when the Group object didn't technically change (for instance, adding or removing a Host to a Group). I don't think that applies to assignment rules, so I'll remove this method.

thearifismail (approved):
LGTM!

jpramos123 (approved):
LGTM!

Tested and table is created correctly  and tests are passing

kruai (approved):

PR #1380, opened 2023-06-06
Replace empty-string host.groups values with empty arrays instead

kruai:
# Overview

This PR is being created to address an issue where some hosts created in the early stages of Inventory Groups have host.groups set to `""`. If a host isn't associated with a group, its default value should be `[]`.
This just creates a new alembic migration to update the existing errant data.


thearifismail (dismissed):
LGTM

thearifismail (approved):
LGTM

fstavela:
@kruai I'll need to adjust IQE tests for this. The DB schema tests don't expect `groups` to not be nullable.

PR #1386, opened 2023-06-09
fix(ESSNTL-4458): Fix too old date on filter

fellipeh:
# Overview

This PR is being created to address [ESSNTL-4458](https://issues.redhat.com/browse/ESSNTL-4458).

Elastic Search doesn't allow date before 1970, returned message from ES:
```
date[0199-03-04T04:56:02Z] is before the epoch in 1970 and cannot be stored in nanosecond resolution
```
so,  If the date args is before 1970, remove from filters query.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -306,19 +306,19 @@ def build_registered_with_filter(registered_with):
     return ({"OR": prs_list},)
 
 
-def _build_modified_on_filter(updated_start: str = None, updated_end: str = None) -> Tuple:
+def _build_modified_on_filter(updated_start: str, updated_end: str) -> Tuple:
     updated_start_date = parser.isoparse(updated_start) if updated_start else None
     updated_end_date = parser.isoparse(updated_end) if updated_end else None
 
     if updated_start_date and updated_end_date and updated_start_date > updated_end_date:
         raise ValueError("updated_start cannot be after updated_end.")
     modified_on_filter = {}
-    if updated_start_date:
+    if updated_start_date and updated_start_date.year > 1970:
         modified_on_filter["gte"] = updated_start_date.isoformat()
-    if updated_end_date:
+    if updated_end_date and updated_end_date.year > 1970:
         modified_on_filter["lte"] = updated_end_date.isoformat()
 
-    return ({"modified_on": modified_on_filter},)
+    return ({"modified_on": modified_on_filter},) if modified_on_filter else ()
 
 
 def build_tag_query_dict_tuple(tags):
@@ -450,13 +450,7 @@ def query_filters(
     if provider_id:
         query_filters += ({"provider_id": {"eq": provider_id.casefold()}},)
     if updated_start or updated_end:
-        # Need to check if date is less then 1970, since ElasticSearch doesn't
-        # allow timestamps before above this year. See ticket #ESSNTL-4458
-        if parser.isoparse(updated_start).year > 1970 or parser.isoparse(updated_end).year > 1970:
-            query_filters += _build_modified_on_filter(
-                updated_start=updated_start if parser.isoparse(updated_start).year > 1970 else None,
-                updated_end=updated_end if parser.isoparse(updated_end).year > 1970 else None,
-            )
+        query_filters += _build_modified_on_filter(updated_start, updated_end)
     if group_name:
         query_filters += ({"group": {"name": string_contains_lc(group_name)}},)
     if group_ids:
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -451,9 +451,12 @@ def query_filters(
         query_filters += ({"provider_id": {"eq": provider_id.casefold()}},)
     if updated_start or updated_end:
         # Need to check if date is less then 1970, since ElasticSearch doesn't
-        # allow query above this year. See ticket #ESSNTL-4458
-        if parser.isoparse(updated_start).year > 1970 and parser.isoparse(updated_end).year > 1970:
-            query_filters += _build_modified_on_filter(updated_start, updated_end)
+        # allow timestamps before above this year. See ticket #ESSNTL-4458
+        if parser.isoparse(updated_start).year > 1970 or parser.isoparse(updated_end).year > 1970:
+            query_filters += _build_modified_on_filter(
+                updated_start=updated_start if parser.isoparse(updated_start).year > 1970 else None,
+                updated_end=updated_end if parser.isoparse(updated_end).year > 1970 else None,
+            )
     if group_name:
         query_filters += ({"group": {"name": string_contains_lc(group_name)}},)
     if group_ids:
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -450,7 +450,10 @@ def query_filters(
     if provider_id:
         query_filters += ({"provider_id": {"eq": provider_id.casefold()}},)
     if updated_start or updated_end:
-        query_filters += _build_modified_on_filter(updated_start, updated_end)
+        # Need to check if date is less then 1970, since ElasticSearch doesn't
+        # allow query above this year. See ticket #ESSNTL-4458
+        if parser.isoparse(updated_start).year > 1970 and parser.isoparse(updated_end).year > 1970:
+            query_filters += _build_modified_on_filter(updated_start, updated_end)
     if group_name:
         query_filters += ({"group": {"name": string_contains_lc(group_name)}},)
     if group_ids:
--- a/tests/test_api_hosts_get.py
+++ b/tests/test_api_hosts_get.py
@@ -474,3 +474,14 @@ def test_get_hosts_contains_invalid_on_string_not_array(patch_xjoin_post, api_ge
 
 def test_get_hosts_timestamp_invalid_value_graceful_rejection(patch_xjoin_post, api_get):
     assert 400 == api_get(build_hosts_url(query="?filter[system_profile][last_boot_time][eq]=foo"))[0]
+
+
+@pytest.mark.parametrize(
+    "query_params",
+    ("?updated_start=0199-03-04T04:56:02.000Z&updated_end=0199-03-04T04:56:02.000Z"),
+)
+def test_get_hosts_too_old_date_params(query_params, api_get):
+    url = build_hosts_url(query=query_params)
+
+    response_code, _ = api_get(url)
+    assert_response_status(response_code, 200)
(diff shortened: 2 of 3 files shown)


tests/test_api_hosts_get.py
```diff
@@ -474,3 +474,14 @@ def test_get_hosts_contains_invalid_on_string_not_array(patch_xjoin_post, api_ge
 
 def test_get_hosts_timestamp_invalid_value_graceful_rejection(patch_xjoin_post, api_get):
     assert 400 == api_get(build_hosts_url(query="?filter[system_profile][last_boot_time][eq]=foo"))[0]
+
+
+@pytest.mark.parametrize(
+    "query_params",
+    ("?updated_start=0199-03-04T04%3A56%3A02.000Z&updated_end=0199-03-04T04%3A56%3A02.000Z"),
```
thearifismail, line 481: For readability, add comment containing the `update_start` and `update_end` times in plain text, i.e. decoded values.

thearifismail (commented):
See the comment

api/filtering/filtering.py
```diff
@@ -435,7 +435,10 @@ def query_filters(
     if provider_id:
         query_filters += ({"provider_id": {"eq": provider_id.casefold()}},)
     if updated_start or updated_end:
-        query_filters += _build_modified_on_filter(updated_start, updated_end)
+        # Need to check if date is less then 1970, since ElasticSearch doesn't
+        # allow query above this year. See ticket #ESSNTL-4458
```
kruai, line 439:
```suggestion
        # allow timestamps before this year. See ticket #ESSNTL-4458
```

api/filtering/filtering.py
```diff
@@ -435,7 +435,10 @@ def query_filters(
...
+        if parser.isoparse(updated_start).year > 1970 and parser.isoparse(updated_end).year > 1970:
+            query_filters += _build_modified_on_filter(updated_start, updated_end)
```
kruai, line 441: If `updated_start` is set to 1965 and `updated_end` is set to 2005, this code won't add either filter. I think that only the invalid filter should be dropped, not both

tests/test_api_hosts_get.py
```diff
@@ -474,3 +474,14 @@ def test_get_hosts_contains_invalid_on_string_not_array(patch_xjoin_post, api_ge
...
+)
+def test_get_hosts_too_old_date_params(query_params, api_get):
+    url = build_hosts_url(query=query_params)
+
+    response_code, _ = api_get(url)
+    assert_response_status(response_code, 200)
```
kruai, line 487: Since our xjoin requests are mocked, this tests nothing - the test would still pass without the new code changes. In order to test this, I recommend using the `graphql_query_empty_response` fixture and validating that only the proper filters were added to the xjoin request. There are a lot of examples of this in `test_xjoin.py`; let me know if you need any more details.

kruai (changes requested):

fstavela:
Based on the tests results, we are getting errors 500 whenever we try to one of the `updated` filters without the other. For example, if I just want to filter by `updated_start` and I don't provide `updated_end`.

tests/test_xjoin.py
```diff
@@ -284,6 +284,65 @@ def test_query_variables_provider_id(mocker, graphql_query_empty_response, api_g
...
+            "fields": mocker.ANY,
+        },
+        mocker.ANY,
+    )
+
+    # testing updated_end with too old.
+    url = build_hosts_url(query="?updated_start=2001-03-04T04:56:02.000Z&updated_end=0199-03-04T04:56:02.000Z")
+    response_status, response_data = api_get(url)
+
+    assert response_status == 200
```
kruai, line 311: If the `updated_end` date is before the `updated_start` date, I'd actually expect it to respond with a 403 saying `updated_start cannot be after updated_end`. Otherwise, we're showing them records from after 2001, when they said they don't want to see records from before the year 199

fellipeh: In this particular case, the `updated_end` is not valid (is less than 1970) so, it's removed from the query, before, the checking if the `updated_end` is before.  If we pass `updated_end` > 1970  and  before `updated_start`  will return and Value error, see here: https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/filtering/filtering.py#L313

thearifismail: A test is needed with end date before the start date and to verify 403 is returned in response.

kruai: @fellipeh I know it's removed from the query on the backend to bypass the elasticsearch restriction, but it introduces inconsistency from the behavior that the user sees. Basically, it would allow the user to set the `updated_end` to an earlier date than `updated_start` if they use a date earlier than 1970, but it would error out otherwise. It's definitely an edge case but I feel like we should be consistent with not allowing the end date to be before the start date

jpramos123: @fellipeh @kruai I agree we need to validate and make sure `update_end` is before `update_start`.

api/filtering/filtering.py
```diff
@@ -435,7 +435,13 @@ def query_filters(
     if provider_id:
         query_filters += ({"provider_id": {"eq": provider_id.casefold()}},)
     if updated_start or updated_end:
-        query_filters += _build_modified_on_filter(updated_start, updated_end)
+        # Need to check if date is less then 1970, since ElasticSearch doesn't
+        # allow timestamps before above this year. See ticket #ESSNTL-4458
+        if parser.isoparse(updated_start).year > 1970 or parser.isoparse(updated_end).year > 1970:
+            query_filters += _build_modified_on_filter(
+                updated_start=updated_start if parser.isoparse(updated_start).year > 1970 else None,
+                updated_end=updated_end if parser.isoparse(updated_end).year > 1970 else None,
```
jpramos123, line 443: Do we need this redundant validation here?
```suggestion
                updated_start=updated_start
                updated_end=updated_end
```

api/filtering/filtering.py
```diff
@@ -435,7 +435,13 @@ def query_filters(
     if provider_id:
         query_filters += ({"provider_id": {"eq": provider_id.casefold()}},)
     if updated_start or updated_end:
-        query_filters += _build_modified_on_filter(updated_start, updated_end)
+        # Need to check if date is less then 1970, since ElasticSearch doesn't
+        # allow timestamps before above this year. See ticket #ESSNTL-4458
+        if parser.isoparse(updated_start).year > 1970 or parser.isoparse(updated_end).year > 1970:
```
jpramos123, line 440: Can we add a validation statement to make sure `update_start` is not before `update_end`?

If it is, maybe return an error message?

jpramos123 (changes requested):
Added some suggestions!

api/filtering/filtering.py
kruai: We already have that validation; it happens inside `_build_modified_on_filter` here: https://github.com/RedHatInsights/insights-host-inventory/blob/sp-update/api/filtering/filtering.py#L314

api/filtering/filtering.py
fellipeh: @jpramos123 yes, because we need to pass `None` to the parameter to avoid to create an filter with that, resulting with and empty: `modified_on: {}` on graphql query (in case both params are None).

tests/test_xjoin.py
fellipeh: @jpramos123 it's already validated here: https://github.com/RedHatInsights/insights-host-inventory/blob/sp-update/api/filtering/filtering.py#L314

api/filtering/filtering.py
kruai: This logic could instead be put inside `_build_modified_on_filter`, which would be a bit cleaner (and the code would be simpler). The dates already get isoparsed at the beginning of that function, so you could just put the 1970 check in [this condition](https://github.com/RedHatInsights/insights-host-inventory/blob/master/api/filtering/filtering.py#L316), or between the `start > end` check and the part where the `modified_on_filter` gets populated

kruai: I still think this suggestion would simplify the code and make the behavior more consistent ^

tests/test_xjoin.py
kruai: My above suggestion still stands; the request in this should return HTTP 400 since `updated_end` is before `updated_start`

kruai (changes requested):

tests/test_xjoin.py
```diff
@@ -284,6 +284,65 @@ def test_query_variables_provider_id(mocker, graphql_query_empty_response, api_g
...
 
 
+def test_query_variables_updated_too_old_timestamp(mocker, graphql_query_empty_response, api_get):
+    # testing both timestamp with too old.
+    url = build_hosts_url(query="?updated_start=0199-03-04T04:56:02.000Z&updated_end=0199-03-04T04:56:02.000Z")
+    response_status, response_data = api_get(url)
+
+    assert response_status == 200
+
+    graphql_query_empty_response.graphql_query_with_response(
```
kruai, line 294:
```suggestion
    graphql_query_empty_response.assert_called_once_with(
```

tests/test_xjoin.py
```diff
@@ -284,6 +284,65 @@ def test_query_variables_provider_id(mocker, graphql_query_empty_response, api_g
...
+        {
+            "order_by": mocker.ANY,
+            "order_how": mocker.ANY,
+            "limit": mocker.ANY,
+            "offset": mocker.ANY,
+            "filter": (mocker.ANY, {"modified_on": {"gte": "2001-03-04T04:56:02+00:00"}}),
+            "fields": mocker.ANY,
+        },
+        mocker.ANY,
+    )
```
kruai, line 324: Since the request resulted in an HTTP 400, the code should not be making any graphql queries.

tests/test_xjoin.py
```diff
@@ -284,6 +284,65 @@ def test_query_variables_provider_id(mocker, graphql_query_empty_response, api_g
...
+
+    # testing updated_start with too old.
+    url = build_hosts_url(query="?updated_start=0199-03-04T04:56:02.000Z&updated_end=2010-03-04T04:56:02.000Z")
+    response_status, response_data = api_get(url)
+
+    assert response_status == 200
+
+    graphql_query_empty_response.graphql_query_with_response(
```
kruai, line 332:
```suggestion
    graphql_query_empty_response.assert_called_once_with(
```

kruai (changes requested):
The implementation looks good! Just had a few more comments on the test; `graphql_query_with_response` doesn't make any assertions

kruai (dismissed):
LGTM!

kruai (approved):

thearifismail (approved):
LGTM

FabriciaDinizRH (approved):

jpramos123 (approved):
LGTM

PR #1409, opened 2023-06-27
ESSNTL-4929: fix for updating time when a group is touched

thearifismail:
# Overview

This PR is being created to address [ESSNTL-4929](https://issues.redhat.com/browse/ESSNTL-4929).
This PR fixes the bug and adds a test to verify the fix.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/host_group.py
+++ b/api/host_group.py
@@ -17,7 +17,6 @@ from lib.feature_flags import get_flag_value
 from lib.group_repository import add_hosts_to_group
 from lib.group_repository import get_group_by_id_from_db
 from lib.group_repository import remove_hosts_from_group
-from lib.group_repository import update_group_update_time
 from lib.host_repository import get_host_list_by_id_list_from_db
 from lib.middleware import rbac
 from lib.middleware import rbac_group_id_check
@@ -55,7 +54,6 @@ def add_host_list_to_group(group_id, body, rbac_filter=None):
     # Next, add the host-group associations
     if host_id_list is not None:
         add_hosts_to_group(group_id, body, current_app.event_producer)
-        update_group_update_time(group_id)
 
     updated_group = get_group_by_id_from_db(group_id)
     log_host_group_add_succeeded(logger, host_id_list, group_id)
@@ -76,5 +74,4 @@ def delete_hosts_from_group(group_id, host_id_list, rbac_filter=None):
     if delete_count == 0:
         abort(status.HTTP_404_NOT_FOUND, "Group or hosts not found.")
 
-    update_group_update_time(group_id)
     return Response(None, status.HTTP_204_NO_CONTENT)
--- a/api/host_group.py
+++ b/api/host_group.py
@@ -17,6 +17,7 @@ from lib.feature_flags import get_flag_value
 from lib.group_repository import add_hosts_to_group
 from lib.group_repository import get_group_by_id_from_db
 from lib.group_repository import remove_hosts_from_group
+from lib.group_repository import update_group_update_time
 from lib.host_repository import get_host_list_by_id_list_from_db
 from lib.middleware import rbac
 from lib.middleware import rbac_group_id_check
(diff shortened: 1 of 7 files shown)


FabriciaDinizRH (commented):
the code looks ok but some tests are breaking

api/host_group.py
```diff
@@ -54,6 +55,7 @@ def add_host_list_to_group(group_id, body, rbac_filter=None):
     # Next, add the host-group associations
     if host_id_list is not None:
         add_hosts_to_group(group_id, body, current_app.event_producer)
+        update_group_update_time(group_id)
```
kruai, line 58: The group's timestamp should be updated in the same transaction as when the hosts are added to it. So, the change should happen within the `session_guard` inside `add_hosts_to_group`.

api/host_group.py
```diff
@@ -74,4 +76,5 @@ def delete_hosts_from_group(group_id, host_id_list, rbac_filter=None):
     if delete_count == 0:
         abort(status.HTTP_404_NOT_FOUND, "Group or hosts not found.")
 
+    update_group_update_time(group_id)
```
kruai, line 79: Same as above: the group's timestamp should be updated within the same transaction as when the hosts are removed, so it should happen within `remove_hosts_from_group`.

lib/group_repository.py
```diff
@@ -255,3 +257,11 @@ def patch_group(group: Group, patch_data: dict, event_producer: EventProducer):
...
         added_host_uuids = [str(host_id) for host_id in new_host_ids]
         _produce_host_update_events(event_producer, added_host_uuids, [group_id])
+
+
+def update_group_update_time(group_id: str) -> Group:
+    with session_guard(db.session):
+        group = get_group_by_id_from_db(group_id)
+        group.modified_on = datetime.now(timezone.utc)
+        db.session.add(group)
+        db.session.commit()
```
kruai, line 267: We already have a function that updates the `modified_on` date. All you need to do is call `group.update_modified_on()`; see the usage [here](https://github.com/RedHatInsights/insights-host-inventory/blob/master/lib/group_repository.py#L250) for reference

(Also, this function doesn't return a `Group` object even though the type hint says it does)

tests/test_api_host_group.py
```diff
@@ -188,6 +188,35 @@ def test_add_host_list_with_one_associated_host_to_group(
...
+    assert updated_group["host_count"] == 4
+
+    # compare the two update times
+    from dateutil.parser import parse
+
+    first_update = parse(group["updated"])
+    second_update = parse(updated_group["updated"])
+    assert second_update > first_update
+
+
```
kruai, line 219: We already have multiple tests that validate adding hosts to a group. Can you just add the "updated" check to one of those tests, instead of duplicating it?
You should also add that check to one of the tests for `delete_hosts_from_group` to make sure that's covered.

kruai (changes requested):

lib/group_repository.py
```diff
@@ -255,3 +265,9 @@ def patch_group(group: Group, patch_data: dict, event_producer: EventProducer):
 
         added_host_uuids = [str(host_id) for host_id in new_host_ids]
         _produce_host_update_events(event_producer, added_host_uuids, [group_id])
+
+
+def _update_group_update_time(group_id: str) -> Group:
+    group = get_group_by_id_from_db(group_id)
+    group.update_modified_on()
+    return group
```
kruai, line 273: A couple things:

- Since we always need to add the group to the session, we might as well refactor that into this function.
- Do we actually need to do `db.session.flush()`? If so, we should put that in this function as well; if not, we should remove it.
- Now that there's a new function for this, we should refactor the code that was originally updating the group timestamp (https://github.com/RedHatInsights/insights-host-inventory/blob/master/lib/group_repository.py#L250-L251)

kruai (changes requested):
Looking better; I've just got a couple more suggestions

kruai (approved):
As I mentioned in my previous review, I'd prefer if we switched out my original timestamp-updating code to reference your new function: https://github.com/RedHatInsights/insights-host-inventory/blob/master/lib/group_repository.py#L250-L251
But, I'm not going to hold up the PR just for that.

PR #1427, opened 2023-07-18
Allow multiple values for group_name filter param

kruai:
# Overview

This PR is being created to address an oversight we noticed in [this comment thread of the onboarding doc](https://docs.google.com/document/d/1O87MHvKlPEoH_DHwZw4htvwYlkGMiS-LwLQ_qKsZ5J0/edit?disco=AAAA0srgq8A). Right now, we only allow filtering on one `group_name`, but we need to allow multiple values for it to work correctly. The UI is already set up to send multiple values for this field, so there shouldn't be any compatibility risk if this is executed correctly.

Basically, if we do `GET /hosts?group_name=group1&group_name=group2`, then it should return hosts that are in either group1 or group2.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -452,7 +452,7 @@ def query_filters(
     if updated_start or updated_end:
         query_filters += _build_modified_on_filter(updated_start, updated_end)
     if group_name:
-        query_filters += ({"group": {"name": string_contains_lc(group_name)}},)
+        query_filters += ({"OR": [{"group": {"name": string_contains_lc(name)}} for name in group_name]},)
     if group_ids:
         query_filters += group_id_list_query_filter(group_ids)
 
--- a/swagger/api.spec.yaml
+++ b/swagger/api.spec.yaml
@@ -26,7 +26,7 @@ paths:
         - $ref: '#/components/parameters/providerType'
         - $ref: '#/components/parameters/updatedStart'
         - $ref: '#/components/parameters/updatedEnd'
-        - $ref: '#/components/parameters/groupNameExplicit'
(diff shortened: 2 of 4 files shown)


api/filtering/filtering.py
```diff
@@ -452,7 +452,7 @@ def query_filters(
     if updated_start or updated_end:
         query_filters += _build_modified_on_filter(updated_start, updated_end)
     if group_name:
-        query_filters += ({"group": {"name": string_contains_lc(group_name)}},)
+        query_filters += ({"OR": [{"group": {"name": string_contains_lc(name)}} for name in group_name]},)
```
thearifismail, line 455: shouldn't the same facility be available for `group_ids`, i.e. using multiple `group_ids` for finding hosts?

thearifismail (commented):
LGTM, except one question/comment.

api/filtering/filtering.py
kruai: I'm not opposed to making a Jira for that if there's a use case for it; when would it be used?
For comparison, we need this for `group_name` because the UI is already trying to filter on multiple values, and it's needed for the host table to be filtered correctly according to the UX mocks.

thearifismail (approved):
LGTM

FabriciaDinizRH (approved):

PR #1439, opened 2023-07-27
feat(ESSNTL-5111): Update OpenAPI schema

fellipeh:
- Update OpenAPI schema
- Create methods scafold

This PR is being created to address [ESSNTL-5111](https://issues.redhat.com/browse/ESSNTL-5111).


tests/helpers/api_utils.py
```diff
@@ -471,6 +472,23 @@ def get_id_list_from_hosts(host_list):
...
 
 
+def build_id_list_for_url(list_or_id):
+    if type(list_or_id) == dict:
+        list_or_id = list(list_or_id.values())
+
+    if type(list_or_id) == list:
+        return ",".join([str(h.id) for h in list_or_id])
+
+    return str(list_or_id)
```
kruai, line 482: This can be refactored with `build_host_id_list_for_url`, since they're functionally identical AFAICT.
Also, is this function needed? None of the endpoints have IDs in the path.

thearifismail (commented):
@fellipeh can you please provide the doc link?  I got a question or two which I should ask after looking at the doc.  Thanks.

tests/test_api_account.py
kruai, line 1: Should we divide this up into multiple verb-focused files, like we do for hosts and groups?

tests/helpers/api_utils.py
```diff
@@ -24,6 +24,7 @@
 SYSTEM_PROFILE_URL = f"{BASE_URL}/system_profile"
 RESOURCE_TYPES_URL = f"{BASE_URL}/resource-types"
 ASSIGN_RULE_URL = f"{BASE_URL}/assignment-rules"
+ACCOUNT_URL = "/api/inventory/v1/account"
```
kruai, line 27: This should build off of `BASE_URL`

tests/helpers/api_utils.py
```diff
@@ -471,6 +472,23 @@ def get_id_list_from_hosts(host_list):
...
+        return ",".join([str(h.id) for h in list_or_id])
+
+    return str(list_or_id)
+
+
+def build_account_staleness_url(account_list_or_id=None, path=None, query=None):
+    if account_list_or_id:
+        account_list_or_id = build_id_list_for_url(account_list_or_id)
+
+    return _build_url(base_url=ACCOUNT_URL, id_list=account_list_or_id, path=path, query=query)
```
kruai, line 489: If this function is specifically for account staleness URLs, then the base URL it uses should include `/account/staleness` instead of just `/account`.

kruai (changes requested):
A few suggestions/questions. Good start on the tests; I assume those will be modified to check the response data as the actual endpoints get implemented.

fellipeh:
> A few suggestions/questions. Good start on the tests; I assume those will be modified to check the response data as the actual endpoints get implemented.

@kruai Fixed part of the comments you made. remembering this is just scaffold for the endpoints, we have others ticket for each endpoint which, we'll handle each point you mentioned above.

tests/test_api_account.py
```diff
@@ -0,0 +1,45 @@
...
+        "edge_stale_warning_delta": "number_of_days",
+        "edge_culling_delta": "number_of_days",
+    }
+    url = build_account_staleness_url(path="/staleness")
+    response_status, response_data = api_patch(url, host_data=query)
+    assert_response_status(response_status, 200)
+
+
+def test_reset_staleness(api_patch, subtests):
+    query = {"system_type": "conventional/edge"}
```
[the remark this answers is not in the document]
jpramos123: I think the way I expressed it in the documentation was confusing.

The system type should be:

```
{
"system_type": "conventional" OR "immutable"
}
```
Not both at the same time!

Thanks

swagger/api.spec.yaml
```diff
@@ -1865,6 +1940,10 @@ components:
             type: array
             items:
               $ref: '#/components/schemas/GroupOutWithHostCount'
+    AccountStalenessOutput:
+      title: Boolean result for POC. TODO Need to change
+      type: boolean
+
```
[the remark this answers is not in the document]
jpramos123: @fellipeh We still need to add the object attributes in this section!

jpramos123 (approved):

jpramos123 (approved):

tests/helpers/api_utils.py
```diff
@@ -24,6 +24,7 @@
 SYSTEM_PROFILE_URL = f"{BASE_URL}/system_profile"
 RESOURCE_TYPES_URL = f"{BASE_URL}/resource-types"
 ASSIGN_RULE_URL = f"{BASE_URL}/assignment-rules"
+ACCOUNT_URL = f"{BASE_URL}/account/staleness"
```
thearifismail, line 27: this variable should be called `staleness_url`

thearifismail (dismissed):

kruai (changes requested):
Normally we'd just return a 501 (not implemented) until the method has been coded, but I suppose returning mock data is fine for now if everyone else wants to go in this direction.
Just a couple more notes from me

swagger/api.spec.yaml
```diff
@@ -1865,6 +1940,45 @@ components:
...
+          maxLength: 36
+        immutable_staleness_delta:
+          type: string
+          maxLength: 36
+        immutable_stale_warning_delta:
+          type: string
+          maxLength: 36
+        immutable_culling_delta:
+          type: string
+          maxLength: 36
```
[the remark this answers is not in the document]
fellipeh: This is related to the DB spec: https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/models.py#L582

thearifismail (approved):

swagger/api.spec.yaml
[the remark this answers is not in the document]
jpramos123: @kruai Let me know if you think we should restrict the size in the DB as well. I just came up with 36.

swagger/api.spec.yaml
```diff
@@ -747,6 +747,77 @@ paths:
...
+        - ApiKeyAuth: []
+      responses:
+        '200':
+          description: Successfully account staleness created.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AccountStalenessOutput'
+    patch:
+      operationId: api.account.update_staleness
```
jpramos123, line 786:
```suggestion
      operationId: api.account_staleness.update_staleness
```

swagger/api.spec.yaml
```diff
@@ -747,6 +747,77 @@ paths:
...
+      responses:
+        '200':
+          description: Successfully account staleness updated.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AccountStalenessOutput'
+  /account/staleness/reset:
+    patch:
+      operationId: api.account.reset_staleness
```
jpramos123, line 804:
```suggestion
      operationId: api.account_staleness.reset_staleness
```

jpramos123 (changes requested):
Below are some comments

swagger/api.spec.yaml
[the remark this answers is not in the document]
kruai: oh, my mistake @fellipeh - I thought we were storing them as integers. Thanks for the clarification!
@jpramos123 Can you remind me why we're going with strings for those fields instead?

swagger/openapi.json
```diff
@@ -1271,6 +1271,110 @@
         }
       }
     },
+    "/account/staleness": {
+      "get": {
+        "operationId": "api.account.get_staleness",
```
jpramos123, line 1276: use `make style` to regenerate this file

```suggestion
        "operationId": "api.account_staleness.get_staleness",
```

jpramos123 (changes requested):

swagger/api.spec.yaml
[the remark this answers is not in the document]
jpramos123: @kruai We are going with string because we can have a value of `Never` in the frontend, see mock: https://www.sketch.com/s/dd09e06a-9bb3-4cda-9bf3-2fe92d949145/a/AZnGpP4

With that, I assumed using a string would be better to manipulate this value

Thanks

swagger/api.spec.yaml
```diff
@@ -747,6 +747,77 @@ paths:
         '403':
             description: Forbidden
 
+  /account/staleness:
```
jpramos123, line 750: Still need to run `make style` or apply the commit to update this json file with the correct module names

Thanks

swagger/api.spec.yaml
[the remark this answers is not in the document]
kruai: OK. We could also translate any sufficiently large number to "never", but we can make it work this way too. I won't hold this up then

thearifismail (approved):

jpramos123 (approved):

PR #1371, opened 2023-05-24
feat(ESSNTL-4945): Resource-level RBAC

kruai:
# Overview

This PR is being created to add resource-level RBAC to HBI, addressing [ESSNTL-4945](https://issues.redhat.com/browse/ESSNTL-4945) So that the API only makes one RBAC call per request, I've altered the RBAC decorator to both determine if an endpoint should be accessible at all, and provide a combined attribute filter that informs the endpoints what resources the user should have access to.

This is a WIP for the sake of demonstration right now - here are the current caveats:
- Some endpoints still need the filter implemented (host-deletion specifically)
- Many tests are needed
- I haven't added any handling for ungrouped hosts yet

## PR Checklist

- [x] Keep PR title short, ideally under 72 characters
- [x] Descriptive comments provided in complex code blocks
- [x] Tests: validate optimal/expected output
- [x] Tests: validate exceptions and failure scenarios
- [x] Tests: edge cases
- [ ] Recovers or fails gracefully during potential resource outages (e.g. DB, Kafka)
- [ ] Uses [type hinting](https://docs.python.org/3/library/typing.html), if convenient
- [ ] Documentation, if this PR changes the way other services interact with host inventory
- [ ] Links to related PRs

## Secure Coding Practices Documentation Reference

You can find documentation on this checklist [here](https://github.com/RedHatInsights/secure-coding-checklist).

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #1454, opened 2023-08-08
feat(ESSNTL-5037): Adding staleness post endpoint

fellipeh:
# Overview

This PR is being created to address [ESSNTL-5037](https://issues.redhat.com/browse/ESSNTL-5037).

Adding `account staleness` POST endpoint

### DIFF
--- a/api/account_staleness.py
+++ b/api/account_staleness.py
@@ -6,7 +6,6 @@ from api import api_operation
 from api import flask_json_response
 from api import json_error_response
 from api import metrics
-from app import InventoryException
 from app import RbacPermission
 from app import RbacResourceType
 from app.instrumentation import log_create_account_staleness_failed
@@ -52,7 +51,7 @@ def create_staleness(body):
     try:
         validated_data = InputAccountStalenessSchema().load(body)
     except ValidationError as e:
-        logger.exception(f"Input validation error while creating account staleness: {body}")
+        logger.exception(f'Input validation error, "{str(e.messages)}", while creating account staleness: {body}')
         return json_error_response("Validation Error", str(e.messages), status.HTTP_400_BAD_REQUEST)
 
     try:
@@ -61,17 +60,11 @@ def create_staleness(body):
 
         log_create_account_staleness_succeeded(logger, created_staleness.id)
     except IntegrityError:
-        org_id = validated_data.get("org_id")
+        error_message = f'An account staleness with org_id {validated_data.get("org_id")} already exists.'
 
-        error_message = f"A account staleness with org_id {org_id} already exists."
-
-        log_create_account_staleness_failed(logger, org_id)
+        log_create_account_staleness_failed(logger, validated_data.get("org_id"))
         logger.exception(error_message)
         return json_error_response("Integrity error", error_message, status.HTTP_400_BAD_REQUEST)
-
-    except InventoryException as inve:
-        logger.exception(inve.detail)
-        return json_error_response(inve.title, inve.detail, status.HTTP_400_BAD_REQUEST)
 
     return flask_json_response(serialize_account_staleness(created_staleness), status.HTTP_201_CREATED)
 
--- a/api/account_staleness.py
+++ b/api/account_staleness.py
@@ -1,9 +1,21 @@
 from flask_api import status
+from marshmallow import ValidationError
+from sqlalchemy.exc import IntegrityError
 
 from api import api_operation
 from api import flask_json_response
+from api import json_error_response
 from api import metrics
+from app import InventoryException
+from app import RbacPermission
+from app import RbacResourceType
+from app.instrumentation import log_create_account_staleness_failed
+from app.instrumentation import log_create_account_staleness_succeeded
 from app.logging import get_logger
+from app.models import InputAccountStalenessSchema
+from app.serialization import serialize_account_staleness
+from lib.account_staleness import add_account_staleness
+from lib.middleware import rbac
 
 logger = get_logger(__name__)
 
@@ -26,24 +38,53 @@ def _get_return_data():
 
 
 @api_operation
+@rbac(RbacResourceType.STALENESS, RbacPermission.READ)
 @metrics.api_request_time.time()
 def get_staleness():
     return flask_json_response(_get_return_data(), status.HTTP_200_OK)
 
 
 @api_operation
+@rbac(RbacResourceType.STALENESS, RbacPermission.WRITE)
 @metrics.api_request_time.time()
-def create_staleness():
-    return flask_json_response(_get_return_data(), status.HTTP_201_CREATED)
+def create_staleness(body):
+    # Validate account staleness input data
+    try:
+        validated_data = InputAccountStalenessSchema().load(body)
+    except ValidationError as e:
+        logger.exception(f"Input validation error while creating account staleness: {body}")
+        return json_error_response("Validation Error", str(e.messages), status.HTTP_400_BAD_REQUEST)
+
+    try:
+        # Create account staleness with validated data
+        created_staleness = add_account_staleness(validated_data)
+
+        log_create_account_staleness_succeeded(logger, created_staleness.id)
+    except IntegrityError:
+        org_id = validated_data.get("org_id")
+
+        error_message = f"A account staleness with org_id {org_id} already exists."
+
+        log_create_account_staleness_failed(logger, org_id)
+        logger.exception(error_message)
+        return json_error_response("Integrity error", error_message, status.HTTP_400_BAD_REQUEST)
+
+    except InventoryException as inve:
+        logger.exception(inve.detail)
+        return json_error_response(inve.title, inve.detail, status.HTTP_400_BAD_REQUEST)
+
+    return flask_json_response(serialize_account_staleness(created_staleness), status.HTTP_201_CREATED)
 
 
 @api_operation
+@rbac(RbacResourceType.STALENESS, RbacPermission.WRITE)
 @metrics.api_request_time.time()
 def update_staleness():
     return flask_json_response(_get_return_data(), status.HTTP_200_OK)
 
 
 @api_operation
+@rbac(RbacResourceType.STALENESS, RbacPermission.WRITE)
 @metrics.api_request_time.time()
 def reset_staleness():
     return flask_json_response(_get_return_data(), status.HTTP_200_OK)
--- a/api/account_staleness.py
+++ b/api/account_staleness.py
@@ -12,7 +12,7 @@ from app.instrumentation import log_create_account_staleness_failed
 from app.instrumentation import log_create_account_staleness_succeeded
 from app.logging import get_logger
 from app.models import InputAccountStalenessSchema
-from app.serialization import serialize_account_staleness
+from app.serialization import serialize_account_staleness_response
 from lib.account_staleness import add_account_staleness
 from lib.middleware import rbac
 
@@ -66,7 +66,7 @@ def create_staleness(body):
         logger.exception(error_message)
         return json_error_response("Integrity error", error_message, status.HTTP_400_BAD_REQUEST)
 
-    return flask_json_response(serialize_account_staleness(created_staleness), status.HTTP_201_CREATED)
+    return flask_json_response(serialize_account_staleness_response(created_staleness), status.HTTP_201_CREATED)
 
 
 @api_operation
--- a/app/instrumentation.py
+++ b/app/instrumentation.py
@@ -308,3 +308,11 @@ def log_post_assignment_rule_succeeded(logger, id):
 
 def log_post_assignment_rule_failed(logger):
     logger.info("Failed to create assignment rule", extra={"access_rule": get_control_rule()})
+
+
+def log_create_account_staleness_succeeded(logger, staleness_id):
+    logger.info("Created account staleness: %s", staleness_id)
+
+
+def log_create_account_staleness_failed(logger, org_id):
+    logger.info("Created account staleness Failed with org_id %s", org_id)
--- a/app/instrumentation.py
+++ b/app/instrumentation.py
@@ -315,4 +315,4 @@ def log_create_account_staleness_succeeded(logger, staleness_id):
 
 
 def log_create_account_staleness_failed(logger, org_id):
-    logger.info("Created account staleness Failed with org_id %s", org_id)
+    logger.info("Failed to create staleness for account with org_id %s", org_id)
--- a/app/models.py
+++ b/app/models.py
@@ -844,3 +844,15 @@ class InputAssignmentRule(MarshmallowSchema):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+
+class InputAccountStalenessSchema(MarshmallowSchema):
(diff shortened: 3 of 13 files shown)


tests/test_api_account_staleness_create.py
```diff
@@ -1,8 +1,33 @@
...
+    saved_data = get_account_staleness_by_org_id(saved_org_id)
+
+    assert saved_data.conventional_staleness_delta == input_data["system_staleness_delta"]
+    assert saved_data.conventional_stale_warning_delta == input_data["system_stale_warning_delta"]
+    assert saved_data.conventional_culling_delta == input_data["system_culling_delta"]
+    assert saved_data.immutable_staleness_delta == input_data["edge_staleness_delta"]
+    assert saved_data.immutable_stale_warning_delta == input_data["edge_stale_warning_delta"]
+    assert saved_data.immutable_culling_delta == input_data["edge_culling_delta"]
+
+
```
jpramos123, line 27: I would add tests to validate the RBAC roles we created.
Refer to: https://github.com/RedHatInsights/insights-host-inventory/pull/1371/files#diff-67132a592b391fd29ab8eedfb9cbe1bd23b82f0728f3ffa7e1b3e9f681edc118

jpramos123 (changes requested):
Thanks for the PR

Below are some suggestions

app/models.py
```diff
@@ -844,3 +844,15 @@ class InputAssignmentRule(MarshmallowSchema):
...
         super().__init__(*args, **kwargs)
+
+
+class InputAccountStalenessSchema(MarshmallowSchema):
+    system_staleness_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
+    system_stale_warning_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
+    system_culling_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
+    edge_staleness_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
+    edge_stale_warning_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
+    edge_culling_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
```
kruai, line 855: Do all of these need to be required? Just wondering because we _do_ have default values for all of these fields. I could see a user wanting to set a custom value for just one of them, and wanting to leave the rest as the defaults

app/instrumentation.py
```diff
@@ -308,3 +308,11 @@ def log_post_assignment_rule_succeeded(logger, id):
...
 def log_post_assignment_rule_failed(logger):
     logger.info("Failed to create assignment rule", extra={"access_rule": get_control_rule()})
+
+
+def log_create_account_staleness_succeeded(logger, staleness_id):
+    logger.info("Created account staleness: %s", staleness_id)
+
+
+def log_create_account_staleness_failed(logger, org_id):
+    logger.info("Created account staleness Failed with org_id %s", org_id)
```
kruai, line 318:
```suggestion
    logger.info("Failed to create staleness for account with org_id %s", org_id)
```

kruai (changes requested):

api/account_staleness.py
```diff
@@ -26,24 +38,53 @@ def _get_return_data():
...
+@rbac(RbacResourceType.STALENESS, RbacPermission.WRITE)
 @metrics.api_request_time.time()
-def create_staleness():
-    return flask_json_response(_get_return_data(), status.HTTP_201_CREATED)
+def create_staleness(body):
+    # Validate account staleness input data
+    try:
+        validated_data = InputAccountStalenessSchema().load(body)
+    except ValidationError as e:
+        logger.exception(f"Input validation error while creating account staleness: {body}")
```
thearifismail, line 55:
```suggestion
        logger.exception(f'Input validation error, "{str(e.messages)}", while creating account staleness: {body}')
```

thearifismail: include the exception in the log message so that some one looking at Kibana can see the reason why it failed.

lib/account_staleness.py
```diff
@@ -0,0 +1,45 @@
...
+            immutable_stale_warning_delta=edge_stale_warning_delta,
+            immutable_culling_delta=edge_culling_delta,
+        )
+        db.session.add(new_staleness)
+        db.session.flush()
+
+    # gets the ID of the staleness after it has been committed
+    created_staleness = AccountStalenessCulling.query.filter(AccountStalenessCulling.org_id == org_id).one_or_none()
+
+    return created_staleness
```
thearifismail, line 38: rename the var to show it is an ID.  The current var suggests it to be a complete staleness object.

lib/account_staleness.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def get_account_staleness_by_org_id(org_id) -> AccountStalenessCulling:
```
thearifismail, line 41: This method is not being called, except, a test.  Is it really needed?

lib/account_staleness.py
```diff
@@ -0,0 +1,45 @@
...
+            conventional_culling_delta=system_culling_delta,
+            immutable_staleness_delta=edge_staleness_delta,
+            immutable_stale_warning_delta=edge_stale_warning_delta,
+            immutable_culling_delta=edge_culling_delta,
+        )
+        db.session.add(new_staleness)
+        db.session.flush()
+
+    # gets the ID of the staleness after it has been committed
+    created_staleness = AccountStalenessCulling.query.filter(AccountStalenessCulling.org_id == org_id).one_or_none()
```
thearifismail, line 36: The comment says ID being returned?  Looks like a staleness object is returned.  Either the comment or the var name is not correct.

api/account_staleness.py
```diff
@@ -26,24 +38,53 @@ def _get_return_data():
...
+
+    try:
+        # Create account staleness with validated data
+        created_staleness = add_account_staleness(validated_data)
+
+        log_create_account_staleness_succeeded(logger, created_staleness.id)
+    except IntegrityError:
+        org_id = validated_data.get("org_id")
+
+        error_message = f"A account staleness with org_id {org_id} already exists."
```
thearifismail, line 66:
```suggestion
        error_message = f'A account staleness with org_id {validated_data.get("org_id")} already exists.'
```

api/account_staleness.py
```diff
@@ -26,24 +38,53 @@ def _get_return_data():
...
+        logger.exception(f"Input validation error while creating account staleness: {body}")
+        return json_error_response("Validation Error", str(e.messages), status.HTTP_400_BAD_REQUEST)
+
+    try:
+        # Create account staleness with validated data
+        created_staleness = add_account_staleness(validated_data)
+
+        log_create_account_staleness_succeeded(logger, created_staleness.id)
+    except IntegrityError:
+        org_id = validated_data.get("org_id")
```
thearifismail, line 64:
```suggestion
```

swagger/api.spec.yaml
```diff
@@ -778,8 +778,16 @@ paths:
         Required permissions: inventory:TODO:write
       security:
         - ApiKeyAuth: []
+        - BearerAuth: []
```
thearifismail, line 781: Why is `BearerAuth` need by a `POST` call but not GET?

thearifismail (commented):
Not done yet

lib/account_staleness.py
fellipeh: `created_staleness` is a complete `AccountStalenessCulling` object.

swagger/api.spec.yaml
fellipeh: Just following as same others endpoint... @kruai  any thought on this?

kruai: ¯\_(ツ)_/¯

lib/account_staleness.py
kruai: If this function is a test helper, it should probably be a DB fixture: https://github.com/RedHatInsights/insights-host-inventory/blob/master/tests/fixtures/db_fixtures.py
I'd leave the actual access function for when that endpoint is implemented, as it's out of scope for this PR, and may end up needing to do things differently

app/models.py
kruai: @fellipeh I'm just unresolving this bc I didn't see any discussion on it. It's more of a requirements question though, so CCing @jpramos123 ^

tests/test_api_account_staleness_create.py
```diff
@@ -1,8 +1,28 @@
...
-    url = build_account_staleness_url()
-    response_status, response_data = api_post(url, host_data=None)
+def test_create_staleness(api_create_account_staleness, event_producer, mocker):
+    input_data = {
+        "system_staleness_delta": "1",
+        "system_stale_warning_delta": "7",
+        "system_culling_delta": "14",
+        "edge_staleness_delta": "7",
+        "edge_stale_warning_delta": "120",
+        "edge_culling_delta": "120",
```
kruai, line 12: These field names need to be updated to match the implementation.

kruai (changes requested):
Got a few more notes in addition to the comments I left before the review.

api/account_staleness.py
```diff
@@ -26,24 +38,51 @@ def _get_return_data():
...
+    except IntegrityError:
+        error_message = f'A account staleness with org_id {validated_data.get("org_id")} already exists.'
+
+        log_create_account_staleness_failed(logger, validated_data.get("org_id"))
+        logger.exception(error_message)
+        return json_error_response("Integrity error", error_message, status.HTTP_400_BAD_REQUEST)
+
+    except InventoryException as inve:
+        logger.exception(inve.detail)
+        return json_error_response(inve.title, inve.detail, status.HTTP_400_BAD_REQUEST)
```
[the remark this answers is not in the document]
fellipeh: You're right, we only have `IntegrityError`  exception.. removing...

app/models.py
jpramos123: @kruai @fellipeh I think they should NOT be required fields, as you mentioned, the user can chose to set just one of this values, and the rest will be created using the default values.

swagger/api.spec.yaml
kruai: From our most recent discussion: it should be fine to remove the bearer auth.

tests/test_api_account_staleness_create.py
jpramos123: @fellipeh I still would like to have some test that validate the RBAC roles:
See this as a reference: https://github.com/RedHatInsights/insights-host-inventory/pull/1371/files#diff-e7c80922b4d5365c140f8fef9fe0925798e1b11b603bce7c8620d4556517ad68R161

jpramos123 (changes requested):

app/models.py
kruai: @fellipeh I'm again unresolving this because I do not see a follow-up; at least one field should be provided, but they should not all be required

app/models.py
```diff
@@ -844,3 +847,15 @@ class InputAssignmentRule(MarshmallowSchema):
...
         super().__init__(*args, **kwargs)
+
+
+class InputAccountStalenessSchema(MarshmallowSchema):
+    conventional_staleness_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
+    conventional_stale_warning_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
+    conventional_culling_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
+    immutable_staleness_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
+    immutable_stale_warning_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
+    immutable_culling_delta = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=36))
```
kruai, line 858: Just so we don't lose track of this, the verdict is that not all of these fields are required (since they have default values). The payload should provide at least one of them, but none of them should be marked as required

swagger/api.spec.yaml
```diff
@@ -1988,6 +1995,31 @@ components:
             type: array
             items:
               $ref: '#/components/schemas/GroupOutWithHostCount'
+    AccountStalenessIn:
+      title: Account Staleness In
+      description: >-
+        Data of a single account staleness.
+      type: object
+      allOf:
```
kruai, line 2003: Should use `anyOf` if one or more of them is required
```suggestion
      anyOf:
```

tests/test_api_account_staleness_create.py
```diff
@@ -1,8 +1,76 @@
...
+    response_status, response_data = api_create_account_staleness(_INPUT_DATA)
+    assert_response_status(response_status, 400)
+
+
+def test_create_staleness_with_missing_input(api_create_account_staleness):
+    input_data = {
+        "test_conventional_staleness_delta": "1",
+    }
+    response_status, response_data = api_create_account_staleness(input_data)
+    assert_response_status(response_status, 400)
```
kruai, line 44: If only one value needs to be provided, then the post should probably succeed and return 201

kruai (changes requested):
A few more comments (mostly about all of the payload fields being required), but there's steady progress! I like the new tests

app/models.py
fellipeh: We'll follow with  should provide at least one, so I removed the require=True.

tests/test_api_account_staleness_create.py
fellipeh: This supposed to test an `invalid` payload. I believe the name I used make the confusion, I'll change it.

app/models.py
fellipeh: Resolved per slack conversation.

kruai (dismissed):

jpramos123 (approved):

app/serialization.py
```diff
@@ -325,3 +325,19 @@ def _deserialize_tags_dict(tags):
 
 def _serialize_tags(tags):
     return [tag.data() for tag in Tag.create_tags_from_nested(tags)]
+
+
+def serialize_account_staleness(account_staleness):
```
jpramos123, line 330:
```suggestion
def serialize_account_staleness_response(account_staleness):
```

jpramos123 (changes requested):
Please, can you change this method name?

This will help me a lot when merging your code to mine

swagger/openapi.json
```diff
@@ -3018,6 +3029,41 @@
...
+      "AccountStalenessIn": {
+        "title": "Account Staleness In",
+        "description": "Data of a single account staleness.",
+        "type": "object",
+        "anyOf": [
+          {
+            "properties": {
+              "conventional_staleness_delta": {
+                "type": "string",
+                "maxLength": 36
```
thearifismail, line 3041: `maxLength: 36` is used for uuid values; even the times expressed in `"created_at": "2023-07-28T14:32:16.353082"` is 25 chars long.  A time delta of a thousand years has `31536000000 seconds`, which is 11 chars. IMO, these lengths should be set to some realistic values.  Over allocation of memory can have negative imact.

thearifismail (commented):
Over all looks good.  One minor comment though

kruai (approved):

jpramos123 (approved):

PR #1472, opened 2023-08-29
Segmentio tracking for inventory API.

roliveri:
Initial POC for adding instrumentation to track inventory API events in Segmentio.


api/segmentio.py
```diff
@@ -0,0 +1,54 @@
...
+            identity.org_id,
+            op_name,
+            {
+                "start_time": start_time,
+                "end_time": end_time,
+                "status_code": contextual_data["status_code"],
+                "user_agent": connexion.request.headers["User-Agent"],
+                "auth_type": identity.auth_type,
+            },
+        )
```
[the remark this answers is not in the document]
roliveri: Yes, I'll add that.
We should probably check to see if there's a standard of preferred name for insights host inventory.

api/segmentio.py
```diff
@@ -0,0 +1,59 @@
...
+
+def segmentio_track(op_name, start_time, end_time, contextual_data):
+    identity = get_current_identity()
+
+    print(f"\n**** analytics.write_key: {analytics.write_key}")
+    if analytics.write_key:
+        print(f"**** Calling analytics.track({identity.org_id}, {op_name}, ...)")
+        analytics.track(
+            identity.org_id,
+            op_name,
```
marleystipich2, line 15: Small suggestion but maybe we could name these 2 fields too for readabilty like
```
user_id=identity.org_id,
event=op_name,
```

api/segmentio.py
```diff
@@ -0,0 +1,59 @@
+import connexion
+import segment.analytics as analytics
+
+from app.auth import get_current_identity
+
+
+def segmentio_track(op_name, start_time, end_time, contextual_data):
+    identity = get_current_identity()
+
+    print(f"\n**** analytics.write_key: {analytics.write_key}")
```
thearifismail, line 10: Should't a `logger` be used in place of `print`?

app/__init__.py
```diff
@@ -85,6 +86,14 @@ def initialize_metrics(config):
     )
 
 
+# TODO: Need to call analytics.flush() on shutdown.
+def initialize_segmentio(config):
+    # Should this be maintained by app/config.py?
+    analytics.write_key = config.segmentio_write_key
+    print(f"\n****** Initializing Segmentio - analytics.write_key: {analytics.write_key}")
```
thearifismail, line 93: Should use logger in place of print

thearifismail (commented):
Why another analytic tracking tool is needed on the top of what's already implemented?

thearifismail:
> @kruai @thearifismail We also need to call `analytics.flush()` on shutdown. I looked, but I'm not sure where that should be done. Any suggestions?

How will `analytics` start or initialize?  The invenvetory MQ service uses ShutdownHandler, which is in `lib/handlers.py`

api/segmentio.py
kruai: You're probably referring to the other print statements, but just in case: We should absolutely not print or log secrets

thearifismail (commented):
One comment; rest LGTM

marleystipich2:
@thearifismail We added some tests to this. Could we get another review when you get a chance? Also, there are some tests failing in the PR check but they pass locally.

kruai (changes requested):

kruai (approved):

marleystipich2 (approved):
Looks good to me! Thanks, @roliveri 👍

thearifismail (approved):

PR #1473, opened 2023-08-29
feat(ESSNTL-5197): RBAC structuring, group data censoring

kruai:
# Overview

This PR is being created to address [ESSNTL-5197](https://issues.redhat.com/browse/ESSNTL-5197).

- Gives `rbac_filter` a rigid structure to make its usage more clear
- Gives `rbac_filter` default values instead of the whole object being `None` if not provided
- Removes group data from host API responses if the user does not have `groups:read` permission for that group
- Add test and modify an existing one

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -60,7 +60,7 @@ logger = get_logger(__name__)
 
 
 @api_operation
-@rbac(RbacResourceType.HOSTS, RbacPermission.READ, additional_permissions=["inventory:groups:read"])
+@rbac(RbacResourceType.HOSTS, RbacPermission.READ, additional_restrictions=["inventory:groups:read"])
 @metrics.api_request_time.time()
 def get_host_list(
     display_name=None,
--- a/api/host.py
+++ b/api/host.py
@@ -60,7 +60,7 @@ logger = get_logger(__name__)
 
 
 @api_operation
-@rbac(RbacResourceType.HOSTS, RbacPermission.READ)
+@rbac(RbacResourceType.HOSTS, RbacPermission.READ, additional_permissions=["inventory:groups:read"])
 @metrics.api_request_time.time()
 def get_host_list(
     display_name=None,
@@ -81,7 +81,7 @@ def get_host_list(
     registered_with=None,
     filter=None,
     fields=None,
-    rbac_filter: RbacFilter = None,
+    rbac_filter: RbacFilter = RbacFilter(),
 ):
     total = 0
     host_list = ()
@@ -112,7 +112,7 @@ def get_host_list(
         log_get_host_list_failed(logger)
         flask.abort(400, str(e))
 
-    json_data = build_paginated_host_list_response(total, page, per_page, host_list, additional_fields)
+    json_data = build_paginated_host_list_response(total, page, per_page, host_list, additional_fields, rbac_filter)
     return flask_json_response(json_data)
 
 
@@ -133,7 +133,7 @@ def delete_hosts_by_filter(
     staleness=None,
     tags=None,
     filter=None,
-    rbac_filter: RbacFilter = None,
+    rbac_filter: RbacFilter = RbacFilter(),
 ):
     if not any(
         [
@@ -226,7 +226,7 @@ def _delete_host_list(host_id_list, rbac_filter: RbacFilter):
 @api_operation
 @rbac(RbacResourceType.HOSTS, RbacPermission.WRITE)
 @metrics.api_request_time.time()
-def delete_all_hosts(confirm_delete_all=None, rbac_filter: RbacFilter = None):
+def delete_all_hosts(confirm_delete_all=None, rbac_filter: RbacFilter = RbacFilter()):
     if not confirm_delete_all:
         logger.error("To delete all hosts, provide confirm_delete_all=true in the request.")
         flask.abort(400, "To delete all hosts, provide confirm_delete_all=true in the request.")
@@ -255,7 +255,7 @@ def delete_all_hosts(confirm_delete_all=None, rbac_filter: RbacFilter = None):
 @api_operation
 @rbac(RbacResourceType.HOSTS, RbacPermission.WRITE)
 @metrics.api_request_time.time()
-def delete_host_by_id(host_id_list, rbac_filter: RbacFilter = None):
+def delete_host_by_id(host_id_list, rbac_filter: RbacFilter = RbacFilter()):
     delete_count = _delete_host_list(host_id_list, rbac_filter)
 
     if not delete_count:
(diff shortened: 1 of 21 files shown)


kruai:
I swear, we keep getting those `RemoteDisconnected` intermittent errors. But you can see (in the required PR build) that it's just that network blip and not caused by a real test failure

FabriciaDinizRH (changes requested):
Awesome work! The middleware got me a bit overwhelmed, but everything looks ok.
Got a small suggestion though

api/host.py
```diff
@@ -266,7 +267,15 @@ def delete_host_by_id(host_id_list, rbac_filter=None):
 @api_operation
 @rbac(RbacResourceType.HOSTS, RbacPermission.READ)
```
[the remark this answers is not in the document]
kruai: good catch!

app/serialization.py
```diff
@@ -76,7 +76,7 @@ def deserialize_canonical_facts(raw_data, all=False):
     return _deserialize_canonical_facts(validated_data)
 
 
-def deserialize_host_xjoin(data):
+def deserialize_host_xjoin(data, rbac_filter=None):
```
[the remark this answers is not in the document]
kruai: Nope, it's leftover from when I was brainstorming and put it in the wrong function. I'll remove it

lib/middleware.py
```diff
@@ -33,6 +37,35 @@
 outbound_http_metric = outbound_http_response_time.labels("rbac")
 
 
+class RbacIdFilter:
+    def __init__(self, resource: RbacResourceType, id_set: Set[str]) -> None:
+        self.resource = resource
+        self.id_set = id_set
+
+
+class RbacFilter:
```
thearifismail, line 46: Since there are no functions in RbacFilter, why couldn't this be a `dictionary`?

lib/middleware.py
```diff
@@ -33,6 +37,35 @@
 outbound_http_metric = outbound_http_response_time.labels("rbac")
 
 
+class RbacIdFilter:
```
thearifismail, line 40: What is the benefit of creating a new class/type for just attributes when there are not functions in the class?  Furthermore, `RbacIdFilter` does not get referenced from outside of `middleware.py`?

lib/middleware.py
```diff
@@ -71,87 +104,142 @@ def get_rbac_permissions(app):
...
-                        return func(*args, **kwargs)
+            # Get the allowed IDs for the required permission
+            allowed_group_ids, is_allowed = _get_allowed_ids(
+                rbac_data, permission_base, resource_type, required_permission
+            )
+
+            # Get the allowed IDs for any additional restrictions
+            data_restrictions = {}
+            for additional_permission in additional_restrictions:
+                additional_ids, match_found = _get_allowed_ids(
```
thearifismail, line 220: `_get_group_ids()` is more clear than `allowed` and easy to understand

lib/middleware.py
```diff
@@ -71,87 +104,142 @@ def get_rbac_permissions(app):
...
+            if gid is not None:
+                UUID(gid)
+    except ValueError:
+        abort(
+            status.HTTP_503_SERVICE_UNAVAILABLE,
+            "Received invalid UUIDs for attributeFilter.value in RBAC response.",
+        )
+
+
+def _get_attribute_filter_value(rbac_permission) -> Set[str]:
```
thearifismail, line 135: The parent filename is too generic, which makes reader wonder about is `attribute` about?  If it is a group's attribute then call it `_get_group_filter_value` and if `rbac` then `_get_rbac_filter_value.

lib/middleware.py
```diff
@@ -33,6 +37,35 @@
...
+    #   "inventory:groups:read": <RbacIdFilter>,
+    #   "inventory:hosts:write": <RbacIdFilter>,
+    #   ...etc...
+    # }
+    def __init__(
+        self,
+        resource: RbacResourceType = RbacResourceType.HOSTS,
+        permission: RbacPermission = RbacPermission.READ,
+        filter_by: RbacIdFilter = None,
+        data_restrictions: Dict[str, RbacIdFilter] = {},
```
thearifismail, line 58: What's the string parameter is for?

lib/middleware.py
kruai: It originally was a huge multi-layer dictionary, and it would _technically_ function implemented that way, but anybody coding with it would be doing so blind. Intellisense can tell you the attributes of a class, but can't guess at dictionary keys. I also think it's cleaner and more organized with a class structure, so instead of code snippets like this:
```
if rbac_filter["filter_by"] and rbac_filter["filter_by"]["resource"] == RbacResourceType.GROUPS:
    filters += (AssignmentRule.group_id.in_(rbac_filter["filter_by"]["id_set"]),)
```
we get things like this:
```
if rbac_filter.filter_by and rbac_filter.filter_by.resource == RbacResourceType.GROUPS:
    filters += (AssignmentRule.group_id.in_(rbac_filter.filter_by.id_set),)
```

lib/middleware.py
kruai: `attributeFilter` is the field on RBAC permissions that limits what resources the user can access. It's not a group's attribute, and `rbac_filter` doesn't make sense because I'm specifically getting the value of `attributeFilter` from the permission's `resourceDefinition`

lib/middleware.py
kruai: I left a multiline comment right above this explaining the dict's structure. It's for quicker and easier access to the data restriction specific to that permission

lib/middleware.py
kruai: I tried to keep it general in as many places as possible, because the value of `attributeFilter` in RBAC's response isn't technically guaranteed to be group IDs. HBI doesn't support other values for that yet, but I figured it would be best to play it safe and flexible for when that changes.

lib/middleware.py
kruai: As for the benefits of using a class instead of a dictionary, see my previous comment.
You're correct, `RbacIDFilter` is only used inside `middleware.py`. It's used twice in `RbacFilter` - once for the main filter, and once to represent the data that's restricted.

lib/middleware.py
thearifismail: I like the code clarity of using class.  At the time of this comment I was thinking about performance and class hierarchy and a quick google search did say that dictionaries had better performance than classes; however I did not find much on class hierarchy structure, like in C++ or Java.

thearifismail (approved):

FabriciaDinizRH:
The same required tests that are breaking in this PR are breaking in João's PR, so the problem is not the PR

FabriciaDinizRH (approved):

thearifismail (approved):

kruai:
> The same required tests that are breaking in this PR are breaking in João's PR, so the problem is not the PR

OK. It also looks like it's unrelated to the changes in this PR, so I'll go ahead and merge.

kruai:
I've put this back into draft status because we're not actually ready for this change yet. We want to make sure that all apps have the same display behavior, so we shouldn't start hiding the names until other apps are ready to do so as well (see the linked Jira ticket for more details).

kruai:
We have no plans to censor the group names, and we have already refactored RBAC. Closing.

PR #1489, opened 2023-09-18
RHINGENG-1751: Enable rebuild_events_topic script to stop after speicified time

thearifismail:
# Overview

This PR is being created to address [RHINGENG-1751](https://issues.redhat.com/browse/RHINGENG-1751).  It enables the `rebuild_events_topic` script to stop after some specified time.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -228,6 +228,7 @@ class Config:
 
         self.host_delete_chunk_size = int(os.getenv("HOST_DELETE_CHUNK_SIZE", "1000"))
         self.script_chunk_size = int(os.getenv("SCRIPT_CHUNK_SIZE", "500"))
+        self.rebuild_events_duration = int(os.getenv("REBUILD_EVENTS_DURATION", "3600"))  # 1 hour
         self.sp_authorized_users = os.getenv("SP_AUTHORIZED_USERS", "tuser@redhat.com").split()
 
         if self._runtime_environment == RuntimeEnvironment.PENDO_JOB:
--- a/app/config.py
+++ b/app/config.py
@@ -228,7 +228,7 @@ class Config:
 
         self.host_delete_chunk_size = int(os.getenv("HOST_DELETE_CHUNK_SIZE", "1000"))
         self.script_chunk_size = int(os.getenv("SCRIPT_CHUNK_SIZE", "500"))
-        self.rebuild_events_duration = int(os.getenv("REBUILD_EVENTS_DURATION", "3600"))  # 1 hour
+        self.rebuild_events_time_limit = int(os.getenv("REBUILD_EVENTS_TIME_LIMIT", "3600"))  # 1 hour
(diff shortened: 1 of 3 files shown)


rebuild_events_topic.py
```diff
@@ -55,30 +57,35 @@ def run(config, logger, session, consumer, event_producer, shutdown_handler):
...
-                    logger.exception("Unable to process message", extra={"incoming_message": message.value()})
-        try:
-            # pace the events production speed as flush completes sending all buffered records.
-            event_producer._kafka_producer.flush(300)
-        except ProduceError:
-            raise ProduceError("ProduceError: Failed to flush produced Kafka messages within 300 seconds")
-
-        num_messages = len(new_messages)
-        total_messages_processed += num_messages
+        if datetime.now() < stop_time:
```
kruai, line 63: I think this would work better as part of the main conditional (L59), i.e.:
```
while num_messages > 0 and not shutdown_handler.shut_down() and datetime.now() < stop_time:
```
Or, if you wanted to keep the condition separate, I'd recommend just making a short block to exit instead, like this:
```
if datetime.now() > stop_time:
  logger.info(f"Event topic rebuild halting after {config.rebuild_events_duration} seconds time.")
  exit(0)  # exit as intended
```

app/config.py
```diff
@@ -228,6 +228,7 @@ def __init__(self, runtime_environment):
 
         self.host_delete_chunk_size = int(os.getenv("HOST_DELETE_CHUNK_SIZE", "1000"))
         self.script_chunk_size = int(os.getenv("SCRIPT_CHUNK_SIZE", "500"))
+        self.rebuild_events_duration = int(os.getenv("REBUILD_EVENTS_DURATION", "3600"))  # 1 hour
```
kruai, line 231: Instead of just DURATION, I'd recommend either MAX_DURATION or TIME_LIMIT to be a bit more precise.
Also we probably wanna default to something higher than an hour... but I guess it's OK to leave that as the default and then put a way larger value for Prod

kruai (changes requested):
Looking good, just a couple suggestions

rebuild_events_topic.py
thearifismail: Keeping conditions separate allowing to log a message, which indicates that there could be more messages.

thearifismail (commented):
comments addressed

kruai (approved):

rebuild_events_topic.py
kruai: OK, makes sense to me

kruai (approved):

PR #1506, opened 2023-09-27
fix(ESSNTL-5422): Improve error message

fellipeh:
# Overview

This PR is being created to address [ESSNTL-5422](https://issues.redhat.com/browse/ESSNTL-5422).

Improving error messages on staleness endpoint

### DIFF
--- a/api/staleness.py
+++ b/api/staleness.py
@@ -58,11 +58,6 @@ def _validate_input_data(body):
     return validated_data
 
 
-def _get_org_id():
-    current_identity = get_current_identity()
-    return current_identity.org_id
-
-
 @api_operation
 @rbac(RbacResourceType.STALENESS, RbacPermission.READ)
 @metrics.api_request_time.time()
@@ -109,9 +104,9 @@ def create_staleness(body):
 
         log_create_staleness_succeeded(logger, created_staleness.id)
     except IntegrityError:
-        error_message = f"An account staleness record for org_id {_get_org_id()} already exists."
+        error_message = f"Staleness record for org_id {get_current_identity().org_id} already exists."
 
-        log_create_staleness_failed(logger, _get_org_id())
+        log_create_staleness_failed(logger, get_current_identity().org_id)
         logger.exception(error_message)
         return json_error_response("Integrity error", error_message, status.HTTP_400_BAD_REQUEST)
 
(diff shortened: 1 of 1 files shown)


api/staleness.py
```diff
@@ -58,6 +58,11 @@ def _validate_input_data(body):
     return validated_data
 
 
+def _get_org_id():
+    current_identity = get_current_identity()
+    return current_identity.org_id
```
kruai, line 63: Since current_identity is only used once, this can just be a one-liner
```suggestion
    return get_current_identity().org_id
```

kruai (changes requested):
Not sure `_get_org_id` it needs to be its own function since it's really simple, but it's not necessarily a problem either. I've just got one suggestion to streamline

api/staleness.py
```diff
@@ -127,7 +129,7 @@ def delete_staleness():
         remove_staleness()
         return flask_json_response(None, status.HTTP_204_NO_CONTENT)
     except NoResultFound:
-        abort(status.HTTP_404_NOT_FOUND, "Account Staleness not found.")
+        abort(status.HTTP_404_NOT_FOUND, f"An account staleness record for org_id {_get_org_id()} does not exist.")
```
thearifismail, line 132: It is a host which goes stale NOT the account the host belongs to.

thearifismail, line 132:
```suggestion
        abort(status.HTTP_404_NOT_FOUND, f"Staleness record for org_id {_get_org_id()} does not exist.")
```

thearifismail:
@fellipeh can you please update your branch now that `account_staleness` have been replaced by `staleness` in the parent repo?

api/staleness.py
fellipeh: I refactored using your suggestion

api/staleness.py
```diff
@@ -149,4 +151,4 @@ def update_staleness(body):
 
         return flask_json_response(serialize_staleness_response(updated_staleness), status.HTTP_200_OK)
     except NoResultFound:
-        abort(status.HTTP_404_NOT_FOUND, "Account Staleness not found.")
+        abort(status.HTTP_404_NOT_FOUND, f"An account staleness record for org_id {_get_org_id()} does not exist.")
```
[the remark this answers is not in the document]
fellipeh: Make sense to have `Staleness`  word instead of Account Staleness

api/staleness.py
fellipeh: Make sense to have `Staleness`  word instead of Account Staleness

fellipeh: Make sense to have `Staleness`  word instead of Account Staleness

thearifismail (approved):

kruai (approved):

PR #1525, opened 2023-10-17
Fixed bug caused when User-Agent header is missing.

roliveri:
# Overview

API requests with no User-Agent header fail with a response code of 500.
This change ensures we don't try to reference a header that doesn't exist (the cause of the error).
And it doesn't attempt to track such requests, as we filter on User-Agent.

### DIFF
--- a/api/segmentio.py
+++ b/api/segmentio.py
@@ -5,9 +5,14 @@ from app.auth import get_current_identity
 
 
 def segmentio_track(op_name, processing_time, contextual_data, logger):
-    identity = get_current_identity()
-
     if analytics.write_key:
+        user_agent_header = connexion.request.headers.get("User-Agent", "")
+        if not user_agent_header:
+            logger.debug("Skipping analytics.track(), User-Agent header not set")
+            return None
+
+        identity = get_current_identity()
+
         logger.debug(f"Calling analytics.track({identity.org_id}, {op_name}, ...)")
         analytics.track(
             identity.org_id,
@@ -15,7 +20,7 @@ def segmentio_track(op_name, processing_time, contextual_data, logger):
             properties={
                 "processing_time": processing_time,
                 "status_code": contextual_data["status_code"],
-                "user_agent": connexion.request.headers["User-Agent"],
+                "user_agent": user_agent_header,
                 "auth_type": identity.auth_type,
             },
             context={
--- a/tests/test_segmentio.py
+++ b/tests/test_segmentio.py
@@ -36,3 +36,15 @@ def test_segmentio_track_without_write_key(mocker, api_get):
 
     assert analytics.write_key == ""
     assert mock_track.call_count == 0
+
+
+def test_segmentio_track_without_user_agent(mocker, api_get):
+    mock_track = mocker.patch("api.segmentio.analytics.track")
+    analytics.write_key = "test_write_key"
+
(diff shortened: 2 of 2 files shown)


kruai (approved):

thearifismail (commented):
Shouldn't there be a test producing status_code 500?

api/segmentio.py
```diff
@@ -5,17 +5,22 @@
 
 
 def segmentio_track(op_name, processing_time, contextual_data, logger):
-    identity = get_current_identity()
-
     if analytics.write_key:
+        user_agent_header = connexion.request.headers.get("User-Agent", "")
```
marleystipich2, line 9: @roliveri Can you explain this line? Will it return the `User-Agent` string or `""` if there is no `User-Agent` header?

roliveri:
@thearifismail The User-Agent header isn't required, so it's absence shouldn't result in 500.
This tracking filters on the user agent, so we just bypass the tracking if the header isn't set.
The API request should behave as usual.

tests/test_segmentio.py
```diff
@@ -36,3 +36,15 @@ def test_segmentio_track_without_write_key(mocker, api_get):
 
     assert analytics.write_key == ""
     assert mock_track.call_count == 0
+
+
+def test_segmentio_track_without_user_agent(mocker, api_get):
+    mock_track = mocker.patch("api.segmentio.analytics.track")
+    analytics.write_key = "test_write_key"
+
+    response_status, response_data = api_get(build_resource_types_url(), extra_headers={"User-Agent": ""})
```
fstavela, line 45: This probably wouldn't catch the bug with the missing "User-Agent" header. We need to make the request without setting that header at all, so no `extra_headers` parameter. But, you can add a new test for it, and we can also keep this test as `test_segmentio_track_with_empty_user_agent`. Having more tests can never hurt :D

api/segmentio.py
marleystipich2: Ah okay, makes sense now. Thanks! 

Do we need the `if not user_agent_header:` condition then? Because if there's no `User-Agent` header and `user_agent_header` is set to `""`, we might be able to still do the `analytics.track` call since the `user_agent` will just be set to `""` in the call. Does this make sense?

tests/test_segmentio.py
roliveri: @fstavela Yes, I know.
Given the way the `api_get` fixture is implemented I couldn't find a way to delete that header.
The underlying `werkzeug` library sets the header to a default value when not set by the caller.

However the `connexion.request.headers.get("User-Agent", "")` call should cover both conditions.

fstavela: Ahh, I didn't know the `werkzeug` library sets that. OK, then I think the test is good :+1:

api/segmentio.py
roliveri: @marleystipich2 It depends on how we filter on user_agent. Would we assume a user_agent of "" is a direct API call or not?

fstavela (approved):

api/segmentio.py
marleystipich2: I feel like it'd be best to assume that it is some kind of direct API call. It seems like if it was a web browser or CLI, the user-agent would be set to something.

thearifismail (approved):

PR #1550, opened 2023-11-15
ESSNTL-5487: Replace culling with deletion in api and swagger

thearifismail:
# Overview

This PR is being created to address [ESSNTL-5487](https://issues.redhat.com/browse/ESSNTL-5487).
It replaces:

1. "conventional_culling_delta" with "conventional_deletion_delta"
2. "immutable_culling_delta" with "immutable_deletion_delta"

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


swagger/api.spec.yaml
```diff
@@ -1669,7 +1669,7 @@ components:
               type: string
               format: date-time
               nullable: true
-            culled_timestamp:
+            deletion_timestamp:
```
fstavela, line 1672: I remember that when we were discussing this change, we agreed that we will rename all of the new culling fields to "deletion", which means all the new fields introduced by custom staleness. We didn't want to rename the old fields, because it theoretically may break some customer automation, if they have any (I don't know if they do). If we want to be proper about this, we would  need to release a new version of API with this kind of change.

I don't think anyone really uses this field in automation, it probably doesn't make much sense, but if there is a potential to break something, then I think all of us need to agree on it.

thearifismail: hmmm... I does not make sense to create a new API version for this then.  When a new API

strider: hmmm, I agree with @fstavela on this. We want to avoid any potential breakage of our backwards compatibility of the API. I know that I didn't participate to this discussion, but do we have any soft deprecation rules and if yes, we should probably like to follow a deprecation period until a new API version should be released.

thearifismail:
This PR has been replaced by https://github.com/RedHatInsights/insights-host-inventory/pull/1573

PR #1559, opened 2023-11-22
fix(RHINENG-4810): filtering logic for multiple  values for the same SP field

FabriciaDinizRH:
# Overview

This PR is being created to address [RHINENG-4810](https://issues.redhat.com/browse/RHINENG-4810).
It changes the logic to always use OR when filtering for multiple values in the same system_profile field.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -113,7 +113,7 @@ def _object_filter_builder(input_object, spec):
                     field_value,
                     child_filter,
                     operation,
-                    is_array=None,
+                    is_array=False,
                     spec=child_spec,
                 )[0]
             )
@@ -198,8 +198,10 @@ def _nullable_wrapper(filter_function, field_name, field_value, field_filter, op
         return filter_function(field_name, field_value, operation)
 
 
-def _get_list_operator(field_name, field_filter):
-    if field_name in OR_FIELDS or field_filter == "object":
+def _get_list_operator(field_name, field_filter, operation, is_array):
+    if (operation == "eq" and not is_array) or field_name in OR_FIELDS or field_filter == "object":
+        # Filtering system_profile fields by multiple values of the same field uses OR logic,
+        # fields with type array should have the default operator
         return "OR"
     else:
         return "AND"
@@ -252,7 +254,9 @@ def _base_object_filter_builder(builder_function, field_name, field_value, field
     return ({"AND": filter_list},)
 
 
-def _base_filter_builder(builder_function, field_name, field_value, field_filter, operation, is_array=None, spec=None):
+def _base_filter_builder(
+    builder_function, field_name, field_value, field_filter, operation, is_array=False, spec=None
+):
     xjoin_field_name = field_name if spec else f"spf_{field_name}"
     base_value = _get_object_base_value(field_value, field_filter)
     if isinstance(base_value, list):
@@ -261,14 +265,9 @@ def _base_filter_builder(builder_function, field_name, field_value, field_filter
         for value in base_value:
             isolated_expression = _isolate_object_filter_expression(field_value, value)
             foo_list.append(builder_function(xjoin_field_name, isolated_expression, field_filter, operation, spec)[0])
-        list_operator = _get_list_operator(field_name, field_filter)
+        list_operator = _get_list_operator(field_name, field_filter, operation, is_array)
 
-        if operation == "eq" and not is_array:
-            # Filtering system_profile fields by multiple values of the same field uses OR logic,
-            # fields with type array should have the default operator
-            field_filter = ({"OR": foo_list},)
-        else:
-            field_filter = ({list_operator: foo_list},)
+        field_filter = ({list_operator: foo_list},)
     elif isinstance(base_value, str):
         logger.debug("filter value is a string")
         isolated_expression = _isolate_object_filter_expression(field_value, base_value)
@@ -281,15 +280,13 @@ def _base_filter_builder(builder_function, field_name, field_value, field_filter
 
 
 def _generic_filter_builder(
-    builder_function, field_name, field_value, field_filter, operation, is_array=None, spec=None
+    builder_function, field_name, field_value, field_filter, operation, is_array=False, spec=None
 ):
     spec_builder_function = partial(builder_function, spec=spec)
     nullable_builder_function = partial(_nullable_wrapper, spec_builder_function)
     if field_filter == "object":
         return _base_object_filter_builder(nullable_builder_function, field_name, field_value, field_filter, spec)
     else:
-        if is_array is None and spec is not None:
-            is_array = spec["is_array"]
         return _base_filter_builder(
             nullable_builder_function, field_name, field_value, field_filter, operation, is_array, spec
         )
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -261,7 +261,11 @@ def _base_filter_builder(builder_function, field_name, field_value, field_filter
             isolated_expression = _isolate_object_filter_expression(field_value, value)
             foo_list.append(builder_function(xjoin_field_name, isolated_expression, field_filter, operation, spec)[0])
         list_operator = _get_list_operator(field_name, field_filter)
-        field_filter = ({list_operator: foo_list},)
+        if len(foo_list) > 1:
+            # Filtering system_profile fields by multiple values of the same field uses OR logic
+            field_filter = ({"OR": foo_list},)
+        else:
+            field_filter = ({list_operator: foo_list},)
     elif isinstance(base_value, str):
         logger.debug("filter value is a string")
         isolated_expression = _isolate_object_filter_expression(field_value, base_value)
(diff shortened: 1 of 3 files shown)


FabriciaDinizRH:
@fstavela I don't understand some of the IQE errors in the build (the `False(...)` ones specifically)

thearifismail (dismissed):

kruai (changes requested):

api/filtering/filtering.py
```diff
@@ -260,7 +260,11 @@ def _base_filter_builder(builder_function, field_name, field_value, field_filter
             isolated_expression = _isolate_object_filter_expression(field_value, value)
             foo_list.append(builder_function(xjoin_field_name, isolated_expression, field_filter, operation, spec)[0])
         list_operator = _get_list_operator(field_name, field_filter)
-        field_filter = ({list_operator: foo_list},)
+        if len(foo_list) > 1:
+            # Filtering system_profile fields by multiple values of the same field uses OR logic
+            field_filter = ({"OR": foo_list},)
+        else:
+            field_filter = ({list_operator: foo_list},)
```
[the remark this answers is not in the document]
FabriciaDinizRH: Yeah, it should work for any comparator. I'll take a look at that

fstavela:
@FabriciaDinizRH There is just one issue and it's on the `sap_sids` filter. This field is a list, so we should keep the AND logic on the filter here. There 2 options for how to use this filter, either with a `contains` comparator or without any comparator. With the `contains` comparator (`filter[system_profile][sap_sids][contains][]=ABC&filter[system_profile][sap_sids][contains][]=C1E`) everything works fine and the filter uses `AND` logic. But, if I don't use any comparator (`filter[system_profile][sap_sids][]=ABC&filter[system_profile][sap_sids][]=C1E`), then it uses `OR` logic. UI uses this filter and it used `AND` logic until now, so I don't think we should change it.

I also found one more issue with the filtering of integer and datetime fields with different than `eq` comparators, like `gt`, `gte`, etc., but it didn't work even before this PR, so I created a new Jira for it and it can be fixed in another PR: https://issues.redhat.com/browse/RHINENG-5911

Otherwise, the PR works fine, thanks.

api/filtering/filtering.py
```diff
@@ -113,6 +113,7 @@ def _object_filter_builder(input_object, spec):
                     field_value,
                     child_filter,
                     operation,
+                    child_is_array,
```
kruai, line 116: `child_is_array` is equal to `child_spec["is_array"]`. Since we're already passing `child_spec` to the function, can we just determine whether it's an array from inside the builder function?

kruai (commented):
Looks good, but just one question before I approve

api/filtering/filtering.py
FabriciaDinizRH: you're right. I fixed it.
I'm not entirely satisfied with my solution but let me know what you think.

kruai (commented):
I played around with this PR a bit, and I don't think the `is_array` param is needed. The changes I'd like to suggest would be too unwieldy for PR comments, so I've put them in this commit: https://github.com/RedHatInsights/insights-host-inventory/commit/ff3b75c3663e94f592115dcbd59dfafd0c26a1d5

Please correct me if I'm missing something though!

FabriciaDinizRH:
> I played around with this PR a bit, and I don't think the `is_array` param is needed. The changes I'd like to suggest would be too unwieldy for PR comments, so I've put them in this commit: [ff3b75c](https://github.com/RedHatInsights/insights-host-inventory/commit/ff3b75c3663e94f592115dcbd59dfafd0c26a1d5)
> 
> Please correct me if I'm missing something though!

@kruai the `sap_sids` fields should remain using the AND logic, as per Fero's last comment. That's why I included a check for 'is_array'.

kruai:
> > I played around with this PR a bit, and I don't think the `is_array` param is needed. The changes I'd like to suggest would be too unwieldy for PR comments, so I've put them in this commit: [ff3b75c](https://github.com/RedHatInsights/insights-host-inventory/commit/ff3b75c3663e94f592115dcbd59dfafd0c26a1d5)
> > Please correct me if I'm missing something though!
> 
> @kruai the `sap_sids` fields should remain using the AND logic, as per Fero's last comment. That's why I included a check for 'is_array'.

Ah, right - please disregard that suggestion, then. I've been re-reviewing and might have another suggestion this afternoon. Since `is_array` is implied by the contents of `spec`, and it feels like the logic for determining OR vs AND should be contained within `_get_list_operator`, I wanna try out a couple more things

api/filtering/filtering.py
```diff
@@ -113,7 +113,8 @@ def _object_filter_builder(input_object, spec):
                     field_value,
                     child_filter,
                     operation,
-                    child_spec,
+                    is_array=None,
```
kruai, line 116: Making a few suggestions for `is_array` to be False instead of None, since the naming and usage implies a boolean
```suggestion
                    is_array=False,
```

api/filtering/filtering.py
```diff
@@ -251,7 +252,7 @@ def _base_object_filter_builder(builder_function, field_name, field_value, field
     return ({"AND": filter_list},)
 
 
-def _base_filter_builder(builder_function, field_name, field_value, field_filter, operation, spec=None):
+def _base_filter_builder(builder_function, field_name, field_value, field_filter, operation, is_array=None, spec=None):
```
kruai, line 255:
```suggestion
def _base_filter_builder(builder_function, field_name, field_value, field_filter, operation, is_array=False, spec=None):
```

api/filtering/filtering.py
```diff
@@ -261,7 +262,13 @@ def _base_filter_builder(builder_function, field_name, field_value, field_filter
...
             foo_list.append(builder_function(xjoin_field_name, isolated_expression, field_filter, operation, spec)[0])
         list_operator = _get_list_operator(field_name, field_filter)
-        field_filter = ({list_operator: foo_list},)
+
+        if operation == "eq" and not is_array:
+            # Filtering system_profile fields by multiple values of the same field uses OR logic,
+            # fields with type array should have the default operator
+            field_filter = ({"OR": foo_list},)
+        else:
+            field_filter = ({list_operator: foo_list},)
```
kruai, line 271: I think this logic should be contained within the `_get_list_operator` function, so that the logic for determining "OR" vs "AND" is done in one place. So L263 would change to:
```
list_operator = _get_list_operator(field_name, field_filter, operation, is_array)
```
and the function itself would change to:
```
def _get_list_operator(field_name, field_filter, operation, is_array):
    if (operation == "eq" and not is_array) or field_name in OR_FIELDS or field_filter == "object":
        # Filtering system_profile fields by multiple values of the same field uses OR logic,
        # fields with type array should have the default operator
        return "OR"
    else:
        return "AND"
```

api/filtering/filtering.py
```diff
@@ -273,13 +280,19 @@ def _base_filter_builder(builder_function, field_name, field_value, field_filter
     return field_filter
 
 
-def _generic_filter_builder(builder_function, field_name, field_value, field_filter, operation, spec=None):
+def _generic_filter_builder(
+    builder_function, field_name, field_value, field_filter, operation, is_array=None, spec=None
```
kruai, line 284:
```suggestion
    builder_function, field_name, field_value, field_filter, operation, is_array=False, spec=None
```

api/filtering/filtering.py
```diff
@@ -273,13 +280,19 @@ def _base_filter_builder(builder_function, field_name, field_value, field_filter
...
+):
     spec_builder_function = partial(builder_function, spec=spec)
     nullable_builder_function = partial(_nullable_wrapper, spec_builder_function)
     if field_filter == "object":
         return _base_object_filter_builder(nullable_builder_function, field_name, field_value, field_filter, spec)
     else:
-        return _base_filter_builder(nullable_builder_function, field_name, field_value, field_filter, operation, spec)
+        if is_array is None and spec is not None:
```
kruai, line 291:
```suggestion
        if not is_array and spec is not None:
```

kruai: But also, is this conditional needed? In what situation would this apply?

kruai (changes requested):
OK, just got a few suggestions for logic consolidation and cleanup.

api/filtering/filtering.py
FabriciaDinizRH: I used it as a way of populating `is_array` in case it is not provided but the `spec` is. But it turns out the code works fine without it.

kruai (approved):

thearifismail (approved):
LGTM

PR #1573, opened 2023-12-04
RHINENG-4700_and_5552: Use deletion for culling and rename attributes in custom staleness

thearifismail:
# Overview

This PR is being created to address [RHINENG-4700](https://issues.redhat.com/browse/RHINENG-4700) and [RHINENG-5552](https://issues.redhat.com/browse/RHINENG-5552) and replace the [PR #1550](https://github.com/RedHatInsights/insights-host-inventory/pull/1550) 

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


fellipeh (changes requested):
How this will work with the current databases? I saw the changes on migrations folder only for the fields name creation..,

thearifismail:
> How this will work with the current databases? I saw the changes on migrations folder only for the fields name creation..,

After making these changes, I successfully ran the make task `make upgrade_db` followed by `pytest`.

fellipeh (approved):
lgtm

fstavela:
> > How this will work with the current databases? I saw the changes on migrations folder only for the fields name creation..,
> 
> After making these changes, I successfully ran the make task `make upgrade_db` followed by `pytest`.

Did you run the DB upgrade on an already existing database or on a new one? This change to the DB migration is my only concern with this PR, I think the proper way would be to create a new migration instead of updating an existing one. I don't know if this approach will work on existing databases in Stage and Prod. But I've never done this, so correct me if I'm wrong.

Otherwise, the PR looks good.

thearifismail:
> > > How this will work with the current databases? I saw the changes on migrations folder only for the fields name creation..,
> > 
> > 
> > After making these changes, I successfully ran the make task `make upgrade_db` followed by `pytest`.
> 
> Did you run the DB upgrade on an already existing database or on a new one? This change to the DB migration is my only concern with this PR, I think the proper way would be to create a new migration instead of updating an existing one. I don't know if this approach will work on existing databases in Stage and Prod. But I've never done this, so correct me if I'm wrong.
> 
> Otherwise, the PR looks good.

Thank you @fstavela; very good point.  I ran it on a new DB but let me run against an older one, i.e. migration.

api/staleness_query.py
```diff
@@ -38,12 +38,12 @@ def _build_staleness_sys_default(org_id, config=None):
...
-            "conventional_culling_delta": config.conventional_culling_seconds,
-            "immutable_staleness_delta": config.immutable_staleness_seconds,
-            "immutable_stale_warning_delta": config.immutable_stale_warning_seconds,
-            "immutable_culling_delta": config.immutable_culling_seconds,
+            "conventional_time_to_stale": config.conventional_staleness_seconds,
+            "conventional_time_to_stale_warning": config.conventional_stale_warning_seconds,
+            "conventional_time_to_delete": config.conventional_culling_seconds,
+            "immutable_time_to_stale": config.immutable_staleness_seconds,
+            "immutable_time_to_stale_warning": config.immutable_stale_warning_seconds,
+            "immutable_time_to_delete": config.immutable_culling_seconds,
```
kruai, line 46: We might as well update the field names in the config file and environment variables too; it'll be confusing if we don't.

kruai (changes requested):
+1 to the point about the migrations - you'll need to make a new migration rather than modify the existing ones.
Also, there are a few places where the field names haven't been updated, and it would make sense to update them all at once.

thearifismail:
> +1 to the point about the migrations - you'll need to make a new migration rather than modify the existing ones. Also, there are a few places where the field names haven't been updated, and it would make sense to update them all at once.

hmmm... I did bulk find and resplace.  Let me look.  Thanks.

api/staleness_query.py
thearifismail: Done!!!

jpramos123 (approved):
LGTM!

Thanks for this PR.

the `_time_to` was a good approach to solve the variables naming we discussed!!

PR #1603, opened 2024-01-25
[RHINENG-6095] - Use custom staleness in per_reporter_staleness

jpramos123:
# Overview

It adds custom staleness to per_reporter_staleness graphql filter.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -31,6 +31,7 @@ logger = get_logger(__name__)
 NIL_STRING = "nil"
 NOT_NIL_STRING = "not_nil"
 OR_FIELDS = ("owner_id", "rhc_client_id", "host_type", "system_update_method")
+STALENESS_VALUES = ["fresh", "stale", "stale_warning"]
 
 
 def _invalid_value_error(field_name, field_value):
@@ -291,12 +292,12 @@ def _generic_filter_builder(
 
 
 def build_registered_with_filter(registered_with, staleness):
-    # if API is calling /api/inventory/v1/tags
-    # staleness is None by default, so we need
-    # to add the staleness values as a expection when
-    # when registered_with is set
+    # If /api/inventory/v1/tags is called,
+    # staleness variable is set to None by default.
+    # To use custom staleness, we to make sure we have
+    # staleness values set
     if staleness is None:
-        staleness = ["fresh", "stale", "stale_warning"]
+        staleness = STALENESS_VALUES
 
     reg_with_copy = deepcopy(registered_with)
     prs_list = []
@@ -311,7 +312,7 @@ def build_registered_with_filter(registered_with, staleness):
                 reg_with_copy.extend(OLD_TO_NEW_REPORTER_MAP[old_reporter])
                 reg_with_copy = list(set(reg_with_copy))  # Remove duplicates
 
-        # I need to get the per_report_staleness check_in value for the reporter
+        # Get the per_report_staleness check_in value for the reporter
         # and build the filter based on it
         for item in reg_with_copy:
             prs_item = {
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -1,6 +1,4 @@
 from copy import deepcopy
-from datetime import datetime
-from datetime import timezone
 from enum import Enum
 from functools import partial
 from typing import Tuple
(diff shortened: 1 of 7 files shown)


api/filtering/filtering.py
```diff
@@ -292,7 +290,14 @@ def _generic_filter_builder(
         )
 
 
-def build_registered_with_filter(registered_with):
+def build_registered_with_filter(registered_with, staleness):
+    # if API is calling /api/inventory/v1/tags
```
thearifismail, line 294: This comment is hard to make sense.  Does this comment pertain to `staleness` or the whole function?  If for the whole function, then move it the outside of the function

tests/fixtures/graphql_fixtures.py
```diff
@@ -156,7 +156,56 @@ def _assert_tag_query_host_filter_for_field(url, field, matcher, value, status=2
             url=url,
             host_filter={
                 "AND": ({field: {matcher: value.casefold() if field in CASEFOLDED_FIELDS else value}},),
-                "OR": mocker.ANY,
+                "OR": [
```
thearifismail, line 159: To me, it looks more like a library function than a fixture

thearifismail (commented):
Minor comments

api/filtering/filtering.py
jpramos123: This to `staleness` only to clarify why this is being added to the code.

I made some changes to the comment, hope it makes more sense now

Thanks

tests/fixtures/graphql_fixtures.py
jpramos123: To me it makes sense to be a fixture because it is the expected value to be received in the test.

api/filtering/filtering.py
```diff
@@ -292,7 +290,14 @@ def _generic_filter_builder(
...
+    # staleness is None by default, so we need
+    # to add the staleness values as a expection when
+    # when registered_with is set
+    if staleness is None:
+        staleness = ["fresh", "stale", "stale_warning"]
```
vkrizan, line 299: Why not handle the `None` value, e.g. by skipping the condition (or adding an always true condition)? Or, moving this down to `build_registered_with_filter`?

If you wanna keep this default set, then this is not a correct place where these values should be kept. Probably move those up and make it as a constant.

jpramos123: @vkrizan I wanna keep it now! As this code is going deprecated with xjoin migration

app/xjoin.py
```diff
@@ -97,6 +97,16 @@ def staleness_filter(staleness):
     return staleness_conditions
 
 
+def per_reporter_staleness_filter(staleness):
+    staleness_obj = serialize_staleness_to_dict(get_staleness_obj())
+    staleness_conditions = tuple()
+    for host_type in HOST_TYPES:
+        staleness_conditions += tuple(
```
vkrizan, line 104: Interesting use of append to an immutable container/type . A bit non-standard, but it works as expected.

vkrizan (approved):
LGTM

app/xjoin.py
kruai: Over time, we started aligning on using tuples to collect filters for some reason. I've come to like it though

thearifismail (approved):
LGTM

app/xjoin.py
vkrizan: I kinda see the purpose of it -- to utilize an immutable type. It's a bit bitter to my taste. Append does not work on an immutable type, hence under the hood a new tuple is created each time. And since this is done couple of times, a more appropriate container might be used (i.e. `list`). The tuple behavior here is implicit.

kruai (approved):
Ideally we'd want to test the xjoin request when staleness is set to one or two values instead of all 3, but LGTM

app/xjoin.py
kruai: Oh, I didn't realize that. I've been using tuples for the rewrite as well; I'll switch those to sets or lists. Thanks for the info!

swagger/api.spec.yaml
```diff
@@ -66,7 +66,7 @@ paths:
         - $ref: '#/components/parameters/updatedEnd'
         - $ref: '#/components/parameters/groupNameListParam'
         - $ref: '#/components/parameters/registered_with'
-        - $ref: '#/components/parameters/stalenessNoDefaultsParam'
```
fstavela, line 69: This needs to be reverted. The different staleness param was here on purpose. We don't want to pass any defaults to the `DELETE /hosts` endpoint, because then it would be possible to use this endpoint without specifying any parameter and thus deleting the entire inventory.

PR #1614, opened 2024-02-05
Convert GET /system_profile/operating_system endpoint from XJoin to DB (Part 1)

chambridge:
# Overview

This PR is being created to address [RHINENG-7075](https://issues.redhat.com/browse/RHINENG-7075).

* Add feature flag call in system profile API to call DB method
* Add DB method to aggregate on operating systems
* Enable get_host_list_by_id_list_from_db to only filter on the host list if provided
* Add test case to check aggregation function
* Add api util helper since no test case existed for this endpoint before
* Clean up limit and offset to reuse pagination_params method
* Clean up host query db usage to pass limit offset

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/host.py
+++ b/api/host.py
@@ -49,6 +49,7 @@ from app.queue.events import message_headers
 from app.serialization import deserialize_canonical_facts
 from app.serialization import serialize_host
 from app.utils import Tag
+from app.xjoin import pagination_params
 from lib.feature_flags import FLAG_INVENTORY_DISABLE_XJOIN
 from lib.feature_flags import get_flag_value
 from lib.host_delete import delete_hosts
@@ -464,8 +465,9 @@ def get_host_tag_count(host_id_list, page=1, per_page=100, order_by=None, order_
     current_identity = get_current_identity()
     if get_flag_value(FLAG_INVENTORY_DISABLE_XJOIN, context={"schema": current_identity.org_id}):
         logger.info(f"{FLAG_INVENTORY_DISABLE_XJOIN} is applied to {current_identity.org_id}")
+        limit, offset = pagination_params(page, per_page)
         host_list, total = get_host_tags_list_by_id_list_postgres(
-            host_id_list, page, per_page, order_by, order_how, rbac_filter
+            host_id_list, limit, offset, order_by, order_how, rbac_filter
         )
     else:
         host_list, total = get_host_tags_list_by_id_list(
@@ -483,8 +485,9 @@ def get_host_tags(host_id_list, page=1, per_page=100, order_by=None, order_how=N
     current_identity = get_current_identity()
     if get_flag_value(FLAG_INVENTORY_DISABLE_XJOIN, context={"schema": current_identity.org_id}):
         logger.info(f"{FLAG_INVENTORY_DISABLE_XJOIN} is applied to {current_identity.org_id}")
+        limit, offset = pagination_params(page, per_page)
         host_list, total = get_host_tags_list_by_id_list_postgres(
-            host_id_list, page, per_page, order_by, order_how, rbac_filter
+            host_id_list, limit, offset, order_by, order_how, rbac_filter
         )
     else:
         host_list, total = get_host_tags_list_by_id_list(
--- a/api/system_profile.py
+++ b/api/system_profile.py
@@ -9,6 +9,7 @@ from api import custom_escape
 from api import flask_json_response
 from api import metrics
 from api.filtering.filtering import query_filters
+from api.host_query_db import get_os_info as get_os_info_db
 from app import RbacPermission
 from app import RbacResourceType
 from app.auth import get_current_identity
@@ -25,6 +26,8 @@ from app.xjoin import check_pagination
 from app.xjoin import graphql_query
 from app.xjoin import pagination_params
 from app.xjoin import staleness_filter
+from lib.feature_flags import FLAG_INVENTORY_DISABLE_XJOIN
+from lib.feature_flags import get_flag_value
 from lib.middleware import rbac
 from lib.system_profile_validate import validate_sp_for_branch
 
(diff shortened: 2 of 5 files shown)


api/system_profile.py
```diff
@@ -211,36 +214,48 @@ def get_operating_system(
     rbac_filter=None,
 ):
     limit, offset = pagination_params(page, per_page)
+    current_identity = get_current_identity()
+    if get_flag_value(FLAG_INVENTORY_DISABLE_XJOIN, context={"schema": current_identity.org_id}):
```
vkrizan, line 218: A bit unrelated to this, how is it with the `context` being repeated on similar flag value retrievals?

api/system_profile.py
```diff
@@ -211,36 +214,48 @@ def get_operating_system(
...
+            limit,
+            offset,
+            staleness=staleness,
+            tags=tags,
+            registered_with=registered_with,
+            filter=filter,
+            rbac_filter=rbac_filter,
+        )
+        return flask_json_response(build_collection_response(results, page, per_page, total))
+    else:
```
vkrizan, line 229: Nit: The `else` block is unnecessary as there is return above in the `if` block. Less changes.

lib/host_repository.py
```diff
@@ -298,10 +298,10 @@ def update_system_profile(input_host, identity, staleness_offset, staleness):
 
 def get_host_list_by_id_list_from_db(host_id_list, rbac_filter=None, columns=None):
     current_identity = get_current_identity()
-    filters = (
-        Host.org_id == current_identity.org_id,
-        Host.id.in_(host_id_list),
-    )
+    filters = (Host.org_id == current_identity.org_id,)
+    if host_id_list:
```
vkrizan, line 302: Observation: This seeks for refactoring of the function, as it changes (adds) the purpose of it. Above this is called with `None` as a first value.

tests/test_system_profile.py
```diff
@@ -330,3 +334,56 @@ def test_validate_sp_for_invalid_days(api_post):
...
+
+    # Create an association between the insights IDs
+    ordered_host_data = dict(zip(ordered_insights_ids, ordered_operating_system_data))
+
+    # Create a shuffled list of insights_ids so we can create the hosts in a random order
+    shuffled_insights_ids = ordered_insights_ids.copy()
+    random.shuffle(shuffled_insights_ids)
+
+    # Create hosts for the above host data (in shuffled order)
+    _ = [
```
vkrizan, line 361: Is this assignment necessary for the linter?

vkrizan (approved):
Functionally it looks good. I haven't verified if the response matches with the XJoin one.

tests/test_system_profile.py
chambridge: yes

api/system_profile.py
chambridge: Can you ask this question differently? I don't quite follow. I will say that context was always `None` for other flags prior to this as they didn't evaluate against any custom strategies.

api/system_profile.py
chambridge: yep, I'll remove the `else`.

chambridge: done

thearifismail (commented):
Minor comments, else LGTM

api/host.py
```diff
@@ -464,8 +465,9 @@ def get_host_tag_count(host_id_list, page=1, per_page=100, order_by=None, order_
     current_identity = get_current_identity()
     if get_flag_value(FLAG_INVENTORY_DISABLE_XJOIN, context={"schema": current_identity.org_id}):
         logger.info(f"{FLAG_INVENTORY_DISABLE_XJOIN} is applied to {current_identity.org_id}")
+        limit, offset = pagination_params(page, per_page)
```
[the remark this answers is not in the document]
chambridge: Allowed me to remove code in each db method that was doing this conversion, so essentially avoiding copy/paste or other bug introduction errors.

lib/host_repository.py
```diff
@@ -298,10 +298,18 @@ def update_system_profile(input_host, identity, staleness_offset, staleness):
...
     current_identity = get_current_identity()
-    filters = (
-        Host.org_id == current_identity.org_id,
-        Host.id.in_(host_id_list),
-    )
+    filters = (Host.org_id == current_identity.org_id, Host.id.in_(host_id_list))
+    return _get_hosts_from_db(filters, rbac_filter, columns)
+
+
+def get_hosts_from_db(rbac_filter=None, columns=None):
```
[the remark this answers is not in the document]
chambridge: Sure, but for consistency I would say that the original function `get_host_list_by_id_list_from_db` would need to change to something like `get_host_list_by_id_list_from_db_query`, which I'm not oppose to, but I chose to be consistent with existing naming convention rather than changing more file.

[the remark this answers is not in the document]
thearifismail: I think it makes future readings of the code easier and easy to understand.  I am approving it.  You can change if you like.

thearifismail (approved):

thearifismail:
@kruai ,  The PR got merged while I was reviewing it. The unleash container does not run when using `podman`.  I have been using `podman` for the last couple of days and got too lazy to enable `docker`.  When using `podman` and `podman-compose`, the `unleash` container does not launch.  Switched to docker and it worked.

lib/host_repository.py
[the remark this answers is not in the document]
chambridge: merged in changes with Asa latest to reuse more of the functionality, so this new function was removed.

kruai (approved):
LGTM :+1:

api/system_profile.py
vkrizan: The `get_flag_value` with the context following the same template is being repeated for other flag retrievals. One can make a mistake using the right keys/data within the context. My suggestion here is to think about creating a separate function that would handle it in DRY fashion. Smth like `flag_value_with_identity(flag_name: string, identity: Identity)`. Of course the name can be improved. cc @kruai

thearifismail (approved):
LGTM

api/system_profile.py
vkrizan: It doesn't have to be a separate one. Extending the current one, e.g. with `identity` keyword arg, might be a better option.

PR #1615, opened 2024-02-06
Move DB filters to their own file; enable 2 feature flags in local cfg

kruai:
# Overview

This PR is being created to move the DB filtering functions to their own file (`db_filters.py`). It also enables the "hbi.group-assignment-rules" and "hbi.custom-staleness" feature flags in the local unleash config.


chambridge (approved):
LGTM

PR #1617, opened 2024-02-07
[RHINENG-6095] - fix staleness filter

jpramos123:
# Overview

It fixes a issue found when filtering hosts with `registered_with=`, before the it was not able to identify the correct staleness for each of the reporters. Now it is able to filter the hosts by the reporter_staleness correctly.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -31,7 +31,7 @@ logger = get_logger(__name__)
 NIL_STRING = "nil"
 NOT_NIL_STRING = "not_nil"
 OR_FIELDS = ("owner_id", "rhc_client_id", "host_type", "system_update_method")
-STALENESS_VALUES = ["fresh", "stale", "stale_warning"]
+DEFAULT_STALENESS_VALUES = ["not_culled"]
 
 
 def _invalid_value_error(field_name, field_value):
@@ -291,14 +291,7 @@ def _generic_filter_builder(
         )
 
 
-def build_registered_with_filter(registered_with, staleness):
-    # If /api/inventory/v1/tags is called,
-    # staleness variable is set to None by default.
-    # To use custom staleness, we to make sure we have
-    # staleness values set
-    if staleness is None:
-        staleness = STALENESS_VALUES
-
+def build_registered_with_filter(registered_with):
     reg_with_copy = deepcopy(registered_with)
     prs_list = []
     if "insights" in reg_with_copy:
@@ -314,19 +307,15 @@ def build_registered_with_filter(registered_with, staleness):
 
         # Get the per_report_staleness check_in value for the reporter
         # and build the filter based on it
-        for item in reg_with_copy:
-            prs_item = {
-                "per_reporter_staleness": {
-                    "reporter": {"eq": item.replace("!", "")},
-                    "AND": per_reporter_staleness_filter(staleness),
-                },
-            }
+        for reporter in reg_with_copy:
+            prs_item = per_reporter_staleness_filter(DEFAULT_STALENESS_VALUES, reporter)
 
             # If registered_with starts with "!", we want to invert the condition.
-            if item.startswith("!"):
+            if reporter.startswith("!"):
                 prs_item = {"NOT": prs_item}
 
-            prs_list.append(prs_item)
+            for n_items in prs_item:
+                prs_list.append(n_items)
 
     return ({"OR": prs_list},)
 
@@ -488,11 +477,12 @@ def query_filters(
         tag_filters = build_tag_query_dict_tuple(tags)
         query_filters += ({"OR": tag_filters},)
     if staleness:
-        staleness_filters = tuple(staleness_filter(staleness))
-        query_filters += ({"OR": staleness_filters},)
+        if not registered_with:
+            staleness_filters = tuple(staleness_filter(staleness))
+            query_filters += ({"OR": staleness_filters},)
 
     if registered_with:
-        query_filters += build_registered_with_filter(registered_with, staleness)
+        query_filters += build_registered_with_filter(registered_with)
     if provider_type:
         query_filters += ({"provider_type": {"eq": provider_type.casefold()}},)
     if provider_id:
--- a/app/xjoin.py
+++ b/app/xjoin.py
@@ -1,3 +1,4 @@
+from functools import partial
 from logging import getLogger
 
 from flask import abort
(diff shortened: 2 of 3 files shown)


api/filtering/filtering.py
```diff
@@ -31,7 +31,7 @@
 NIL_STRING = "nil"
 NOT_NIL_STRING = "not_nil"
 OR_FIELDS = ("owner_id", "rhc_client_id", "host_type", "system_update_method")
-STALENESS_VALUES = ["fresh", "stale", "stale_warning"]
+STALENESS_VALUES = ["not_culled"]
```
vkrizan, line 34: Doesn't this change the semantics of this constant to be default staleness values?

api/filtering/filtering.py
```diff
@@ -488,10 +484,12 @@ def query_filters(
...
         query_filters += ({"OR": tag_filters},)
     if staleness:
-        staleness_filters = tuple(staleness_filter(staleness))
-        query_filters += ({"OR": staleness_filters},)
+        if not registered_with:
+            staleness_filters = tuple(staleness_filter(staleness))
+            query_filters += ({"OR": staleness_filters},)
 
     if registered_with:
+        staleness = STALENESS_VALUES
```
vkrizan, line 492: Since the default was moved out of the `build_registered_with_filter` function, is the `None` handling of `staleness`  arg still necessary?

app/culling.py
```diff
@@ -87,3 +87,12 @@ def staleness_to_conditions(staleness, staleness_states, host_type, timestamp_fi
     condition = Conditions(staleness, host_type)
     filtered_states = (state for state in staleness_states if state != "unknown")
     return (timestamp_filter_func(*getattr(condition, state)(), host_type=host_type) for state in filtered_states)
+
+
+def staleness_to_conditions_per_reporter(staleness, staleness_states, host_type, reporter, timestamp_filter_func):
```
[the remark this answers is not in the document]
jpramos123: Changed!

api/filtering/filtering.py
jpramos123: Yeah, make sense naming it as `DEFAULT_STALENESS_VALUES`

api/filtering/filtering.py
jpramos123: Not necessary anymore! Thanks for catching this

app/culling.py
```diff
@@ -83,7 +83,12 @@ def _culled_timestamp(self):
         return self.now - offset
 
 
-def staleness_to_conditions(staleness, staleness_states, host_type, timestamp_filter_func):
+def staleness_to_conditions(staleness, staleness_states, host_type, reporter, timestamp_filter_func):
```
vkrizan, line 86: The reporter might have been last with `None` as a default. :slightly_smiling_face:

app/xjoin.py
```diff
@@ -139,10 +141,16 @@ def _stale_timestamp_filter(gt=None, lte=None, host_type=None):
     return {"AND": ({"modified_on": filter_, "spf_host_type": {"eq": host_type}})}
 
 
-def _stale_timestamp_per_reporter_filter(gt=None, lte=None, host_type=None):
+def _stale_timestamp_per_reporter_filter(gt=None, lte=None, host_type=None, reporter=None):
```
kruai, line 144: Doesn't make sense to default this to `None`, since `per_reporter_staleness.reporter` can't be None/null

app/culling.py
```diff
@@ -83,7 +83,12 @@ def _culled_timestamp(self):
...
     condition = Conditions(staleness, host_type)
     filtered_states = (state for state in staleness_states if state != "unknown")
+    if reporter:
+        return (
+            timestamp_filter_func(*getattr(condition, state)(), host_type=host_type, reporter=reporter)
+            for state in filtered_states
+        )
```
kruai, line 93: This logic is iffy to me. Now `reporter` and `timestamp_filter_func` are tightly coupled and can be in conflict: if you call this function with `reporter="puptoo"` and `timestamp_filter_func=_stale_timestamp_filter`, it'll error out. It would also return unexpected results if you use `reporter=None` with `timestamp_filter_func=_stale_timestamp_per_reporter_filter`.

The cleanest way I can think to resolve this right now is:

1. Revert this function's changes (remove the reporter param, condition, etc), and remove the `None` reporter value from all the calls to `staleness_to_conditions`
2. In `per_reporter_staleness_filter`, use a partial function for the timestamp_filter_func param: `partial(_stale_timestamp_per_reporter_filter, reporter=reporter)`

Alternatively, we could add a `reporter` param to `_stale_timestamp_filter` that simply does nothing, but that feels iffy too. @vkrizan thoughts?

kruai (changes requested):

api/filtering/filtering.py
```diff
@@ -488,10 +477,12 @@ def query_filters(
...
     if staleness:
-        staleness_filters = tuple(staleness_filter(staleness))
-        query_filters += ({"OR": staleness_filters},)
+        if not registered_with:
+            staleness_filters = tuple(staleness_filter(staleness))
+            query_filters += ({"OR": staleness_filters},)
 
     if registered_with:
+        staleness = DEFAULT_STALENESS_VALUES
         query_filters += build_registered_with_filter(registered_with, staleness)
```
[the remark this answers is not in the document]
jpramos123: Done! 

Thanks for the review

app/culling.py
jpramos123: @kruai Loved this suggestion, I applied it to the code.

Now I'm using `partial` and not duplication the function

app/xjoin.py
jpramos123: I was forcing the null because of the following error:
`SyntaxError: non-default argument follows default argument`

But I changed the place of the `reporter` attribute and we are good now

Thanks

kruai: Ah OK, that amount of rewriting isn't worth it. Thanks for clarifying!

kruai (approved):

app/culling.py
vkrizan: Good idea with the `partial`.

PR #1619, opened 2024-02-08
Convert GET /system_profile/sap_system endpoint from XJoin to DB

chambridge:
# Overview

This PR is being created to address [RHINENG-7077](https://issues.redhat.com/browse/RHINENG-7077).

* Add feature flag and new DB based call to API
* Create new DB query aggregation on sap_system value
* Add new test case to create sap_system entries and compare aggregation for true/false values

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/host_query_db.py
+++ b/api/host_query_db.py
@@ -1,6 +1,7 @@
 from typing import List
 from typing import Tuple
 
+from sqlalchemy import Boolean
 from sqlalchemy import func
 from sqlalchemy.orm import Query
 from sqlalchemy.sql.expression import ColumnElement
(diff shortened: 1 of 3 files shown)


tests/test_system_profile.py
```diff
@@ -387,3 +387,47 @@ def test_system_profile_operating_system(mq_create_or_update_host, api_get):
...
+        if datum:
+            sap_list.append(datum)
+        else:
+            not_sap_list.append(datum)
+
+    with patch("api.system_profile.get_flag_value", return_value=True):
+        # Validate the basics, i.e. response code and results size
+        response_status, response_data = api_get(url)
+        assert response_status == 200
+        assert len(list(set(ordered_sap_system_data))) == len(response_data["results"])
```
kruai, line 425: Does it need to be converted to a list?
```suggestion
        assert len(set(ordered_sap_system_data)) == len(response_data["results"])
```

tests/test_system_profile.py
```diff
@@ -387,3 +387,47 @@ def test_system_profile_operating_system(mq_create_or_update_host, api_get):
...
+
+    for index in range(len(response_data["results"])):
+        item = response_data["results"][index]
+        item_count = item["count"]
+        if item["value"]:
+            assert item_count == len(sap_list)
+        else:
+            assert item_count == len(not_sap_list)
```
kruai, line 433: You know, I didn't think about this for `test_system_profile_operating_system`, but... I did all this shuffling and ordering for `test_get_hosts_order_by_operating_system` because I was validating the sorting of operating_system objects.  If ordering isn't important to these tests, all this might be kinda overkill. wdyt?

kruai (changes requested):

tests/test_system_profile.py
chambridge: ok, pushed

tests/test_system_profile.py
chambridge: sure. removed.

api/host_query_db.py
```diff
@@ -210,3 +211,25 @@ def get_os_info(limit, offset, staleness, tags, registered_with, filter, rbac_fi
...
+
+
+def get_sap_system_info(limit, offset, staleness, tags, registered_with, filter, rbac_filter):
+    columns = [
+        Host.system_profile_facts["sap_system"].label("value"),
+    ]
+    sap_query = _find_all_hosts(columns)
+    sap_filter = [
+        func.jsonb_typeof(Host.system_profile_facts["sap_system"]) == "boolean",
+        Host.system_profile_facts["sap_system"].astext.cast(Boolean) is not None,
```
vkrizan, line 223: Isn't the `is not None` evaluated as statement here? AFAIK identity check with `is` cannot be overloaded (o be used by frameworks). But I might be wrong.

chambridge: Searched at the Sqlalchemy on this topic. Looks like the `is not None` sequence results in just a `True` instead of adding a clause for `IS NOT NULL`. I've updated the code to use a `!= None` which looks like it is equivalent to the usage of `is_not(None)`.

vkrizan (approved):

kruai (approved):

thearifismail (approved):

PR #1623, opened 2024-02-14
perf(RHINENG-6139) Add pg_repack cronjob

kruai:
# Overview

This PR is being created to address [RHINENG-6139](https://issues.redhat.com/browse/RHINENG-6139).
Adds pg_repack and all its dependencies, as well as a cronjob to repack the DB weekly.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


Dockerfile
```diff
@@ -30,7 +31,8 @@ ENV PIPENV_VENV_IN_PROJECT=1
 
 RUN python3 -m pip install --upgrade pip setuptools wheel && \
     python3 -m pip install pipenv && \
-    pipenv install --system --dev
+    pipenv install --system --dev && \
+    pgxn install pg_repack
```
vkrizan, line 35: Do we want to install unvetted postgresql extensions to the application container? I think that this might pose some risk(s). What about having a separate (sub)container for the repack, or would that be an overkill?
cc @chambridge

kruai: The container would need all of the usual dependencies for the app, but also the extra packages you see above. The other containers wouldn't need all of that, so having a different container would streamline the main containers, at least.

WDYM by subcontainer, though - do you mean creating a new image stream that uses this one as a base?

kruai:
Tested in EEs. The pg_repack client successfully installs and connects to the DB.
It complains that the extension isn't installed, but that much is expected. The extension is already installed in Stage and Prod by default, and we're going to manually add it to the DBs before allowing the job to run.

Dockerfile
vkrizan: For subcontainer I meant to use multistage builds. One base, one for app, one for pg_repack. Unless it is a bad idea. I know that usually by default it operates on last build container (i.e. for tagging during build).

deploy/clowdapp.yml
```diff
@@ -635,6 +635,36 @@ objects:
...
+        resources:
+          limits:
+            cpu: ${CPU_LIMIT_PG_REPACK}
+            memory: ${MEMORY_LIMIT_PG_REPACK}
+          requests:
+            cpu: ${CPU_REQUEST_PG_REPACK}
+            memory: ${MEMORY_REQUEST_PG_REPACK}
     database:
       name: ${DB_NAME}
       version: 12
```
thearifismail, line 670: hasn't the DB version upgraded to 13?  App interface has https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/host-inventory/namespaces/prod-host-inventory-prod.yml?ref_type=heads#L97

Dockerfile
vkrizan: Good Q. @thearifismail. In normal scenarios of multistage builds, the result is a single image. However, there are at two other possibilities: building a specific stage or tagging a layer (not ideal). Probably multistage would not the be a right fit here.

kruai: @thearifismail Yep, I built locally a number of times, which is how I determined that all all of these are required in order to install pg_repack (unfortunately).
If we wanna play it safe, we could make a new image for this called host-inventory-repack or something, and use FROM insights-host-inventory:latest, or something like that. If that's the direction we wanna go, would it be all right if I separated it out in a follow-up Jira @vkrizan @chambridge?

Dockerfile
```diff
@@ -17,9 +17,10 @@ WORKDIR $APP_ROOT
 
 RUN microdnf module enable postgresql:13 python39:3.9 && \
     microdnf upgrade -y && \
-    microdnf install --setopt=tsflags=nodocs -y postgresql python39 rsync tar procps-ng make snappy && \
+    microdnf install --setopt=tsflags=nodocs -y postgresql python39 rsync tar procps-ng make snappy gcc postgresql-devel \
```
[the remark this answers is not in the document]
kruai: It appears to be used by `unit_test.sh`.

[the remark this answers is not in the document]
kruai: ([Right here, specifically](https://github.com/RedHatInsights/insights-host-inventory/blob/master/unit_test.sh#L11))

Dockerfile
chambridge: I'm fine with a follow up PR. I appreciate others trying to reduce risk that could be introduced the package addition.

kruai: Cool. I opened [RHINENG-8356](https://issues.redhat.com/browse/RHINENG-8356) to address this.

run_pg_repack.py
```diff
@@ -0,0 +1,50 @@
...
+            str(cfg._db_user),
+            "-k",
+            str(cfg._db_name),
+        ],
+        capture_output=True,
+    )
+    logger.info(result.stdout)
+    if result.returncode != 0 or result.stderr:
+        logger.error(f"Call to pg_repack failed with error: {result.stderr}")
+        exit(1)
```
vkrizan, line 43:
```
sys.exit(result.returncode)
```

Alternatively there can be an `exec` call instead of `run` that would take over from Python to pg_repack.

deploy/clowdapp.yml
```diff
@@ -635,9 +635,39 @@ objects:
           requests:
             cpu: ${CPU_REQUEST_SYNCHRONIZER}
             memory: ${MEMORY_REQUEST_SYNCHRONIZER}
+    - name: pg-repack
+      schedule: '@weekly'
+      suspend: ${{PG_REPACK_SUSPEND}}
+      podSpec:
+        image: ${IMAGE}:${IMAGE_TAG}
```
vkrizan, line 642: Here is where I'd put a different image so the repack would not be together with the app, to lower the surface of possibilities (if you know what I mean).

vkrizan (approved):

deploy/clowdapp.yml
kruai: Yep - I opened [RHINENG-8356](https://issues.redhat.com/browse/RHINENG-8356) to tackle that, since it'll require setting up a new image repo in quay, creating all the app-interface config, etc.

run_pg_repack.py
kruai: Fixed!

Dockerfile
vkrizan: This can land up on production before the followup is done, so I'd suggest to do it sooner or before the production deploy. Otherwise the "surface of possibilities" opens up.

kruai: I'm holding off on merging this until our next prod deploy is successful. After we give that time to settle, I'll merge this and we can get started on that ticket as well

kruai:
I've made a separate Dockerfile for this job, and removed the changes to the original one.

build_deploy.sh
```diff
@@ -3,6 +3,7 @@
 set -exv
 
 IMAGE="quay.io/cloudservices/insights-inventory"
+PG_REPACK_IMAGE="quay.io/cloudservices/insights-inventory-pg-repack"
```
vkrizan, line 6: Just a reminder to add it to the internal `app.yml` configuration.

Dockerfile_pg_repack
```diff
@@ -0,0 +1,52 @@
...
+RUN python3 -m pip install --upgrade pip setuptools wheel && \
+    python3 -m pip install pipenv && \
+    python3 -m pip install dumb-init && \
+    pipenv install --system --dev && \
+    pgxn install pg_repack
+
+# allows pre-commit and unit tests to run successfully within the container if image is built in "test" environment
+RUN if [ "$TEST_IMAGE" = "true" ]; then \
+        microdnf install --setopt=tsflags=nodocs -y git && \
+        chgrp -R 0 $APP_ROOT && \
```
vkrizan, line 41: Is the root group necessary (even for a test image)?

(Note, that I've opened a discussion on our internal mailing list regarding this).

Dockerfile_pg_repack
```diff
@@ -0,0 +1,52 @@
...
+ENV APP_ROOT=/opt/app-root/src
+WORKDIR $APP_ROOT
+
+RUN microdnf module enable postgresql:13 python39:3.9 && \
+    microdnf upgrade -y && \
+    microdnf install --setopt=tsflags=nodocs -y postgresql python39 rsync tar procps-ng make snappy gcc postgresql-devel \
+    redhat-rpm-config openssl-devel postgresql-static.x86_64 readline-devel lz4-devel && \
+    rpm -qa | sort > packages-before-devel-install.txt && \
+    microdnf install --setopt=tsflags=nodocs -y libpq-devel python39-devel gcc && \
+    rpm -qa | sort > packages-after-devel-install.txt
```
vkrizan, line 24: Just a note to utilize these files within Github Action that checks for new image base and packages.

Dockerfile_pg_repack
```diff
@@ -0,0 +1,52 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
```
vkrizan, line 1: Regarding file name, consider using `pg_repack.Dockerfile` or `Dockerfile.pg_repack`. I'd choose the one that supports automatic syntax highlighting in IDEs.

vkrizan (approved):
Thank you Asa!

vkrizan:
I think more teams could benefit from using this image with pg_repack. AFAIK Vulnerability is also using pg_repack.
cc @chambridge

chambridge:
> I think more teams could benefit from using this image with pg_repack. AFAIK Vulnerability is also using pg_repack. cc @chambridge

If found to have a profound impact it might be worth reaching out to the platform team and presenting around this on the community townhall or pulling a slide deck together to share this information out.

build_deploy.sh
kruai: Thanks for the reminder - [already did it yesterday](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/101814) :smile:

Dockerfile_pg_repack
kruai: I'm not sure! I was following the setup we do in our main Dockerfile. I know it needs some amount of elevated permissions, but I don't know if there's another way to go about it.
(also, which mailing list? It doesn't look like I received the email)

Dockerfile_pg_repack
kruai: Done - renamed it to pg_repack.dockerfile (lowercased for consistency with `dev.dockerfile`)

chambridge (approved):

PR #1633, opened 2024-02-28
[RHINENG-6095] - Fix registered_with and staleness filters

jpramos123:
# Overview

It fixes a reported issue that was found in production where the hosts were being marked as stale and it was not able to query hosts using `registered_with` and `staleness` filters together.

Also, it standardize the Tags and System profile graphQL query filters to match the same structure of Hosts query.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/filtering/filtering.py
+++ b/api/filtering/filtering.py
@@ -472,13 +472,14 @@ def query_filters(
     if tags:
         tag_filters = build_tag_query_dict_tuple(tags)
         query_filters += ({"OR": tag_filters},)
+
     if staleness:
-        if not registered_with:
-            staleness_filters = tuple(staleness_filter(staleness))
-            query_filters += ({"OR": staleness_filters},)
+        staleness_filters = tuple(staleness_filter(staleness))
+        query_filters += ({"OR": staleness_filters},)
 
     if registered_with:
         query_filters += build_registered_with_filter(registered_with)
+
     if provider_type:
         query_filters += ({"provider_type": {"eq": provider_type.casefold()}},)
     if provider_id:
--- a/api/system_profile.py
+++ b/api/system_profile.py
@@ -37,12 +37,14 @@ logger = get_logger(__name__)
 
 SAP_SYSTEM_QUERY = """
     query hostSystemProfile (
-        $hostFilter: HostFilter
+        $hostFilter: [HostFilter!],
         $limit: Int
         $offset: Int
     ) {
         hostSystemProfile (
-            hostFilter: $hostFilter
+            hostFilter: {
+                AND: $hostFilter,
+            },
         ) {
             sap_system (
                 limit: $limit
@@ -63,13 +65,15 @@ SAP_SYSTEM_QUERY = """
 
 SAP_SIDS_QUERY = """
     query hostSystemProfile (
-        $hostFilter: HostFilter
+        $hostFilter: [HostFilter!],
         $filter: SapSidFilter
         $limit: Int
         $offset: Int
     ) {
         hostSystemProfile (
-            hostFilter: $hostFilter
+            hostFilter: {
+                AND: $hostFilter,
+            },
         ) {
             sap_sids (
                 filter: $filter
@@ -141,18 +145,17 @@ def get_sap_system(
         return flask_json_response(build_collection_response(results, page, per_page, total))
 
     variables = {
-        "hostFilter": {
-            # we're not indexing null timestamps in ES
-            "OR": list(staleness_filter(staleness))
-        },
+        "hostFilter": {},
         "limit": limit,
         "offset": offset,
     }
 
-    hostfilter_and_variables = query_filters(tags=tags, registered_with=registered_with, filter=filter)
+    hostfilter_and_variables = query_filters(
+        tags=tags, registered_with=registered_with, filter=filter, staleness=staleness
+    )
 
     if hostfilter_and_variables != ():
-        variables["hostFilter"]["AND"] = hostfilter_and_variables
+        variables["hostFilter"] = hostfilter_and_variables
 
     response = graphql_query(SAP_SYSTEM_QUERY, variables, log_get_sap_system_failed)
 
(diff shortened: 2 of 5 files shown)


api/tag.py
```diff
@@ -134,7 +131,7 @@ def get_tags(
             group_name,
             None,
             tags,
-            None,  # Staleness to not use the default values
+            staleness,  # Staleness to not use the default values
```
thearifismail, line 134: what does this comment mean?  What is lost when this comment is not provided?

thearifismail (commented):
@jpramos123 the PR looks good as programmed but I have questions that will be easier to ask over Google Meet than phrasing the questions properly here.

api/tag.py
jpramos123: It is an old comment, I will remove it

thearifismail (approved):
LGTM

api/system_profile.py
```diff
@@ -63,13 +65,15 @@
...
+        $hostFilter: [HostFilter!],
         $filter: SapSidFilter
         $limit: Int
         $offset: Int
     ) {
         hostSystemProfile (
-            hostFilter: $hostFilter
+            hostFilter: {
+            AND: $hostFilter,
+        },
```
kruai, line 76:
```suggestion
            hostFilter: {
                AND: $hostFilter,
        },
```

api/system_profile.py
```diff
@@ -37,12 +37,14 @@
...
-        $hostFilter: HostFilter
+        $hostFilter: [HostFilter!],
         $limit: Int
         $offset: Int
     ) {
         hostSystemProfile (
-            hostFilter: $hostFilter
+                    hostFilter: {
+            AND: $hostFilter,
+        },
```
kruai, line 47:
```suggestion
            hostFilter: {
                AND: $hostFilter,
            },
```

api/tag.py
```diff
@@ -16,25 +16,25 @@
...
         $order_by: HOST_TAGS_ORDER_BY,
         $order_how: ORDER_DIR,
         $limit: Int,
         $offset: Int
     ) {
         hostTags (
-            hostFilter: $hostFilter,
+        hostFilter: {
+            AND: $hostFilter,
+        },
```
kruai, line 37:
```suggestion
            hostFilter: {
                AND: $hostFilter,
            },
```

api/filtering/filtering.py
```diff
@@ -473,12 +473,20 @@ def query_filters(
...
             staleness_filters = tuple(staleness_filter(staleness))
             query_filters += ({"OR": staleness_filters},)
 
-    if registered_with:
-        query_filters += build_registered_with_filter(registered_with)
+    else:
+        if registered_with:
+            registered_with_filters = build_registered_with_filter(registered_with)
+            query_filters += (registered_with_filters,)
+
```
kruai, line 489: I know I bring this up every time, but I'm concerned about the staleness filter to be dependent on `registered_with`. As a user, I'd expect `registered_with` to filter the hosts in the exact same way regardless of whether I'm already using the `staleness` filter (and vice-versa).

Isn't the above logic functionally equivalent to the following?

```
    if staleness:
        staleness_filters = tuple(staleness_filter(staleness))
        query_filters += ({"OR": staleness_filters},)

    if registered_with:
        query_filters += build_registered_with_filter(registered_with)
```

The main difference is that your version has an extra "AND" that contains both filters, but the whole list of filters is ANDed on the top level anyway

kruai (changes requested):

api/filtering/filtering.py
jpramos123: True, just realized that I made a change in the query it self, so it won't be necessary to add another AND to wrap the filter! 

Thanks for this!

I made some tests locally, and looks like I have the same behavior

api/filtering/filtering.py
```diff
@@ -313,7 +313,7 @@ def build_registered_with_filter(registered_with):
             for n_items in prs_item:
                 prs_list.append(n_items)
 
-    return ({"OR": prs_list},)
+    return {"OR": prs_list}
```
kruai, line 316: Is there a reason for this change? Doing it this way makes it more cumbersome to add it to the filter list, i.e. why you had to change `query_filters += build_registered_with_filter(registered_with)` to `query_filters += (build_registered_with_filter(registered_with),)`. Also, all the other filter functions return a tuple rather than a dict

api/system_profile.py
kruai: The indentation is still off here

api/system_profile.py
kruai: The indentation is still off here as well

api/tag.py
kruai: Indentation still needs to be fixed here as well

kruai (changes requested):
Still needs a few tweaks but it's almost there. And I just wanna shout out your work on the SAP stuff - great job standardizing that even more :smile:

api/filtering/filtering.py
jpramos123: Actually, there is no reason for that.

I probably tested it with some code that was breaking it (I remember It was failing to concatenate something).

Just adjusted it back

kruai (approved):
LGTM assuming the build passes, and after @fstavela gives some kinda general thumbs up!

fstavela (approved):
I ran the tests that failed in the PR check again and they passed, so I wouldn't worry about that. I also ran some tests created by @msager27 for `registered_with` filter and a combination of that with `staleness`, and it seems to work fine. I would still like to let @msager27 test it more before we release it to Prod, but I didn't find any obvious issue, so I think we can safely merge this and then do additional testing in Stage.

FabriciaDinizRH (approved):

PR #1640, opened 2024-03-06
fix(RHINENG-8730): Change max values of time_to_delete to 2 years

jpramos123:
# Overview

This PR is being created to address [RHINENG-8730](https://issues.redhat.com/browse/RHINENG-8730).
Change the max value of `time_to_delete` to 2 years. It fixes an issue where xjoin-search was not accepting values larger than 54 years.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -904,12 +904,22 @@ class InputAssignmentRule(MarshmallowSchema):
 
 
 class StalenessSchema(MarshmallowSchema):
-    conventional_time_to_stale = fields.Integer(validate=marshmallow_validate.Range(min=1, max=2147483647))
-    conventional_time_to_stale_warning = fields.Integer(validate=marshmallow_validate.Range(min=1, max=2147483647))
-    conventional_time_to_delete = fields.Integer(validate=marshmallow_validate.Range(min=1, max=2147483647))
-    immutable_time_to_stale = fields.Integer(validate=marshmallow_validate.Range(min=1, max=2147483647))
-    immutable_time_to_stale_warning = fields.Integer(validate=marshmallow_validate.Range(min=1, max=2147483647))
-    immutable_time_to_delete = fields.Integer(validate=marshmallow_validate.Range(min=1, max=2147483647))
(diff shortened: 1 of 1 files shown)


kruai (approved):
I'm fine with this change. Is there a test for this schema, though? We should probably have a unit test that makes sure that the staleness values get validated if not

jpramos123:
> I'm fine with this change. Is there a test for this schema, though? We should probably have a unit test that makes sure that the staleness values get validated if not

Yes, we do! We test the CRUD for the staleness

app/models.py
```diff
@@ -904,12 +904,22 @@ def __init__(self, *args, **kwargs):
...
+    conventional_time_to_stale = fields.Integer(
+        validate=marshmallow_validate.Range(min=1, max=604800)
+    )  # Max of 7 days
+    conventional_time_to_stale_warning = fields.Integer(
+        validate=marshmallow_validate.Range(min=1, max=15552000)
+    )  # Max of 180 days
+    conventional_time_to_delete = fields.Integer(
+        validate=marshmallow_validate.Range(min=1, max=63072000)
+    )  # Max of 2 years
+    immutable_time_to_stale = fields.Integer(validate=marshmallow_validate.Range(min=1, max=604800))  # Max of 7 days
```
fstavela, line 916: I don't think we should limit the `time_to_stale` to 7 days, that seems way too low to me, especially for edge systems. I'm quite surprised that UI currently doesn't allow us to set it to more than 7 days TBH. In case that changes, we would need to change the backend too, so I think limiting all of these fields to the same 2 years value would be OK.

PR #1653, opened 2024-03-15
Update System Profile schema

kruai:
# Overview

This PR is being created to keep the System Profile schema in sync with the source of truth. Updated with this PR's changes: https://github.com/RedHatInsights/inventory-schemas/pull/127


kruai:
This is urgent, and is extremely unlikely that it will break anything, so I'm going to merge it.

PR #1656, opened 2024-03-19
feat(RHINENG-7070): System Profile postgres filter

kruai:
# Overview

This PR is being created to address [RHINENG-7070](https://issues.redhat.com/browse/RHINENG-7070) (and affects some other tickets too).
It implements the postgresql version of the System Profile filter. It allows for filtering using the same deep object notation used in the xjoin version, but uses a simpler implementation. It has generic support for system profile fields, and also the custom logic of the operating_system filter.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


api/filtering/db_custom_filters.py
```diff
@@ -0,0 +1,201 @@
...
+    os_filter_list = []
+    separated_filters = separate_operating_system_filters(filter_param["operating_system"])
+
+    for os_comparison in separated_filters:
+        comparator = POSTGRES_COMPARATOR_LOOKUP.get(os_comparison.comparator)
+        comparator_no_eq = comparator[:1]
+
+        if os_comparison.comparator == "eq":
+            os_filter_text = (
+                f"(system_profile_facts->'operating_system'->>'name' = '{os_comparison.name}' AND "
```
chambridge, line 104: Since you are building SQL here on the fly, I feel like perhaps we should do some kind of sanity check that  `os_comparison.name` is a member of some valid inputs. Just don't want an opportunity for SQL injection via filter parameter, if that makes sense.  So perhaps have some type checking as part of the OS comparison class. 

There may be some other opportunities where we need to guard against SQL injection opportunities in the "non-operating_system" filter flow. Perhaps some validation that there isn't some of the key words used in the value `SELECT`, `DELETE`, etc. Unless I'm reading this wrong.

chambridge (dismissed):
LGTM. Nice work with the complex filter here. :tada:

api/filtering/db_custom_filters.py
```diff
@@ -0,0 +1,237 @@
...
+# For instance, {"foo": {"bar": "val1", "baz": "val2"}} becomes:
+# [{"foo": {"bar": "val1"}}, {"foo": {"baz": "val2"}}]
+#
+# When multiple values are provided for one filter, the paths are grouped in a list,
+# making them easier to logically group.
+# For instance, [{"foo": {"bar": "val1"}, "baz": ["val2", "val3"]}] becomes:
+# [{"foo": {"bar": "val1"}}, [{"foo": {"baz": "val2"}}, {"foo": {"baz": "val3"}}]]
+def _unique_paths(
+    node: dict,
+    ignore_nodes: list = [],
```
thearifismail, line 158: Use the word "ignore" implies that the node is not to be used for filtering, but instead recurse on the node.   How about calling it "not_leaf_nodes"?

api/filtering/filtering_common.py
```diff
@@ -39,6 +39,27 @@ class OPERATION_SETS(Enum):
     "integer": OPERATION_SETS.eq.value[0],
 }
 
+POSTGRES_COMPARATOR_LOOKUP = {
```
thearifismail, line 42: Need a comment stating the values of these keys are standard SQL operators. Since bang(!) is often used for NOT, I thought the value of "neq" should have been "!=".

api/filtering/filtering_common.py
```diff
@@ -39,6 +39,27 @@ class OPERATION_SETS(Enum):
...
+    "lt": "<",
+    "lte": "<=",
+    "gt": ">",
+    "gte": ">=",
+    "eq": "=",
+    "neq": "<>",
+    "is": "IS",
+}
+
+POSTGRES_DEFAULT_COMPARATOR = {
```
thearifismail, line 52: In Postgres SQL, `ILIKE` is ignore_case comparison but it is being used for "wildcard".  A comment will be helpful to no-so-familiar to SQL code readers.

thearifismail (commented):
Great work @kruai on figuring out the complex logic and covering all bases for testing.  Got minor comments for helping future readers of the code.

api/filtering/db_custom_filters.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def _check_field_in_spec(spec, field_name):
+    if field_name not in spec.keys():
+        raise ValidationException(f"invalid filter field: {field_name}")
+
+
+# Takes a filter dict and converts it into:
+#   jsonb_path: The jsonb path, i.e. system_profile_facts->sap->>sap_system
+#   pg_op: The comparison to use (e.g. =, >, <)
```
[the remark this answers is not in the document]
kruai: I didn't include all the comparators in this list; I just wanted to give some examples for each output the function returns.

api/filtering/db_custom_filters.py
kruai: `not_leaf_nodes` doesn't make sense for this either. The function recurses through all nodes except ones in `ignore_nodes`; if the node is in `ignore_nodes`, it skips recursion and the whole node is added to the output.

api/filtering/filtering_common.py
kruai: Doesn't the name `POSTGRES_COMPARATOR_LOOKUP` imply that? I kinda thought the naming made it clear. Would it be worth renaming it to `SQL_COMPARATOR_LOOKUP` or `SQL_COMPARISON_OPERATOR_LOOKUP`?

thearifismail: The name did communicate to some extent but upon the first glance, I thought `!=` would have been better.  Then I realized that `<>` might have some meaning in SQL, which turned out to be the case.

api/filtering/filtering_common.py
kruai: `ILIKE` is used for case-insensitive pattern matching (i.e. wildcard support, basically). I could add an inline comment that says `# ILIKE is the SQL operator that provides wildcard support`? But I also feel like that's implied by the dict. Correct me if I'm wrong though

thearifismail: if this is the only thing then that's okay because Chris already approved it and it may require re-approval.

thearifismail (approved):

thearifismail (dismissed):

api/filtering/db_custom_filters.py
kruai: I could call it something like `skip_recursion_for_nodes`? Bit more of a mouthful but more precise

kruai: I opted to keep the naming, but added a comment about its usage to the relevant code block.

api/filtering/filtering_common.py
kruai: I added a comment further clarifying the values in the dict.

api/filtering/filtering_common.py
kruai: I added a comment further clarifying the values in the dict.

thearifismail (approved):
LGTM!

PR #1704, opened 2024-04-15
fix(RHINENG-6139) Use pg_repack 1.4.6 specifically

kruai:
# Overview

This PR is being created so that the pg_repack job uses pg_repack v 1.4.6 specifically. This is required because the job will not run unless the version installed on the DB exactly matches the version installed on the client.


kruai:
I'm going to admin-merge this, since it's a tiny change and it can only be verified in the Stage environment.

PR #1714, opened 2024-04-17
Specify pg_repack tables again

kruai:
# Overview

This PR is being created to update pg_repack by telling it which specific tables to repack. This is to avoid the issue that happens sometimes, where it tries to repack a different table and fails.


kruai:
This only affects the pg_repack CronJob, so I'm going to merge.

PR #1715, opened 2024-04-17
Change minReplicas to replicas

kruai:
# Overview

This PR is being created to rename `minReplicas` to `replicas` in our deployment config. Currently, reducing minReplicas has no effect; this appears to be because the actual field we need to modify is `replicas`.


thearifismail (approved):
LGTM

chambridge (approved):

PR #1800, opened 2024-06-20
fix(RHINENG-10816) Handle nil/not_nil values for operating_system filter

kruai:
# Overview

This PR is being created to address [RHINENG-10816](https://issues.redhat.com/browse/RHINENG-10816).
It enables the use of "nil" and "not_nil" values on the operating_system system profile filter. For instance, these work:

- `?filter[system_profile][operating_system]=nil`
- `?filter[system_profile][operating_system][]=nil`
- `?filter[system_profile][operating_system][]=nil&filter[system_profile][operating_system][]=not_nil`

Note that you cannot combine a nil/not_nil filter with a filter on a specific OS version due to a technical limitation (the deep-object query param structure). So queries like this will not work:

- `?filter[system_profile][operating_system][]=nil&filter[system_profile][operating_system][RHEL][version][gt][]=8`

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/filtering/db_custom_filters.py
+++ b/api/filtering/db_custom_filters.py
@@ -15,7 +15,7 @@ logger = get_logger(__name__)
 
 # Utility class to facilitate OS filter comparison
 class OsComparison:
-    def __init__(self, name, comparator, major, minor):
+    def __init__(self, name="", comparator="", major=0, minor=0):
         self.name = name
         self.comparator = comparator
         self.major = major
@@ -99,6 +99,15 @@ def _get_valid_os_names() -> list:
 # Has a similar purpose to _unique_paths, but the OS filter works a bit differently.
 def separate_operating_system_filters(filter_param: dict) -> list[OsComparison]:
     os_filter_list = []
+
+    # Handle filter_param if a list is passed in
+    if isinstance(filter_param, list):
+        return [OsComparison(comparator=param) for param in filter_param]
+
+    # Handle filter_param if a str is passed in
+    if isinstance(filter_param, str):
+        return [OsComparison(comparator=filter_param)]
+
     for os_name in filter_param.keys():
         if os_name not in (os_names := _get_valid_os_names()):
             raise ValidationException(f"operating_system filter only supports these OS names: {os_names}.")
@@ -160,7 +169,9 @@ def build_operating_system_filter(filter_param: dict) -> tuple:
         comparator = POSTGRES_COMPARATOR_LOOKUP.get(os_comparison.comparator)
         comparator_no_eq = comparator[:1]
 
-        if os_comparison.comparator == "eq":
+        if os_comparison.comparator in ["nil", "not_nil"]:
+            os_filter_list.append(f"(system_profile_facts->>'operating_system' {comparator})")
+        elif os_comparison.comparator == "eq":
             os_filter_text = (
                 f"(system_profile_facts->'operating_system'->>'name' = '{os_comparison.name}' AND "
                 f"(system_profile_facts->'operating_system'->>'major')::int = {os_comparison.major} AND "
--- a/api/filtering/db_custom_filters.py
+++ b/api/filtering/db_custom_filters.py
@@ -106,7 +106,7 @@ def separate_operating_system_filters(filter_param) -> list[OsComparison]:
         return [OsComparison(comparator=param) for param in filter_param]
 
     # Handle filter_param if a str is passed in
-    if isinstance(filter_param, str):
+    elif isinstance(filter_param, str):
         return [OsComparison(comparator=filter_param)]
 
     # filter_param is a dict
--- a/api/filtering/db_custom_filters.py
+++ b/api/filtering/db_custom_filters.py
@@ -14,6 +14,7 @@ logger = get_logger(__name__)
 
 
 # Utility class to facilitate OS filter comparison
+# The list of comparators can be seen in POSTGRES_COMPARATOR_LOOKUP
 class OsComparison:
     def __init__(self, name="", comparator="", major=0, minor=0):
         self.name = name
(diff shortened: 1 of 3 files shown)


jpramos123 (approved):

api/filtering/db_custom_filters.py
```diff
@@ -15,7 +15,7 @@
 
 # Utility class to facilitate OS filter comparison
 class OsComparison:
-    def __init__(self, name, comparator, major, minor):
+    def __init__(self, name="", comparator="", major=0, minor=0):
```
thearifismail, line 18: All comparison operators should be listed in a comment block before the class name

thearifismail: "os_name" for the input parameter "name" sounds more natural and easy to understand, like done in the code below.

api/filtering/db_custom_filters.py
```diff
@@ -99,6 +99,15 @@ def _get_valid_os_names() -> list:
 # Has a similar purpose to _unique_paths, but the OS filter works a bit differently.
 def separate_operating_system_filters(filter_param: dict) -> list[OsComparison]:
     os_filter_list = []
+
+    # Handle filter_param if a list is passed in
+    if isinstance(filter_param, list):
```
thearifismail, line 104: Why filter_param is being check check for type `list` or `str`, when in the input parameter type is `dict`?

api/filtering/db_custom_filters.py
```diff
@@ -160,7 +169,9 @@ def build_operating_system_filter(filter_param: dict) -> tuple:
         comparator = POSTGRES_COMPARATOR_LOOKUP.get(os_comparison.comparator)
         comparator_no_eq = comparator[:1]
 
-        if os_comparison.comparator == "eq":
+        if os_comparison.comparator in ["nil", "not_nil"]:
+            os_filter_list.append(f"(system_profile_facts->>'operating_system' {comparator})")
```
thearifismail, line 173: Is the `append` statement correct?  Should there be `=` or `is` between 'operating_system' and '{comparator}'?

tests/test_api_hosts_get.py
```diff
@@ -1564,6 +1564,86 @@ def test_query_all_sp_filters_operating_system(db_create_host, api_get, sp_filte
...
+            [
+                {
+                    "operating_system": {
+                        "name": "RHEL",
+                        "major": "8",
+                        "minor": "1",
+                    },
+                }
+            ],
+            ["[operating_system][]=nil", "[operating_system]=nil"],
```
thearifismail, line 1581: For easy of reading, one param per line should be used.

tests/test_api_hosts_get.py
```diff
@@ -1564,6 +1564,86 @@ def test_query_all_sp_filters_operating_system(db_create_host, api_get, sp_filte
...
+                {
+                    "operating_system": {
+                        "name": "RHEL",
+                        "major": "8",
+                        "minor": "1",
+                    },
+                }
+            ],
+            [None],
+            ["[operating_system][]=not_nil", "[operating_system]=not_nil"],
```
thearifismail, line 1594: Same as above

tests/test_api_hosts_get.py
thearifismail: Should there parameters using  "major": "" and/or "minor": ""?

api/filtering/db_custom_filters.py
kruai: Wouldn't it be redundant to call it `OsComparison.os_name`? All the properties of OsComparison are OS-related

api/filtering/db_custom_filters.py
kruai: Good point - we generally expect a dict, but it can be passed in as a list or a str, depending on the value of the query param. I can remove or update the type hint accordingly

api/filtering/db_custom_filters.py
kruai: `append` is correct; it's adding a value to the list. And no, there should not be an `=` or `is`, because it's already in the comparator (see the change to `filtering_common.py`)

tests/test_api_hosts_get.py
kruai: OK, I can space it out further.
No, there should not be parameters using major/minor; the PR's description has an explanation for why.

jpramos123 (dismissed):

api/filtering/db_custom_filters.py
thearifismail: Yes, I was reluctant in making this comment upon seeing the class name but then seeing it used that down below in the code made reading os_name much easier.  This is fine as-is now.

api/filtering/db_custom_filters.py
thearifismail: yes but then checking for list and string should be done in `if else` blocks and NOT both separately and independent of each other

api/filtering/db_custom_filters.py
thearifismail: Got it!

thearifismail (commented):
LGTM, except one minor comment about checking parameter type

api/filtering/db_custom_filters.py
kruai: They're functionally the same here because each conditional block returns. I can change to elif though if you'd prefer

thearifismail (approved):
LGTM

PR #1816, opened 2024-07-03
fix(RHINENG-10977) Handle MQ exceptions more gracefully

kruai:
# Overview

This PR is being created to address [RHINENG-10977](https://issues.redhat.com/browse/RHINENG-10977).
There are some scenarios that that are not handled gracefully in our MQ service, which can lead to the pods crashing/restarting. This PR handles these:

- **InventoryException when SP pod cannot find the host described in the incoming message**: this was actually already handled, but it appeared to be crashing the pod because it was the last exception logged before the pod stopped. I added a test case just to make sure this was covered.
- **Malformed SQL when incoming host has no canonical facts**: this was only able to happen on the SP pods because it doesn't require canonical facts. However, when the host could not be found via its host ID and it also contained no canonical facts, it would generate malformed SQL for the canonical facts filter. I've now added a check to handle this (admittedly rare) scenario.
- **StaleDataError when attempting to commit a batch of hosts**: I don't think we can avoid this error while we're using the batch approach for MQ. However, we can handle it so that only the current batch gets rolled back, and the pod does not crash. I also made it so the failure count is incremented by the amount of messages that were in the batch, so the metric should stay reliable.

Here's an example of the malformed SQL it would attempt to generate (for the second issue):
```
[SQL: SELECT hosts.system_profile_facts -> %(param_1)s AS anon_1, hosts.id AS hosts_id, hosts.account AS hosts_account, hosts.org_id AS hosts_org_id, hosts.display_name AS hosts_display_name, hosts.ansible_host AS hosts_ansible_host, hosts.created_on AS hosts_created_on, hosts.modified_on AS hosts_modified_on, hosts.facts AS hosts_facts, hosts.tags AS hosts_tags, hosts.canonical_facts AS hosts_canonical_facts, hosts.system_profile_facts AS hosts_system_profile_facts, hosts.groups AS hosts_groups, hosts.stale_timestamp AS hosts_stale_timestamp, hosts.reporter AS hosts_reporter, hosts.per_reporter_staleness AS hosts_per_reporter_staleness FROM hosts WHERE hosts.org_id = %(org_id_1)s AND NOT  AND (hosts.modified_on > %(modified_on_1)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_1)s) AS VARCHAR) = %(param_2)s OR hosts.modified_on > %(modified_on_2)s AND hosts.modified_on <= %(modified_on_3)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_2)s) AS VARCHAR) = %(param_3)s OR hosts.modified_on > %(modified_on_4)s AND hosts.modified_on <= %(modified_on_5)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_3)s) AS VARCHAR) = %(param_4)s OR hosts.modified_on > %(modified_on_6)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_4)s) AS VARCHAR) IS NULL OR hosts.modified_on > %(modified_on_7)s AND hosts.modified_on <= %(modified_on_8)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_5)s) AS VARCHAR) IS NULL OR hosts.modified_on > %(modified_on_9)s AND hosts.modified_on <= %(modified_on_10)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_6)s) AS VARCHAR) IS NULL) ORDER BY hosts.modified_on DESC  LIMIT %(param_5)s]
```
Note the `AND NOT  AND` near the beginning of the `WHERE` clause; this happened because it was adding the "no unmatching canonical facts" filter, but its canonical facts were empty. After the fix, the filter looks like this:

```
[SQL: SELECT hosts.system_profile_facts -> %(param_1)s AS anon_1, hosts.id AS hosts_id, hosts.account AS hosts_account, hosts.org_id AS hosts_org_id, hosts.display_name AS hosts_display_name, hosts.ansible_host AS hosts_ansible_host, hosts.created_on AS hosts_created_on, hosts.modified_on AS hosts_modified_on, hosts.facts AS hosts_facts, hosts.tags AS hosts_tags, hosts.canonical_facts AS hosts_canonical_facts, hosts.system_profile_facts AS hosts_system_profile_facts, hosts.groups AS hosts_groups, hosts.stale_timestamp AS hosts_stale_timestamp, hosts.reporter AS hosts_reporter, hosts.per_reporter_staleness AS hosts_per_reporter_staleness FROM hosts WHERE hosts.org_id = %(org_id_1)s AND (hosts.modified_on > %(modified_on_1)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_1)s) AS VARCHAR) = %(param_2)s OR hosts.modified_on > %(modified_on_2)s AND hosts.modified_on <= %(modified_on_3)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_2)s) AS VARCHAR) = %(param_3)s OR hosts.modified_on > %(modified_on_4)s AND hosts.modified_on <= %(modified_on_5)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_3)s) AS VARCHAR) = %(param_4)s OR hosts.modified_on > %(modified_on_6)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_4)s) AS VARCHAR) IS NULL OR hosts.modified_on > %(modified_on_7)s AND hosts.modified_on <= %(modified_on_8)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_5)s) AS VARCHAR) IS NULL OR hosts.modified_on > %(modified_on_9)s AND hosts.modified_on <= %(modified_on_10)s AND CAST((hosts.system_profile_facts ->> %(system_profile_facts_6)s) AS VARCHAR) IS NULL) ORDER BY hosts.modified_on DESC  LIMIT %(param_5)s]
```

## PR Checklist

- [x] Include raw query examples in the PR description, if adding/modifying SQL query
- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


lib/host_repository.py
```diff
@@ -164,6 +164,10 @@ def find_host_by_multiple_canonical_facts(identity, canonical_facts):
     """
     logger.debug("find_host_by_multiple_canonical_facts(%s)", canonical_facts)
 
+    # If no canonical facts are supplied, it can't match anything.
+    if not canonical_facts:
+        return None
```
vkrizan, line 169: Are these changes related? Looks like out of context.

kruai: This change addresses the second bullet point from the description. The malformed SQL was being generated due to it trying to filter out hosts that don't match the provided canonical facts, when no canonical facts are provided. This can't happen in most scenarios because we require canonical facts for our main MQ and API functionality, but the MQ-SP pods don't require them because they expect to receive a host ID (and they use canonical facts as a backup).

vkrizan: Thanks for explanation! So this fixes a bug then. Please prefer to separate concerns.

kruai: Just to clarify, you mean we should create separate Jiras and PRs, right? If so, thanks for the reminder - I was investigating them all at once and forgot to separate them

PR #1834, opened 2024-07-16
chore(RHINENG-11401): Refactor RBAC

kruai:
# Overview

This PR is being created to address [RHINENG-11401](https://issues.redhat.com/browse/RHINENG-11401).
It refactors the RBAC implementation, extracts logical blocks of the code into their own functions, and simplifies its usage within export_service.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/lib/middleware.py
+++ b/lib/middleware.py
@@ -1,6 +1,7 @@
 from functools import partial
 from functools import wraps
 from http import HTTPStatus
+from typing import Tuple
 from uuid import UUID
 
 from app_common_python import LoadedConfig
@@ -18,7 +19,7 @@ from app import RbacPermission
 from app import RbacResourceType
 from app import REQUEST_ID_HEADER
 from app.auth import get_current_identity
-from app.auth.identity import create_mock_identity_with_org_id
+from app.auth.identity import Identity
 from app.auth.identity import IdentityType
 from app.common import inventory_config
 from app.instrumentation import rbac_failure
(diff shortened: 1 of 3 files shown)


lib/middleware.py
```diff
@@ -75,96 +70,139 @@ def get_rbac_permissions(app, request_header=None):
...
+    rbac_request_headers: dict,
+    permission_base: str = "inventory",
+) -> Tuple[str, dict]:
+    # Returns a 2-item tuple:
+    # 1. Whether or not the request should be allowed at all
+    # 2. The filter that should be applied to the data (if allowed)
+    if inventory_config().bypass_rbac:
+        return True, None
+
+    current_identity = get_current_identity()
```
jpramos123, line 137: I'm afraid it will fail in the export-service. As we are not receiving an API call, the application does not create a new context, so the `connexion.context.context["user"]` attribute will not exist. The app runs once, in a infinite loop, and it leaves the normal API pipeline that Flask creates when receiving API calls

That is the reason we were using `current_identity = create_mock_identity_with_org_id(org_id)` to create the identity, because we needed to mock it every time we received an message from Kafka.

I would suggest validation if the `identity` exists, if not get it using `get_current_identity()`

```suggestion
   if not identity:
       current_identity = get_current_identity()
```

jpramos123 (changes requested):
Just one comment

lib/middleware.py
kruai: My mistake; this was leftover from the original implementation. This line doesn't need to be here. When coming from `rbac()`, the identity is already acquired from the connexion context. When coming from the export service, it's already set to the mock identity.

jpramos123 (approved):

thearifismail (approved):

PR #1825, opened 2024-07-08
refactor(RHINENG-11207): refactor notifications.py

FabriciaDinizRH:
# Overview

This PR is being created to address [RHINENG-11207](https://issues.redhat.com/browse/RHINENG-11207).
Refactor of `notifications.py` to remove repeated code

## PR Checklist

- [ ] Keep PR title short, ideally under 72 characters
- [ ] Descriptive comments provided in complex code blocks
- [ ] Include raw query examples in the PR description, if adding/modifying SQL query
- [ ] Tests: validate optimal/expected output
- [ ] Tests: validate exceptions and failure scenarios
- [ ] Tests: edge cases
- [ ] Recovers or fails gracefully during potential resource outages (e.g. DB, Kafka)
- [ ] Uses [type hinting](https://docs.python.org/3/library/typing.html), if convenient
- [ ] Documentation, if this PR changes the way other services interact with host inventory
- [ ] Links to related PRs

## Secure Coding Practices Documentation Reference

You can find documentation on this checklist [here](https://github.com/RedHatInsights/secure-coding-checklist).

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Authentication and Password Management
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Error Handling and Logging
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [ ] General Coding Practices

PR #1844, opened 2024-07-22
feat(RHINENG-7840): add logic to newly registered system notifications

FabriciaDinizRH:
# Overview

This PR is being created to address [RHINENG-7840](https://issues.redhat.com/browse/RHINENG-7840).

- add the logic necessary to send the notifications for newly registered hosts.
- add a few changes to `notifications.py` that will be overwritten by the [refactoring PR](https://github.com/RedHatInsights/insights-host-inventory/pull/1825)

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/queue/notifications.py
+++ b/app/queue/notifications.py
@@ -11,6 +11,7 @@ from app.common import inventory_config
 from app.logging import threadctx
 from app.queue.events import hostname
 from app.queue.metrics import notification_serialization_time
+from app.serialization import _deserialize_tags
 from app.serialization import build_rhel_version_str
 from app.serialization import deserialize_canonical_facts
 
@@ -34,7 +35,7 @@ class BaseContextSchema(MarshmallowSchema):
     hostname = fields.Str(required=True)
     display_name = fields.Str(required=True)
     rhel_version = fields.Str(required=True)
-    tags = fields.Raw()
+    tags = fields.Dict()
 
 
 class BaseEventListSchema(MarshmallowSchema):
@@ -210,7 +211,7 @@ def build_base_notification_obj(notification_type, host):
             "hostname": canonical_facts.get("fqdn", ""),
             "display_name": host.get("display_name"),
             "rhel_version": build_rhel_version_str(system_profile),
-            "tags": host.get("tags"),
+            "tags": _deserialize_tags(host.get("tags")),
         },
         "events": [],
     }
--- a/app/queue/notifications.py
+++ b/app/queue/notifications.py
@@ -228,7 +228,7 @@ def populate_events(base_notification_obj, host_list, extra_fields=[]):
                 "insights_id": canonical_facts.get("insights_id", ""),
                 "subscription_manager_id": canonical_facts.get("subscription_manager_id", ""),
                 "satellite_id": canonical_facts.get("satellite_id", ""),
-                "groups": [{"id": group.get("id"), "name": group.get("name")} for group in host.get("groups")],
+                "groups": [{"id": group.get("id"), "name": group.get("name")} for group in host.get("groups", [])],
             },
         }
 
--- a/app/queue/queue.py
+++ b/app/queue/queue.py
@@ -351,6 +351,13 @@ def handle_message(message, notification_event_producer, message_operation=add_h
             )
             staleness_timestamps = Timestamps.from_config(inventory_config())
             event_type = operation_results_to_event_type(operation_result)
+
+            if message_operation == add_host:
+                send_notification(
+                    notification_event_producer,
+                    notification_type=NotificationType.new_system_registered,
+                    host=host,
+                )
             return OperationResult(
                 host_row,
                 platform_metadata,
--- a/app/queue/queue.py
+++ b/app/queue/queue.py
@@ -43,6 +43,7 @@ from app.queue import metrics
 from app.queue.event_producer import EventProducer
 from app.queue.events import build_event
 from app.queue.events import EventType
+from app.queue.events import HOST_EVENT_TYPE_CREATED
 from app.queue.events import message_headers
 from app.queue.events import operation_results_to_event_type
 from app.queue.export_service import create_export
@@ -315,12 +316,6 @@ def add_host(host_data, platform_metadata, notification_event_producer, operatio
         host_row, add_result = host_repository.add_host(input_host, identity, operation_args=operation_args)
         success_logger = partial(log_add_update_host_succeeded, logger, add_result)
 
-        send_notification(
-            notification_event_producer,
-            notification_type=NotificationType.new_system_registered,
-            host=host_data,
-        )
-
         return host_row, add_result, identity, success_logger
     except ValidationException:
         metrics.add_host_failure.labels("ValidationException", host_data.get("reporter", "null")).inc()
@@ -568,6 +563,20 @@ def event_loop(consumer, flask_app, event_producer, notification_event_producer,
                     # The above session is automatically committed or rolled back.
                     # Now we need to send out messages for the batch of hosts we just processed.
                     write_message_batch(event_producer, processed_rows)
+
+                    # Find the operation related to 'host_created' and send the notification
+                    for operation_result_obj in processed_rows:
+                        if operation_result_obj and operation_result_obj.event_type.name == HOST_EVENT_TYPE_CREATED:
+                            send_notification(
+                                notification_event_producer,
+                                notification_type=NotificationType.new_system_registered,
+                                host=serialize_host(
+                                    operation_result_obj.host_row,
+                                    staleness_timestamps=operation_result_obj.staleness_timestamps,
+                                    staleness=operation_result_obj.staleness_object,
+                                    omit_null_facts=True,
+                                ),
+                            )
                 except StaleDataError as exc:
                     metrics.ingress_message_handler_failure.inc(amount=len(messages))
                     logger.error(
--- a/app/queue/queue.py
+++ b/app/queue/queue.py
@@ -280,7 +280,7 @@ def sync_event_message(message, session, event_producer):
     return
 
 
-def update_system_profile(host_data, platform_metadata, operation_args={}):
+def update_system_profile(host_data, platform_metadata, notification_event_producer=None, operation_args={}):
     try:
         input_host = deserialize_host(host_data, schema=LimitedHostSchema)
         input_host.id = host_data.get("id")
@@ -303,7 +303,7 @@ def update_system_profile(host_data, platform_metadata, operation_args={}):
         raise
 
 
-def add_host(host_data, platform_metadata, operation_args={}):
+def add_host(host_data, platform_metadata, notification_event_producer, operation_args={}):
     try:
         identity = _get_identity(host_data, platform_metadata)
         # basic-auth does not need owner_id
@@ -314,6 +314,12 @@ def add_host(host_data, platform_metadata, operation_args={}):
         log_add_host_attempt(logger, input_host)
         host_row, add_result = host_repository.add_host(input_host, identity, operation_args=operation_args)
         success_logger = partial(log_add_update_host_succeeded, logger, add_result)
+
+        send_notification(
+            notification_event_producer,
+            notification_type=NotificationType.new_system_registered,
+            host=host_data,
+        )
 
         return host_row, add_result, identity, success_logger
     except ValidationException:
@@ -347,17 +353,11 @@ def handle_message(message, notification_event_producer, message_operation=add_h
         try:
             host = validated_operation_msg["data"]
             host_row, operation_result, identity, success_logger = message_operation(
-                host, platform_metadata, validated_operation_msg.get("operation_args", {})
+                host, platform_metadata, notification_event_producer, validated_operation_msg.get("operation_args", {})
             )
             staleness_timestamps = Timestamps.from_config(inventory_config())
             event_type = operation_results_to_event_type(operation_result)
 
-            if message_operation == add_host:
-                send_notification(
-                    notification_event_producer,
-                    notification_type=NotificationType.new_system_registered,
-                    host=host,
-                )
             return OperationResult(
                 host_row,
                 platform_metadata,
--- a/app/serialization.py
+++ b/app/serialization.py
@@ -144,6 +144,7 @@ def serialize_host(
     staleness=None,
     system_profile_fields=None,
     for_export_svc=False,
+    omit_null_facts=False,
 ):
     # TODO: In future, this must handle groups staleness
 
@@ -170,7 +171,7 @@ def serialize_host(
             host.modified_on, staleness["conventional_time_to_delete"]
         )
 
-    serialized_host = {**serialize_canonical_facts(host.canonical_facts)}
+    serialized_host = {**serialize_canonical_facts(host.canonical_facts, omit_null_facts=omit_null_facts)}
 
     fields = DEFAULT_FIELDS + additional_fields
     if for_mq:
@@ -326,8 +327,11 @@ def _deserialize_all_canonical_facts(data):
     return {field: _recursive_casefold(data[field]) if data.get(field) else None for field in _CANONICAL_FACTS_FIELDS}
 
 
-def serialize_canonical_facts(canonical_facts):
-    return {field: canonical_facts.get(field) for field in _CANONICAL_FACTS_FIELDS}
+def serialize_canonical_facts(canonical_facts, omit_null_facts=False):
+    if omit_null_facts:
+        return {field: canonical_facts.get(field) for field in _CANONICAL_FACTS_FIELDS if field in canonical_facts}
+    else:
+        return {field: canonical_facts.get(field) for field in _CANONICAL_FACTS_FIELDS}
 
 
 def _deserialize_facts(data):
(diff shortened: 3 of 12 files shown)


thearifismail (commented):
@FabriciaDinizRH I am not done but have comment on the code.  One question is when a host does not have an `insights_id`, like this [line](https://github.com/RedHatInsights/insights-host-inventory/blob/d830962a7de6adafc8c04c981242376874c6e7c0/app/queue/notifications.py#L137), does it mean that no notification will be sent?

app/queue/notifications.py
```diff
@@ -129,7 +129,7 @@ class SystemRegisteredContextSchema(MarshmallowSchema):
     display_name = fields.Str(required=True)
     rhel_version = fields.Str(required=True)
     host_url = fields.Str(required=True)
-    tags = fields.Dict()
+    tags = fields.Raw()
```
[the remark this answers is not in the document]
FabriciaDinizRH: Using Dict() gives me the following error in several tests:
`ValueError: dictionary update sequence element #0 has length 3; 2 is required`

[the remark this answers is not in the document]
vkrizan: Can you decode the error for Dict?

app/queue/notifications.py
```diff
@@ -246,8 +246,16 @@ def system_registered_notification(notification_type, host):
     base_notification_obj = build_base_notification_obj(notification_type, host)
 
     host_id = host.get("id")
-    canonical_facts = host.get("canonical_facts")
-    system_profile = host.get("system_profile_facts")
+    # these changes will be overwritten by the cleanup PR (#1825)
+    canonical_facts = host.get("canonical_facts", deserialize_canonical_facts(host))
+    groups = []
+    if "groups" in host.keys():
+        groups = host.get("groups")
```
vkrizan, line 253: Second parameter of `dict.get` is a default value. Can it be used here instead of doing the key check?

app/queue/notifications.py
```diff
@@ -246,8 +246,16 @@ def system_registered_notification(notification_type, host):
...
+    system_profile = {}
+    if "system_profile_facts" in host.keys():
+        system_profile = host.get("system_profile_facts")
+    elif "system_profile" in host.keys():
+        system_profile = host.get("system_profile")
```
vkrizan, line 258: Similar to these two above...

tests/test_mq_service.py
```diff
@@ -316,28 +316,7 @@ def test_handle_message_verify_message_headers(mocker, add_host_result, mq_creat
...
-@pytest.mark.parametrize(
-    "expected_value, system_profile",
-    [
-        ("False", {}),
-        ("False", {"bootc_status": {}}),
-        ("True", {"bootc_status": {"booted": {}}}),
-        ("True", {"bootc_status": {"booted": {"image": "192.168.0.1:5000/foo/foo:latest"}}}),
-    ],
-)
-def test_verify_bootc_in_headers(expected_value, system_profile, mq_create_or_update_host):
```
vkrizan, line 328: Why is this test being removed?

tests/test_notifications.py
```diff
@@ -45,7 +44,7 @@ def test_new_system_notification_fields(mq_create_or_update_host, notification_e
     assert_system_registered_notification_is_valid(notification_event_producer_mock, host)
 
     assert notification["context"]["rhel_version"] == "8.6"
-    assert notification["account_id"] == SYSTEM_IDENTITY["account_number"]
```
vkrizan: Please keep the check for account_id as it still might be used on event consumers.

app/serialization.py
```diff
@@ -516,8 +516,8 @@ def _serialize_per_reporter_staleness(host, staleness, staleness_timestamps):
 def build_rhel_version_str(system_profile: dict) -> str:
     os = system_profile.get("operating_system")
     if os:
-        if os.get("name") == "rhel":
+        if os.get("name").lower() == "rhel":
```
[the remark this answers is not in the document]
FabriciaDinizRH: the default is "". Since the field is rhel version, if the os is not rhel this field would be an empty string.

tests/test_notifications.py
FabriciaDinizRH: Mikel Barcina from the notification team said I could remove it because they just ignore it as well

tests/test_mq_service.py
FabriciaDinizRH: ouch, that's a merge fix that went wrong

app/serialization.py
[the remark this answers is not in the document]
FabriciaDinizRH: Yeah, you're right. I see, `lower()` would raise an error there, I'll fix it

app/queue/notifications.py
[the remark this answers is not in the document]
FabriciaDinizRH: done

app/queue/queue.py
```diff
@@ -346,6 +346,13 @@ def handle_message(message, notification_event_producer, message_operation=add_h
             )
             staleness_timestamps = Timestamps.from_config(inventory_config())
             event_type = operation_results_to_event_type(operation_result)
+
+            if message_operation == add_host:
```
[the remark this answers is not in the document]
FabriciaDinizRH: I didn't want to pass the `notification_event_producer` to the other operations, but since there's just one other, I did it

app/queue/queue.py
```diff
@@ -315,6 +315,12 @@ def add_host(host_data, platform_metadata, operation_args={}):
         host_row, add_result = host_repository.add_host(input_host, identity, operation_args=operation_args)
         success_logger = partial(log_add_update_host_succeeded, logger, add_result)
 
+        send_notification(
+            notification_event_producer,
+            notification_type=NotificationType.new_system_registered,
+            host=host_data,
+        )
+
```
kruai, line 323: The main issue I foresee with this is that the message is sent before the transaction has been committed to the DB. This could lead to situations where the notification is sent, but then the transaction is rolled back (and the host is not created). The host events are produced here upon a successful commit: https://github.com/RedHatInsights/insights-host-inventory/blob/search-tag-dollar-sign/app/queue/queue.py#L546

tests/helpers/mq_utils.py
```diff
@@ -230,6 +230,49 @@ def assert_patch_event_is_valid(
...
+    }
+
+    expected_payload_keys = {
+        "insights_id",
+        "subscription_manager_id",
+        "satellite_id",
+        "groups",
+        "reporter",
+        "system_check_in",
+    }
```
kruai, line 266: Should these be of type `set`, since they're being compared to sets directly below this?

tests/test_mq_service.py
```diff
@@ -134,7 +134,7 @@ def test_handle_message_happy_path(identity, mocker, flask_app):
     assert result.event_type == EventType.created
     assert result.host_row.canonical_facts["insights_id"] == expected_insights_id
 
-    mock_notification_event_producer.write_event.assert_not_called()
+    mock_notification_event_producer.write_event.assert_called_once()
```
kruai, line 137: IIRC, the original intent of this `assert_not_called()` was to assert that there was no notification produced due to validation errors. Can we make a more specific assertion so we know that this is producing a "host added" notification rather than a "validation error" notification?

jpramos123 (approved):
One comment

jpramos123 (changes requested):
Please take a look above question

app/queue/queue.py
```diff
@@ -280,7 +280,7 @@ def sync_event_message(message, session, event_producer):
     return
 
 
-def update_system_profile(host_data, platform_metadata, operation_args={}):
+def update_system_profile(host_data, platform_metadata, notification_event_producer=None, operation_args={}):
```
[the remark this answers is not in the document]
FabriciaDinizRH: It fails because I added the notification producer [here](https://github.com/RedHatInsights/insights-host-inventory/pull/1844/files/8062b2eeebb20d0c7c9db6553d4cc55f844ed37d#diff-08d3eb76a0fcb1dc61b0bfa4ca74d97315116073476ff033029a69ac180cf008R356) and that calls this function as well, so it needs to be there

tests/helpers/mq_utils.py
FabriciaDinizRH: I thought sets containing values could be initialized using curly braces. Is it outdated/unused currently?

app/queue/queue.py
jpramos123: Understood!

Now we need to figure out how to:
1. Know when the message receive is related to add_host
2. A way to of having `host_data` outside of this function

jpramos123 (dismissed):

tests/helpers/mq_utils.py
jpramos123: @FabriciaDinizRH is correct here! There are sets, not dicts

app/queue/queue.py
```diff
@@ -72,8 +73,9 @@ class OperationSchema(Schema):
 
 # Helper class to facilitate batch operations
 class OperationResult:
-    def __init__(self, hr, pm, st, so, et, sl):
+    def __init__(self, hr, pm, st, so, et, sl, sh):
         self.host_row = hr
+        self.serialized_host = sh
```
kruai, line 78: We can make this change without adding a `serialized_host` field to this class (redundant since we have host_row). You can just call the serialization function when it's needed. For example: https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/queue/queue.py#L416

app/serialization.py
```diff
@@ -517,10 +517,10 @@ def _serialize_per_reporter_staleness(host, staleness, staleness_timestamps):
 def build_rhel_version_str(system_profile: dict) -> str:
     os = system_profile.get("operating_system")
     if os:
-        if os.get("name") == "rhel":
+        if os.get("name", "").lower() == "rhel":
             major = os.get("major")
             minor = os.get("minor")
-            return f"{major:03}.{minor:03}"
+            return f"{major}.{minor}"
```
kruai, line 523: This change seems irrelevant to the PR's purpose

tests/helpers/mq_utils.py
kruai: My mistake! Thank you for the information.

tests/test_mq_service.py
kruai: This one is still open ^

kruai (changes requested):

tests/test_mq_service.py
jpramos123: Using `assert_called_once` is not the expected behavior? We generate the notification in line 134 then we use the assertion

tests/test_mq_service.py
```diff
@@ -130,11 +131,12 @@ def test_handle_message_happy_path(identity, mocker, flask_app):
 
     message = wrap_message(host.data(), "add_host", get_platform_metadata(identity))
     result = handle_message(json.dumps(message), mock_notification_event_producer)
+    send_notification(mock_notification_event_producer, NotificationType.new_system_registered, host.data())
```
kruai, line 134: Just noticed this - is this manually sending a notification from the test? If that's the case, then won't `mock_notification_event_producer.write_event` always have been called in the assertion?

thearifismail (approved):
LGTM

tests/test_mq_service.py
jpramos123: Yes it is calling it, otherwise `mock_notification_event_producer.write_event` will never be called, as it is not sending the notification inside the `handle_menssage`/`add_host` function. 

Is not this test expected to handle the happy_path for the host creation?

tests/test_mq_service.py
jpramos123: Added in line 143 the event_type validation.

tests/test_mq_service.py
kruai: It is expected to handle the happy path; I just don't understand how the assertion is useful if the notification is sent manually within the test, and not within the code we're testing. Since the notification is no longer sent within `handle_message`, perhaps this notification should be validated in a different test (one that uses `event_loop()`, for instance)?

tests/test_mq_service.py
```diff
@@ -1796,7 +1826,27 @@ def test_add_host_with_invalid_identity(mocker, event_datetime_mock, mq_create_o
...
+                mac_addresses=[
+                    "02:00:00:{:02x}:{:02x}:{:02x}".format(
+                        random.randint(0, 255), random.randint(0, 255), random.randint(0, 255)
+                    )
+                ],
+            ).data(),
+            "add_host",
+            get_platform_metadata(),
+        )
+    )
```
kruai, line 1849: I'm seeing this kind of change in a number of places. Why is it needed?
Can it be refactored, or does `minimal_host()` need to be updated or something?

tests/test_notifications.py
```diff
@@ -0,0 +1,90 @@
...
+        },
+    )
+
+    mq_create_or_update_host(host, return_all_data=True)
+    send_notification(notification_event_producer_mock, NotificationType.new_system_registered, host.data())
+    notification = json.loads(notification_event_producer_mock.event)
+    assert_system_registered_notification_is_valid(notification_event_producer_mock, host)
+
+    assert notification["context"]["rhel_version"] == "8.6"
+    assert notification["org_id"] == SYSTEM_IDENTITY["org_id"]
```
kruai, line 49: These tests also don't assert that a notification was produced upon host creation, since the notification is sent within the test and not the code. It does validate the notification itself, though

kruai (changes requested):
A few more questions and concerns

tests/test_notifications.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+# this test should pass, RHINENG-11348 was created to fix this behavior
+
+# def test_add_host_fail(mocker):
```
vkrizan, line 54: Commented code. Either remove it or mark the test as xfail.

app/serialization.py
```diff
@@ -517,10 +517,10 @@ def _serialize_per_reporter_staleness(host, staleness, staleness_timestamps):
 def build_rhel_version_str(system_profile: dict) -> str:
     os = system_profile.get("operating_system")
     if os:
-        if os.get("name", "").lower() == "rhel":
+        if os.get("name") == "rhel":
```
vkrizan: Constant `"rhel"`

app/serialization.py
```diff
@@ -517,10 +517,10 @@ def _serialize_per_reporter_staleness(host, staleness, staleness_timestamps):
...
             major = os.get("major")
             minor = os.get("minor")
-            return f"{major}.{minor}"
+            return f"{major:03}.{minor:03}"
```
vkrizan: Why is this formatted with 0 fill. It does not look good: `008.001`.

app/queue/queue.py
```diff
@@ -568,6 +563,20 @@ def event_loop(consumer, flask_app, event_producer, notification_event_producer,
                     # The above session is automatically committed or rolled back.
                     # Now we need to send out messages for the batch of hosts we just processed.
                     write_message_batch(event_producer, processed_rows)
+
+                    # Find the operation related to 'host_created' and send the notification
+                    for operation_result_obj in processed_rows:
+                        if operation_result_obj:
+                            if operation_result_obj.event_type.name == HOST_EVENT_TYPE_CREATED:
+                                send_notification(
```
vkrizan, line 571: Aesthetically this looks slanted. The two `if`s can be merged together. Or alternatively, you can prepare a list (or generator) that would get the necessary objects, and then do the loop.

vkrizan (changes requested):

app/queue/queue.py
jpramos123: Done!

app/serialization.py
```diff
@@ -144,6 +144,7 @@ def serialize_host(
     staleness=None,
     system_profile_fields=None,
     for_export_svc=False,
+    for_create_host_notification=False,
```
kruai, line 147: Might be good to rename this to something more descriptive to its function, like `omit_null_facts` or something similar. Honestly we might wanna restructure this function in a follow-up to make it easier to maintain

app/serialization.py
```diff
@@ -326,8 +329,11 @@ def _deserialize_all_canonical_facts(data):
     return {field: _recursive_casefold(data[field]) if data.get(field) else None for field in _CANONICAL_FACTS_FIELDS}
 
 
-def serialize_canonical_facts(canonical_facts):
-    return {field: canonical_facts.get(field) for field in _CANONICAL_FACTS_FIELDS}
+def serialize_canonical_facts(canonical_facts, for_create_host_notification=False):
```
kruai, line 332: Same here; probably want a clearer name for the `for_create_host_notification` param

kruai (commented):
Looking good! It's really close. I just have a nitpick about the naming of a param.

_(I also think we should restructure the serialization function, but that shouldn't be done in this PR)_

app/serialization.py
jpramos123: Done! Created [RHINENG-12028](https://issues.redhat.com/browse/RHINENG-12028) to address the refactoring/restructure

kruai (approved):

thearifismail (approved):
LGTM

PR #1846, opened 2024-07-23
Add rate limiting for segmentio track calls.

roliveri:
# Overview

This PR is being created to address [RHINENG-11321](https://issues.redhat.com/browse/RHINENG-11321).
Prevent a flood of Segmentio track calls when we experience a spike in API calls.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/tests/test_segmentio.py
+++ b/tests/test_segmentio.py
@@ -1,5 +1,10 @@
+import os
+from importlib import reload
+from unittest import mock
+
 import segment.analytics as analytics
 
+import api.segmentio
 from tests.helpers.api_utils import assert_response_status
 from tests.helpers.api_utils import build_resource_types_url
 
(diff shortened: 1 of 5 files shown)


tests/test_segmentio.py
```diff
@@ -93,3 +98,45 @@ def test_segmentio_track_insights_client(mocker, api_get):
...
+
+    #
+    # Lower the maximum call limit.
+    #
+    with mock.patch.dict(os.environ, {"SEGMENTIO_LIMIT_CALLS": str(max_calls)}):
+        reload(api.segmentio)
+        reload(api)
+
+    #
+    # Test calls that do exceed the limit.
```
marleystipich2, line 130: @roliveri Can we test the exception message `segmentio_track: rate limit exceeded.`?

roliveri: > @roliveri Can we test the exception message `segmentio_track: rate limit exceeded.`?

Given the exception is caught and ignored, the only way the test code can check this is probably to mock the log.debug call, but that would apply to all the debug calls made in the tested code path. So, I'm not sure there's much value in doing so.

roliveri:
@marleystipich2 
Can we (all interested parties) come to consensus on the proper rate limit parameters?
Is 500 calls per minute acceptable to everyone?

tests/test_segmentio.py
marleystipich2: Ah right but I wonder if it'd be useful to send some kind of message to Segment when the rate limit was exceeded? So that we'll know when there was a possible spike? This might not be too important right now but something to think about for the future?

marleystipich2:
> @marleystipich2 Can we (all interested parties) come to consensus on the proper rate limit parameters? Is 500 calls per minute acceptable to everyone?

I'll send another message tomorrow to make sure everyone is okay with the limit but I think it'll be mostly up to whoever watches the data the closest which I guess would be us, Heiko, and Rob. I'll also take a look at the data again to make sure these parameters make sense.

tests/test_segmentio.py
roliveri: We would have to check the logs to identify when rate limiting occurs. The problem is, the log message is only generated in debug.

marleystipich2: That's probably fine anyway. The most important thing is that we aren't tracking the spikes anymore which this will do.

kruai (approved):
LGTM - I appreciate the tests as well!

PR #1857, opened 2024-07-30
bugfix(mq): Remove cache set mechanism in payload processing

chambridge:
# Overview

* Caching of cert-auth flow will only occur in API layer after issues with payload processing

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


kruai (dismissed):

fstavela:
Does this mean that the cache will not get updated when the host is updated via archive upload? I see 2 potential issues with it:
- Generally, HBI won't return correct data after the upload. This doesn't look right, although it may not be that big of an issue if we know that only insights-client uses these cert auth calls and it only looks for host ID in the response.
- Since multiple hosts can have the same insights_id (it shouldn't happen too often, hopefully), we are risking that the insights-client will get wrong host ID with the `GET /hosts?insights_id=...` call when a new host is created with the same insights_id. For example, if there is only one host with `insights_id=x`, then we create a new host with the same insights_id and then we do `GET /hosts?insights_id=x`, we will get only the old one.

chambridge:
> Does this mean that the cache will not get updated when the host is updated via archive upload? I see 2 potential issues with it:
> 
> * Generally, HBI won't return correct data after the upload. This doesn't look right, although it may not be that big of an issue if we know that only insights-client uses these cert auth calls and it only looks for host ID in the response.
> * Since multiple hosts can have the same insights_id (it shouldn't happen too often, hopefully), we are risking that the insights-client will get wrong host ID with the `GET /hosts?insights_id=...` call when a new host is created with the same insights_id. For example, if there is only one host with `insights_id=x`, then we create a new host with the same insights_id and then we do `GET /hosts?insights_id=x`, we will get only the old one.

@fstavela So now we are only caching data in the API layer instead of the upload flow, essentially recalling the data we returned the previous upload (assuming on invalidation occurred). We only cache the data if the resulting DB response had a single entry so we wouldn't cache if some duplicate was present. Your edge case of creating a new host with the same insights_id would also have to have the same owner_id in order for it to match in the cache.

kruai (approved):
LGTM

PR #1858, opened 2024-07-30
Detect and filter zero mac addresses in canonical facts.

roliveri:
# Overview

This PR is being created to address [RHINENG-11698](https://issues.redhat.com/browse/RHINENG-11698).

It appears that some reporters are sending  00:00:00:00:00:00 as one of the reported mac_addresses in the canonical facts. It also seems there are instances where this value is the only reported mac address. Given this is not a unique value, this can lead to duplication issues.

The CanonicalFactsSchema has been updated to check for this. In all cases, the offending reporter is logged, so it can be identified and fixed. The values are removed when they are received, so they won't be saved to the database.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -71,6 +71,8 @@ OLD_TO_NEW_REPORTER_MAP = {"yupana": ("satellite", "discovery")}
 MIN_CANONICAL_FACTS_VERSION = 0
 MAX_CANONICAL_FACTS_VERSION = 1
 
+ZERO_MAC_ADDRESS = "00:00:00:00:00:00"
+
 
 class ProviderType(str, Enum):
     ALIBABA = "alibaba"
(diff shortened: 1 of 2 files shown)


app/models.py
```diff
@@ -732,6 +734,14 @@ class Meta:
     def validate_schema(self, data, **kwargs):
         schema_version = data.get("canonical_facts_version")
 
+        if "mac_addresses" in data:
+            mac_addresses = data["mac_addresses"]
+            while ZERO_MAC_ADDRESS in mac_addresses:
+                logger.warning(f"Zero MAC address reported by: {data.get('reporter', 'Not Available')}")
+                mac_addresses.remove(ZERO_MAC_ADDRESS)
+            if not mac_addresses:
```
jpramos123, line 742: Is there a scenario where the host being reported/checked in only contains zero MAC addresses and it is the only possibility for that host?

jpramos123 (commented):
One comment

app/models.py
roliveri: I don't think so. If that were the case, then the host couldn't exist on the network (not even the local network).

jpramos123 (dismissed):

thearifismail (dismissed):

thearifismail (approved):

PR #1866, opened 2024-08-02
feat(cache): Avoid blocking execution on cache invalidation

chambridge:
# Overview

* Enable cache invalidation to avoid blocking by spawning a thread
* Slower cache invalidation has caused kafka heartbeat timeouts
* Slower cache invalidation could potentially lead to longer API responses
* Enable efficient cache deletion when full key is known, avoid scanning keys when possible

# Testing
Used local flow to test MQ caching and check System based API query for cached data

Added the following the the compose file:
```
    redis:
      image: docker.io/library/redis:latest
      restart: always
      ports:
        - "6379:6379"
      volumes:
        - ./cache:/root/redis
      environment:
        - REDIS_PORT=6379
        - REDIS_DATABASES=2
```
Added the following to the environment:
```
INVENTORY_API_CACHE_TIMEOUT_SECONDS=30
INVENTORY_API_CACHE_TYPE="RedisCache"
IS_INSIGHTS_CLIENT="True"
```
You must also enable the `hbi.api.use-cached-insights-client-system` feature flag in Unleash.

Run the following to populate the Cache:
```
make run_inv_mq_service
```
And in another terminal
```
make run_inv_mq_service_test_producer
```
This will make a system with an insights_id value that will be stored in the cache.

Review the created redis key:
```
docker exec -it insights-host-inventory-redis-1 redis-cli keys "*"
```

Next launch the web server:
```
python3 run_gunicorn.py
```

Run the following call to obtain the data from the cache instead of via the API. You will need to alter the value of the insights_id to match the value generated by the test producer:
```
curl --location 'http://localhost:8080/api/inventory/v1/hosts?insights_id=56f6c4fc-250f-4b01-8255-6719b2e88eb4' \
--header 'x-rh-identity: eyJpZGVudGl0eSI6ewogICAgIm9yZ19pZCI6ICI1ODk0MzAwIiwKICAgICJ0eXBlIjogIlN5c3RlbSIsCiAgICAiYXV0aF90eXBlIjogImNlcnQtYXV0aCIsCiAgICAic3lzdGVtIjogeyJjbiI6ICIxYjM2YjIwZi03ZmEwLTQ0NTQtYTZkMi0wMDgyOTRlMDYzNzgiLCAiY2VydF90eXBlIjogInN5c3RlbSJ9LAogICAgImludGVybmFsIjogeyJhdXRoX3RpbWUiOiA2MzAwfQp9fQ=='
```
You should see the response from the cache, with a log entry that the feature flag had been applied.

Next test cache deletion via curl or Postman (update the `bios_uuid` to a valid value for an existing system) when a change occurs to the system:
```
curl --location 'http://localhost:8080/api/inventory/v1/hosts/checkin' \
--header 'x-rh-identity: eyJpZGVudGl0eSI6eyJvcmdfaWQiOiI1ODk0MzAwIiwidHlwZSI6IlVzZXIiLCJhdXRoX3R5cGUiOiJiYXNpYy1hdXRoIiwidXNlciI6eyJ1c2VyX2lkIjogNzE0LCAidXNlcm5hbWUiOiJ0dXNlckByZWRoYXQuY29tIiwiZW1haWwiOiJ0dXNlckByZWRoYXQuY29tIiwiZmlyc3RfbmFtZSI6InRlc3QiLCJsYXN0X25hbWUiOiJ1c2VyIiwiaXNfYWN0aXZlIjp0cnVlLCJpc19vcmdfYWRtaW4iOmZhbHNlLCJpc19pbnRlcm5hbCI6dHJ1ZSwibG9jYWxlIjoiZW5fVVMifX19' \
--header 'Content-Type: application/json' \
--data '{"bios_uuid": "58563b5f-5004-4e0a-a8cb-1bc757c60eb4"}'
```
Review the existing redis keys:
```
docker exec -it insights-host-inventory-redis-1 redis-cli keys "*"
```
The key should have been deleted.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


api/cache.py
```diff
@@ -35,31 +38,37 @@ def init_cache(app_config, flask_app):
...
 
-def delete_keys(prefix):
+def delete_keys(cache_key, wildcard=True):
     global CACHE_CONFIG
 
-    if CACHE_CONFIG and CACHE_CONFIG.get("CACHE_TYPE") == "RedisCache" and prefix:
-        _delete_keys_redis(prefix)
+    if CACHE_CONFIG and CACHE_CONFIG.get("CACHE_TYPE") == "RedisCache" and cache_key:
+        thread = threading.Thread(target=_delete_keys_redis, args=(cache_key, wildcard))
+        thread.start()
```
vkrizan, line 62: These threads do not seem to be managed. There is no limit of how many of them would be spawned. Have you considered using a thread worker, or, better an asyncio (if possible)?

api/cache.py
```diff
@@ -35,31 +38,37 @@ def init_cache(app_config, flask_app):
...
         logger.exception("Cache deletion failed", exc_info=exec)
 
 
-def delete_keys(prefix):
+def delete_keys(cache_key, wildcard=True):
     global CACHE_CONFIG
 
-    if CACHE_CONFIG and CACHE_CONFIG.get("CACHE_TYPE") == "RedisCache" and prefix:
-        _delete_keys_redis(prefix)
+    if CACHE_CONFIG and CACHE_CONFIG.get("CACHE_TYPE") == "RedisCache" and cache_key:
```
vkrizan, line 60: `"RedisCache"` constant.

app/queue/queue.py
```diff
@@ -396,7 +396,8 @@ def write_delete_event_message(event_producer: EventProducer, result: OperationR
     event_producer.write_event(event, str(result.host_row.id), headers, wait=True)
     delete_keys(result.host_row.org_id)
     insights_id = result.host_row.canonical_facts.get("insights_id")
-    delete_cached_system_keys(insights_id=insights_id)
+    owner_id = result.host_row.system_profile_facts.get("owner_id")
+    delete_cached_system_keys(insights_id=insights_id, org_id=result.host_row.org_id, owner_id=owner_id)
```
vkrizan, line 400: Is the `owner_id` required to be present? Because this can end up calling `delete_keys(f"insights_id={insights_id}_org={org_id}", wildcard=True)`

api/cache.py
```diff
@@ -35,31 +38,37 @@ def init_cache(app_config, flask_app):
...
     try:
         if not REDIS_CLIENT:
             REDIS_CLIENT = Redis(host=CACHE_CONFIG.get("CACHE_REDIS_HOST"), port=CACHE_CONFIG.get("CACHE_REDIS_PORT"))
-        # Use SCAN to find keys to delete that start with the prefix; default prefix is flask_cache_
-        for key in REDIS_CLIENT.scan_iter(f"flask_cache_{prefix}*"):
-            REDIS_CLIENT.delete(key)
+        if wildcard:
+            # Use SCAN to find keys to delete that start with the prefix; default prefix is flask_cache_
+            for key in REDIS_CLIENT.scan_iter(f"{CACHE_PREFIX}{cache_key}*"):
+                REDIS_CLIENT.delete(key)
```
vkrizan, line 50: Is there a better removal of keys by pattern, that would not require iterating over existing and would be atomic?

app/queue/queue.py
chambridge: `owner_id` isn't required, but having the value will make for a faster deletion.

api/cache.py
chambridge: I'm not aware of a better way.

api/cache.py
chambridge: updated.

api/cache.py
chambridge: I looked at creating a ThreadPoolExecutor, but it seems that I can't create the executor on the init of the app, it has to be on the request thread otherwise I see a runtime error.

vkrizan: If it uses gunicorn to run the app, then look into the right initialization function to extend/override. I think it needs to be created per process.

vkrizan: I thought that this is being used on kafka processors. If this is also done on the webapp, then this should be kept to be blocking on the webapp (API).

chambridge: I've updated so only the mq workers spawn a thread and the max number of threads is configurable via variable.

api/cache.py
```diff
@@ -6,18 +8,23 @@
...
 
 
 def init_cache(app_config, flask_app):
     global CACHE
     global CACHE_CONFIG
+    global CACHE_SEMAPHORE
     cache_type = "NullCache"
     logger.info("Initializing Cache")
 
+    CACHE_SEMAPHORE = threading.Semaphore(app_config.api_cache_max_thread_pool_workers)
```
kruai, line 25: Should we use a `BoundedSemaphore`, as the max number of threads is fixed?

api/staleness_query.py
```diff
@@ -19,7 +19,7 @@ def get_staleness_obj(identity=None):
         logger.info("Using custom account staleness")
         staleness = _build_serialized_acc_staleness_obj(staleness)
     except NoResultFound:
-        logger.info(f"No data found for user {org_id}, using system default values")
+        logger.debug(f"No data found for user {org_id}, using system default values")
```
kruai, line 22: Thanks for this - this will clear up our Kibana logs a fair bit.

api/cache.py
kruai: (fwiw, I can't think of a better way either)

kruai (commented):
I'm trying to learn more about caching as I go, but I wanted to check that we're using the right semaphore class at least.

api/cache.py
chambridge: done

kruai (dismissed):

api/cache.py
vkrizan: To reiterate, the issue that this tries to fix is that the invalidation was a blocking operation that had an impact on performance (especially during peak). If the semaphore would be at 0, this would be blocking again, limited by initial value of the semaphore and (average) time for cache invalidation. On peak, this could pose the same problem...

I'd still recommend going either with a worker and queue architecture, or async.

api/cache.py
vkrizan: There is at least a support to remove more keys with one call (with O(n)).

api/cache.py
```diff
@@ -49,9 +49,12 @@ def _delete_keys_redis(cache_key, wildcard=True):
         if not REDIS_CLIENT:
             REDIS_CLIENT = Redis(host=CACHE_CONFIG.get("CACHE_REDIS_HOST"), port=CACHE_CONFIG.get("CACHE_REDIS_PORT"))
         if wildcard:
+            keys_to_delete = []
             # Use SCAN to find keys to delete that start with the prefix; default prefix is flask_cache_
             for key in REDIS_CLIENT.scan_iter(f"{CACHE_PREFIX}{cache_key}*"):
-                REDIS_CLIENT.delete(key)
+                keys_to_delete.append(key)
```
vkrizan, line 55: I guess calling the list on the iterator would do the same.
Be sure tho check whether there isn't an upper limit of how many keys it can accept. Probably it does not matter, as the delete is O(n).

api/cache.py
```diff
@@ -35,31 +42,43 @@ def init_cache(app_config, flask_app):
...
-    if CACHE_CONFIG and CACHE_CONFIG.get("CACHE_TYPE") == "RedisCache" and prefix:
-        _delete_keys_redis(prefix)
-
-
-def delete_cached_system_keys(insights_id=None, org_id=None):
-    if insights_id:
-        delete_keys(f"insights_id={insights_id}")
-    if org_id:
-        delete_keys(f"insights_id=*_org={org_id}")
+    global CACHE_EXECUTOR
```
vkrizan, line 66: The `global` keyword isn't needed, as there is no assignment in these functions other than in the init one.

kruai (approved):

PR #1871, opened 2024-08-07
feat(performance): Use batched queries to improve DB call performance 

jpramos123:
# Overview

This PR is being created to address [RHINENG-11833](https://issues.redhat.com/browse/RHINENG-11833).
This PR aims to improve the DB call that is made by the export service.

Today if we have a big amount of hosts for in a account, and we make an export request, the pods are killed because of Out Of Memory (OOM) issue.

Using batched queries would reduce the chances of breaking the pods.

I'm proposing using `yield_per` from [SQLAlchemy](https://docs.sqlalchemy.org/en/20/core/connections.html#streaming-with-a-fixed-buffer-via-yield-per)  to address this change and the use of generators for serialize each host individually. This would reduce the memory used, as the serialized data is reduced compared to the data that is fetch from the DB.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/host_query_db.py
+++ b/api/host_query_db.py
@@ -1,5 +1,6 @@
 import re
 from itertools import islice
+from typing import Iterator
 from typing import List
 from typing import Tuple
 
@@ -593,22 +594,20 @@ def get_host_ids_list(
 
 def get_hosts_to_export(
     identity: object, filters: object = {}, export_format: str = "json", rbac_filter: dict = {}, batch_size: int = 0
-) -> list:
+) -> Iterator[dict]:
     st_timestamps = staleness_timestamps()
     staleness = get_staleness_obj(identity)
 
     q_filters, _ = query_filters(filter=filters, rbac_filter=rbac_filter)
     export_host_query = _find_all_hosts(identity=identity).filter(*q_filters)
     export_host_query = export_host_query.execution_options(yield_per=batch_size)
-    serialized_hosts_list = []
-    for partition in db.session.scalars(export_host_query).partitions():
-        for host in partition:
-            serialized_hosts_list.append(
-                serialize_host(
-                    host, for_mq=False, for_export_svc=True, staleness_timestamps=st_timestamps, staleness=staleness
-                )
-            )
+
+    num_hosts = export_host_query.count()
+    logger.debug(f"Number of hosts to be exported: {num_hosts}")
+
+    for host in db.session.scalars(export_host_query):
+        yield serialize_host(
+            host, for_mq=False, for_export_svc=True, staleness_timestamps=st_timestamps, staleness=staleness
+        )
 
     db.session.close()
-
-    return serialized_hosts_list
--- a/api/host_query_db.py
+++ b/api/host_query_db.py
@@ -28,6 +28,7 @@ from app.models import HostGroupAssoc
 from app.serialization import serialize_host
 from lib.host_repository import update_query_for_owner_id
 
+
 __all__ = (
     "get_all_hosts",
     "get_host_list",
@@ -591,23 +592,23 @@ def get_host_ids_list(
 
 
 def get_hosts_to_export(
-    identity: object,
-    filters: object = {},
-    export_format: str = "json",
-    rbac_filter: dict = {},
+    identity: object, filters: object = {}, export_format: str = "json", rbac_filter: dict = {}, batch_size: int = 0
 ) -> list:
     st_timestamps = staleness_timestamps()
     staleness = get_staleness_obj(identity)
 
     q_filters, _ = query_filters(filter=filters, rbac_filter=rbac_filter)
     export_host_query = _find_all_hosts(identity=identity).filter(*q_filters)
-    db.session.close()
+    export_host_query = export_host_query.execution_options(yield_per=batch_size)
+    serialized_hosts_list = []
+    for partition in db.session.scalars(export_host_query).partitions():
+        for host in partition:
+            serialized_hosts_list.append(
+                serialize_host(
+                    host, for_mq=False, for_export_svc=True, staleness_timestamps=st_timestamps, staleness=staleness
+                )
+            )
 
-    serialized_hosts_list = [
-        serialize_host(
-            host, for_mq=False, for_export_svc=True, staleness_timestamps=st_timestamps, staleness=staleness
-        )
-        for host in export_host_query.all()
-    ]
+    db.session.close()
 
     return serialized_hosts_list
--- a/app/queue/export_service.py
+++ b/app/queue/export_service.py
@@ -54,7 +54,12 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={}, 
         rbac_filter = get_rbac_filter(
             RbacResourceType.HOSTS, RbacPermission.READ, identity=identity, rbac_request_headers=rbac_request_headers
         )
-        data_to_export = get_hosts_to_export(identity, export_format=exportFormat, rbac_filter=rbac_filter)
+        data_to_export = get_hosts_to_export(
+            identity,
+            export_format=exportFormat,
+            rbac_filter=rbac_filter,
+            batch_size=inventory_config.export_svc_batch_size,
+        )
 
         if data_to_export:
             logger.debug(f"Trying to upload data using URL:{request_url}")
--- a/app/queue/export_service.py
+++ b/app/queue/export_service.py
@@ -98,7 +98,7 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={}, 
 def _handle_export_response(response, exportFormat, exportUUID):
     if response.status_code != HTTPStatus.ACCEPTED:
         raise Exception(response.text)
-    else:
+    elif response.text != "":
         logger.info(f"{response.text} for export ID {exportUUID} in {exportFormat.upper()} format")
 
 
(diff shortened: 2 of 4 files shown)


api/host_query_db.py
```diff
@@ -571,18 +579,22 @@ def get_hosts_to_export(
...
     staleness = get_staleness_obj(identity)
 
     q_filters, _ = query_filters(filter=filters, rbac_filter=rbac_filter)
     export_host_query = _find_all_hosts(identity=identity).filter(*q_filters)
-    db.session.close()
+    export_host_query = export_host_query.execution_options(yield_per=config.export_svc_batch_size)
+    serialized_hosts_list = []
+    for partition in db.session.scalars(export_host_query).partitions():
+        for host in partition:
+            serialized_hosts_list.append(
```
vkrizan, line 592: This would still have the issue of reaching a memory limit. Each batch should be pushed out, clearing the memory for next batch.

api/host_query_db.py
```diff
@@ -571,18 +579,22 @@ def get_hosts_to_export(
...
     st_timestamps = staleness_timestamps()
     staleness = get_staleness_obj(identity)
 
     q_filters, _ = query_filters(filter=filters, rbac_filter=rbac_filter)
     export_host_query = _find_all_hosts(identity=identity).filter(*q_filters)
-    db.session.close()
+    export_host_query = export_host_query.execution_options(yield_per=config.export_svc_batch_size)
+    serialized_hosts_list = []
+    for partition in db.session.scalars(export_host_query).partitions():
+        for host in partition:
```
vkrizan, line 591: Although for-loops may be more readable, a (list) comprehension might be faster, especially on lager sets.

api/host_query_db.py
```diff
@@ -53,6 +56,11 @@
 ]
 
 
+def _init_config():
+    config = Config(RuntimeEnvironment.SERVICE)
+    return config
+
+
```
strider, line 63: João, there is no need to get this here as the `create_export` function gets the `inventory_config` by parameter. You may just need to pass the batch size value to the `get_hosts_to_export(...)` as parameter as well instead of instantiate all the Config object here.

My two cents.

strider (commented):
It works pretty well as is.

However, we may address my comment about the config object.

api/host_query_db.py
thearifismail: I agree with @strider because the app has already been created and config available.

api/host_query_db.py
thearifismail: @jpramos123 This is what I meant yesterday to process a chunk, manually set the object to None before iterating again.

jpramos123: I see! I need to validate with export-svc team, but I'm I doubt we can send chunks of data to be handled by the service now.

At the end, we need to send all the data that is being serialized at the same time via  POST request.

But... still there is a possibility to improve and implement this feature on export-service side.

Thanks!

api/host_query_db.py
jpramos123: Thanks, I will do this change!

api/host_query_db.py
jpramos123: Thanks! I will address this ASAP

api/host_query_db.py
thearifismail: @jpramos123 , it is a host-inventory pod that's crashing, right?  Unless a request for data export comes with an ID, which must go back with the response, there should be no problem with how many chunks host-inventory sends back.

jpramos123: @vkrizan I made a try, but looks like that the go server needs to handle the chunked requests on their side.
Still, can you evaluate my last draft commit changes? 

The export-service-go returns 410 with the message: `his resource has already been processed`

vkrizan: @jpramos123 I'll check out out... Feel free to merge the database optimization first, and then we can do chunking next.

vkrizan: Update the `get_hosts_to_export` to yield a chunk of hosts. It still prepares the full list!

vkrizan: If chunks aren't supported, then it needs to be done using stream (as in file upload). https://requests.readthedocs.io/en/latest/user/advanced/#streaming-uploads It does not necessarily have to be a file, a file-like object would do (https://docs.python.org/3/library/io.html).

app/queue/export_service.py
```diff
@@ -54,15 +54,29 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+            rbac_filter=rbac_filter,
+            batch_size=inventory_config.export_svc_batch_size,
+        )
+
+        try:
+            host_data = [next(data_to_export)]
+            while True:
+                host_data.append(next(data_to_export))
+        except StopIteration:
+            logger.debug("End of iterator")
```
vkrizan, line 71: @jpramos123 Why can't we use `for` (each) loop? This reimplements iterator pattern that is already part of the Python language.

vkrizan (changes requested):

app/queue/export_service.py
```diff
@@ -54,15 +54,29 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+            identity,
+            export_format=exportFormat,
+            rbac_filter=rbac_filter,
+            batch_size=inventory_config.export_svc_batch_size,
+        )
+
+        try:
+            host_data = [next(data_to_export)]
+            while True:
+                host_data.append(next(data_to_export))
```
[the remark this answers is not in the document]
vkrizan: We did serialized it one at the time before, so fundamentally this is the same.

> We decided to not to use the batch POST calls for now.

Can you elaborate on this one? 
If we don't want to do a batched POST call, then lets keep this PR clean from this serialization batching and only have the db query batch improvement.

app/queue/export_service.py
```diff
@@ -54,15 +54,26 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+        data_to_export = get_hosts_to_export(
+            identity,
+            export_format=exportFormat,
+            rbac_filter=rbac_filter,
+            batch_size=inventory_config.export_svc_batch_size,
+        )
+
+        host_data = []
+        for host in data_to_export:
+            host_data.append(host)
```
vkrizan, line 68: This can be shortened by calling `list()` on the generator.

vkrizan (approved):

api/host_query_db.py
```diff
@@ -566,23 +567,18 @@ def get_host_ids_list(
 
 
 def get_hosts_to_export(
-    identity: object,
-    filters: object = {},
-    export_format: str = "json",
-    rbac_filter: dict = {},
-) -> list:
+    identity: object, filters: object = {}, export_format: str = "json", rbac_filter: dict = {}, batch_size: int = 0
+):
```
kruai, line 571:
```suggestion
) -> Iterator[dict]:
```

Note that you'll need to add this to your imports:
`from typing import Iterator`

kruai (commented):
Mostly looks good, I just wanna keep an accurate type hint (and also I agree with Viliam's suggestion)

api/host_query_db.py
jpramos123: Added this suggestion!

app/queue/export_service.py
jpramos123: Done!

vkrizan (approved):

api/host_query_db.py
```diff
@@ -566,23 +568,18 @@ def get_host_ids_list(
...
     staleness = get_staleness_obj(identity)
 
     q_filters, _ = query_filters(filter=filters, rbac_filter=rbac_filter)
     export_host_query = _find_all_hosts(identity=identity).filter(*q_filters)
-    db.session.close()
+    export_host_query = export_host_query.execution_options(yield_per=batch_size)
 
-    serialized_hosts_list = [
-        serialize_host(
+    for host in db.session.scalars(export_host_query):
```
vkrizan, line 580: What would `.scalars()` do here? And, what's the difference between that and `.all()`?

vkrizan (changes requested):
I'm unsure about the use of `.scalars()`. By the docs, it would return all rows on the first column. Which does seem wrong and would likely bring back the memory issue, if that's true. Can you please double check?

vkrizan (approved):

app/queue/export_service.py
```diff
@@ -54,15 +54,24 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
             RbacResourceType.HOSTS, RbacPermission.READ, identity=identity, rbac_request_headers=rbac_request_headers
         )
-        data_to_export = get_hosts_to_export(identity, export_format=exportFormat, rbac_filter=rbac_filter)
 
-        if data_to_export:
+        # create a generator with serialized host data
+        host_data = list(
+            get_hosts_to_export(
+                identity,
+                export_format=exportFormat,
```
thearifismail, line 62: How many different export formats are possible?  It looks like `json` is the only option.  Can it work without even specifying format?

kruai (dismissed):

app/queue/export_service.py
jpramos123: 2 types, CSV and JSON.

The default export is JSON

api/host_query_db.py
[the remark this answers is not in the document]
jpramos123: Added this suggestion as a debug log

Thanks

kruai (approved):

thearifismail (approved):

PR #1880, opened 2024-08-13
feat(RHINENG-11709) Implement GET /host_exists

kruai:
# Overview

This PR is being created to address [RHINENG-11709](https://issues.redhat.com/browse/RHINENG-11709).
Implements the `GET /host_exists` endpoint. I was not able to call it `GET /hosts/exists` because it interprets that as the `GET /hosts/{host_id_list}` endpoint with an invalid ID.

Given an Insights ID, it fetches only the Host ID for a matching host. If no host matches, it returns 404. If more than one host matches, it returns 409.

Example SQL:
```
SELECT hosts.id AS hosts_id 
FROM hosts 
WHERE hosts.org_id = %(org_id_1)s AND lower((hosts.canonical_facts ->> %(canonical_facts_1)s)) = %(lower_1)s ORDER BY hosts.modified_on DESC
```

## PR Checklist

- [x] Include raw query examples in the PR description, if adding/modifying SQL query
- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


api/host.py
```diff
@@ -602,3 +605,20 @@ def host_checkin(body, rbac_filter=None):
...
         flask.abort(404, "No hosts match the provided canonical facts.")
+
+
+@api_operation
+@rbac(RbacResourceType.HOSTS, RbacPermission.READ)
+@metrics.api_request_time.time()
+def get_host_exists(insights_id, rbac_filter=None):
+    try:
+        host_id = get_host_id_by_insights_id(insights_id, rbac_filter)
+    except ValueError as e:
```
vkrizan, line 616: When is a ValueError raised?

api/filtering/db_filters.py
```diff
@@ -38,7 +38,7 @@
 DEFAULT_STALENESS_VALUES = ["not_culled"]
 
 
-def _canonical_fact_filter(canonical_fact: str, value, case_insensitive: bool = False) -> List:
+def canonical_fact_filter(canonical_fact: str, value, case_insensitive: bool = False) -> List:
```
vkrizan, line 41: Small but good refactoring, that can be separated by commit.

api/host_query_db.py
```diff
@@ -155,6 +156,21 @@ def get_host_list_by_id_list(
...
+        [Host.org_id == identity.org_id]
+        + canonical_fact_filter("insights_id", insights_id, True)
+        + rbac_permissions_filter(rbac_filter)
+    )
+
+    query = db.session.query(Host).filter(*all_filters)
+    query = update_query_for_owner_id(identity, query)
+
+    found_id = query.with_entities(Host.id).order_by(Host.modified_on.desc()).first()
+    return str(found_id[0]) if found_id else None
```
vkrizan, line 171: Wouldn't calling `.scalar()` (instead of `first()`) do the trick here without the need to do the check and pull the first element?

api/host_query_db.py
```diff
@@ -155,6 +156,21 @@ def get_host_list_by_id_list(
     return items, total, additional_fields, system_profile_fields
 
 
+def get_host_id_by_insights_id(insights_id: str, rbac_filter=None) -> str:
+    identity = get_current_identity()
+    all_filters = (
+        [Host.org_id == identity.org_id]
+        + canonical_fact_filter("insights_id", insights_id, True)
```
vkrizan, line 163: I'd advise to use keyword argument for the third parameter. When looking at it, it isn't obvious what the `True` is.

api/host_query_db.py
```diff
@@ -155,6 +156,21 @@ def get_host_list_by_id_list(
...
+        + rbac_permissions_filter(rbac_filter)
+    )
+
+    query = db.session.query(Host).filter(*all_filters)
+    query = update_query_for_owner_id(identity, query)
+
+    found_id = query.with_entities(Host.id).order_by(Host.modified_on.desc()).first()
```
vkrizan, line 170: Can you share an example of SQL here?

vkrizan (approved):
Looks good so far. I've left some small comments inline.

api/host_query_db.py
vkrizan: Does it have to be case insensitive if it pulls an id? I wonder if the index for this canonical fact is also insensitive to have a fast lookup. Can you validate that please with `EXPLAIN`, if this needs to stay insensitive?

api/host.py
kruai: Good point, it's not raised on this path. Removing this block

api/host_query_db.py
kruai: Yep, `scalar` is the better call here

api/filtering/db_filters.py
kruai: Please clarify - do you mean I should make a separate PR for it?

api/host_query_db.py
kruai: Good point - we shouldn't need the case insensitivity anyway. Ran a couple `EXPLAIN ANALYZE` and you're right, the insights_id index was not used when using the `lower()` function.

api/filtering/db_filters.py
vkrizan: A commit would do.

jpramos123 (approved):

api/filtering/db_filters.py
thearifismail, line 41: Shouldn't `canonical_fact_filter` be added to `__all__`?

api/host_query_db.py
```diff
@@ -155,6 +158,28 @@ def get_host_list_by_id_list(
     return items, total, additional_fields, system_profile_fields
 
 
+def get_host_id_by_insights_id(insights_id: str, rbac_filter=None) -> str:
```
thearifismail, line 161: Hard coding `insights_id` in the function name?  Does it mean that each `canonical_fact` will have its own function?  Wouldn't renaming the function to `get_host_id_by_canonical_facts()` be better and create filter using the canonical_facts provided.

thearifismail, line 161: Hard coding `insights_id` in the function name?  Does it mean that each `canonical_fact` will have its own function?  Wouldn't renaming the function to `get_host_id_by_canonical_facts()` be better and create filter using the canonical_facts provided.

api/host.py
```diff
@@ -37,6 +38,7 @@
 from app.auth.identity import to_auth_header
 from app.common import inventory_config
 from app.instrumentation import get_control_rule
+from app.instrumentation import log_get_host_exists_succeeded
```
thearifismail, line 41: No possibility of `log_get_host_exists_failed`?

swagger/api.spec.yaml
```diff
@@ -350,6 +350,29 @@ paths:
...
+      responses:
+        '200':
+          description: Found a matching host.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HostIdOut'
+        '400':
+          description: Invalid request.
+        '404':
```
thearifismail, line 374: The exception in the function returns `409`

tests/helpers/api_utils.py
```diff
@@ -463,6 +464,10 @@ def build_hosts_url(host_list_or_id=None, path=None, query=None):
     return _build_url(id_list=host_list_or_id, path=path, query=query)
 
 
+def build_host_exists_url(insights_id):
+    return _build_url(base_url=HOST_EXISTS_URL, query=f"?insights_id={insights_id}")
```
thearifismail, line 468: Same question regarding other canonical_facts?

api/filtering/db_filters.py
kruai: Good call, added

swagger/api.spec.yaml
kruai: You're right! Added it to the expected response codes.

api/host_query_db.py
kruai: No, we only need it for `insights_id`. The use case is specific to insights-client's usage, which looks up a single host by the Insights ID, and we need it to be as lightweight as possible. Introducing other filters adds complexity that we don't need to support.
If, in the future, we need this endpoint to support other ID types, we can modify this implementation to reflect that.

kruai: Same response as https://github.com/RedHatInsights/insights-host-inventory/pull/1880#discussion_r1725497330

tests/helpers/api_utils.py
kruai: Same response as this https://github.com/RedHatInsights/insights-host-inventory/pull/1880#discussion_r1725497330

api/host.py
kruai: There's no use case I can think of where it would be useful. Got one in mind?

thearifismail: Since finding more than one hosts is an error, should there be a message logged stating more than one hosts found.  I okay with not logging multiple hosts found.

thearifismail (approved):
LGTM!

kruai:
Just to be safe, I'm gonna hold off on merging this until https://issues.redhat.com/browse/RHINENG-11710 is complete (FYI @fstavela)

FabriciaDinizRH (approved):

PR #1894, opened 2024-08-27
bugfix(rhineng-12147): Perform finer grain key deletions from cache

chambridge:
# Overview

This PR is being created to address [RHINENG-12147](https://issues.redhat.com/browse/RHINENG-12147).
* Optimize cache key deletion to avoid search and delete cases
* Allow thread spawning for staleness APIs

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/lib/group_repository.py
+++ b/lib/group_repository.py
@@ -1,5 +1,6 @@
 from typing import List
 
+from api.cache import delete_cached_system_keys
 from api.host_query import staleness_timestamps
 from api.staleness_query import get_staleness_obj
 from app.auth import get_current_identity
@@ -59,6 +60,11 @@ def _produce_host_update_events(event_producer, host_id_list, group_id_list=[], 
         )
         event = build_event(EventType.updated, serialized_host, platform_metadata=metadata)
         event_producer.write_event(event, serialized_host["id"], headers, wait=True)
+
+    for host in host_list:
+        insights_id = host.canonical_facts.get("insights_id")
+        owner_id = host.system_profile_facts.get("owner_id")
+        delete_cached_system_keys(insights_id=insights_id, org_id=identity.org_id, owner_id=owner_id)
 
 
 def _add_hosts_to_group(group_id: str, host_id_list: List[str]):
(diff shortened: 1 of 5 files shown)


lib/group_repository.py
```diff
@@ -60,6 +61,11 @@ def _produce_host_update_events(event_producer, host_id_list, group_id_list=[],
         event = build_event(EventType.updated, serialized_host, platform_metadata=metadata)
         event_producer.write_event(event, serialized_host["id"], headers, wait=True)
 
+    for host in host_list:
+        insights_id = host.canonical_facts.get("insights_id")
+        owner_id = host.system_profile_facts.get("owner_id")
+        delete_cached_system_keys(insights_id=insights_id, org_id=identity.org_id, owner_id=owner_id)
```
vkrizan, line 67: Could you clarify the reason for calling this cache key delete within the produce function? It might be more appropriate to have this run on a different layer (above) that aligns better with the intended functionality.

chambridge: We are attempting to minimize usage of the search for key deletion as this is causing a response delay in the API leading to some 504s.

This produce event flow is a common code flow between the group change APIs that also already loads the list of hosts from the database. If I were to move this to another section of the code within this library function we would need to either re-query for the hosts or pass the host list as a return object in order to avoid the look up which would in turn just be used to call the same for loop.

kruai (approved):

fstavela (approved):

lib/group_repository.py
vkrizan: Can we do the latter then (pass the host list as a return object)? Refactor the code as necessary to accommodate the change?

The loop is duplicated here nevertheless. However, that shouldn't be an issue to loop it again.

chambridge: Yes. I can make that change.

PR #1911, opened 2024-09-09
feat(RHINENG-11833): Reduce the number of columns in the DB call

jpramos123:
# Overview

This PR is being created to address [RHINENG-11833](https://issues.redhat.com/browse/RHINENG-11833).

This PR reduces the number of columns that is fetch from the DB when using the export-service in order to reduce the amount of data that is kept in memory when making the DB calls.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


api/host_query_db.py
```diff
@@ -247,7 +250,10 @@ def _find_all_hosts(query_base=None, columns: List[ColumnElement] = None, identi
 
     query = query_base.filter(Host.org_id == identity.org_id)
     if columns:
-        query = query.with_entities(*columns)
+        if load_only_option:
+            query = query.options(load_only(*columns))
+        else:
+            query = query.with_entities(*columns)
```
kruai, line 256: Is the load_only approach necessary here? `query.with_entities()` already makes it so only the specified columns are loaded

jpramos123: @kruai When I was using the `query.with_entities()` I was receiving  an strange behavior (only the first column was being returned). Also, the `query.with_entities` returns a list of tuples with the data from the object, but I need the actual Host object to be returned, so it is serialized correctly.

Thanks

kruai: For posterity, we're going to revisit `_find_all_hosts` because the query it returns can currently return different data types depending on whether `columns` is provided. @jpramos123 created [RHINENG-12481](https://issues.redhat.com/browse/RHINENG-12481) to address this.
And just to be clear, the addition of `load_only` does not exacerbate the issue

kruai (approved):

PR #1912, opened 2024-09-09
fix(RHINENG-12407) Add org_id and account to threadctx, if available

kruai:
# Overview

This PR is being created to address [RHINENG-12407](https://issues.redhat.com/browse/RHINENG-12407).
It fixes the batch MQ logging issue. The logger automatically pulls its org_id and account from threadctx, so even though these fields were correct on the logged host, they weren't correct on the top level of the log. This is fixed by setting the org_id and account on threadctx each time we produce an add/update message.

The reason I don't pass org_id and account to `initialize_thread_local_storage` during `handle_message` is that the incoming message hasn't been loaded in yet.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/queue/queue.py
+++ b/app/queue/queue.py
@@ -407,7 +407,7 @@ def write_delete_event_message(event_producer: EventProducer, result: OperationR
 def write_add_update_event_message(event_producer: EventProducer, result: OperationResult):
     # The request ID in the headers is fetched from threadctx.request_id
     request_id = result.platform_metadata.get("request_id")
-    initialize_thread_local_storage(request_id)
+    initialize_thread_local_storage(request_id, result.host_row.org_id, result.host_row.account)
 
     payload_tracker = get_payload_tracker(request_id=request_id)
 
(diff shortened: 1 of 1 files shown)


FabriciaDinizRH (approved):

app/queue/queue.py
```diff
@@ -585,5 +585,9 @@ def event_loop(consumer, flask_app, event_producer, notification_event_producer,
                     db.session.rollback()
 
 
-def initialize_thread_local_storage(request_id):
+def initialize_thread_local_storage(request_id: str, org_id: str = None, account: str = None):
     threadctx.request_id = request_id
+    if org_id:
+        threadctx.org_id = org_id
+    if account:
```
thearifismail, line 592: Since `org_id` is the primary attribute and account is not used rather ignored,, should it even check for `account`?

kruai: org_id is the primary attribute, but account can still be useful as it's used in some places like customer cases. It's  also automatically added to a bunch of our logs, so we want to make sure it's correct at the very least.

If you want to remove account from the logs then I think that's a separate conversation.

jpramos123 (approved):

PR #1922, opened 2024-09-12
chore(RHINENG-9244) Remove XJoin code

kruai:
# Overview

This PR is being created to address [RHINENG-9244](https://issues.redhat.com/browse/RHINENG-9244).
Removes all code relating to XJoin and GraphQL from the codebase, including tests.
I also had to add a couple more validation checks to `db_custom_filters.py` due to some test failures.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


api/__init__.py
```diff
@@ -83,3 +83,9 @@ def custom_escape(expression):
 
 def json_error_response(title, detail, status=HTTPStatus.BAD_REQUEST):
     return flask_json_response({"title": title, "detail": detail}, status)
+
+
+def pagination_params(page, per_page):
```
jpramos123, line 88: This is adding new code, can be split into another PR?

api/filtering/db_custom_filters.py
```diff
@@ -270,6 +270,9 @@ def build_single_text_filter(filter_param: dict) -> str:
         if not pg_op:
             pg_op = POSTGRES_DEFAULT_COMPARATOR.get(field_filter)
 
+        if pg_op == "?" and field_filter != "array":
```
jpramos123, line 273: same question as above

api/filtering/db_custom_filters.py
```diff
@@ -286,8 +289,10 @@ def build_single_text_filter(filter_param: dict) -> str:
         # Put value in quotes if appropriate for the field type and operation
         if field_filter in ["wildcard", "string", "timestamp", "array"] and pg_op != "IS":
             value = f"'{value}'"
-        elif value == "" or (
-            field_filter == "integer" and (not value.isdigit() and value not in ["NULL", "NOT NULL"])
+        elif (
```
jpramos123, line 292: same question as above

api/host.py
```diff
@@ -224,49 +212,25 @@ def delete_hosts_by_filter(
...
-                updated_start,
-                updated_end,
-                group_name,
-                registered_with,
-                staleness,
-                tags,
-                filter,
-                rbac_filter,
-            )
+        ids_list = get_host_ids_list(
```
jpramos123, line 215: same question as above.

This feel more like a renaming, and code standardization.

jpramos123 (commented):
I think some parts of the code can be split into another PR to make it atomic.

Otherwise all good!

Thakns

api/__init__.py
kruai: It's actually not new code; I just moved it from app.xjoin

api/host.py
kruai: I figured it made sense to rename it in this PR, since we're no longer differentiating between xjoin and postgres. Would you rather I save that for a separate PR?

api/filtering/db_custom_filters.py
kruai: The changes to this function are what I was referring to in the description. Removing the xjoin implementation actually caused a few existing tests to fail, indicating that there were some gaps in the postgres-specific implementation. I can add these checks to a separate PR, though we'll have to merge it before this one

api/__init__.py
vkrizan: Such moves deserve to be separated with commits (before the file is deleted).

fstavela:
Most failing tests in the "All tests" pipeline are failing because they need to be updated. I have the updates ready, but I can't merge them until this PR is merged, so we can ignore the majority of these failures for now.
However, there are a few valid failures that indicate a bug. If we try to filter hosts or tags with the `filter` parameter and use `nil` on some fields, I get error 400 saying "'NULL' is an invalid value for field <field>". These are the failing fields: katello_agent_running, satellite_managed, sap_system, is_marketplace, greenboot_fallback_detected

api/filtering/db_custom_filters.py
jpramos123: Nah, in this case I would prefer to keep it here, but separate this changes at least into atomic commits in the same branch.

No need to open a new PR, but let try to keep the code changes atomic!

thanks

kruai: OK, I'll remember that for next time. I tend to forget that kind of thing and see the PR comparison as the end result.

api/filtering/db_custom_filters.py
kruai: Gonna resolve this as well, since the above was resolved

api/__init__.py
kruai: OK, I'll remember that for next time. I suppose I need to pay more attention to what goes into specific commits. When I'm reviewing PRs, I tend to focus on the final code comparison rather than comparing by commit, which I think has led me to be less meticulous with separating the contents/purpose of each commit.

kruai:
> Most failing tests in the "All tests" pipeline are failing because they need to be updated. I have the updates ready, but I can't merge them until this PR is merged, so we can ignore the majority of these failures for now. However, there are a few valid failures that indicate a bug. If we try to filter hosts or tags with the `filter` parameter and use `nil` on some fields, I get error 400 saying "'NULL' is an invalid value for field ". These are the failing fields: katello_agent_running, satellite_managed, sap_system, is_marketplace, greenboot_fallback_detected

Good catch. I made a tiny fix to address this, and will keep an eye on the test results. Thanks for pointing me to the specific failures - makes it much easier to address the real issues.

fstavela (approved):

api/system_profile.py
```diff
@@ -184,51 +60,23 @@ def get_sap_sids(
...
-    response = graphql_query(SAP_SIDS_QUERY, variables, log_get_sap_sids_failed)
-
-    data = response["hostSystemProfile"]
-
-    check_pagination(offset, data["sap_sids"]["meta"]["total"])
-
-    log_get_sap_sids_succeeded(logger, data)
-    return flask_json_response(
-        build_collection_response(data["sap_sids"]["data"], page, per_page, data["sap_sids"]["meta"]["total"])
+    results, total = get_sap_sids_info_db(
```
thearifismail, line 68: Following aliasing should be removed.  If not now, in a separate PR.  Else future reader would think, what's other than DB?
```
from api.host_query_db import get_os_info as get_os_info_db
from api.host_query_db import get_sap_sids_info as get_sap_sids_info_db
from api.host_query_db import get_sap_system_info as get_sap_system_info_db
```

thearifismail (approved):
Following aliasing should be removed. If not now, in a separate PR. Else future reader would think, what's other than DB?
```
from api.host_query_db import get_os_info as get_os_info_db
from api.host_query_db import get_sap_sids_info as get_sap_sids_info_db
from api.host_query_db import get_sap_system_info as get_sap_system_info_db
```

PR #1926, opened 2024-09-13
fix(RHINENG-11980): Do not export culled hosts

jpramos123:
# Overview

This PR is being created to address [RHINENG-11980](https://issues.redhat.com/browse/RHINENG-11980).

- It fixes the behavior of exporting culled hosts when using the export-service.
- Changes the way we create the identity. Now using `x-rh-identity` header and `from_auth_header` to create it.
- Updated Makefile to contain `auth_type` key in the base64 string.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/queue/export_service.py
```diff
@@ -42,6 +41,34 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+            batch_size=inventory_config.export_svc_batch_size,
+        )
+    )
+
+    return host_data
+
+
+@metrics.create_export_processing_time.time()
+def create_export(export_svc_data, org_id, inventory_config, operation_args={}, rbac_filter={}):
+    identity = create_mock_identity_with_org_id(org_id)
```
thearifismail, line 63: `create_mock_identity_with_org_id()` assumes `USER` as the `IdentityType`.  What happens when the provided `org_id` belongs to `IdentityType.SYSTEM`?

thearifismail:
running `podman-compose -f dev.yml up` appears to get stuck
```
$ pc -f dev.yml up
0fdb0af8b1787341167e96de929873698b089fb29c5b94ad7d9168c6d9c62e4f
Error: creating named volume "42688411761f3b0b6ad9565de6460e1d63ec823c957c69681467ac8d023159b0": allocating lock for new volume: allocation failed; exceeded num_locks (2048)
Error: creating named volume "2ea0e60c6498d7a0865415d25d81721a33ed60117e965f54872ae18b1160e201": allocating lock for new volume: allocation failed; exceeded num_locks (2048)
fb831c1d844b3cd989e8bbb17f8fe61813c14066a6e4e265810e5a428f1ca651
Error: "hbi-db" is not a valid container, cannot be used as a dependency: no container with name or ID "hbi-db" found: no such container
Error: "zookeeper" is not a valid container, cannot be used as a dependency: no container with name or ID "zookeeper" found: no such container
Error: "hbi-db" is not a valid container, cannot be used as a dependency: no container with name or ID "hbi-db" found: no such container
Error: allocating lock for new container: allocation failed; exceeded num_locks (2048)
[export-service]  | Error: no container with name or ID "export-service" found: no such container
[db]              | Error: no container with name or ID "hbi-db" found: no such container
[kafka]           | Error: no container with name or ID "kafka" found: no such container
[s3-createbucket] | Error: no container with name or ID "s3-createbucket" found: no such container
[zookeeper]       | Error: no container with name or ID "zookeeper" found: no such container
[unleash]         | Error: no container with name or ID "unleash" found: no such container
[minio]           | MinIO Object Storage Server
[minio]           | Copyright: 2015-2024 MinIO, Inc.
[minio]           | License: GNU AGPLv3 - https://www.gnu.org/licenses/agpl-3.0.html
[minio]           | Version: RELEASE.2024-08-03T04-33-23Z (go1.22.5 linux/amd64)
[minio]           | 
[minio]           | API: http://0.0.0.0:9099 
[minio]           | WebUI: http://0.0.0.0:9991 
[minio]           | 
[minio]           | Docs: https://min.io/docs/minio/linux/index.html
[minio]           | INFO: 
[minio]           |  You are running an older version of MinIO released 1 month before the latest release 
[minio]           |  Update: Run `mc admin update ALIAS` 
[minio]           | 
[minio]           | 
```

thearifismail:
Before the latest push to this branch, `podman-compose` could not launch only the Kafka container but this one has problems with many more

jpramos123:
> running `podman-compose -f dev.yml up` appears to get stuck
> 
> ```
> $ pc -f dev.yml up
> 0fdb0af8b1787341167e96de929873698b089fb29c5b94ad7d9168c6d9c62e4f
> Error: creating named volume "42688411761f3b0b6ad9565de6460e1d63ec823c957c69681467ac8d023159b0": allocating lock for new volume: allocation failed; exceeded num_locks (2048)
> Error: creating named volume "2ea0e60c6498d7a0865415d25d81721a33ed60117e965f54872ae18b1160e201": allocating lock for new volume: allocation failed; exceeded num_locks (2048)
> fb831c1d844b3cd989e8bbb17f8fe61813c14066a6e4e265810e5a428f1ca651
> Error: "hbi-db" is not a valid container, cannot be used as a dependency: no container with name or ID "hbi-db" found: no such container
> Error: "zookeeper" is not a valid container, cannot be used as a dependency: no container with name or ID "zookeeper" found: no such container
> Error: "hbi-db" is not a valid container, cannot be used as a dependency: no container with name or ID "hbi-db" found: no such container
> Error: allocating lock for new container: allocation failed; exceeded num_locks (2048)
> [export-service]  | Error: no container with name or ID "export-service" found: no such container
> [db]              | Error: no container with name or ID "hbi-db" found: no such container
> [kafka]           | Error: no container with name or ID "kafka" found: no such container
> [s3-createbucket] | Error: no container with name or ID "s3-createbucket" found: no such container
> [zookeeper]       | Error: no container with name or ID "zookeeper" found: no such container
> [unleash]         | Error: no container with name or ID "unleash" found: no such container
> [minio]           | MinIO Object Storage Server
> [minio]           | Copyright: 2015-2024 MinIO, Inc.
> [minio]           | License: GNU AGPLv3 - https://www.gnu.org/licenses/agpl-3.0.html
> [minio]           | Version: RELEASE.2024-08-03T04-33-23Z (go1.22.5 linux/amd64)
> [minio]           | 
> [minio]           | API: http://0.0.0.0:9099 
> [minio]           | WebUI: http://0.0.0.0:9991 
> [minio]           | 
> [minio]           | Docs: https://min.io/docs/minio/linux/index.html
> [minio]           | INFO: 
> [minio]           |  You are running an older version of MinIO released 1 month before the latest release 
> [minio]           |  Update: Run `mc admin update ALIAS` 
> [minio]           | 
> [minio]           | 
> ```

This is working just fine on my side. But I'm using docker-compose.

app/queue/export_service.py
jpramos123: Good question. When trying to use an different IDENTITY with the system type, I was not able to even pass through the export-service API.

@dehort do you have an example of using the Identity with the system type?

Thanks

api/filtering/db_filters.py
```diff
@@ -153,8 +153,11 @@ def per_reporter_staleness_filter(staleness, reporter, host_type_filter):
     return staleness_conditions
 
 
-def _staleness_filter(staleness: List[str], host_type_filter: Set[str]) -> List:
-    staleness_obj = serialize_staleness_to_dict(get_staleness_obj())
+def _staleness_filter(staleness: List[str], host_type_filter: Set[str], identity=None) -> List:
+    if identity:
+        staleness_obj = serialize_staleness_to_dict(get_staleness_obj(identity))
+    else:
+        staleness_obj = serialize_staleness_to_dict(get_staleness_obj())
```
kruai, line 160: `get_staleness_obj()` already has handling for `identity=None`, and that's its default value, so:
```suggestion
    staleness_obj = serialize_staleness_to_dict(get_staleness_obj(identity))
```

api/filtering/db_filters.py
```diff
@@ -321,7 +325,10 @@ def query_filters(
         sp_filter, host_type_filter = _system_profile_filter(filter)
         filters += sp_filter
     if staleness:
-        filters += _staleness_filter(staleness, host_type_filter)
+        if identity:
+            filters += _staleness_filter(staleness, host_type_filter, identity)
+        else:
+            filters += _staleness_filter(staleness, host_type_filter, identity)
```
kruai, line 331: Both paths do the same thing
```suggestion
        filters += _staleness_filter(staleness, host_type_filter, identity)
```

api/host_query_db.py
```diff
@@ -599,8 +599,11 @@ def get_hosts_to_export(
 ) -> Iterator[dict]:
     st_timestamps = staleness_timestamps()
     staleness = get_staleness_obj(identity)
+    staleness_states = ["fresh", "stale", "stale_warning"]
```
kruai, line 602: These are already defined by `ALL_STALENESS_STATES` in host_repository.py; we could use that instead of re-defining them

api/filtering/db_filters.py
```diff
@@ -283,6 +286,7 @@ def query_filters(
     filter: dict = None,
     rbac_filter: dict = None,
     order_by: str = None,
+    identity=None,
```
kruai, line 289: I don't love the idea of adding `identity` to this function, but... it seems it's needed for `get_staleness_obj()`, right?

kruai (changes requested):

api/host_query_db.py
jpramos123: Thanks, updated!

api/filtering/db_filters.py
jpramos123: Correct, that was needed or it was going to break when the `identity` was equal to `None`.

kruai: OK. Would probably be best if we then made `identity` required, and removed calls to `get_current_identity` within the flow, but that should be a separate PR; I'll play around with it after this goes in

app/queue/export_service.py
dehort: @jpramos123 I believe the SYSTEM identity type is generated when the user/system authenticates with a cert?  So I believe what you are seeing is related to this https://issues.redhat.com/browse/RHCLOUD-34799

kruai (approved):

app/queue/export_service.py
jpramos123: @dehort Thanks! That is correct!

@thearifismail Can we move on as this is an open issue on export-service side?

Thanks

tests/test_export_service.py
```diff
@@ -171,3 +175,45 @@ def test_handle_rbac_prohibited(mock_post, subtests, flask_app, db_create_host,
...
+            "immutable_time_to_delete": 1,
+        }
+
+        db_create_staleness_culling(**CUSTOM_STALENESS_DELETE_CONVENTIONAL_IMMUTABLE)
+        db_create_host()
+
+        # This is not ideal, but there no other way that I
+        # found to not use time.sleep
+
+        time.sleep(1)
```
kruai, line 200: We do need to avoid using this whenever possible, since each one adds a second to the runtime of the unit test suite. I know you left a comment above this, though, and I suppose we can address it with [RHINENG-3408](https://issues.redhat.com/browse/RHINENG-3408)...

tests/test_export_service.py
```diff
@@ -171,3 +175,45 @@ def test_handle_rbac_prohibited(mock_post, subtests, flask_app, db_create_host,
...
+        identity = Identity(USER_IDENTITY)
+        host_list = get_host_list(
+            identity=identity, exportFormat="json", rbac_filter=None, inventory_config=inventory_config
+        )
+
+        assert len(host_list) == 0
+
+
+@mock.patch("requests.Session.post", autospec=True)
+def test_export_one_host(mock_post, flask_app, db_create_host, db_create_staleness_culling, mocker):
```
kruai, line 210: It doesn't look like `@mock.patch`, `mock_post`, or `mocker` are used in this test

tests/test_export_service.py
```diff
@@ -171,3 +175,45 @@ def test_handle_rbac_prohibited(mock_post, subtests, flask_app, db_create_host,
                 mock_post.return_value.status_code = 202
                 resp = handle_export_message(message=export_message, inventory_config=config)
                 assert resp is False
+
+
+@mock.patch("requests.Session.post", autospec=True)
+def test_do_not_export_culled_hosts(mock_post, flask_app, db_create_host, db_create_staleness_culling, mocker):
```
kruai, line 181: It doesn't look like `@mock.patch`, `mock_post`, or `mocker` are used in this test

kruai (changes requested):

app/queue/export_service.py
```diff
@@ -42,10 +46,37 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+    host_data = list(
+        get_hosts_to_export(
+            identity,
+            export_format=exportFormat,
+            rbac_filter=rbac_filter,
+            batch_size=inventory_config.export_svc_batch_size,
+        )
+    )
+
+    return host_data
```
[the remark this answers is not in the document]
jpramos123: I would like to keep it like this as the `get_hosts_to_exports` is serializing the hosts using `yield`, at least we optimize that portion of the code using the generator.
Does it make sense?

app/queue/export_service.py
```diff
@@ -42,10 +48,43 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
     }
 
     rbac_request_headers = {
-        IDENTITY_HEADER: export_svc_data["data"]["resource_request"]["x_rh_identity"],
+        IDENTITY_HEADER: x_rh_identity,
         REQUEST_ID_HEADER: str(exportUUID),
```
kruai, line 52: Just noting that the `exportUUID` type hint says that it's already a str. If it's actually a UUID, the type hint should be updated; if it's a str, then the cast is not necessary

app/queue/export_service.py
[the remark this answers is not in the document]
kruai: I don't think I quite understand... `get_hosts_to_exports` is returning a generator, which is great. The part that confuses me is that it's always immediately converted into a list, which loads the whole dataset into memory anyway. It feels like the original intent was to feed the generator into a chunk-encoded POST request.
Anyway, I suppose this isn't strictly related to exporting culled hosts, so it can be a follow-up or something like that. Seems closer to [RHINENG-11833](https://issues.redhat.com/browse/RHINENG-11833).

app/queue/export_service.py
```diff
@@ -42,10 +48,43 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
     }
 
+    return rbac_request_headers, request_headers
+
+
+def get_host_list(identity: Identity, exportFormat: str, rbac_filter: dict, inventory_config: Config):
```
kruai, line 58:
```suggestion
def get_host_list(identity: Identity, exportFormat: str, rbac_filter: dict, inventory_config: Config) -> List[dict]:
```

app/queue/export_service.py
```diff
@@ -42,10 +48,43 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+
+
+@metrics.create_export_processing_time.time()
+def create_export(
+    export_svc_data: dict,
+    base64_x_rh_identity: str,
+    inventory_config: Config,
+    operation_args={},
+    rbac_filter: dict = {},
+):
```
kruai, line 78:
```suggestion
) -> bool:
```

app/queue/export_service.py
```diff
@@ -152,7 +184,7 @@ def _handle_export_response(response: Response, exportUUID: UUID, exportFormat:
         logger.info(f"{response.text} for export ID {str(exportUUID)} in {exportFormat.upper()} format")
 
 
-def _format_export_data(data, exportFormat):
+def _format_export_data(data: dict, exportFormat: str):
```
kruai, line 187:
```suggestion
def _format_export_data(data: dict, exportFormat: str) -> str:
```

Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: init test run_inv_mq_service
 
-IDENTITY_HEADER="eyJpZGVudGl0eSI6IHsiYWNjb3VudF9udW1iZXIiOiAiYWNjb3VudDEyMyIsICJvcmdfaWQiOiAiNTg5NDMwMCIsICJ0eXBlIjogIlVzZXIiLCAidXNlciI6IHsiaXNfb3JnX2FkbWluIjogdHJ1ZSwgInVzZXJuYW1lIjogImZyZWQifSwgImludGVybmFsIjogeyJvcmdfaWQiOiAib3JnMTIzIn19fQ=="
+IDENTITY_HEADER="eyJpZGVudGl0eSI6IHsiYWNjb3VudF9udW1iZXIiOiAiYWNjb3VudDEyMyIsICJvcmdfaWQiOiAiNTg5NDMwMCIsICJ0eXBlIjogIlVzZXIiLCAiYXV0aF90eXBlIjogImJhc2ljLWF1dGgiLCAidXNlciI6IHsiaXNfb3JnX2FkbWluIjogdHJ1ZSwgInVzZXJuYW1lIjogImZyZWQifSwgImludGVybmFsIjogeyJvcmdfaWQiOiAib3JnMTIzIn19fQ=="
```
thearifismail, line 3: what is the reason for adding `"auth_type": "basic-auth "` to Identity?  Why not  any of the following:
```
    BASIC = "basic-auth"
    CERT = "cert-auth"
    JWT = "jwt-auth"
    UHC = "uhc-auth"
    SAML = "saml-auth"
    X509 = "X509"
```

app/queue/export_service.py
```diff
@@ -23,17 +25,21 @@
...
 
 
-@metrics.create_export_processing_time.time()
-def create_export(export_svc_data, org_id, inventory_config, operation_args={}, rbac_filter={}):
-    identity = create_mock_identity_with_org_id(org_id)
-
-    metrics.create_export_count.inc()
-    logger.info("Creating export for HBI")
-
+def extract_export_svc_data(export_svc_data: dict):
```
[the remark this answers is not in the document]
thearifismail: @kruai suggestion had 4 members in the Tuple. I added one more because because this function returns 5 attributes.

app/queue/export_service.py
```diff
@@ -23,17 +25,21 @@
...
+    x_rh_identity = export_svc_data["data"]["resource_request"]["x_rh_identity"]
+
+    return exportFormat, exportUUID, applicationName, resourceUUID, x_rh_identity
+
+
+def build_headers(x_rh_identity: str, exportUUID: str, inventory_config: Config, exportFormat: str):
+    rbac_request_headers = {
+        IDENTITY_HEADER: x_rh_identity,
+        REQUEST_ID_HEADER: str(exportUUID),
+    }
```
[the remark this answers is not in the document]
thearifismail: yes `rbac_request_headers` defined twice.

app/queue/export_service.py
```diff
@@ -42,10 +48,43 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+            rbac_filter=rbac_filter,
+            batch_size=inventory_config.export_svc_batch_size,
+        )
+    )
+
+    return host_data
+
+
+@metrics.create_export_processing_time.time()
+def create_export(
```
thearifismail, line 72: `create_export()` has 4 return statements sprinkled throughout the function.  This is confusing.  It is preferred to have only one `return` statement. So please provide a variable and set its value to True or False replacing the return statements, and return the value at the end.

tests/helpers/export_service_utils.py
```diff
@@ -89,10 +89,10 @@ def create_export_message_mock():
                     "resource": "urn:redhat:application:inventory:export:systems",
                     "uuid": "2844f3a1-e047-45b1-b0ce-fb9812ad6a6f",
                     "x-rh-identity": (
-                        "eyJpZGVudGl0eSI6IHsib3JnX2lkIjogIjU4OTQzMD"
-                        "AiLCAidHlwZSI6ICJVc2VyIiwgImF1dGhfdHlwZSI6"
-                        "ICJiYXNpYy1hdXRoIiwgInVzZXIiOiB7ImVtYWlsIj"
-                        "ogImpyYW1vc0ByZWRoYXQuY29tIiwgImZpcnN0X25hbWUiOiAidGVzdCJ9fX0="
+                        "eyJpZGVudGl0eSI6IHsib3JnX2lkIjogInRlc3QiL"
```
thearifismail, line 92: linter gone mad... Is it confusing or I am the only one find it confusing?

Makefile
jpramos123: I'm using `basic-auth` which is the first in the list you sent

app/queue/export_service.py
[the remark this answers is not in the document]
jpramos123: Agreed.

Would be ideal to address this specifically there, I will start a follow up PR so we can discuss this optimization there.

Thanks

tests/helpers/export_service_utils.py
jpramos123: Tried to make it prettier.

app/queue/export_service.py
jpramos123: I don't like this idea, because, depending where we are in the code, thing are going to be processed unnecessarily.

Take this example:

```
    export_service_endpoint = inventory_config.export_service_endpoint

    if not allowed:
        session = Session()
        request_url = _build_export_request_url(
            export_service_endpoint, exportUUID, applicationName, resourceUUID, "error"
        )
        _handle_export_error(
            "You don't have the permission to access the requested resource.",
            "403",
            request_url,
            session,
            request_headers,
            exportUUID,
            exportFormat,
        )
        return False

    session = Session()
```

I don't want the code to continue to process without necessity. So the `return` statements are based on the code behavior.

Thanks

thearifismail: if `if not allowed: ` has a corresponding `else` block with `try except` in it, then there should be no unnecessary execution.

jpramos123: Done, added the else statement and the variable. Good idea for debugging the code in future...

thearifismail: with the new latest implementation, `if not allowed` block is executed the function returns without closing the sesssion.  Here is what I think should be done.

1. initialize session in once before checking if allowed or not.
2. Using finally for closing the session is good but it requires `try{}` block.
3. first start `try` block, then `if allowed` and performa the work.  Current `if not allowed` should be done in `else` block inside the `try` block.  Then handle exception and finally `finally` to wrap up

jpramos123: 1. Thanks, good catch! Added the `session.close()`
2. It was a matter of adding `session.close()`  to the `if not allowed`
3. The way it is now is cleaner for reading, as it just validate the `if not` statement, otherwise continues the flow.

Thanks

thearifismail (dismissed):

tests/test_export_service.py
```diff
@@ -171,3 +175,43 @@ def test_handle_rbac_prohibited(mock_post, subtests, flask_app, db_create_host,
                 mock_post.return_value.status_code = 202
                 resp = handle_export_message(message=export_message, inventory_config=config)
                 assert resp is False
+
+
+def test_do_not_export_culled_hosts(flask_app, db_create_host, db_create_staleness_culling):
+    with flask_app.app.app_context():
+        inventory_config = flask_app.app.config.get("INVENTORY_CONFIG")
```
kruai, line 182: We have an `inventory_config` fixture that automatically handles this

tests/test_export_service.py
```diff
@@ -171,3 +175,43 @@ def test_handle_rbac_prohibited(mock_post, subtests, flask_app, db_create_host,
...
+        host_list = get_host_list(
+            identity=identity, exportFormat="json", rbac_filter=None, inventory_config=inventory_config
+        )
+
+        assert len(host_list) == 0
+
+
+def test_export_one_host(flask_app, db_create_host):
+    with flask_app.app.app_context():
+        inventory_config = flask_app.app.config.get("INVENTORY_CONFIG")
```
kruai, line 210: Same as above - you should be able to use the `inventory_config` fixture

kruai (commented):
Looks good, just got a tiny suggestion for the tests

kruai (approved):

PR #1929, opened 2024-09-16
fix(RHINENG-11863): User without host read permission is able to export hosts

jpramos123:
# Overview

This PR is being created to address [RHINENG-11863](https://issues.redhat.com/browse/RHINENG-11863).

This fixes the issue that was allowing user without host-read to export hosts from inventory. Now it raises an exception when the user does not have the correct permission and fails the export request. 

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/queue/export_service.py
+++ b/app/queue/export_service.py
@@ -32,29 +32,39 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={}, 
     applicationName = export_svc_data["data"]["resource_request"]["application"]
     resourceUUID = export_svc_data["data"]["resource_request"]["uuid"]
 
-    rbac_request_headers = {
-        IDENTITY_HEADER: export_svc_data["data"]["resource_request"]["x_rh_identity"],
-        REQUEST_ID_HEADER: str(exportUUID),
-    }
-
     # x-rh-exports-psk must be an env variable
     request_headers = {
         "x-rh-exports-psk": inventory_config.export_service_token,
         "content-type": HEADER_CONTENT_TYPE[exportFormat.lower()],
     }
+
+    rbac_request_headers = {
+        IDENTITY_HEADER: export_svc_data["data"]["resource_request"]["x_rh_identity"],
+        REQUEST_ID_HEADER: str(exportUUID),
+    }
+
+    allowed, rbac_filter = get_rbac_filter(
+        RbacResourceType.HOSTS, RbacPermission.READ, identity=identity, rbac_request_headers=rbac_request_headers
+    )
+
+    export_service_endpoint = inventory_config.export_service_endpoint
+
+    if not allowed:
+        session = Session()
+        request_url = _build_export_request_url(
+            export_service_endpoint, exportUUID, applicationName, resourceUUID, "error"
+        )
+        _handle_export_error(
+            "You don't have the permission to access the requested resource.",
+            "403",
+            request_url,
+            session,
+            request_headers,
+        )
+        return False
+
     session = Session()
     try:
-        export_service_endpoint = inventory_config.export_service_endpoint
-        request_url = f"{export_service_endpoint}/app/export/v1/{exportUUID}/{applicationName}/{resourceUUID}/upload"
-
-        session.mount(request_url, HTTPAdapter(max_retries=3))
-
-        logger.info(f"Trying to get data for org_id: {identity.org_id}")
-
-        rbac_filter = get_rbac_filter(
-            RbacResourceType.HOSTS, RbacPermission.READ, identity=identity, rbac_request_headers=rbac_request_headers
-        )
-
         # create a generator with serialized host data
         host_data = list(
             get_hosts_to_export(
@@ -64,6 +74,14 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={}, 
                 batch_size=inventory_config.export_svc_batch_size,
             )
         )
+
+        request_url = _build_export_request_url(
+            export_service_endpoint, exportUUID, applicationName, resourceUUID, "upload"
+        )
+
+        session.mount(request_url, HTTPAdapter(max_retries=3))
+
+        logger.info(f"Trying to get data for org_id: {identity.org_id}")
 
         if host_data:
             logger.debug(f"Trying to upload data using URL:{request_url}")
@@ -77,8 +95,8 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={}, 
             return True
         else:
             logger.debug(f"No data found for org_id: {identity.org_id}")
-            request_url = (
-                f"{export_service_endpoint}/app/export/v1/{exportUUID}/{applicationName}/{resourceUUID}/error"
+            request_url = _build_export_request_url(
+                export_service_endpoint, exportUUID, applicationName, resourceUUID, "error"
             )
             response = session.post(
                 url=request_url,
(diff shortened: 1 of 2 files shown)


app/queue/export_service.py
```diff
@@ -51,19 +53,21 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+                get_hosts_to_export(
+                    identity,
+                    export_format=exportFormat,
+                    rbac_filter=rbac_filter,
+                    batch_size=inventory_config.export_svc_batch_size,
+                )
             )
-        )
+        else:
+            abort(HTTPStatus.FORBIDDEN)
```
kruai, line 70: `create_export()` is MQ, not API. Doesn't make sense to return HTTP 403 to a kafka consumer

app/queue/export_service.py
```diff
@@ -87,17 +91,34 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+            session,
+            request_headers,
+        )
     except Exception as e:
-        logger.error(e)
-        request_url = f"{export_service_endpoint}/app/export/v1/{exportUUID}/{applicationName}/{resourceUUID}/error"
-        response = session.post(
-            url=request_url, headers=request_headers, data=json.dumps({"message": str(e), "error": 500})
+        return _handle_export_error(
+            e, 500, export_service_endpoint, exportUUID, applicationName, resourceUUID, session, request_headers
```
kruai, line 107: A few notes/questions for this line:
- Why the hardcoded 500 for the status code?
- It would be cleaner to just pass in the `request_url` instead of all of its pieces separately.
- It seems a bit odd to return `_handle_export_error()` when it's always gonna be False. Honestly instead of returning a boolean, I think it would make more sense for `create_export` to just raise an exception if it fails

kruai (changes requested):

app/queue/export_service.py
```diff
@@ -87,17 +91,34 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
         session.close()
 
 
+def _handle_export_error(
+    e, status_code, endpoint, exportUUID, applicationName, resourceUUID, session, request_headers
+):
+    logger.error(e)
+    request_url = f"{endpoint}/app/export/v1/{exportUUID}/{applicationName}/{resourceUUID}/error"
+    session.post(url=request_url, headers=request_headers, data=json.dumps({"message": str(e), "error": status_code}))
+    return False
```
vkrizan, line 119: Minor: This return value is suppose to be an outcome of the parent, hence it shouldn't be part of this handle function (this one should be kept without return value).

app/queue/export_service.py
```diff
@@ -51,19 +53,21 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
         )
-
-        # create a generator with serialized host data
-        host_data = list(
-            get_hosts_to_export(
-                identity,
-                export_format=exportFormat,
-                rbac_filter=rbac_filter,
-                batch_size=inventory_config.export_svc_batch_size,
+        if allowed:
```
vkrizan, line 59: The allowed should be checked prior to getting into the `try-except` block of the export, and before generating a session to the export endpoint service: right after `rbac_request_headers`.

app/queue/export_service.py
jpramos123: My idea was to use it to raise the exception and use the status code/message that is sent to the export-service API via POST request. I can make it by hand as well.

Do you think would be clear/correct this way?

Thanks

app/queue/export_service.py
jpramos123: Thanks I will try to figure out something better.

app/queue/export_service.py
jpramos123: Built it by hand!

Let me know if it fits

app/queue/export_service.py
```diff
@@ -32,39 +32,60 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+                get_hosts_to_export(
+                    identity,
+                    export_format=exportFormat,
+                    rbac_filter=rbac_filter,
+                    batch_size=inventory_config.export_svc_batch_size,
+                )
             )
+        else:
+            raise Exception("You don't have the permission to access the requested resource.")
+    except Exception as e:
```
vkrizan, line 65: This is not very ideal, as this catches (almost) everything and claims it to be a `403 Forbidden` (as per `status_code` passed to the handler). Move the check of `allowed` outside of the `try-except`, and handle other errors with differently. If an error can be handled still within the same function, then it does not make sense to throw it and then catch it later on (at least not how it is done in this case).

app/queue/export_service.py
```diff
@@ -32,38 +32,56 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
-        # create a generator with serialized host data
-        host_data = list(
-            get_hosts_to_export(
-                identity,
-                export_format=exportFormat,
-                rbac_filter=rbac_filter,
-                batch_size=inventory_config.export_svc_batch_size,
-            )
+    # create a generator with serialized host data
+    host_data = list(
```
vkrizan, line 67: Consider keeping this under the try-except block, in case the db fetch or serialization fail it would be reported back.

jpramos123: Done!

vkrizan (approved):
Thank you!

app/queue/export_service.py
```diff
@@ -88,16 +106,26 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
 
+def _build_export_request_url(
+    export_service_endpoint, exportUUID, applicationName, resourceUUID, request_type: str
+) -> str:
+    return f"{export_service_endpoint}/app/export/v1/{exportUUID}/{applicationName}/{resourceUUID}/{request_type}"
+
+
+def _handle_export_error(e, status_code, request_url, session, request_headers):
+    logger.error(e)
+    session.post(url=request_url, headers=request_headers, data=json.dumps({"message": str(e), "error": status_code}))
```
kruai, line 126: This POST's response isn't considered; seems like it should be evaluated using `_handle_export_response`

app/queue/export_service.py
```diff
@@ -88,16 +106,26 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+            export_service_endpoint, exportUUID, applicationName, resourceUUID, "error"
         )
+        _handle_export_error(e, 500, request_url, session, request_headers)
         return False
     finally:
         session.close()
 
 
+def _build_export_request_url(
+    export_service_endpoint, exportUUID, applicationName, resourceUUID, request_type: str
```
kruai, line 119: this could use some more type hints

app/queue/export_service.py
```diff
@@ -88,16 +106,26 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
...
+) -> str:
+    return f"{export_service_endpoint}/app/export/v1/{exportUUID}/{applicationName}/{resourceUUID}/{request_type}"
+
+
+def _handle_export_error(e, status_code, request_url, session, request_headers):
```
kruai, line 124: this could use type hints

kruai (changes requested):
A few small suggestions. I also think `create_export()` could be DRY'd a bit more... I'm gonna take a stab at it to make sure I'm not off base, though, and I don't wanna block on it

app/queue/export_service.py
jpramos123: Done

app/queue/export_service.py
jpramos123: Done

app/queue/export_service.py
jpramos123: Done

kruai (approved):

PR #1939, opened 2024-09-19
fix(RHINENG-11863): Fix error handling

jpramos123:
# Overview

This PR is being created to address [RHINENG-11863](https://issues.redhat.com/browse/RHINENG-11863).

It fixes an error handling that was failing in the code.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/queue/export_service.py
+++ b/app/queue/export_service.py
@@ -2,7 +2,6 @@ import json
 from http import HTTPStatus
 from typing import List
 from typing import Tuple
-from typing import Union
 from uuid import UUID
 
 from requests import Response
@@ -99,7 +98,7 @@ def create_export(
         )
         _handle_export_error(
             "You don't have the permission to access the requested resource.",
-            "403",
+            403,
             request_url,
             session,
             request_headers,
@@ -147,7 +146,7 @@ def create_export(
         request_url = _build_export_request_url(
             export_service_endpoint, exportUUID, applicationName, resourceUUID, "error"
         )
-        _handle_export_error(e, 500, request_url, session, request_headers, exportUUID, exportFormat)
+        _handle_export_error(str(e), 500, request_url, session, request_headers, exportUUID, exportFormat)
         export_created = False
     finally:
         session.close()
@@ -161,8 +160,8 @@ def _build_export_request_url(
 
 
 def _handle_export_error(
-    error_message: Union[str, Exception],
-    status_code: Union[str, int],
+    error_message: str,
+    status_code: int,
     request_url: str,
     session: Session,
     request_headers: dict,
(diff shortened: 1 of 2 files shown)


FabriciaDinizRH (dismissed):

app/queue/export_service.py
```diff
@@ -59,7 +59,7 @@ def create_export(export_svc_data, org_id, inventory_config, operation_args={},
         )
         _handle_export_error(
             "You don't have the permission to access the requested resource.",
-            "403",
+            403,
```
kruai, line 62: With this change, the `status_code` arg in `_handle_export_error` can just be type "int" rather than the Union (which I didn't notice before).
Ideally, I think its `error_message` arg should be type "str" as well, though you'll have to pass in `str(e)` instead of `e` for the instance where the whole error is passed in

kruai (commented):
I don't have a problem with these changes; I do have a suggestion on how to simplify/clean up a bit if you wanna go one step further

app/queue/export_service.py
jpramos123: I like this idea, and made the changes!

Thanks

kruai (dismissed):

thearifismail (commented):
``

app/queue/queue.py
```diff
@@ -468,10 +468,15 @@ def handle_export_message(message, inventory_config):
         logger.debug("parsed_message: %s", validated_msg)
         base64_x_rh_identity = validated_msg["data"]["resource_request"]["x_rh_identity"]
 
-        if create_export(validated_msg, base64_x_rh_identity, inventory_config):
-            metrics.export_service_message_handler_success.inc()
-            return True
-        else:
+        try:
```
thearifismail, line 471: Same comment as PR 1926 about multiple return statements.  A function should not have more than one return statement.

jpramos123: Thanks for this.

Already addressed it

thearifismail: hmmm, did I miss this before?
There should be only return statement.  I apologize for being a stickler here.  Using `multiple return` and `goto` statements has always been discouraged.  
One reason I see is the use of `finally`.  I wonder if `finally` is really needed here.  If you must, then I suggest expanding the try{} block to include the first `if` statement so that finally is available to all execution paths.

kruai (approved):

thearifismail (approved):

kruai (approved):

PR #1962, opened 2024-09-27
fix(RHINENG-12212): fix host_url in notification

FabriciaDinizRH:
# Overview

This PR is being created to address [RHINENG-12212](https://issues.redhat.com/browse/RHINENG-12212).

- sets `CONSOLEDOT_HOSTNAME` for creating the host url

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ BYPASS_UNLEASH="true"
 
 # If you want to use the legacy prefix, otherwise don't set PATH_PREFIX
 # PATH_PREFIX="/r/insights/platform"
+# For testing notifications in stage/ephemeral, set this to the appropriate url
+# PLATFORM_HOSTNAME="https://console.redhat.com"
 
 APP_NAME="inventory"
 INVENTORY_DB_USER="insights"
--- a/README.md
+++ b/README.md
@@ -44,8 +44,6 @@ BYPASS_UNLEASH="true"
 
 # If you want to use the legacy prefix, otherwise don't set PATH_PREFIX
 # PATH_PREFIX="/r/insights/platform"
-# For testing notifications in stage/ephemeral, set this to the appropriate url
-# PLATFORM_HOSTNAME="https://console.redhat.com"
 
 APP_NAME="inventory"
 INVENTORY_DB_USER="insights"
--- a/app/config.py
+++ b/app/config.py
@@ -115,8 +115,6 @@ class Config:
             self.unleash_url = os.getenv("UNLEASH_URL")
             self.unleash_token = os.getenv("UNLEASH_TOKEN")
 
-        self.platform_hostname = os.getenv("PLATFORM_HOSTNAME", "https://console.redhat.com")
-
     def non_clowder_config(self):
         self.is_clowder = False
         self.metrics_port = 9126
@@ -151,8 +149,6 @@ class Config:
         self._cache_host = os.environ.get("CACHE_HOST", "localhost")
         self._cache_port = os.environ.get("CACHE_PORT", "6379")
         self.export_service_token = os.environ.get("EXPORT_SERVICE_TOKEN", "testing-a-psk")
-
-        self.platform_hostname = os.getenv("PLATFORM_HOSTNAME", "https://console.redhat.com")
 
     def days_to_seconds(self, n_days):
         factor = 86400
@@ -207,6 +203,9 @@ class Config:
 
         self.prometheus_pushgateway = os.environ.get("PROMETHEUS_PUSHGATEWAY", "localhost:9091")
         self.kubernetes_namespace = os.environ.get("NAMESPACE")
+
+        self.consoledot_hostname = os.getenv("CONSOLEDOT_HOSTNAME", "localhost")
+        self.base_ui_url = f"https://{self.consoledot_hostname}/insights/inventory"
 
         self.kafka_ssl_configs = {
             "security.protocol": self.kafka_security_protocol,
--- a/app/config.py
+++ b/app/config.py
@@ -115,6 +115,8 @@ class Config:
             self.unleash_url = os.getenv("UNLEASH_URL")
             self.unleash_token = os.getenv("UNLEASH_TOKEN")
 
+        self.platform_hostname = os.getenv("PLATFORM_HOSTNAME", "https://console.redhat.com")
+
     def non_clowder_config(self):
         self.is_clowder = False
         self.metrics_port = 9126
@@ -149,6 +151,8 @@ class Config:
         self._cache_host = os.environ.get("CACHE_HOST", "localhost")
         self._cache_port = os.environ.get("CACHE_PORT", "6379")
         self.export_service_token = os.environ.get("EXPORT_SERVICE_TOKEN", "testing-a-psk")
+
+        self.platform_hostname = os.getenv("PLATFORM_HOSTNAME", "https://console.redhat.com")
 
     def days_to_seconds(self, n_days):
         factor = 86400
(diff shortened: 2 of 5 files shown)


jpramos123 (changes requested):
Please, review the approach in this PR. I don't feel comfortable introducing a bug that we can already solve in this PR.

Thanks

app/queue/notifications.py
```diff
@@ -26,6 +25,8 @@
 )
 EventSeverity = Enum("EventSeverity", ("warning", "error", "critical"))
 
+PROD_URL = "https://console.redhat.com/insights/inventory/"
```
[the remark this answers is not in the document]
FabriciaDinizRH: @jpramos123 this is going to fix the existing behavior for the customers. After that is fixed for them I can take a bit more time to figure out about the env variable and setting it everywhere it is needed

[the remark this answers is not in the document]
thearifismail: I aggree with @jpramos123.  Furhermore, target environment URL called `PROD_URL` does not look right.  Type of environment and its URL should be provided using environment variables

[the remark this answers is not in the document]
FabriciaDinizRH: Ok, so I'll wait for advisor's response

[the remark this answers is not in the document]
FabriciaDinizRH: @kruai should I change the config variable name in this PR to `PLATFORM_HOSTNAME`?

app/config.py
```diff
@@ -115,6 +115,8 @@ def topic(t):
             self.unleash_url = os.getenv("UNLEASH_URL")
             self.unleash_token = os.getenv("UNLEASH_TOKEN")
 
+        self.consoledot_hostname = os.getenv("CONSOLEDOT_HOSTNAME", "https://console.redhat.com")
```
jpramos123, line 118: I would rather set the default value to the one that is used `locally` or in `ephemeral`.
It should be `localhost`.

We can set the environment variable in stage and prod files in gitlab!

app/config.py
```diff
@@ -150,6 +152,8 @@ def non_clowder_config(self):
         self._cache_port = os.environ.get("CACHE_PORT", "6379")
         self.export_service_token = os.environ.get("EXPORT_SERVICE_TOKEN", "testing-a-psk")
 
+        self.consoledot_hostname = os.getenv("CONSOLEDOT_HOSTNAME", "https://console.redhat.com")
```
jpramos123, line 155: I would rather set the default value to the one that is used locally or in ephemeral.
It should be localhost.

We can set the environment variable in stage and prod files in gitlab!

deploy/clowdapp.yml
```diff
@@ -1381,6 +1387,8 @@ parameters:
   value: host-inventory
 - name: PATH_PREFIX
   value: api
+- name: CONSOLEDOT_HOSTNAME
+  value: 'https://console.redhat.com'
```
jpramos123, line 1391: I would rather set the default value to the one that is used locally or in ephemeral.
It should be localhost.

We can set the environment variable in stage and prod files in gitlab!

jpramos123 (changes requested):
Looks good, but I have a suggestion.

Thanks for the PR

README.md
```diff
@@ -44,6 +44,8 @@ BYPASS_UNLEASH="true"
 
 # If you want to use the legacy prefix, otherwise don't set PATH_PREFIX
 # PATH_PREFIX="/r/insights/platform"
+# For testing notifications in stage/ephemeral, set this to the appropriate url
```
kruai, line 47: These instructions are for local setup; do we care about the hostname used for notifications produced locally? If we do, then my guess is that the hostname should be set to `localhost` or `0.0.0.0:8080`, since that's what we've configured for the local env.
(Note that the value for Stage is automatically configured, so no need to specify that here).

README.md
```diff
@@ -44,6 +44,8 @@ BYPASS_UNLEASH="true"
...
+# CONSOLEDOT_HOSTNAME="https://console.redhat.com"
```
kruai: If we do want developers to configure this field for their local setup, then this should be uncommented and probably put at the end of the list of env vars. Also, the protocol (`https://`) should be left off

deploy/clowdapp.yml
kruai: It only contains the hostname, not the protocol
```suggestion
  value: 'console.redhat.com'
```

app/config.py
FabriciaDinizRH: done

app/config.py
FabriciaDinizRH: done

deploy/clowdapp.yml
FabriciaDinizRH: done

README.md
jpramos123, line 47:
```suggestion
# Set the appropriate URL value  to be used in stage/prod environments. Default value is:
```

jpramos123 (changes requested):
Small change in the READM.md file

app/config.py
```diff
@@ -115,6 +115,8 @@ def topic(t):
             self.unleash_url = os.getenv("UNLEASH_URL")
             self.unleash_token = os.getenv("UNLEASH_TOKEN")
 
+        self.consoledot_hostname = os.getenv("CONSOLEDOT_HOSTNAME", "localhost")
+
```
kruai, line 119: I see this line in both `clowder_config()` and `non_clowder_config()`. We set this the same way regardless of whether we're using clowder, so this value should be set within `__init__()`.

Also, you can remove `get_base_url()` and just set it here to simplify a bit:
```
self.consoledot_hostname = os.getenv("CONSOLEDOT_HOSTNAME", "localhost")
self.base_ui_url = f"https://{self.consoledot_hostname}/insights/inventory"
```

app/config.py
```diff
@@ -150,6 +152,8 @@ def non_clowder_config(self):
         self._cache_port = os.environ.get("CACHE_PORT", "6379")
         self.export_service_token = os.environ.get("EXPORT_SERVICE_TOKEN", "testing-a-psk")
 
+        self.consoledot_hostname = os.getenv("CONSOLEDOT_HOSTNAME", "localhost")
+
```
kruai, line 156: Should be removed (see above)

kruai (changes requested):

README.md
FabriciaDinizRH: I see. I still struggle with things that go "under the hood". I wanted to let everything written out but I agree it can be left out here

FabriciaDinizRH: I'll take Asa's suggestion and remove this bit to make things simpler

kruai (dismissed):

deploy/clowdapp.yml
```diff
@@ -1381,6 +1387,8 @@ parameters:
   value: host-inventory
 - name: PATH_PREFIX
   value: api
+- name: CONSOLEDOT_HOSTNAME
+  value: 'console.redhat.com'
```
jpramos123, line 1391: Shouldn't me localhost as well here? 
This must be overwritten in stage/prod deployment files

```suggestion
  value: 'localhost'
```

jpramos123 (changes requested):
Last review

deploy/clowdapp.yml
kruai: Oh, for some reason I thought it was already "localhost". Yes I agree that it should be

thearifismail (dismissed):

kruai (approved):

jpramos123 (approved):

PR #1968, opened 2024-10-01
Revert "feat(rhineng-12237): Update container files for ubi9 minimal"

chambridge:
Reverts RedHatInsights/insights-host-inventory#1944


kruai (approved):
For context, there is an issue with psycopg2 that occurs on RHEL9 when using FIPS. We need to roll back to UBI8 temporarily until that issue is resolved. We plan to upgrade back to UBI9 after that

PR #2050, opened 2024-11-04
chore(RHINENG-13421): Migrate to ruff

fstavela:
# Overview

This PR is being created to address [RHINENG-13421](https://issues.redhat.com/browse/RHINENG-13421).

Ruff can fully replace flake8, black, pyupgrade, and more. For now I enabled a few basic rules which will make sure that everything that was checked by our previous linters will be checked by ruff too and I also added a few more which make our code even better. We can continue to enable more and more rules and make the code more beautiful, but I think it should be handled in separate PRs, because this one is pretty big already. For example, one thing that I noticed is that we have a lot of unused parameters in our functions, so I created a followup Jira to get rid of them: [RHINENG-14059](https://issues.redhat.com/browse/RHINENG-14059)

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


lib/check_org.py
```diff
@@ -21,10 +20,9 @@ def check_org_id(result):
         data = result
     if isinstance(data, dict):
         message = f"Response contained data for another org_id. Path={request.path}, parameters={request.args}."
-        if "org_id" in data:
-            if data.get("org_id") and (data.get("org_id") != current_identity.org_id):
-                logger.error(message)
-                abort(HTTPStatus.FORBIDDEN)
+        if "org_id" in data and data["org_id"] != current_identity.org_id:
+            logger.error(message)
+            abort(HTTPStatus.FORBIDDEN)
```
fstavela, line 25: Please look carefully at this during your PR reviews, because this is also changing the behavior a little bit. I removed the `if data.get("org_id")` check. What this means is that if the response includes `org_id` but it is `None` or `""` (empty string) or `0` or anything that converts to `False` bool value, this would pass before, but it will not pass with this change. In my opinion it is better this way, because having such a weird value on `org_id` field may indicate some serious issues and we wouldn't know what org this data belongs to, but I wanted to check with you all if you don't know about any valid scenario where org_id could have this value and if you also think it is safe to make this change.

kruai (approved):
I see a lot of good changes in here, including some best practices I had forgotten about (like `raise from`). Also, thanks for fixing that typo - I assume that was you and not the linter :)

thearifismail (approved):
LGTM

PR #2069, opened 2024-11-12
feat(rhineng-12924): Proposed mechanism for replicated table migrations

chambridge:
# Overview

This PR is being created to address [RHINENG-12924](https://issues.redhat.com/browse/RHINENG-12924).
* Define migration runner script that would be utilized by the application team
* Create app migrations folder that will house the SQL files for a migration
* Capture migration strategy in a README document in the new folder

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


thearifismail:
@chambridge Reference to this working should be provided in the main/first README.  Should this wait until logical replication has been enabled by the host-inventory first?

chambridge:
@fstavela I've updated the main README to link to this documentation. I've also added a make command for generating the SQL script.

I like what you are proposing for something more in depth. I may check in a SQL file that only performs a read of the current schema structure, just for testability. The app-interface MRs might be follow on items.

chambridge:
@thearifismail I've referenced this information in the top level README. I created the README file in the `app_migrations` directory initially as a placeholder mechanism as you cannot check-in empty directories to git and I like the argument that a README in the directory is better than a `.gitkeep` file.

I don't know of any reason why this would need to wait to go in.

thearifismail (dismissed):
LGTM!

thearifismail (commented):
What is provided looks good to me but finding hard to picture without having a migration script, i.e. running a migration script

fstavela (approved):
LGTM! Thank you for the makefile changes and the additional documentation.

README.md
```diff
@@ -350,6 +350,18 @@ Run this command to generate a new revision in `migrations/versions`
...
 ```
 
+When creating a new migration before merging your pull request make sure you have created a migration for replicated tables. If the migration affects tables that are being replicated the migration must be performed on the replicated tables first. For details on creating and performing migrations against replicated tables read the following [document](app_migrations/README.md).
+
+
+## Capturing the current HBI schema state for replicaiton subscribers
+When migrations are being made it makes sense to capture the HBI schema state for replication subscribers to simplify their procedure for onboarding or schema recreation. Run the following command to capture the updated schema state:
+
+```bash
+make gen_hbi_schema_dump
```
fstavela, line 360: We should mention that we need to install `postgresql16` to get the `pg_dump` command. We can add it at the [beginning of the README](https://github.com/RedHatInsights/insights-host-inventory?tab=readme-ov-file#pg_config) (the change is probably required only on Fedora and Debian docs)

fstavela (commented):
I had a couple of comments, but otherwise, it looks good, thanks!

Just to make sure I understand the migration procedure correctly. First, we need to create a new DB migration in `migrations/versions` by `make migrate_db` as we would normally do. Then, we can use this migration and the revs in the `make gen_offline_sql` to generate an SQL migration script for the apps. Then (this is what I'm not sure about) we have to delete the original migration file from `migrations/versions` before committing because otherwise the migration would be executed on the publisher DB in Stage before apps get a chance to upgrade their subscriber DBs? And once all of the subscribers ran the migration, we can add the file back to `migrations/versions`.

app_migrations/README.md
```diff
@@ -0,0 +1,42 @@
...
+Provide the dry run environment variable at the command line, in the case below we are using the `query_hbi_schema.sql` migration file present in the `app_migrations` directory.
+```
+INVENTORY_MIGRATION_DRYRUN="true" INVENTORY_MIGRATION_FILE="query_hbi_schema.sql" python inv_migration_runner.py
+```
+When this runs you will only see the SQL that would have been executed on the database.
+
+### Execution
+In the case below we are using the `query_hbi_schema.sql` migration file present in the `app_migrations` directory.
+```
+INVENTORY_MIGRATION_DRYRUN="true" INVENTORY_MIGRATION_FILE="query_hbi_schema.sql" python inv_migration_runner.py
```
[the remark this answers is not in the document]
chambridge: yeah. Copy paste 😏

README.md
chambridge: I've got a Mac. What should the update look like if you wouldn't mind?

chambridge: I pushed a couple changes.

fstavela: When I installed it on a Fedora I used `sudo dnf install postgresql16` to make sure I have v16, but I just tried that installing `postgresql` also installs v16 on Fedora 40, so it's OK as well.

fstavela (dismissed):
Looks good now, thanks.

inv_migration_runner.py
```diff
@@ -0,0 +1,78 @@
...
+def _init_db(config):
+    engine = create_engine(config.db_uri)
+    return sessionmaker(bind=engine)
+
+
+def _excepthook(logger, type, value, traceback):
+    logger.exception("Inventory migration failed", exc_info=value)
+
+
+def run(config, logger, session, shutdown_handler, application):
```
thearifismail, line 39: Is the config parameter still needed, if not called in this function?

inv_migration_runner.py
```diff
@@ -0,0 +1,78 @@
...
+        dry_run = os.environ.get("INVENTORY_MIGRATION_DRYRUN", "false")
+        migration_file = os.environ.get("INVENTORY_MIGRATION_FILE")
+        file_path = os.path.join(basedir, "app_migrations", migration_file)
+        if not os.path.isfile(file_path):
+            logger.warning(f"No migration file found for given {file_path}.")
+            return
+        with open(file_path) as file:
+            sql_statements = file.read()
+            logger.info(f"Executing the following SQL statements:\n{sql_statements}")
+            if dry_run.lower() != "true":
```
thearifismail, line 51: Is a `dry_run` really needed if it does not execute any SQL statements?  Shouldn't a dry run execute SQL statements but NOT commit them or roll them back.

app_migrations/hbi_schema_2024-11-15.sql
```diff
@@ -0,0 +1,357 @@
+--
```
thearifismail, line 1: Should this file be included?

app/config.py
```diff
@@ -78,7 +78,11 @@ def clowder_config(self):
         self.export_service_token = os.environ.get("EXPORT_SERVICE_TOKEN", "testing-a-psk")
 
         def topic(t):
-            return app_common_python.KafkaTopics[t].name if t else None
+            if t:
```
thearifismail, line 81: Is it related to this PR?

inv_migration_runner.py
chambridge: The SQL file selected could contain a COMMIT statement in it, so there would be no way of avoiding that. This is just an operational mechanism to perform the JOB and validate the SQL that would be run without running it. Allows you to prove out the deployment step and that the image of the container has the version of the SQL you intended to run.

app/config.py
chambridge: Yes. This ClowdJobInvocation fails if this isn't changed as no kafka topics are defined or needed for it.

app_migrations/hbi_schema_2024-11-15.sql
chambridge: The intention was to supply the existing HBI schema, yes.

thearifismail (approved):

PR #2103, opened 2024-12-02
chore(RHINENG-13957, RHINENG-13958) Squash migrations; migrate version table

kruai:
# Overview

This PR is being created to address [RHINENG-13957](https://issues.redhat.com/browse/RHINENG-13957).
Squashes all of our previous migrations into one. Since I gave it the same ID as the last migration, Stage and Prod will not try to upgrade to this new version. The new SQL dump modifies the schema creation to use `IF NOT EXISTS` and omits the owner/role-related directives, but is otherwise the same as `hbi_schema_2024-11-15.sql`.

This PR additionally migrates the alembic version table from the "public" schema to the "hbi" schema. It determines whether it needs to by checking the current revision on the desired ("hbi") schema; if that value does not exist, it migrates from the "public" schema. It does this by getting the current revision in the version table on the public schema, and stamping that to the version table in the hbi schema. After that, it runs migrations as normal. So, if the DB is spinning up from scratch, it switches to the hbi schema and runs the migration; if the DB is already on the latest revision, it switches to the hbi schema but does not run the migration (because it's already on that revision).

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


chambridge:
Does it make sense to do this squash before moving the alembic_versions table out of the public schema?

kruai:
> Does it make sense to do this squash before moving the alembic_versions table out of the public schema?

From what I can tell, it should be about the same amount of work regardless of what order we do this in. I might have missed something, though - what's your concern?

chambridge:
> > Does it make sense to do this squash before moving the alembic_versions table out of the public schema?
> 
> From what I can tell, it should be about the same amount of work regardless of what order we do this in. I might have missed something, though - what's your concern?

The end result is we want a fresh DB to start off with all data in the `hbi` schema. If we do this now, we'll need to do it again after migrating the alembic_versions table out of the public schema so that we don't have some data in both schemas and a migration happening to move the data out of the public schema.

kruai:
> > > Does it make sense to do this squash before moving the alembic_versions table out of the public schema?
> > 
> > 
> > From what I can tell, it should be about the same amount of work regardless of what order we do this in. I might have missed something, though - what's your concern?
> 
> The end result is we want a fresh DB to start off with all data in the `hbi` schema. If we do this now, we'll need to do it again after migrating the alembic_versions table out of the public schema so that we don't have some data in both schemas and a migration happening to move the data out of the public schema.

OK, I'll make this a draft and focus on the version table migration first.

chambridge:
Is there anything we need to update here: https://github.com/RedHatInsights/insights-host-inventory/blob/master/Makefile#L25-L28 so a future pg_dump has the output we want?

Or do we need to capture any updates in this section of the README? https://github.com/RedHatInsights/insights-host-inventory/blob/master/README.md#capturing-the-current-hbi-schema-state-for-replicaiton-subscribers

chambridge (approved):
LGTM

PR #2112, opened 2024-12-04
CHAINSAW-234 Add environments to schema

Tomboyo:
# Overview

Adds the rhsm.environments field to the system profile schema, per CONTRIBUTING guidelines of inventory-schemas. See https://github.com/RedHatInsights/inventory-schemas/pull/145. Let me know if I'm issing anything!

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


swagger/system_profile.spec.yaml
```diff
@@ -592,6 +592,15 @@ $defs:
             example: "8.1, 7.5, 9.9"
             type: string
             maxLength: 255
+          environments:
```
thearifismail, line 595: Can this attribute be renamed to indicate its purpose?  "environment" appears to generic and can raise questions in future, if not now?

swagger/system_profile.spec.yaml
```diff
@@ -592,6 +592,15 @@ $defs:
...
+            description: UUIDs of environments ("content templates") the system is subscribed to.
+            example: ["262e621d10ae4475ab5732b39a9160b2"]
+            type: array
+            items:
+              description: An environment UUID.
+              type: string
```
thearifismail, line 601: Try setting type to UUID as it is a type of fields supported `mashmallow` but not sure if `openapi specs` allow it.  If not then use:
```
type: string
format: uuid
required: true | false
```

swagger/system_profile.spec.yaml
```diff
@@ -592,6 +592,15 @@ $defs:
...
+              example: "262e621d10ae4475ab5732b39a9160b2, 01fd642e02de4e6da2da6172081a971e"
```
thearifismail, line 602: These examples need hyphens as required by the pattern specified.

kruai (changes requested):
Let's discuss the changes on [this PR](https://github.com/RedHatInsights/inventory-schemas/pull/145) instead. The process is that after the inventory-schemas PR is merged, we open an insights-host-inventory PR with the automated changes from `make update_schema`.

swagger/system_profile.spec.yaml
kruai: The pattern does not specify hyphens; `[0-9a-f]{32}` means a 32 character string, where each character is a number or a letter from "a" to "f"

swagger/system_profile.spec.yaml
kruai: Just to clarify, this field is under `rhsm`, so it's `rhsm.environments`. I'm not against a more specific field name, but it at least won't conflict with fields from other workloads

swagger/system_profile.spec.yaml
kruai: UUID is a valid value for `format`, but not for `type` ([source](https://swagger.io/docs/specification/v3_0/data-models/data-types/)). However, I think that format requires hyphenated UUIDs.

swagger/system_profile.spec.yaml
Tomboyo: The other option for a name would be `rhsm.content_template`, which aligns with what Insights refers to them as (as I understand it). Hosted Candlepin (the source of truth for this data) refers to them as environments, though. Is either of those clearer than the others?

Tomboyo: I also see other schema entries tend to suffix `_uuid` or `_id`, but not all of them. Do we have a preference on way or the other?

swagger/system_profile.spec.yaml
Tomboyo: To the best of my understanding, UUID isn't a formally defined OpenAPI type. Other uuid-type fields in the spec file seemed to use no-format string types and a pattern regex.

thearifismail: As @kruai said, UUID is NOT a type but `format: uuid` is.  Use of inventory schema has grown since its inception and we have are starting to ask for new attributes.  That's why it should be specified

thearifismail: Since a `UUID` without hyphens is a valid value, I thought it might pass the validation test.  If not then should hyphenation be forced?  @Tomboyo your thoughts?

Tomboyo: I've gone back and spoken with the Hosted Candlepin team, and they would prefer that this schema treat the IDs as opaque strings (no format, no pattern) rather than try to codify the specific representation, which Candlepin considers to be an implementation detail. Does that sound okay?

Tomboyo:
Per Candlepin maintainer feedback, I've updated the environment ID schema to be an opaque string, and removed any reference to it being a UUID in favor of just calling it an ID. I've also renamed rhsm.environments to rhsm.environment_ids so it's clear that the field contains identifiers.  The same is done in the https://github.com/RedHatInsights/inventory-schemas/pull/145/files PR, so per Kruai we can continue conversation over there if that's preferred. Thanks all.

Tomboyo:
@kruai Locally I `reset --hard` this branch to the tip of master and ran `make update-schema`, but the redocly hook fails because "openapi.dev.json does not exist or is invalid." I am not familiar with this repo and the redocly setup in familiar, but it looked like the redocly file was just pointing to the wrong relative path (missing ./swagger/ prefix), so I updated it to resolve the issue.

Tomboyo:
Closing as obsoleted by bc61958d7f1392f11bd092a36fc62b19e67fa179, which already has the desired updates in it.

PR #2181, opened 2025-01-14
chore(RHINENG-14059): Add Ruff ARG rule (Part 1)

kruai:
# Overview

This PR is being created to address [RHINENG-14059](https://issues.redhat.com/browse/RHINENG-14059).
Adds [the ARG rule](https://docs.astral.sh/ruff/rules/#flake8-unused-arguments-arg) to Ruff, and fixes the code that breaks the rule.

NOTE: This PR is a collaborative work-in-progress. In order to contribute, you must do the following:

1. In `pyproject.toml`, uncomment lines 21 and 38 ("ARG" and "ignore-variadic-names")
2. Run `make style`
3. Correct any errors raised by the linter
4. Re-comment the lines from step 1
5. Commit

When we are ready to merge the PR, we will need to uncomment the lines from step 1.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


tests/test_api_assignment_rules_create.py
```diff
@@ -103,7 +103,7 @@ def test_create_assignemnt_rule_same_group(api_create_assign_rule, db_create_gro
     assert group in response_data["detail"]
 
 
-def test_create_assignment_rule_RBAC_denied(subtests, mocker, api_create_assign_rule, db_create_group, enable_rbac):
+def test_create_assignment_rule_RBAC_denied(subtests, mocker, api_create_assign_rule, db_create_group, _enable_rbac):
```
fstavela, line 106: If a test needs to use a fixture, but it doesn't use its return value, you can decorate the test with `@pytest.mark.usefixtures("fixture_name")` and then you don't have to include this fixture in the test arguments.

kruai: Fantastic, thanks a ton for the info! I'll switch to using that, as it's simpler

kruai:
There were initially 295 lint errors; now we're down to 141

FabriciaDinizRH (approved):
LGTM

app/auth/__init__.py
```diff
@@ -12,6 +12,7 @@
 
 
 def authentication_header_handler(apikey, required_scopes=None):
+    _ = (required_scopes,)  # Unused, but needed for the security scheme (apikeyInfoFunc)
```
fstavela, line 15: This is just a suggestion, not a request for a change.

You can use this comment: `# noqa: ARG001`. This will turn off the ruff validation for this rule on the line where the comment is, and you won't have to use the argument in the function. This should be used as little as possible of course, as the goal is to get rid of useless arguments. 

I'm not sure if the comment is better than what you did here though, because with your way we at least have an explaining comment. But maybe it is possible to add the explanation to the `noqa` comment as well, for example: `# noqa: ARG001, 'required_scopes' is needed for the security scheme (apikeyInfoFunc)`

api/staleness.py
```diff
@@ -49,6 +49,7 @@ def _validate_input_data(body):
 @rbac(RbacResourceType.HOSTS, RbacPermission.READ)
 @metrics.api_request_time.time()
 def get_staleness(rbac_filter=None):
+    _ = (rbac_filter,)  # Unused: group-level RBAC does not apply to account-level staleness
```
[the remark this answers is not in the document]
kruai: Nope; group-level custom staleness would probably be a different endpoint. Similar to the other one you commented on, this arg is just unused in this context. For this one, it's because we need `rbac_filter` on all API endpoints, but the RBAC filter only tells us what groups and hosts the user has access to; it isn't used for this endpoint, since it's account-level (i.e. it doesn't matter which groups the user has access to).

It would likely be cleaner to just use `# noqa` here as well.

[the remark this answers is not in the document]
kruai: OK, I've updated it to use `# noqa`.

app/auth/__init__.py
kruai: OK, I've updated it to use `# noqa`.

fstavela (approved):
LGTM, thank you!

thearifismail (approved):
LGTM

FabriciaDinizRH (approved):

PR #2280, opened 2025-02-18
RHINENG-15929: Give subman_id higher priority in deduplication

fstavela:
# Overview

This PR is being created to address [RHINENG-15929](https://issues.redhat.com/browse/RHINENG-15929).

If `hbi.deduplication-elevate-subman_id` feature flag is turned on, we will give `subscription_manager_id` the highest priority (except immutable `provider_id`) in deduplication. Also, we will remove `mac_addresses` from the elevated facts.

When we added `mac_addresses` to the elevated facts we had 2 assumptions. One was that MAC addresses are very unlikely, almost impossible to change. This however isn't true if we the collectors run in containers. The other assumption was that MAC addresses will never be duplicated on multiple machines. However, the recent duplication issues show us that this also isn't true, so it probably doesn't make sense to keep MAC addresses in the elevated facts anymore.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


lib/host_repository.py
```diff
@@ -37,20 +42,22 @@
 
 # These are the "elevated" canonical facts that are
 # given priority in the host deduplication process.
-# NOTE: The order of this tuple is important.  The order defines
-# the priority.
+# NOTE: The order of this tuple is important. The order defines the priority.
 ELEVATED_CANONICAL_FACT_FIELDS = ("provider_id", "mac_addresses", "insights_id", "subscription_manager_id")
+# These "v2" elevated facts are used when "hbi.deduplication-elevate-subman_id" FF is turned on.
+ELEVATED_CANONICAL_FACT_FIELDS_V2 = ("provider_id", "subscription_manager_id", "insights_id")
```
kruai, line 48: Should `mac_addresses` still be elevated (but lowest prio)?

kruai (approved):
Looks good, just had a Q about mac_addresses.

lib/host_repository.py
fstavela: The priority has a very low importance now. It is enough if one of the elevated canonical facts matches to update the host, it doesn't matter which one. So, even if subman_id and insights_id is different and MAC addresses match, we would update the host even if it has the lowest priority. The only scenario where priority matters is when we match multiple hosts on different elevated facts. For example, if the payload matches host1 on insights_id and host2 on mac_addresses, we would update host1 if mac_addresses has the lowest priority.

As I wrote in the description, our assumptions around MAC addresses which we had when we added it to the elevated facts were not correct, so I don't think it should keep being in the elevated facts. My long term goal is to simplify the entire dedup logic to rely on much less fields, ideally just one, but for that I need more information from other teams, so this PR is just an attempt to fix the current dedup issues quickly.

PR #2300, opened 2025-02-26
chore(RHINENG-4592, RHINENG-15638) order_how refactoring & case-insensitivy

kruai:
# Overview

This PR is being created to address [RHINENG-4592](https://issues.redhat.com/browse/RHINENG-4592) and [RHINENG-15638](https://issues.redhat.com/browse/RHINENG-15638).
Does the following:
- Fixes the pre-commit & redocly config
- Re-removes `openapi.dev.json`
- Refactors the order_how params. The descriptions for each order_how are different, so I couldn't entirely replace the parameter definitions. However, I refactored the schemas for each.
- Makes order_how case-insensitive. There is no support for case-insensitive enums or the case-insensitive regex flag, so I used a regex that simply accounts for all capital/lowercase combinations.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


strider (approved):
LGTM

FabriciaDinizRH (approved):
LGTM

jpramos123 (approved):
LGTM!

One quick question, isn't better to have just `OrderHow` query parameter, and apply it to the resource path as needed, instead of creating multiple objects to use the same reference?

kruai:
> LGTM!
> 
> One quick question, isn't better to have just `OrderHow` query parameter, and apply it to the resource path as needed, instead of creating multiple objects to use the same reference?

@jpramos123  I would like to do that too, but the description has to be different for each one. This is because order_how defaults to a different value depending on what you're using for order_by, and order_by has different options for each of these endpoints. For example, hosts can be ordered by display_name, group_name, updated, and operating_system; groups can be ordered by name, host_count, and updated.

In my first attempt, I tried to use `$ref` for the whole parameter and override the description for each, but that doesn't work bc `$ref` causes sibling nodes to be ignored. I also tried using yaml anchors, but those aren't supported by OpenAPI. So our two choices are:

- Use the same `OrderHow` query param and ditch the unique descriptions (or make one big complicated one)
- Use different `OrderHow` query params, but refactor the `schema` part of the param

PR #2311, opened 2025-02-28
feat(RHINENG-15447): tags data transformation

FabriciaDinizRH:
# Overview

This PR is being created to address [RHINENG-15447](https://issues.redhat.com/browse/RHINENG-15447).

- add a `tags_alt` column to the hosts table
- migrate data from `tags` to `tags_alt`
- alter create/update host functions to also populate `tags_alt`
- update a couple of tests to verify the `tags_alt` behavior is consistent with `tags`

**DO NOT MERGE IT YET EVEN IF APPROVED**

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/models.py
```diff
@@ -430,6 +447,20 @@ def _delete_tags_namespace(self, namespace):
 
         orm.attributes.flag_modified(self, "tags")
 
+    def _populate_tags_alt_from_tags(self, tags):
+        transformed_tags = []
+
+        if type(tags) is dict:
```
strider, line 453: Why not using `isinstance` built-in function here like you do in the `_deserialize_tags_alt` function below ?

FabriciaDinizRH: no particular reason, changing it

tests/test_tags.py
```diff
@@ -357,3 +358,17 @@ def test_get_tags_with_dollar_signs_via_db(api_get, db_create_host, tag, flatten
...
     assert response_status == 200
     assert flattened_tag == response_data["results"][0]["tag"]
+
+
+def test_tag_alt_is_created(api_get, db_create_host):
+    tag = {"insights-client": {"test-key": "test-value"}}
+    flattened_tag = {"namespace": "insights-client", "key": "test-key", "value": "test-value"}
+    db_create_host(extra_data={"tags": tag})
+
+    url = build_tags_url()
```
FabriciaDinizRH, line 368: I'm aware I'm using the wrong url function here

app/models.py
```diff
@@ -430,6 +447,20 @@ def _delete_tags_namespace(self, namespace):
...
+            transformed_tags_obj = Tag.create_tags_from_nested(tags)
+        elif type(tags) is list:
+            transformed_tags_obj = Tag.to_structured(tags)
```
kruai, line 456: If the tags are of type "list", they're already in the correct format, right? seems unnecessary to convert them to structured and then back to flat afterwards

app/models.py
```diff
@@ -189,11 +190,13 @@ def __init__(
         org_id=None,
         facts=None,
         tags=None,
+        tags_alt=None,
         system_profile_facts=None,
         groups=None,
     ):
         if tags is None:
             tags = {}
+            tags_alt = []
```
kruai, line 199: What about if `tags` is populated, but `tags_alt` is None?

FabriciaDinizRH: good catch

app/models.py
```diff
@@ -287,6 +293,11 @@ def __init__(
     ):
         if tags is None:
             tags = {}
+            if tags_alt is None:
+                tags_alt = []
```
chambridge, line 297: Is this if condition necessary or shouldn't we always just set `tags_alt` to `[]` if we are setting `tags` to `{}`

app/models.py
```diff
@@ -287,6 +293,11 @@ def __init__(
...
+        elif tags_alt is None:
```
chambridge, line 298: Do we need this to be an `elif` and check for `None`, instead of just an `else`?

app/models.py
```diff
@@ -836,7 +897,7 @@ def _normalize_system_profile(normalize, data):
         return {**data, "system_profile": system_profile}
 
     @staticmethod
-    def build_model(data, canonical_facts, facts, tags):
+    def build_model(data, canonical_facts, facts, tags, tags_alt=None):
```
chambridge, line 900: Why not set the default for `tags_alt` to `[]`?

app/models.py
```diff
@@ -879,7 +941,9 @@ class Meta:
     reporter = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=255))
 
     @staticmethod
-    def build_model(data, canonical_facts, facts, tags):
+    def build_model(data, canonical_facts, facts, tags, tags_alt=None):
```
chambridge, line 944: Why not set the default for `tags_alt` to `[]`?

migrations/versions/6f44b7ecd7be_add_tags_alt_column_and_migrate_tags_.py
```diff
@@ -0,0 +1,47 @@
...
+                            )
+                        )
+                    FROM JSONB_OBJECT_KEYS(tags) AS ns(namespace),
+                        JSONB_EACH(tags -> ns.namespace) AS k(key, value),
+                        JSONB_ARRAY_ELEMENTS_TEXT(k.value) AS v(value)),
+                '[]'::jsonb
+            ) AS tags_alt
+            FROM hbi.hosts
+        ) AS sub
+        WHERE h.id = sub.id;
```
chambridge, line 42: So this migration would run against every host in the hosts table. We should be able to filter out the UPDATE to run against only hosts that have tag data and I believe during our standup we discussed only updating those hosts that are of host_type edge, assuming the rest would get updated via check-ins to minimize the impact of the migration.

tests/test_models.py
```diff
@@ -267,6 +267,7 @@ def test_create_host_with_tags(tags, db_create_host):
     )
 
     assert created_host.tags == tags
+    assert created_host.tags_alt == tags
```
chambridge, line 270: Is this an accurate assertion? Shouldn't these be different formats?

chambridge (changes requested):
A couple code questions and one adjustment around what the migration is attempting to update.

migrations/versions/6f44b7ecd7be_add_tags_alt_column_and_migrate_tags_.py
FabriciaDinizRH: I saw that coalesce was supposed to do this filtering of only getting hosts that had the tags column

app/models.py
chambridge: I see.

migrations/versions/6f44b7ecd7be_add_tags_alt_column_and_migrate_tags_.py
chambridge: The above will execute on all host records. There is nothing in the where clauses to filter down to hosts that have tags content (meaning more than `{}`) and there is nothing to filter this down based on host type either.

tests/test_models.py
chambridge: Oh, I actually think its because the test is written badly to begin with here as the input tags value seen above on lines 252-257. The format here is in the list structure which isn't the format tags would normally be stored in.

app_migrations/6f44b7ecd7be.sql
```diff
@@ -0,0 +1,73 @@
...
+[2025-03-07 03:45:53,207] [2887625] [140501292804480] [inventory.app.payload_tracker] [INFO] Using injected producer object (<app.payload_tracker.NullProducer object at 0x7fc8eebee0d0>) for PayloadTracker
+[2025-03-07 03:45:53,208] [2887625] [140501292804480] [inventory.app] [INFO] Initializing Segmentio
+[2025-03-07 03:45:53,208] [2887625] [140501292804480] [inventory.app] [INFO] Initializing Segmentio
+[2025-03-07 03:45:53,208] [2887625] [140501292804480] [inventory.app] [INFO] Registering Segmentio flush on shutdown
+[2025-03-07 03:45:53,208] [2887625] [140501292804480] [inventory.app] [INFO] Registering Segmentio flush on shutdown
+[2025-03-07 03:45:53,212] [2887625] [140501292804480] [alembic.runtime.migration] [INFO] Context impl PostgresqlImpl.
+[2025-03-07 03:45:53,212] [2887625] [140501292804480] [alembic.runtime.migration] [INFO] Generating static SQL
+[2025-03-07 03:45:53,212] [2887625] [140501292804480] [alembic.runtime.migration] [INFO] Will assume transactional DDL.
```
kruai, line 38: Why are all these logs here - was that a side effect of the schema dump?

app/models.py
```diff
@@ -288,6 +307,10 @@ def __init__(
     ):
         if tags is None:
             tags = {}
+            tags_alt = []
+        else:
+            tags_alt = super()._populate_tags_alt_from_tags(tags)
+
```
kruai, line 313: This check is automatically handled by the parent class LimitedHost, so this shouldn't be needed

kruai (changes requested):
Just a couple of small suggestions

app_migrations/6f44b7ecd7be.sql
FabriciaDinizRH: side effect of `gen_offline_sql`

FabriciaDinizRH: I can just delete these?

FabriciaDinizRH: "You will need to clean up this file to remove the log entries and the update to the alembic version table."
Guess I skipped this line on the README

FabriciaDinizRH: fixed

chambridge (approved):

thearifismail (approved):
LGTM

app/utils.py
```diff
@@ -312,6 +312,15 @@ def to_nested(self):
 
         return {self.namespace: {self.key: []}}
 
+    @staticmethod
+    def to_structured(tags_dict_list):
```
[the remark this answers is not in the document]
FabriciaDinizRH: Last I checked I was being used, but apparently I removed that code. I'll open a following PR to removed it

kruai (approved):

PR #2332, opened 2025-03-06
chore(RHINENG-16421) Use org_id rather than Identity for staleness functions

kruai:
# Overview

This PR is being created to address [RHINENG-16421](https://issues.redhat.com/browse/RHINENG-16421).
Does some refactoring that's blocking Kessel Phase 0. Includes the following:

- Removes the `get_current_identity()` fallback from`get_staleness_obj()`, because that breaks unless you're in an API context
- Refactors staleness-related functions to require org_id rather than Identity
- Remove some uses of `create_mock_identity_with_org_id`, which was only required for structural reasons (the org_id is the only part of the identity that was actually used)
- Sets `"org_id": "test"` in payloads.py to match the README so as not to confuse new developers. The currently-used org_id does not match the b64 identity header in the README, so contributors currently create hosts using the kafka producer util but then can't fetch them from the API

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/utils/payloads.py
+++ b/utils/payloads.py
@@ -8,13 +8,13 @@ from datetime import timedelta
 from datetime import timezone
 
 # complete system identity
-# IDENTITY = {
-#    "org_id": "test",
-#    "type": "System",
-#    "auth_type": "cert-auth",
-#    "system": {"cn": "1b36b20f-7fa0-4454-a6d2-008294e06378", "cert_type": "system"},
-#    "internal": {"auth_time": 6300},
-# }
+IDENTITY = {
+    "org_id": "test",
+    "type": "System",
+    "auth_type": "cert-auth",
+    "system": {"cn": "1b36b20f-7fa0-4454-a6d2-008294e06378", "cert_type": "system"},
+    "internal": {"auth_time": 6300},
+}
 
 # system identity: invalid or incomplete for testing
 # IDENTITY = {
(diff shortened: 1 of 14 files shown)


utils/payloads.py
```diff
@@ -27,7 +27,7 @@
 
 # complete user identity
 IDENTITY = {
-    "org_id": "5894300",
+    "org_id": "test",
```
thearifismail, line 30: I think the first commented [identity](https://github.com/RedHatInsights/insights-host-inventory/blob/a0f67d26090654809d8df176d166193a6bc2db0c/utils/payloads.py#L11C1-L18C1) , which uses `system` identity type which is more relevant than `basic-auth`.

thearifismail (approved):
LGTM

utils/payloads.py
kruai: Should I comment this one out, and uncomment the first identity instead?

thearifismail: I think so. if you want you can add an email address to that one.  It used be the test identity we used before but Joao opted to create a new one for some reason.  I never bothered to revert back to the original one.

kruai: Email address is only valid for identities of type User, while the first identity is of type System. I've uncommented that one though.

thearifismail (approved):
LGTM

PR #2378, opened 2025-03-24
adhoc: prevent rhsm-system-profile-bridge to be culled an deleted

jpramos123:
# Overview

This PR is being created to address [RHINENG-16901](https://issues.redhat.com/browse/RHINENG-16901).

Meanwhile, prevent rhsm-system-profile-bridge hosts to be culled and deleted by host-reaper.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/host_reaper.py
+++ b/host_reaper.py
@@ -76,7 +76,22 @@ def run(config, logger, session, event_producer, notification_event_producer, sh
     with application.app.app_context():
         filter_hosts_to_delete = find_hosts_in_state(logger, session, ["culled"])
 
-        query = session.query(Host).filter(or_(False, *filter_hosts_to_delete))
+        # Adhoc fix for RHINENG-16901
+        # hosts reporter by rhsm-system-profile-bridge are not being deleted
+        # when marked as culled, and are also prevented to stay culled as w
+        # we are forcing its modified_on and last_check_in to be updated.
+
+        rhsm_bridge_hosts_query = session.query(Host).filter(
+            and_(or_(False, *filter_hosts_to_delete), Host.reporter == "rhsm-system-profile-bridge")
+        )
+
+        for host in rhsm_bridge_hosts_query.yield_per(config.host_delete_chunk_size):
+            host._update_modified_date()
+            host._update_last_check_in_date()
+
+        query = session.query(Host).filter(
+            and_(or_(False, *filter_hosts_to_delete), Host.reporter != "rhsm-system-profile-bridge")
+        )
         hosts_processed = config.host_delete_chunk_size
         deletions_remaining = query.count()
 


host_reaper.py
```diff
@@ -76,6 +76,23 @@ def run(config, logger, session, event_producer, notification_event_producer, sh
...
+        # we are forcing its modified_on and last_check_in to be updated.
+
+        rhsm_briedge_hosts = (
+            session.query(Host)
+            .filter(and_(or_(False, *filter_hosts_to_delete), Host.reporter == "rhsm-system-profile-bridge"))
+            .all()
+        )
+        filter_hosts_to_delete = filter_hosts_to_delete + [Host.reporter != "rhsm-system-profile-bridge"]
+        rhsm_briedge_hosts_ids = []
+        for host in rhsm_briedge_hosts:
```
strider, line 91: Same as Arif's comment above here.

host_reaper.py
```diff
@@ -76,6 +76,23 @@ def run(config, logger, session, event_producer, notification_event_producer, sh
...
+            host._update_modified_date()
+            host._update_last_check_in_date()
+            rhsm_briedge_hosts_ids.append(host.id)
```
kruai, line 94:
```suggestion
            rhsm_bridge_hosts_ids.append(host.id)
```
Also, what's the purpose of creating this huge list of IDs? It doesn't appear to be used anywhere

host_reaper.py
```diff
@@ -76,6 +76,23 @@ def run(config, logger, session, event_producer, notification_event_producer, sh
...
+        # when marked as culled, and are also prevented to stay culled as w
+        # we are forcing its modified_on and last_check_in to be updated.
+
+        rhsm_briedge_hosts = (
+            session.query(Host)
+            .filter(and_(or_(False, *filter_hosts_to_delete), Host.reporter == "rhsm-system-profile-bridge"))
+            .all()
+        )
+        filter_hosts_to_delete = filter_hosts_to_delete + [Host.reporter != "rhsm-system-profile-bridge"]
+        rhsm_briedge_hosts_ids = []
```
kruai, line 90:
```suggestion
        rhsm_bridge_hosts_ids = []
```

host_reaper.py
```diff
@@ -76,6 +76,23 @@ def run(config, logger, session, event_producer, notification_event_producer, sh
...
+        # Adhoc fix for RHINENG-16901
+        # hosts reporter by rhsm-system-profile-bridge are not being deleted
+        # when marked as culled, and are also prevented to stay culled as w
+        # we are forcing its modified_on and last_check_in to be updated.
+
+        rhsm_briedge_hosts = (
+            session.query(Host)
+            .filter(and_(or_(False, *filter_hosts_to_delete), Host.reporter == "rhsm-system-profile-bridge"))
+            .all()
+        )
```
kruai, line 88: I recommend not querying all hosts and storing them in a variable - that's how we get OOMs. Instead, I recommend using `yield_per` to go through the hosts a batch at a time. We have plenty of examples in the codebase at this point to use as a reference

host_reaper.py
```diff
@@ -76,6 +76,23 @@ def run(config, logger, session, event_producer, notification_event_producer, sh
...
+        filter_hosts_to_delete = filter_hosts_to_delete + [Host.reporter != "rhsm-system-profile-bridge"]
+        rhsm_briedge_hosts_ids = []
+        for host in rhsm_briedge_hosts:
+            host._update_modified_date()
+            host._update_last_check_in_date()
```
kruai, line 93: I don't think we should force the "modified" and "check-in" dates to be updated, since the host isn't actually being updated. Isn't it enough to add the extra "rhsm-system-profile-bridge" clause to the filter?

kruai (changes requested):

host_reaper.py
jpramos123: If we don't do that, the host will stay marked as culled, and won't be able to be fetched by our APIs. That is why I'm forcing it to be updated.

kruai: OK. I feel like it might lead to some confusion, but if it's as a band-aid so their systems stop being removed while they figure out a fix, we might not have a choice here

kruai (approved):

PR #2400, opened 2025-03-28
feat(RHINENG-15648): Async update per_reporter_staleness

jpramos123:
# Overview

This PR is being created to address [RHINENG-15648](https://issues.redhat.com/browse/RHINENG-15648).

It updates PRS asynchronously when user CREATE/UPDATE/DELETE its custom staleness.

A Thread is created every time a user perform the mentioned interactions with our API.

This feature is behind the `hbi.create_last_check_in_update_per_reporter_staleness` feature flag.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/staleness.py
+++ b/api/staleness.py
@@ -1,4 +1,5 @@
 from http import HTTPStatus
+from threading import Thread
 
 from flask import abort
 from marshmallow import ValidationError
@@ -18,14 +19,18 @@ from app.instrumentation import log_create_staleness_failed
 from app.instrumentation import log_create_staleness_succeeded
 from app.instrumentation import log_patch_staleness_succeeded
 from app.logging import get_logger
+from app.models import Host
 from app.models import StalenessSchema
 from app.serialization import serialize_staleness_response
 from app.serialization import serialize_staleness_to_dict
 from app.staleness_serialization import get_sys_default_staleness_api
+from lib.feature_flags import FLAG_INVENTORY_CREATE_LAST_CHECK_IN_UPDATE_PER_REPORTER_STALENESS
+from lib.feature_flags import get_flag_value
 from lib.middleware import rbac
 from lib.staleness import add_staleness
 from lib.staleness import patch_staleness
 from lib.staleness import remove_staleness
+from run import app as application
 
 logger = get_logger(__name__)
 
@@ -42,6 +47,25 @@ def _validate_input_data(body):
     except ValidationError as e:
         logger.exception(f'Input validation error, "{str(e.messages)}", while creating account staleness: {body}')
         abort(HTTPStatus.BAD_REQUEST, f"Validation Error: {str(e.messages)}")
+
+
+def _update_hosts_staleness_async(org_id):
+    with application.app.app_context():
+        logger.debug("Starting host staleness update thread")
+        try:
+            logger.debug(f"Querying hosts for org_id: {org_id}")
+            hosts_query = Host.query.filter(Host.org_id == org_id)
+            num_hosts = hosts_query.count()
+            if num_hosts > 0:
+                logger.debug(f"Found {num_hosts} hosts for org_id: {org_id}")
+                for host in hosts_query.yield_per(500):
+                    host._update_all_per_reporter_staleness()
+                hosts_query.session.commit()
+
+                delete_cached_system_keys(org_id=org_id, spawn=True)
+            logger.debug("Leaving host staleness update thread")
+        except Exception as e:
+            raise e
 
 
 @api_operation
@@ -89,7 +113,11 @@ def create_staleness(body):
     try:
         # Create account staleness with validated data
         created_staleness = add_staleness(validated_data)
-        delete_cached_system_keys(org_id=org_id, spawn=True)
+        if get_flag_value(FLAG_INVENTORY_CREATE_LAST_CHECK_IN_UPDATE_PER_REPORTER_STALENESS):
+            update_hosts_thread = Thread(target=_update_hosts_staleness_async, daemon=True, args=(org_id,))
+            update_hosts_thread.start()
+        else:
+            delete_cached_system_keys(org_id=org_id, spawn=True)
         log_create_staleness_succeeded(logger, created_staleness.id)
     except IntegrityError:
         error_message = f"Staleness record for org_id {org_id} already exists."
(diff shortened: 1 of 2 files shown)


tests/test_custom_staleness.py
```diff
@@ -187,3 +193,36 @@ def test_no_hosts_to_delete(
...
+                # Call the thread
+                host_update_thread.start()
+
+                hosts_before_update = db_get_hosts(host_ids).all()
+                assert hosts_before_update == 1
+
+                # Wait for the thread to finish running
+                host_update_thread.join()
+
+                # hosts_after_update = db_get_hosts(host_ids).all()
```
jpramos123, line 228: My expectation is that after `host_update_thread.join()` is called, the DB is up to date with the new `per_reporter_staleness`

kruai: Makes sense to me. I would have suggested just calling the create_staleness API endpoint, but idk how to determine whether the thread has finished. What issue are you running into with this approach?

api/staleness.py
```diff
@@ -44,6 +49,25 @@ def _validate_input_data(body):
         abort(HTTPStatus.BAD_REQUEST, f"Validation Error: {str(e.messages)}")
 
 
+def _update_hosts_staleness_async(org_id):
+    with application.app.app_context():
```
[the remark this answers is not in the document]
jpramos123: Yes, it works, in this case we do not create a hole new Flask server again, and use the same that is already running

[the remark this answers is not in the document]
jpramos123: As we spoke, changed a little bit how we use this.

api/staleness.py
```diff
@@ -89,7 +113,18 @@ def create_staleness(body):
...
+                target=_update_hosts_staleness_async,
+                daemon=True,
+                args=(
+                    org_id,
+                    current_app._get_current_object(),
+                ),
+            )
+            update_hosts_thread.start()
+        else:
+            delete_cached_system_keys(org_id=org_id, spawn=True)
```
kruai, line 127: Shouldn't we delete the cache regardless of the flag value? Or is there a reason we don't want to do that when the flag is True?

tests/test_custom_staleness.py
```diff
@@ -187,3 +191,96 @@ def test_no_hosts_to_delete(
...
+                stale_timestamp = _now + timedelta(
+                    seconds=CUSTOM_STALENESS_NO_HOSTS_TO_DELETE["conventional_time_to_stale"]
+                )
+                stale_timestamp = stale_timestamp.isoformat()
+                assert hosts_before_update[0].per_reporter_staleness[reporter]["stale_timestamp"] == stale_timestamp
+            db_create_staleness_culling(**CUSTOM_STALENESS_HOST_BECAME_STALE)
+
+        host_update_thread = Thread(target=_update_hosts_staleness_async, daemon=True, args=("test", flask_app.app))
+        host_update_thread.start()
+        host_update_thread.join()
```
kruai, line 227: I think it would be better if we created the Staleness settings via the API here. Otherwise we're recreating the functionality within the test rather than testing the actual functionality

tests/test_custom_staleness.py
```diff
@@ -187,3 +191,96 @@ def test_no_hosts_to_delete(
...
+
+        with flask_app.app.app_context():
```
kruai, line 229: Not a big deal, but all this could be within the same app context, rather than making another one here

tests/test_custom_staleness.py
```diff
@@ -187,3 +191,96 @@ def test_no_hosts_to_delete(
...
+                stale_timestamp = stale_timestamp.isoformat()
+                assert hosts_before_update[0].per_reporter_staleness[reporter]["stale_timestamp"] == stale_timestamp
+
+            db_delete_staleness_culling(hosts_before_update[0].org_id)
+
+        host_update_thread = Thread(target=_update_hosts_staleness_async, daemon=True, args=("test", flask_app.app))
+        host_update_thread.start()
+        host_update_thread.join()
+
+        with flask_app.app.app_context():
```
kruai, line 279: Same thing here as the previous test

kruai (changes requested):

api/staleness.py
jpramos123: When the flag is enabled, host updates run in the background. We need to ensure cache cleanup happens only after the thread completes (not before) to maintain cached responses during bulk updates. This way, all hosts for an org_id stay cached until the entire update finishes, then get purged together.

tests/test_custom_staleness.py
jpramos123: I see, what I was not able to figure out is how I can make sure the thread finished, which is made by using `.join()` in the thread...

tests/test_custom_staleness.py
jpramos123: That is something really, strange... before If I did not added these 2 context, the tests failed. Now, it works inside the same context... (3 dots for ever)

tests/test_custom_staleness.py
jpramos123: I will try with putting a sleep, and lets see how it behaves

jpramos123: It worked!

api/staleness.py
kruai: OK, thanks for the clarification.

kruai, line 127: This would be good to refactor into a new function, since I see the same thing line-for-line 3 times in this file

kruai (commented):
Just got one recommendation

api/staleness.py
jpramos123: Done!

Thanks

kruai (approved):

app/models.py
```diff
@@ -448,6 +456,19 @@ def _update_stale_timestamp(self, stale_timestamp, reporter):
...
 
+    def _update_all_per_reporter_staleness(self):
+        st = _create_staleness_timestamps_values(self, self.org_id)
+
+        for reporter in self.per_reporter_staleness:
+            self.per_reporter_staleness[reporter].update(
+                stale_timestamp=st["stale_timestamp"].isoformat(),
+                culled_timestamp=st["culled_timestamp"].isoformat(),
+                stale_warning_timestamp=st["stale_warning_timestamp"].isoformat(),
+                last_check_in=self.last_check_in.isoformat(),
```
FabriciaDinizRH, line 467: what counts as checking in?

FabriciaDinizRH (approved):
LGTM!
Just left a general question :)

app/models.py
jpramos123: This is the value that is already in the Host object, it was created using `datetime.now()` when the host made a check-in in our MQ pod.

PR #2433, opened 2025-04-10
fix: handle tags key name with slash

jpramos123:
# Overview

This PR is being created to address [RHINENG-16224](https://issues.redhat.com/browse/RHINENG-16224).
It handles the tag filtering when the key contains a slash `/` in the name.

It will accepted the following (but not limited to) examples:
`tags=/aws/user=jramos`
`tags=namespace/aws/user=jramos`
`tags=namespace/aws/user/id=123`

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


ezr-ondrej (commented):
Can you add test for the leading slash as well?
I mean empty namespace and slash in the key. That sounds like an edge case we should cover :)

jpramos123:
> Can you add test for the leading slash as well? I mean empty namespace and slash in the key. That sounds like an edge case we should cover :)

Added it!

Thanks

kruai (approved):
Seems fine, but is there a way to determine whether additional slashes are in the key name or namespace? Is the assumption that if there are multiple slashes then they're in the key and not the namespace?

PR #2435, opened 2025-04-10
feat: Send host update event after updating host staleness async

jpramos123:
# Overview

This PR is being created to address [RHINENG-17423](https://issues.redhat.com/browse/RHINENG-17423) and [RHINENG-17581](https://issues.redhat.com/browse/RHINENG-17581).

Creates an array of events that is used to use to send `update host events` after an successful commit to the DB.
Also fix the bug reported on [RHINENG-17581](https://issues.redhat.com/browse/RHINENG-17581). The cause was that the modified_on/updated value of the host was updated by the asyc call that was made to update the staleness values of the host after a change in the custom staleness.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


api/staleness.py
```diff
@@ -49,26 +56,54 @@ def _validate_input_data(body):
...
+                    app.event_producer.write_event(event, host_id, headers, wait=True)
+
+                delete_cached_system_keys(org_id=identity.org_id, spawn=True)
             logger.info("Leaving host staleness update thread")
         except Exception as e:
             raise e
 
 
-def _validate_flag_and_call_thread(org_id):
+def _emit_host_updated_event(serialized_host, host, identity):
```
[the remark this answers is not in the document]
jpramos123: 1. Agreed! Just changed the name, it was emitting the event at the begging, but I changed the behavior.
2. Created X for us to change all `EventTypes` headers and events generically.

ezr-ondrej (dismissed):
Looks great to me! 👍

api/staleness.py
```diff
@@ -49,26 +56,54 @@ def _validate_input_data(body):
         abort(HTTPStatus.BAD_REQUEST, f"Validation Error: {str(e.messages)}")
 
 
-def _update_hosts_staleness_async(org_id, app):
+def _update_hosts_staleness_async(identity, app, staleness):
```
kruai, line 59: Type hints would help here, especially bc `staleness` can mean a few different things structurally throughout the codebase - would be good to standardize that, maybe make a class or two to make it a bit more rigid. Different PR for that though

api/staleness.py
```diff
@@ -49,26 +56,54 @@ def _validate_input_data(body):
...
+        threadctx.request_id = None
         logger.debug("Starting host staleness update thread")
         try:
-            logger.info(f"Querying hosts for org_id: {org_id}")
-            hosts_query = Host.query.filter(Host.org_id == org_id)
+            logger.info(f"Querying hosts for org_id: {identity.org_id}")
+            hosts_query = Host.query.filter(Host.org_id == identity.org_id)
             num_hosts = hosts_query.count()
+            st = staleness_timestamps()
+            staleness = serialize_staleness_to_dict(staleness)
```
kruai, line 68: Better to make a new var instead of overwriting the param imo, maybe something like this
```suggestion
            staleness_dict = serialize_staleness_to_dict(staleness)
```

api/staleness.py
```diff
@@ -49,26 +56,54 @@ def _validate_input_data(body):
...
+                    app.event_producer.write_event(event, host_id, headers, wait=True)
+
+                delete_cached_system_keys(org_id=identity.org_id, spawn=True)
             logger.info("Leaving host staleness update thread")
         except Exception as e:
             raise e
 
 
-def _validate_flag_and_call_thread(org_id):
+def _build_host_updated_event_params(serialized_host, host, identity):
```
kruai, line 92: Type hints would help here as well

FabriciaDinizRH (dismissed):
LGTM!

api/staleness.py
```diff
@@ -49,26 +56,54 @@ def _validate_input_data(body):
...
                 hosts_query.session.commit()
 
-                delete_cached_system_keys(org_id=org_id, spawn=True)
+                # After a successful commit to the db
+                # call all the events in the list
+                for event, headers, host_id in list_of_events_params:
+                    app.event_producer.write_event(event, host_id, headers, wait=True)
+
+                delete_cached_system_keys(org_id=identity.org_id, spawn=True)
             logger.info("Leaving host staleness update thread")
```
kruai, line 87: I know this wasn't done in this PR, but I just realized that the "Starting" log is debug-level, and this one should be as well
```suggestion
            logger.debug("Leaving host staleness update thread")
```

tests/test_custom_staleness.py
```diff
@@ -227,7 +227,7 @@ def test_async_update_host_create_custom_staleness(
             assert status == 201
 
             # Wait for thread to finish
-            time.sleep(0.1)
+            time.sleep(1)
```
kruai, line 230: Is this necessary? Makes the test 10x longer and seems to pass either way

tests/test_custom_staleness.py
```diff
@@ -277,7 +277,7 @@ def test_async_update_host_delete_custom_staleness(
             assert status == 204
 
             # Wait for thread to finish
-            time.sleep(0.1)
+            time.sleep(0.2)
```
kruai, line 280: Same Q here - does it not work for you using 0.1?

tests/test_custom_staleness.py
```diff
@@ -328,7 +328,7 @@ def test_async_update_host_update_custom_staleness(
             assert status == 200
 
             # Wait for thread to finish
-            time.sleep(0.1)
+            time.sleep(0.2)
```
kruai, line 331: Same Q here - does it not work for you using 0.1?

kruai (changes requested):
Good implementation. I just have a few suggestions for improved readability and test performance

strider (approved):
LGTM

api/staleness.py
jpramos123: Done!

api/staleness.py
jpramos123: Done!

tests/test_custom_staleness.py
jpramos123: Not really! It was a typo here.

Thank you

tests/test_custom_staleness.py
jpramos123: Somehow, it was not working with 0.1 locally, now it is... Threads are not that consistent locally as far as I'm seeing...

lib/feature_flags.py
```diff
@@ -21,7 +21,7 @@
     FLAG_INVENTORY_KESSEL_WORKSPACE_MIGRATION: False,
     FLAG_INVENTORY_API_READ_ONLY: False,
     FLAG_INVENTORY_DEDUPLICATION_ELEVATE_SUBMAN_ID: True,
-    FLAG_INVENTORY_CREATE_LAST_CHECK_IN_UPDATE_PER_REPORTER_STALENESS: False,
+    FLAG_INVENTORY_CREATE_LAST_CHECK_IN_UPDATE_PER_REPORTER_STALENESS: True,
```
ezr-ondrej, line 24: Is this changer related?

jpramos123: No! Sorry, it was a mistype!

ezr-ondrej (approved):

kruai (changes requested):
Implementation looks good, but a test should be updated to verify that the update event is produced

fstavela:
@jpramos123 One of the issues I found is that the produced update events don't include `system_profile`.

thearifismail (commented):
Why keep the whole `identity` object around when only the `org_id` string is needed.   How about instantiating only the org_id from identity as `org_id = get_current_identity().org_id`?

ezr-ondrej (commented):
The code still looks good to me, but I'm not sure what we are testing in the test 🤔
Read: I do not understand the event_producer enough to understand if it's testing the code that was added :)

jpramos123:
> Why keep the whole `identity` object around when only the `org_id` string is needed. How about instantiating only the org_id from identity as `org_id = get_current_identity().org_id`?

Full identity object is used here: https://github.com/RedHatInsights/insights-host-inventory/pull/2435/files#diff-386ab3475fad2bd4bd5fb75e9c7183a0ef700528419d2a01c66e559cb64f4e59R106

jpramos123:
> The code still looks good to me, but I'm not sure what we are testing in the test 🤔 Read: I do not understand the event_producer enough to understand if it's testing the code that was added :)

The `event_producer` is the object used to produce events—specifically HostUpdate events in this case. The test ensures it is called once per host. I also added some parameters to make it clearer.

fstavela (changes requested):
I found 2 more issues:

- The `request_id` from the staleness API request isn't forwarded to the `updated` events. In the headers, the `request_id` is an empty string and in metadata it is `null`.
- The `updated` timestamp in the kafka events is inconsistent. Should the `updated` timestamp on a host change when we change the custom staleness settings? Right now the `updated` timestamp stays the same (it doesn't change) when I create new staleness config, but the `updated` timestamp changes (at least it is changed in the events) when I update or delete the staleness config.

kruai (approved):

jpramos123:
> I found 2 more issues:
> 
> * The `request_id` from the staleness API request isn't forwarded to the `updated` events. In the headers, the `request_id` is an empty string and in metadata it is `null`.

For the `request_id`, when is it used? Should we actually forward it? Meaning that the request it self is not to update the hosts, but to create/update/delete the custom staleness. Update of hosts are a reflection of this. Happy to add this anyways!

> * The `updated` timestamp in the kafka events is inconsistent. Should the `updated` timestamp on a host change when we change the custom staleness settings? Right now the `updated` timestamp stays the same (it doesn't change) when I create new staleness config, but the `updated` timestamp changes (at least it is changed in the events) when I update or delete the staleness config.

No, I don't think we should change it. I will make sure it does not happen

kruai (dismissed):

api/staleness.py
```diff
@@ -49,26 +61,65 @@ def _validate_input_data(body):
         abort(HTTPStatus.BAD_REQUEST, f"Validation Error: {str(e.messages)}")
 
 
-def _update_hosts_staleness_async(org_id, app):
+@sa.event.listens_for(Host, "before_update")
```
thearifismail, line 64: Please provide a comment  stating what is this function does?  Why is it needed?  Is `target` a host or something else.  This function is not being explicitly called; does it mean that its a callback function used by `sqlalchemy`?

ezr-ondrej: Thank you for addressing this 🙏 
I'm missing why here, we don't learn why we need to skip modified_on from the comment.

ezr-ondrej:
> For the `request_id`, when/why is it used? Should we actually forward it? Meaning that the request it self is not to update the hosts, but to create/update/delete the custom staleness. Update of hosts are a reflection of this. Happy to add this anyways!

`request_id` should be ideally passed on to all the actions that happend on this given user "input". So whatever happens when user does a given action (one API call), we want to track what happens as a consequence of that call.

FabriciaDinizRH (dismissed):
LGTM

kruai (dismissed):

api/staleness.py
```diff
@@ -49,26 +63,77 @@ def _validate_input_data(body):
...
+
+
+def _update_hosts_staleness_async(identity: Identity, app: Flask, staleness: Staleness, request_id):
     with app.app_context():
+        threadctx.request_id = request_id
         logger.debug("Starting host staleness update thread")
         try:
-            logger.info(f"Querying hosts for org_id: {org_id}")
-            hosts_query = Host.query.filter(Host.org_id == org_id)
+            logger.info(f"Querying hosts for org_id: {identity.org_id}")
```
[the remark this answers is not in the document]
jpramos123: Definitely

kruai (approved):

thearifismail (dismissed):
LGTM

api/staleness.py
jpramos123: Updated it.

This was generated with AI btw! I just made some adjustments :)

kruai (approved):

thearifismail (dismissed):

fstavela:
> > * The `updated` timestamp in the kafka events is inconsistent. Should the `updated` timestamp on a host change when we change the custom staleness settings? Right now the `updated` timestamp stays the same (it doesn't change) when I create new staleness config, but the `updated` timestamp changes (at least it is changed in the events) when I update or delete the staleness config.
> 
> No, I don't think we should change it. I will make sure it does not happen

This is still an issue. When I create a new staleness config, the `updated` timestamp on hosts stays the same. But, when I update or delete the staleness, the `updated` timestamp changes.

PR #2443, opened 2025-04-15
Automerge dependabot PRs

ezr-ondrej:
Add GH action to automerge dependabot PRs.

To work, we'll need to add a Personal access token or create an Github Application.


.github/workflows/automerge.yaml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  dependabot-automerge:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Enable auto-merge for Dependabot PRs
+        if: ${{ contains(github.event.pull_request.labels.*.name, 'dependencies') &&
+                 (contains(github.event.pull_request.title, 'bump') &&
+                 (contains(github.event.pull_request.title, 'patch') ||
+                  contains(github.event.pull_request.title, 'minor')))
```
kruai, line 18: Do you have an example of this? I don't think our dependabot PRs ever have "patch" or "minor" in the title

.github/workflows/automerge.yaml
```diff
@@ -0,0 +1,25 @@
...
+             }}
+        run: |
+          gh pr review --approve "$PR_URL"
+          gh pr merge --auto --merge "$PR_URL"
```
kruai, line 22: Should be fine, but we also have automerge set up in the repo now; I assume these won't conflict, but I do wonder if it's needed

ezr-ondrej: this is set to work with automerge. Without automerge this action would need to wait for the CI, with automerge the action only sets it to merge once CI passes :)

ezr-ondrej:
That was lazy engineering by me, I've copied action from place where update type is in the title 😮‍💨
I've fixed that, [fetch-metadata action](https://github.com/dependabot/fetch-metadata?tab=readme-ov-file) helped out to get info about the type of update :)
I've requested a fineGrain token and IMO this is ready once the token is approved ;)

.github/workflows/automerge.yaml
ezr-ondrej: yup, just my lazy engineering, I've improved that :)

thearifismail:
When is this one needed, when repo level automerge is already in place?

ezr-ondrej:
> When is this one needed, when repo level automerge is already in place?

Automerge is bit confusing term chosen by GitHub, but the repo level automerge only gives a reviewer option to mark the PR as "good to merge once CI is green".
In contrast this adds a workflow to autoapprove dependabot PRs without a person manual intervention.

kruai (approved):

FabriciaDinizRH (approved):
The PR builds are breaking for a couple of PRs, so this one is LGTM

PR #2466, opened 2025-04-24
Add subscription_manager_id filter to DELETE /hosts endpoint (RHINENG-17385)

adarshdubey-star:
# Overview

This PR is being created to address [RHINENG-17385](https://issues.redhat.com/browse/RHINENG-17385).
* The subscription_manager_id can now be used to fetch and delete a host as a filter

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Add support for filtering hosts by subscription_manager_id in the DELETE and GET /hosts endpoints

New Features:
- Introduce a new query parameter 'subscription_manager_id' to filter hosts in both GET and DELETE host endpoints

Enhancements:
- Update API specification, query filtering, and host list retrieval to support subscription_manager_id as a filtering option


ezr-ondrej:
The code looks great to me, but the failures are strange :thinking: Do we need to adjust IQE to know about the new parameter?

FabriciaDinizRH (dismissed):
LGTM

adarshdubey-star:
I missed some ref the test cases caught , I have updated it

jpramos123 (commented):
Should we write unit tests or are we handling everything to IQE?

I like having unit tests!

adarshdubey-star:
> Should we write unit tests or are we handling everything to IQE?
> 
> I like having unit tests!

Well I have written a a test case to handle the scenario in IQE , I can add a unit test case as well if needed

tests/test_api_hosts_delete.py
```diff
@@ -137,6 +137,14 @@ def test_create_then_delete_without_insights_id(
...
 
 
+def test_create_then_delete_with_subscription_manager_id(api_delete_filtered_hosts):
+    host = db_host()
+    response_status, _ = api_delete_filtered_hosts(
+        query_parameters={"subscription_manager_id": host.canonical_facts["subscription_manager_id"]}
+    )
+    assert_response_status(response_status, expected_status=202)
+
+
```
fstavela, line 147: I think I would rename this test to something like `test_delete_hosts_filtered_by_subscription_manager_id`. Also, Can we add something that will check that the host was actually deleted? Maybe assertion of that some DB method was called, or something else, hopefully some other tests are doing something like this as well. I think we could also add a check (in the same, or in a different new test) for the MQ delete event produced by this action.

I just noticed that in some other tests we also check the response of this API: https://github.com/RedHatInsights/insights-host-inventory/blob/439fbd39341161f3d538835c7a181ecfef0ef2d6/tests/test_api_hosts_delete.py#L567

We can add a similar check here as well, but for me it's very surprising that this endpoint returns any data. The return data are not in the API docs, so I didn't even know that this endpoint returns anything until now. I think the expected response should be added to the documentation (probably in a different PR).

jpramos123: Agreed! It will be beneficial to have the tests validating the deletion actually happened, and maybe validate that at least the delete notification was called

adarshdubey-star: I have updated the test to check all the above

fstavela: Thank you! Can you please create one more host at the beginning of the test, with different subscription_manager_id, to make sure that only the one with matching ID is deleted?

jpramos123 (approved):
LGTM!

I would implement Feros suggestion, but not blocking

ezr-ondrej (approved):
Thank you @adarshdubey-star ! 🧡 Will wait with merge for tomorrow morning, or feel free to merge yourself once online.

tests/test_api_hosts_delete.py
adarshdubey-star: done

adarshdubey-star:
> LGTM!
> 
> I would implement Fero's suggestion, but not blocking

I have updated it

PR #2517, opened 2025-05-09
fix: Update job avoid OOM

jpramos123:
# Overview

This PR is being created to address [RHINENG-15659](https://issues.redhat.com/browse/RHINENG-15659).

This code get all hosts ids that need update, loop through them updating and commiting in batches of 500 hosts per iteration.

Using `yield_per` and `session.commit` was causing the OOM issue, because, all changes are held in memory, and are released only after `session.commit()` is called. So `yield_per` was querying in batches correctly, but the ORM session was holding all the Host changes in memory until the commit.

Now using `yield_per` with `session.flush()` seems to work just fine. The max memory I reached for 14k hosts locally was 95mb.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Update job to limit the number of hosts processed to avoid out-of-memory (OOM) issues

Bug Fixes:
- Prevent potential memory exhaustion by processing hosts in batches

Enhancements:
- Added configurable limit for number of hosts processed in a single job run

Chores:
- Added error handling for individual host updates

### DIFF
--- a/update_edge_hosts_prs.py
+++ b/update_edge_hosts_prs.py
@@ -23,7 +23,7 @@ PROMETHEUS_JOB = "inventory-update-edge-hosts-prs"
 LOGGER_NAME = "update-edge-hosts-prs"
 RUNTIME_ENVIRONMENT = RuntimeEnvironment.JOB
 
-UPDATE_EDGE_HOSTS_STALENESS = int(os.getenv("UPDATE_EDGE_HOSTS_STALENESS", 100))
+NUM_EDGE_HOSTS_STALENESS = int(os.getenv("NUM_EDGE_HOSTS_STALENESS", 500))
 
 
 @sa.event.listens_for(Host, "before_update")
@@ -44,8 +44,9 @@ def receive_before_host_update(mapper: Mapper, connection: Connection, host: Hos
     :param host: The Host object being updated.
     """
     host_details = sa.inspect(host)
-    flag_changed, _, _ = host_details.attrs.per_reporter_staleness.history
-    if flag_changed:
+    prs_changed, _, _ = host_details.attrs.per_reporter_staleness.history
+    staleness_changed, _, _ = host_details.attrs.deletion_timestamp.history
+    if prs_changed or staleness_changed:
         orm.attributes.flag_modified(host, "modified_on")
 
 
@@ -64,19 +65,47 @@ def run(logger: Logger, session: Session, application: FlaskApp):
             "discovery",
             "satellite",
         ]
-        query_filter = and_(
-            Host.system_profile_facts.has_key("host_type"),
-            or_(~Host.per_reporter_staleness[reporter].has_key("culled_timestamp") for reporter in reporters_list),
+        query_filter = or_(
+            and_(
+                Host.system_profile_facts.has_key("host_type"),
+                or_(~Host.per_reporter_staleness[reporter].has_key("culled_timestamp") for reporter in reporters_list),
+            ),
+            and_(Host.system_profile_facts.has_key("host_type"), Host.stale_warning_timestamp == None),  # noqa: E711
         )
-        query = session.query(Host).filter(query_filter)
-        num_edge_hosts = query.count()
 
-        if num_edge_hosts > 0:
-            logger.info(f"There are still {num_edge_hosts} to be updated")
-            for host in query.order_by(Host.org_id).yield_per(UPDATE_EDGE_HOSTS_STALENESS):
+        hosts_query = session.query(Host).filter(query_filter).order_by(Host.org_id)
+        hosts_count = hosts_query.count()
+
+        logger.info(f"Found {hosts_count} to be updated")
+        processed_in_current_batch = 0
+        total_hosts_updated = 0
+
+        for host in hosts_query.yield_per(NUM_EDGE_HOSTS_STALENESS):
+            try:
                 host._update_all_per_reporter_staleness()
                 host._update_staleness_timestamps()
-            session.commit()
+
+                processed_in_current_batch += 1
+                total_hosts_updated += 1
+                if processed_in_current_batch >= NUM_EDGE_HOSTS_STALENESS:
+                    logger.info(f"Updating batch of {processed_in_current_batch} hosts...")
+                    session.flush()  # Flush current session so we free memory
+                    logger.info(f"Flushed. Total updated so far: {total_hosts_updated}")
+                    processed_in_current_batch = 0  # Reset counter for the next batch
+
+            except Exception as e:
+                logger.error(f"Error committing batch: {e}", exc_info=True)
+                session.rollback()
+
+        if total_hosts_updated > 0:
+            try:
+                logger.info(f"Committing {total_hosts_updated} hosts.")
+                session.commit()
+                logger.info(f"Job finished. Successfully updated staleness for {total_hosts_updated} hosts.")
+            except Exception:
+                session.rollback()
+        else:
+            logger.info("No hosts to be updated. Finishing job")
 
 
 if __name__ == "__main__":


jpramos123:
@sourcery-ai review

update_edge_hosts_prs.py
```diff
@@ -73,11 +73,20 @@ def run(logger: Logger, session: Session, application: FlaskApp):
 
         if num_edge_hosts > 0:
             logger.info(f"There are still {num_edge_hosts} to be updated")
-            for host in query.order_by(Host.org_id).yield_per(UPDATE_EDGE_HOSTS_STALENESS):
-                host._update_all_per_reporter_staleness()
-                host._update_staleness_timestamps()
+            updated_count = 0
+            for host in query.order_by(Host.org_id).limit(NUM_EDGE_HOSTS_STALENESS):
```
thearifismail, line 77: What happens when there more edge hosts than `NUM_EDGE_HOSTS_STALENESS`?  
Shouldn't line 74 be `while num_edge_hosts > 0`?
After processing the `NUM_EDGE_HOSTS_STALENESS` hosts, get more hosts and go into the for another batch

update_edge_hosts_prs.py
```diff
@@ -23,7 +23,7 @@
 LOGGER_NAME = "update-edge-hosts-prs"
 RUNTIME_ENVIRONMENT = RuntimeEnvironment.JOB
 
-UPDATE_EDGE_HOSTS_STALENESS = int(os.getenv("UPDATE_EDGE_HOSTS_STALENESS", 100))
+NUM_EDGE_HOSTS_STALENESS = int(os.getenv("NUM_EDGE_HOSTS_STALENESS", 4000))
```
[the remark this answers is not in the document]
jpramos123: The problem was actually with the `session.commit()` that accumulates the updated hosts in memory until it is called.
So independently of the chuck size, it was holding all updated hosts in memory until calling `session.commit()`.

Tried using `session.commit()` inside a loop using `yield_per`, but got a error raising `error closing cursor`, which means that the cursor that was created by `yield_per` in the DB side, was lost after the `session.commit()` was called.

This approach, make sure we get the `NUM_EDGE_HOSTS_STALENESS`, update it with a memory safe approach, making sure all hosts are updated by different CronJobs

kruai (changes requested):
By using `limit` instead of `yield_per`, this will now only update 4000 hosts and then exit. Also, it will load all 4000 hosts in at once, so I think the memory issue would actually be worse, unless I'm missing something

thearifismail:
> By using `limit` instead of `yield_per`, this will now only update 4000 hosts and then exit. Also, it will load all 4000 hosts in at once, so I think the memory issue would actually be worse, unless I'm missing something

I thought the OOM arose because more than 4000 hosts were updated by setting the size as environment variable

update_edge_hosts_prs.py
jpramos123: Using this approach, and using chunks of 500 hosts per iteration

update_edge_hosts_prs.py
```diff
@@ -68,15 +69,31 @@ def run(logger: Logger, session: Session, application: FlaskApp):
...
                 host._update_all_per_reporter_staleness()
                 host._update_staleness_timestamps()
-            session.commit()
+                updated_count += 1
+            try:
+                session.commit()
+                logger.info(f"Committed batch of {len(hosts_to_process)} hosts.")
+            except Exception as e:
+                logger.error(f"Error committing batch: {e}", exc_info=True)
+                session.rollback()
```
FabriciaDinizRH, line 94: Just a question: when you call `session.rollback()`, where will the session go back to? Like, will it discard all previously updated hosts? And how will it pick the work back up after the rollback?

jpramos123: when we call rollback, it means the job will finish with an error, and a new one will start from scratch.
So, to answer your questions:
1) The session does not go back in the same job
2) Yes, we will discard the previously updated hosts of the CURRENT iteration in the for loop
3) As new job will run in this case, we get back from the last successfully updated hosts, or the missing hosts to be udpated.

kruai, line 94: This is fine, but be aware that we have the `session_guard` context manager to automatically take care of this kind of commit-and-rollback behavior

update_edge_hosts_prs.py
```diff
@@ -68,15 +69,31 @@ def run(logger: Logger, session: Session, application: FlaskApp):
...
             or_(~Host.per_reporter_staleness[reporter].has_key("culled_timestamp") for reporter in reporters_list),
         )
-        query = session.query(Host).filter(query_filter)
-        num_edge_hosts = query.count()
 
-        if num_edge_hosts > 0:
-            logger.info(f"There are still {num_edge_hosts} to be updated")
-            for host in query.order_by(Host.org_id).yield_per(UPDATE_EDGE_HOSTS_STALENESS):
+        host_ids_query = session.query(Host.id).filter(query_filter).order_by(Host.org_id, Host.id)
+        all_host_ids = [item[0] for item in host_ids_query.all()]
```
kruai, line 74: This is going to be a huge list; how many records do you expect this will query in Prod?

kruai (changes requested):
This is more complex than the `yield_per` approach, but feels like it's using similar logic, except that it keeps a big list of IDs in memory. If the issue was just that we shouldn't wait until the end to commit all of the changes at once, you could also just add a counter and commit after each batch. For instance:
```
count = 0
for host in query.order_by(Host.org_id).yield_per(UPDATE_EDGE_HOSTS_STALENESS):
    host._update_all_per_reporter_staleness()
    host._update_staleness_timestamps()
    count += 1
    if count % UPDATE_EDGE_HOSTS_STALENESS == 0:
        session.commit()

session.commit()
```

thearifismail:
@jpramos123 I have similiar concerns as @kruai but feel that a lot of time has been spent based on reviewing this PR.  IMO the best way is to create an image and then use this to deploy host inventory in ephemeral cluster.  Before creating an image, update the code to log the size of memory in each iteration.  Add each iteration memory to the previous iternation until session.commit()  is called.  Based on this number, you may be able to guess the batch size to use in Prod deployment.

update_edge_hosts_prs.py
ezr-ondrej: Yup, even if it's just ids, it's still 4million integers.

ezr-ondrej (commented):
Sorry for being late to the party, but why aren't we using offset and limit for batching? 🤔 
It sounds like a perfect tool for the job.

update_edge_hosts_prs.py
jpramos123: It is a small set, remembering this is only for edge hosts.

In prod, I see 14094 hosts

update_edge_hosts_prs.py
jpramos123: Thanks for letting me know

jpramos123:
> This is more complex than the `yield_per` approach, but feels like it's using similar logic, except that it keeps a big list of IDs in memory. If the issue was just that we shouldn't wait until the end to commit all of the changes at once, you could also just add a counter and commit after each batch. For instance:
> 
> ```
> count = 0
> for host in query.order_by(Host.org_id).yield_per(UPDATE_EDGE_HOSTS_STALENESS):
>     host._update_all_per_reporter_staleness()
>     host._update_staleness_timestamps()
>     count += 1
>     if count % UPDATE_EDGE_HOSTS_STALENESS == 0:
>         session.commit()
> 
> session.commit()
> ```

What made me not use `yield_per` is that after the commit, the session is closed on the client side (HBI) and is kept open in the DB side, and when I try to perform another query in the loop, the session is already closed. That is why I'm not using it over here.

jpramos123:
> > > This is more complex than the `yield_per` approach, but feels like it's using similar logic, except that it keeps a big list of IDs in memory. If the issue was just that we shouldn't wait until the end to commit all of the changes at once, you could also just add a counter and commit after each batch. For instance:
> > > ```
> > > count = 0
> > > for host in query.order_by(Host.org_id).yield_per(UPDATE_EDGE_HOSTS_STALENESS):
> > >     host._update_all_per_reporter_staleness()
> > >     host._update_staleness_timestamps()
> > >     count += 1
> > >     if count % UPDATE_EDGE_HOSTS_STALENESS == 0:
> > >         session.commit()
> > > 
> > > session.commit()
> > > ```
> > 
> > 
> > What made me not use `yield_per` is that after the commit, the session is closed on the client side (HBI) and is kept open in the DB side, and when I try to perform another query in the loop, the session is already closed. That is why I'm not using it over here.
> 
> Oh, true - you'd have to use `session.flush()` and `session.expunge_all()` instead of `session.commit()` because we want to keep the session open

Worked this way! Thanks

update_edge_hosts_prs.py
```diff
@@ -64,19 +65,48 @@ def run(logger: Logger, session: Session, application: FlaskApp):
...
+                    logger.info(f"Flushed. Total updated so far: {total_hosts_updated}")
+                    processed_in_current_batch = 0  # Reset counter for the next batch
+
+            except Exception as e:
+                logger.error(f"Error committing batch: {e}", exc_info=True)
+                session.rollback()
+
+        if total_hosts_updated > 0:
+            try:
+                logger.info(f"Commiting {total_hosts_updated} hosts.")
```
kruai, line 102:
```suggestion
                logger.info(f"Committing {total_hosts_updated} hosts.")
```

update_edge_hosts_prs.py
```diff
@@ -64,19 +65,48 @@ def run(logger: Logger, session: Session, application: FlaskApp):
...
+                session.commit()
+                session.expunge_all()
```
kruai, line 104: I had meant we could call this after `flush()` to free up memory after a batch. I don't think it would do anything after we've already committed

jpramos123: Got it, using `expunge_all()` after the `flush()` led to 
```
Object cannot be converted to ‘persistent’ state, as this identity map is no longer valid.
```

But looks like, just using `flush` we are able to free memory.

update_edge_hosts_prs.py
```diff
@@ -64,19 +65,48 @@ def run(logger: Logger, session: Session, application: FlaskApp):
...
         ]
-        query_filter = and_(
-            Host.system_profile_facts.has_key("host_type"),
-            or_(~Host.per_reporter_staleness[reporter].has_key("culled_timestamp") for reporter in reporters_list),
+        query_filter = or_(
+            and_(
+                Host.system_profile_facts.has_key("host_type"),
+                or_(~Host.per_reporter_staleness[reporter].has_key("culled_timestamp") for reporter in reporters_list),
+            ),
+            and_(Host.system_profile_facts.has_key("host_type"), or_(Host.stale_warning_timestamp == None)),  # noqa: E711
```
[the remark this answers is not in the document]
jpramos123: True! Thanks

update_edge_hosts_prs.py
kruai: Ah OK, hopefully that will be enough

kruai (approved):

FabriciaDinizRH (approved):
LGTM!

PR #2624, opened 2025-06-02
fix(RHINENG-18473): allow RBAC equal operation

ezr-ondrej:
equal operation was not used until now and we have not allowed its usage. This was wrong per the spec https://github.com/RedHatInsights/insights-rbac/blob/master/docs/source/specs/openapi.json#L3854-L3856

This commit is fixing this and handles equal operation gracefully.

This PR is being created to address [RHINENG-18473](https://issues.redhat.com/browse/RHINENG-18473).

## Summary by Sourcery

Add support for the RBAC “equal” operation in permission filtering and ensure single values are handled correctly.

New Features:
- Allow RBAC attributeFilter.operation “equal” alongside “in”.

Bug Fixes:
- Prevent abort when encountering an “equal” operation by treating its value as a singleton list.

Enhancements:
- Extract the operation into a variable and update validation logic to include “equal”.
- Convert the single value from an equal operation into a list before UUID validation.

Tests:
- Add a test case to verify that invalid UUIDs in an equal operation return a 503 error.


tests/test_rbac.py
```diff
@@ -81,6 +81,22 @@ def test_RBAC_invalid_UUIDs(mocker, api_get):
...
+    mock_rbac_response = create_mock_rbac_response(
+        "tests/helpers/rbac-mock-data/inv-hosts-read-resource-defs-template.json"
+    )
+    mock_rbac_response[0]["resourceDefinitions"][1]["attributeFilter"]["value"] = "not a UUID"
+
+    get_rbac_permissions_mock.return_value = mock_rbac_response
+
+    response_status, _ = api_get(build_hosts_url())
+
+    assert_response_status(response_status, 503)
```
kruai, line 97: This test works for catching invalid UUIDs, but we also need a test that confirms that using `equal` with a valid UUID grants permission to that group, and doesn't result in a 503.

kruai (changes requested):

tests/test_rbac.py
ezr-ondrej: correct, we don't have that test for any groups today if I'm reading correctly. I can try to come up with a test for it :)

ezr-ondrej: A'right finally got around to fixing this :)

kruai (approved):

PR #2729, opened 2025-07-02
RHINENG-18151: Check hosts first before patching a group

thearifismail:
# Overview

This PR is being created to address [RHINENG-18151](https://issues.redhat.com/browse/RHINENG-18151).  Though the defect could not be reproduced using known, this PR updates the patch group  method to avoid the situation that led to casual occurrences of the problem.  This update should be made even if the problem was not encountered by QE.

cc: @msager27 

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/api/group.py
+++ b/api/group.py
@@ -44,6 +44,7 @@ from lib.group_repository import patch_group
 from lib.group_repository import remove_hosts_from_group
 from lib.group_repository import validate_add_host_list_to_group_for_group_create
 from lib.group_repository import wait_for_workspace_event
+from lib.host_repository import find_existing_host_by_id
 from lib.metrics import create_group_count
 from lib.middleware import delete_rbac_workspace
 from lib.middleware import patch_rbac_workspace
@@ -183,12 +184,18 @@ def patch_group_by_id(group_id, body, rbac_filter=None):
         log_patch_group_failed(logger, group_id)
         abort(HTTPStatus.NOT_FOUND)
 
-    try:
-        if get_flag_value(FLAG_INVENTORY_KESSEL_WORKSPACE_MIGRATION):
-            # Update group on Kessel
-            new_group_name = validated_patch_group_data.get("name")
+    # Do not patch the group if bad hosts provided
+    if body.get("host_ids") is not None:
+        for host_id in body["host_ids"]:
+            if not find_existing_host_by_id(identity, host_id):
+                log_patch_group_failed(logger, group_id, f"not finding the host with id: '{host_id}'")
+                abort(HTTPStatus.BAD_REQUEST, f"The host with id: '{host_id}' not found")
 
-            if new_group_name:
+    # group and hosts in it are valid, let's patch the group
+    try:
+        if get_flag_value(FLAG_INVENTORY_KESSEL_WORKSPACE_MIGRATION):  # noqa: SIM102
+            # Update group on Kessel
+            if new_group_name := validated_patch_group_data.get("name"):  # noqa: SIM102
                 patch_rbac_workspace(group_id, name=new_group_name)
 
         # Separate out the host IDs because they're not stored on the Group
--- a/app/instrumentation.py
+++ b/app/instrumentation.py
@@ -288,8 +288,11 @@ def log_patch_group_success(logger, group_id):
     logger.info(f"Patched group: {group_id}")
 
 
-def log_patch_group_failed(logger, group_id):
-    logger.debug(f"Failed to find group during patch operation: {group_id}")
+def log_patch_group_failed(logger, group_id, message=None):
+    if not message:
+        logger.debug(f"Failed to find group during patch operation: {group_id}")
+    else:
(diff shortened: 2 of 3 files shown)


api/group.py
```diff
@@ -183,12 +184,18 @@ def patch_group_by_id(group_id, body, rbac_filter=None):
...
+
+    # group and hosts in it are valid, let's patch the group
     try:
-        if get_flag_value(FLAG_INVENTORY_KESSEL_WORKSPACE_MIGRATION):
+        if get_flag_value(FLAG_INVENTORY_KESSEL_WORKSPACE_MIGRATION):  # noqa: SIM102
             # Update group on Kessel
-            new_group_name = validated_patch_group_data.get("name")
-
-            if new_group_name:
+            if new_group_name := validated_patch_group_data.get("name"):  # noqa: SIM102
```
kruai, line 198: Why the `noqa`s? Best to address those when possible

api/group.py
```diff
@@ -183,12 +184,18 @@ def patch_group_by_id(group_id, body, rbac_filter=None):
         log_patch_group_failed(logger, group_id)
         abort(HTTPStatus.NOT_FOUND)
 
+    # Do not patch the group if bad hosts provided
+    if body.get("host_ids") is not None:
+        for host_id in body["host_ids"]:
+            if not find_existing_host_by_id(identity, host_id):
+                log_patch_group_failed(logger, group_id, f"not finding the host with id: '{host_id}'")
+                abort(HTTPStatus.BAD_REQUEST, f"The host with id: '{host_id}' not found")
```
kruai, line 192: A couple of things:

- I think we should add this check in `lib.group_repository._add_hosts_to_group`, so adding a host to a group always fails if the host doesn't exist. Putting the change here only fixes it specifically when patching the group via API.
- This makes a SQL query for each host_id in the body. It would be more efficient to query the list of host IDs together and raise an exception if the count doesn't match.

app/instrumentation.py
```diff
@@ -288,8 +288,11 @@ def log_patch_group_success(logger, group_id):
     logger.info(f"Patched group: {group_id}")
 
 
-def log_patch_group_failed(logger, group_id):
-    logger.debug(f"Failed to find group during patch operation: {group_id}")
+def log_patch_group_failed(logger, group_id, message=None):
+    if not message:
+        logger.debug(f"Failed to find group during patch operation: {group_id}")
+    else:
+        logger.debug(f"Failed to patch group with id: {group_id} due to {message}")
```
kruai, line 295: I'm not a fan of the log's intent changing based on whether a message was provided. Instead, you could just set the message's default value:
```suggestion
def log_patch_group_failed(logger, group_id, message="Group not found."):
    logger.debug(f"Failed to patch group with id {group_id}: {message}")
```

tests/test_api_groups_update.py
```diff
@@ -337,6 +338,25 @@ def test_patch_group_same_hosts(
...
+    # Create a group with hosts
+    mocker.patch.object(event_producer, "write_event")
+    group = db_create_group_with_hosts("test_group", 5)
+    group_id = group.id
+    host_id_list = [str(host.id) for host in db_get_hosts_for_group(group_id)]
+
+    import uuid
+
+    phantom_host_id = uuid.uuid4()
+    host_id_list.append(str(phantom_host_id))
```
kruai, line 353:
```suggestion
    host_id_list.append(generate_uuid())
```

tests/test_api_groups_update.py
```diff
@@ -337,6 +338,25 @@ def test_patch_group_same_hosts(
         assert host["groups"][0]["id"] == str(group_id)
 
 
+def test_patch_group_with_a_bad_host(
```
kruai, line 341:
```suggestion
def test_patch_group_with_a_nonexistent_host_id(
```

kruai (changes requested):

api/group.py
thearifismail: > * `_add_hosts_to_group`

@kruai, what you're asking is already handled by https://github.com/RedHatInsights/insights-host-inventory/blob/ee367fda644cad8c63322df53d656662f98e4d68/lib/group_repository.py#L124

api/group.py
thearifismail: The linter wants to combine the both `if` statements into one but both must be evaluated separately.

app/instrumentation.py
thearifismail: Much better, thanks.

thearifismail (commented):
Addressed Asa's comments

kruai (changes requested):

PR #2740, opened 2025-07-06
PoC: Batch Insert Hosts

dkuc:
(This is not production code.  It will break a lot of stuff)
### Summary
This PR demonstrates how hosts can be updated and created in batches to greatly reduce database resource consumption. The current approach in production is very ORM-centric: a single host is queried for, marshaled into a Python object, mutated, and saved back to the database. This is wasteful in two significant ways - extra round trips and unnecessary data being loaded from the database into the service. In the new approach, we skip loading data into the service and instead send batches of hosts to be upserted.

| Old    | New |
| -------- | ------- |
| host = Host.query.filter(...).first()<br> host.modified = now() <br>host.save() | insert(Host).values(hosts).on_conflict_do_update(...).returning(Host.id)   |

There are over 450 lines changed, but most of it consists of incidental changes to support the core difference shown above in the table. There are two commits. The first - `Benchmark` - is just code that can be used to measure the change. The second commit is where the actual message processing change is implemented.

### Measuring the Performance
Run the mq service:
```
PYTHONUNBUFFERED=1 BYPASS_TENANT_TRANSLATION=true INVENTORY_LOG_LEVEL=INFO KAFKA_CONSUMER_AUTO_COMMIT_INTERVAL_MS=500 KAFKA_EVENT_TOPIC=platform.inventory.events MQ_DB_BATCH_MAX_MESSAGES=100 PAYLOAD_TRACKER_SERVICE_NAME=inventory-mq-service python inv_mq_service.py
```

Seed a realistic number of hosts into the database:
```
make run_fast_seed_hosts NUM_HOSTS=1000000
```

Measure how long it takes to process hosts. This command will print out a duration for processing 4,000 hosts:
```
make run_inv_mq_service_test_producer NUM_HOSTS=4000 USE_EXISTING_HOSTS=true
```

Compare it to without the change:
```
git checkout HEAD^
```
Restart the mq service and run the benchmark again.

### Results
On my laptop, here's the time it takes to process 4,000 hosts:

| Old    | New |
| -------- | ------- |
| 31s  |  19s  |

It comes out to about 35% faster. On its own, this is not that impressive. However, if you look more closely, this doesn't tell the whole story. It turns out the mq service is highly CPU bottlenecked. If you comment out the database inserts in both methods, you can measure just how much time it takes to read the message from Kafka, validate the data, turn it into model objects, and create the outgoing events. It turns out these non-database activities consume a majority of the time. You can then calculate how much time is spent in the database by subtracting the two.

| | Old    | New |
| ------ | -------- | ------- |
| Total Time | 31s  |  20s  |
| Non-DB Time |  15s | 15s |
| DB Time | 16s | 5s |

So if you subtract out the non-database time-consuming activities, this method is actually over 3 times faster.

## Summary by Sourcery

Replace per-host ORM calls with bulk upsert logic for hosts, refactor the MQ consumer to handle message batches, and add supporting tooling and scripts for benchmarking and seeding performance tests

New Features:
- Handle and process messages in batches in the message consumer
- Introduce `add_hosts` repository method to perform bulk upsert of hosts
- Add a SQL function `find_host_by_canonical_facts` to resolve host IDs by canonical facts
- Enhance Kafka test producer to support existing hosts, consumer lag tracking, and processing completion

Enhancements:
- Optimize the event loop to batch validate, commit, and notify messages reducing database round trips
- Improve metrics for batch success/failure counts and message sizes

Build:
- Update Makefile with `USE_EXISTING_HOSTS` flag, fast seeding, and improved mq service commands
- Adjust dev container configuration for increased shared memory and explicit Kafka image version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 
 IDENTITY_HEADER="eyJpZGVudGl0eSI6IHsiYWNjb3VudF9udW1iZXIiOiAiYWNjb3VudDEyMyIsICJvcmdfaWQiOiAiNTg5NDMwMCIsICJ0eXBlIjogIlVzZXIiLCAiYXV0aF90eXBlIjogImJhc2ljLWF1dGgiLCAidXNlciI6IHsiaXNfb3JnX2FkbWluIjogdHJ1ZSwgInVzZXJuYW1lIjogImZyZWQifSwgImludGVybmFsIjogeyJvcmdfaWQiOiAib3JnMTIzIn19fQ=="
 NUM_HOSTS=1
+USE_EXISTING_HOSTS=false
 
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 current_dir := $(dir $(mkfile_path))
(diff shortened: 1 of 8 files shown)


dkuc:
Closing to stop the rebuilds

PR #2747, opened 2025-07-08
feat(RHINENG-18981, RHINENG-18738): Adapt HostGroupAssoc, LimitedHost and Host classes to process new columns  

jpramos123:
# Overview

This PR is being created to address [RHINENG-18981](https://issues.redhat.com/browse/RHINENG-18981) and [RHINENG-18738](https://issues.redhat.com/browse/RHINENG-18738)

It was found that we needed to add org_id to the hosts_groups table (HostGroupAssoc class), in order to make the Host table partitioning. This PR adjust our code base to handle these changes. Also, it has a migration that renames the old and new tables.

LimitedHost and Host classes are adapted to process the new canonical facts columns. These columns are:

- insights_id
- subscription_manager_id
- satellite_id
- fqdn
- bios_uuid
- ip_addresses
- mac_addresses
- provider_id
- provider_type

There was no need to touch Host MQ, as we serialize the canonical facts into a dictionary already.
If these fields are not needed in the future, we can remove it.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


app/models/host.py
```diff
@@ -61,6 +61,15 @@ def __init__(
         system_profile_facts=None,
         groups=None,
         id=None,
+        insights_id=None,
```
thearifismail, line 64: what's the reason for keeping `canonical_facts` around?  Shouldn't `canonical_facts` be replaced by insights_id, subscription_manager_id, satellite_id, and the rest of the canonical_facts?

app/models/host.py
```diff
@@ -132,7 +152,7 @@ def operating_system(cls):
 
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
     account = db.Column(db.String(10))
-    org_id = db.Column(db.String(36))
+    org_id = db.Column(db.String(36), primary_key=True)
```
thearifismail, line 155: does this mean that a `host_id` and `org_id` combination will make primary keys in the hosts table?

app/models/host.py
jpramos123: This is kept because we want to make small changes, during time. The removal of `canonical_facts` and `system_profile_fatcs` JSONBs will be made on DB Refactoring Phase 2.

app/models/host.py
jpramos123: Correct! And allow the composite foreign key relationship with HostGroupAssoc

app/models/schemas.py
```diff
@@ -102,23 +102,23 @@ class Meta:
...
     is_virtual = fields.Boolean(required=False)
 
-    insights_id = fields.Str(validate=verify_uuid_format)
-    subscription_manager_id = fields.Str(validate=verify_uuid_format)
-    satellite_id = fields.Str(validate=verify_satellite_id)
-    fqdn = fields.Str(validate=marshmallow_validate.Length(min=1, max=255))
-    bios_uuid = fields.Str(validate=verify_uuid_format)
-    ip_addresses = fields.List(fields.Str(validate=verify_ip_address_format))
+    insights_id = fields.Raw(validate=verify_uuid_format, allow_none=True)
+    subscription_manager_id = fields.Str(validate=verify_uuid_format, allow_none=True)
```
kruai, line 106: Why `Raw` and `Str` instead of UUID?

lib/host_remove_duplicates.py
```diff
@@ -95,7 +95,9 @@ def unique(host_list: list[Host]) -> None:
             logger.info(f"Found {len(duplicate_host_ids)} duplicates for org_id: {actual_org_id}")
             total_deleted += len(duplicate_host_ids)
         else:
-            hosts_by_ids_query = misc_session.query(Host).filter(Host.id.in_(duplicate_host_ids))
+            hosts_by_ids_query = misc_session.query(Host).filter(
+                Host.org_id == actual_org_id, Host.id.in_(duplicate_host_ids)
+            )
```
kruai, line 100: Doesn't feel like this is related to the rest of the changes. Why is the org_id filter needed here?

tests/fixtures/db_fixtures.py
```diff
@@ -56,16 +57,18 @@ def database(database_name: None) -> Generator[str]:  # noqa: ARG001
...
 
 
 @pytest.fixture(scope="function")
-def db_get_hosts(flask_app: FlaskApp) -> Callable[[list[str]], Query]:  # noqa: ARG001
-    def _db_get_hosts(host_ids: list[str]) -> Query:
-        return Host.query.filter(Host.id.in_(host_ids))
+def db_get_hosts(flask_app: FlaskApp) -> Callable[[list[str], str | None], Query]:  # noqa: ARG001
+    def _db_get_hosts(host_ids: list[str], org_id: str | None = None) -> Query:
+        org_id = org_id or SYSTEM_IDENTITY["org_id"]
+        return Host.query.filter(Host.org_id == org_id, Host.id.in_(host_ids))
```
kruai, line 71: Why is an org_id filter being added to the fixtures that fetch Hosts based on their Host IDs?

tests/fixtures/db_fixtures.py
```diff
@@ -128,8 +131,13 @@ def _db_get_hosts_for_group(group_id):
...
 def db_get_groups_for_host(flask_app):  # noqa: ARG001
-    def _db_get_groups_for_host(host_id):
-        return Group.query.join(HostGroupAssoc).filter(HostGroupAssoc.host_id == host_id).all()
+    def _db_get_groups_for_host(host_id, org_id=None):
+        org_id = org_id or SYSTEM_IDENTITY["org_id"]
+        return (
+            Group.query.join(HostGroupAssoc)
+            .filter(HostGroupAssoc.org_id == org_id, HostGroupAssoc.host_id == host_id)
+            .all()
+        )
```
kruai, line 140: Same Q here

tests/fixtures/db_fixtures.py
```diff
@@ -150,6 +158,24 @@ def _db_create_host(
...
+        extra_data = extra_data or {}
+        org_id = extra_data.pop("org_id", None) or identity["org_id"]
+        host = host or db_host_with_custom_canonical_facts(
+            org_id=org_id, account=identity["account_number"], **extra_data
+        )
+        db.session.add(host)
+        db.session.commit()
+        return host
+
+    return _create_host_custom_canonical_facts
```
kruai, line 176: This fixture, `db_host_with_custom_canonical_facts`, and `db_host_with_custom_canonical_facts_dict` are almost exact copies of `db_create_host`, `minimal_db_host`, and `minimal_db_host_dict`, respectively. As far as I can tell, the only difference is that `minimal_db_host_dict` generates a UUID for `canonical_facts.insights_id`.
Instead of this duplication, how about just updating `minimal_db_host_dict` so it doesn't generate an insights ID if at least one canonical fact is provided?

tests/test_synchronizer.py
```diff
@@ -99,7 +99,9 @@ def test_synchronize_grouped_host_event(
     db_create_host_group_assoc(created_host_id, created_group_id)
 
     # Overwrite Host.groups data
-    db.session.query(Host).filter(Host.id == created_host_id).update({"groups": []})
+    db.session.query(Host).filter(Host.org_id == created_host.org_id, Host.id == created_host_id).update(
+        {"groups": []}
+    )
```
kruai, line 104: Why is filtering on the org_id necessary here?

tests/test_models.py
```diff
@@ -1352,3 +1352,59 @@ def test_create_host_validate_staleness(db_create_host, db_get_host):
...
+def test_create_host_with_missing_canonical_facts(db_create_host_custom_canonical_facts, db_get_host):
+    canonical_facts = {
+        "insights_id": generate_uuid(),
+        "subscription_manager_id": generate_uuid(),
+        "satellite_id": generate_uuid(),
+        "fqdn": "test.fqdn",
+        "bios_uuid": generate_uuid(),
+        "provider_id": "test_provider",
+        "provider_type": "test_provider_type",
+    }
```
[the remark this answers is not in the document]
kruai: I agree with this suggestion

kruai (changes requested):
A few questions and concerns

lib/host_remove_duplicates.py
rodrigonull: Without the org_id filter, PostgreSQL has no idea which one of the partitions to look in for the host_ids. It would be forced to search every single partition.

app/models/schemas.py
jpramos123: I made it like this so we are able to handle the validation when the host is already deserialized (in a dict, with all values as string), and also when we are validating a Host object, and in this case this value is a UUID.

tests/fixtures/db_fixtures.py
rodrigonull: Same answer applies to all org_id filter questions.

tests/test_synchronizer.py
rodrigonull: Same answer applies to all org_id filter questions.

kruai (approved):

PR #2840, opened 2025-08-01
feat(RHINENG-19061): Migration to recreate the new system profile partitioned tables

rodrigonull:
# Overview

This PR was created to address some issues found during the analysis to create a new publication to enable Postgres Logical replication using the new system_profile_* partitioned tables.

The PR uses the same env var used for the hosts table to determine the number of partitions that will be created `HOSTS_TABLES_NUM_PARTITIONS`. In the previous migration we introduced a new env var for that purpose and we forgot to add it to clowdapp which made it create only a single partition on stage (it should be 4 to keep it sync with the other partitioned tables).

The PR adds the insights_id column to both system_profile tables as we discovered that this is a requirement to create a filtered publication. This column will be ONLY used for filtering, so we don't need to add it to the system profile models.

A new trigger was introduced to synchronize the insights_id value from the hosts table with the two new system_Profile tables.

A new index was added to the workloads column in the system_profile_dynamic table.

New indexes were added for the logical replication.

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Recreate the system_profiles_static and system_profiles_dynamic tables with the correct number of partitions driven by HOSTS_TABLES_NUM_PARTITIONS, add an insights_id column with synchronization triggers for filtered logical replication, and create supporting indexes.

Enhancements:
- Drop any empty legacy system_profiles tables and recreate them as HASH-partitioned tables based on the HOSTS_TABLES_NUM_PARTITIONS environment variable
- Introduce an insights_id column to both static and dynamic profile tables and synchronize its value from the hosts table via a PL/pgSQL trigger function
- Add new indexes including a GIN index on dynamic.workloads and unique replica identity indexes on both tables to support logical replication

### DIFF
--- a/migrations/versions/3b60b7daf0f2_partitioned_system_profiles_tables_fix.py
+++ b/migrations/versions/3b60b7daf0f2_partitioned_system_profiles_tables_fix.py
@@ -298,7 +298,6 @@ def upgrade():
         $$ LANGUAGE plpgsql;
     """)
 
-    # The WHEN clause is removed, and the logic is moved into the function.
     op.execute(f"""
         CREATE TRIGGER trigger_sync_insights_id
         AFTER INSERT OR UPDATE OF insights_id ON {INVENTORY_SCHEMA}.hosts
--- a/migrations/versions/3b60b7daf0f2_partitioned_system_profiles_tables_fix.py
+++ b/migrations/versions/3b60b7daf0f2_partitioned_system_profiles_tables_fix.py
@@ -308,8 +308,9 @@ def upgrade():
 
 def downgrade():
     """
-    Removes the system_profiles_static and system_profiles_dynamic tables
-    and their partitions, as well as the synchronization trigger.
+    Removes only the synchronization trigger. Drop table commands are omitted because
+    previous migrations still have the system_profiles_static and system_profiles_dynamic tables.
+    Downgrading 2 revisions would result in it trying to drop these tables twice.
     """
     op.execute(f"DROP TRIGGER IF EXISTS trigger_sync_insights_id ON {INVENTORY_SCHEMA}.hosts;")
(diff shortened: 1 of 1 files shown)


migrations/versions/3b60b7daf0f2_partitioned_system_profiles_tables_fix.py
```diff
@@ -0,0 +1,317 @@
...
+
+def downgrade():
+    """
+    Removes the system_profiles_static and system_profiles_dynamic tables
+    and their partitions, as well as the synchronization trigger.
+    """
+    op.execute(f"DROP TRIGGER IF EXISTS trigger_sync_insights_id ON {INVENTORY_SCHEMA}.hosts;")
+    op.execute(f"DROP FUNCTION IF EXISTS {INVENTORY_SCHEMA}.sync_insights_id_to_profiles();")
+    op.drop_table(SP_STATIC_TABLE_NAME, schema=INVENTORY_SCHEMA)
+    op.drop_table(SP_DYNAMIC_TABLE_NAME, schema=INVENTORY_SCHEMA)
```
kruai, line 317: I don't think these should be dropped. Migration `d705de909597`, which this revises, still contains those tables. So downgrading 2 revisions would result in it trying to drop these tables twice
```suggestion
```

kruai (changes requested):

kruai (dismissed):

jpramos123 (commented):
One doubt

kruai (approved):

PR #2925, opened 2025-08-18
fix(RHINENG-20208): operating_system endpoint only fetch distinct items

ezr-ondrej:
# Overview

This PR is being created to address performance issues of calculating OS counts in python over the DB.
Lets try moving back to DB.
We've moved to python calculation in #1723 this should be more performant than a subquery, plus our DB is much more performant now.

## Summary by Sourcery

Move operating_system count computation to the database by leveraging SQL aggregation, grouping, and ordering to improve performance and remove Python-level post-processing.

Enhancements:
- Compute and fetch distinct operating_system items with counts using SQL aggregation and grouping
- Remove Python-level post-processing for OS counts and parse results directly from query results
- Include identity parameter in the host entities DB query to ensure proper filtering and authorization

### DIFF
--- a/api/host_query_db.py
+++ b/api/host_query_db.py
@@ -6,7 +6,6 @@ from itertools import islice
 from typing import Any
 
 from sqlalchemy import Boolean
-from sqlalchemy import String
 from sqlalchemy import case
 from sqlalchemy import func
 from sqlalchemy import select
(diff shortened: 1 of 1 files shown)


fstavela (approved):
I don't know if a DB query will be more performant than Python, but I think it makes sense. LGTM!

rodrigonull (approved):

kruai (approved):
Nice, LGTM!

kruai:
The RBAC checks are failing on all PRs, so I force-merged this one.

PR #2983, opened 2025-09-10
fix(RHINENG-1896): Fix os kernel length size

jpramos123:
# Overview

This PR is being created to address [RHINENG-18969](https://issues.redhat.com/browse/RHINENG-18969).
Revert the os kernel length size to 20.
Fix the test helper

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

## Summary by Sourcery

Revert the os_kernel_version length to 20 across the database model, OpenAPI spec, and test data, and enhance string field validation by applying Regexp for non-UUID patterns alongside the existing UUID format check.

Bug Fixes:
- Revert os_kernel_version column size from 50 to 20 in the database model
- Adjust OpenAPI spec maxLength for os_kernel_version to 20
- Update sample static profile test helper to use a kernel version within 20 characters

Enhancements:
- Apply marshmallow Regexp validator for pattern definitions that aren’t UUIDs while preserving the UUID format validator

### DIFF
--- a/dashboards/grafana-dashboard-insights-inventory-general.configmap.yaml
+++ b/dashboards/grafana-dashboard-insights-inventory-general.configmap.yaml
@@ -1761,6 +1761,168 @@ data:
           "valueName": "current"
         },
         {
+          "id": 93,
+          "type": "stat",
+          "title": "Outbox successes",
+          "gridPos": {
+            "x": 12,
+            "y": 18,
+            "h": 2,
+            "w": 3
+          },
+          "fieldConfig": {
+            "defaults": {
+              "mappings": [
+                {
+                  "type": "special",
+                  "options": {
+                    "match": "null",
+                    "result": {
+                      "text": "0",
+                      "index": 0
+                    }
+                  }
+                }
+              ],
+              "thresholds": {
+                "mode": "absolute",
+                "steps": [
+                  {
+                    "value": null,
+                    "color": "#299c46"
+                  },
+                  {
+                    "value": 0.1,
+                    "color": "rgba(237, 129, 40, 0.89)"
+                  },
+                  {
+                    "color": "#d44a3a"
+                  }
+                ]
+              },
+              "unit": "locale",
+              "decimals": 0,
+              "noValue": "0"
+            },
+            "overrides": []
+          },
+          "pluginVersion": "11.6.3",
+          "targets": [
+            {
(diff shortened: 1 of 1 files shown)


app/models/system_profile_normalizer.py
```diff
@@ -270,10 +270,11 @@ def _create_string_field(self, field_def: dict[str, Any]) -> fields.Field:
             validators.append(marshmallow_validate.Length(min=field_def["minLength"]))
 
         # Add pattern validation for UUIDs
-        if "pattern" in field_def and (
-            "uuid" in field_def.get("description", "").lower() or field_def["pattern"].startswith("[0-9a-f]{8}-")
-        ):
-            validators.append(verify_uuid_format)
+        if "pattern" in field_def:
+            if "uuid" in field_def.get("description", "").lower() or field_def["pattern"].startswith("[0-9a-f]{8}-"):
```
kruai, line 274: I know this conditional already existed prior to your change, but I must have missed it. I disagree with using different logic based on the field's `description`; we should check the `format` instead.
This can be done in a follow-up PR if you'd prefer; I don't wanna block this PR based on a change you didn't make.

jpramos123: We are already checking the format, that for UUID is `string`. Then, we need to decide what will be the validator that we're going to use. And we do have a custom validator for UUIDs.

kruai: Oh - I didn't realize we weren't using the `format` property for string fields in the System Profile spec. I was referring to how we do it in our API spec ([like this](https://github.com/RedHatInsights/insights-host-inventory/blob/master/swagger/api.spec.yaml#L1839), for example).
It looks like all the UUID fields in the SP spec are using that `pattern` property though, so we can omit the description check altogether, right?

rodrigonull (dismissed):
LGTM

app/models/system_profile_normalizer.py
jpramos123: Yes, we can remove the description check!

rodrigonull (approved):

PR #3009, opened 2025-09-15
chore(deps): Add Kessel dependencies

ezr-ondrej:
Add deps needed by #2989 in separate PR to keep the deps update cleanly separated

## Summary by Sourcery

Update dependencies for Kessel integration by bumping existing package versions in hermetic requirements and adding new Kessel gRPC and proto libraries, and synchronize these changes in the Pipfile.

Chores:
- Bump multiple Python package versions in .hermetic_builds/requirements.txt (e.g., Alembic, boto3, botocore, jsonschema, sqlalchemy, Starlette, typing-extensions)
- Add Kessel-specific gRPC and proto dependencies (inventory-grpc-clients-python-kessel-project, googleapis-common-protos, grpcio, protobuf) and pin unleashclient in the Pipfile


kruai (approved):

ezr-ondrej:
If Konflux goes green, I'll force merge this to unblock Kessel

ezr-ondrej:
> @ezr-ondrej This PR is just updating version numbers and it is failing `Required` steps. When or what step verifies that the version updates of packages, particularly those NOT in Pipfile do not encounter problem. The reason for asking this question is that sometimes package updates requires some migration or the update is NOT straight forward.

The best verification is Konflux, which is the strictest build we have and it's the one we use in deployments. So for package updates the most important is Konflux build plus enterprise contract.

Yet I expect all required pipelines to pass on every PR (at least once).
Which is now true, so this is safe to merge :)

PR #3044, opened 2025-09-25
RHINENG-20966: Add groups when none provided for creating new hosts

thearifismail:
# Overview

This PR is being created to address [RHINENG-20966](https://issues.redhat.com/browse/RHINENG-20966).

## PR Checklist

- [ ] Tests: edge cases
- [ ] Links to related PRs

## Secure Coding Checklist

- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/app/queue/host_mq.py
+++ b/app/queue/host_mq.py
@@ -64,11 +64,14 @@ from app.queue.notifications import NotificationType
 from app.queue.notifications import send_notification
 from app.serialization import deserialize_host
 from app.serialization import remove_null_canonical_facts
+from app.serialization import serialize_group_without_host_count
 from app.serialization import serialize_host
 from app.staleness_serialization import AttrDict
 from lib import group_repository
 from lib import host_repository
 from lib.db import session_guard
+from lib.feature_flags import FLAG_INVENTORY_KESSEL_WORKSPACE_MIGRATION
+from lib.feature_flags import get_flag_value
 from utils.system_profile_log import extract_host_dict_sp_to_log
 
 logger = get_logger(__name__)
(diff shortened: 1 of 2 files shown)


app/queue/host_mq.py
```diff
@@ -409,6 +410,11 @@ def process_message(
             # New hosts don't have an id, so create one
             input_host.id = uuid.uuid4() if input_host.id is None else input_host.id
 
+            # New hosts don't have a group, so create one as it is needed for rbac_v2 (workspaces)
+            if input_host.groups is None or len(input_host.groups) == 0:
```
kruai, line 414: I don't expect `input_host.groups` is ever set, since `input_host` is sent by the reporters; the host's associated group is stored in the DB.

tests/test_tags.py
```diff
@@ -223,10 +224,27 @@ def test_get_host_tags_with_RBAC_denied(subtests, mocker, api_get):
...
+                    "subscription_manager_id": generate_uuid(),
+                    "fqdn": fqdn,
+                },
+                "display_name": f"host{i}",
+                "facts": FACTS,
+                "tags": {"ns1": {"key1": ["val1", "val2"], "key2": ["val1"]}, "SPECIAL": {"tag": ["ToFind"]}},
+            }
+        )
+        host_list.append(host)
+
```
kruai, line 247: Why is this change needed?

tests/test_tags.py
```diff
@@ -242,11 +260,27 @@ def test_get_host_tag_count_RBAC_allowed(mocker, api_get, subtests, mq_create_th
...
+                    "insights_id": generate_uuid(),
+                    "subscription_manager_id": generate_uuid(),
+                    "fqdn": fqdn,
+                },
+                "display_name": f"host{i}",
+                "facts": FACTS,
+                "tags": {"ns1": {"key1": ["val1", "val2"], "key2": ["val1"]}, "SPECIAL": {"tag": ["ToFind"]}},
+            }
+        )
+        created_hosts.append(host)
```
kruai, line 283: Same Q here

kruai (changes requested):
- A bunch of unit tests are failing
- The main change in `host_mq.py` doesn't make sense to me

wscalf:
I wanted to put this out here too- if we reach the [else clause in this statement](https://github.com/RedHatInsights/insights-host-inventory/blob/add_group_to_new_host/lib/outbox_repository.py#L42), which is what's actually injecting the empty 'common' object, it's going to cause an outage every time. Making sure there are groups is good, but I wonder if we should do something there, like maybe raise an exception if we do get there in a bad state - that way we can capture information about what's happening (parameters, stack trace, etc) and prevent both the host being saved and the outbox record being written so we don't have a blocked queue or orphaned data.

We talked about this some last week, and Arif had concerns about how that would impact the mocks, I think, but it seems worth at least mentioning here.

PR #3052, opened 2025-10-01
ci(Konflux): Bump konflux pipeline versions

ezr-ondrej:
# Overview

This PR is being created to address failures in the Konflux Conforma.

We didn't get the automated bump for a while now, I wonder why :/


ezr-ondrej:
I'll go ahead and merge, this is simple bump of dependency and it fixes the Conforma (it's 🍏 now)

PR #3954, opened 2026-04-09
ci: Add konflux checks as well for autolabeling

adarshdubey-star:
## Jira
[RHINENG-25556](https://issues.redhat.com/browse/RHINENG-25556)

## What
Fix the PR labeling workflow to detect Konflux CI results in addition to GitHub Actions check suites.

## Why
Konflux reports results via commit statuses (the older GitHub API), not check suites. The existing workflow only triggered on `check_suite` events and queried the Check Suites API, so it never saw Konflux failures or successes labels were never applied.

## How
- Added `status` event trigger alongside `check_suite` so the workflow fires when Konflux reports
- Added PR lookup from commit SHA for `status` events (which don't include PR info directly)
- Query both the Check Suites API and the Commit Status API, then combine results
- Pending checks result in no label changes to prevent flickering
- Self-filtering: excludes the label workflow's own check suite from the evaluation

## Testing
- YAML validated via `check-yaml` pre-commit hook

[RHINENG-25556]: https://redhat.atlassian.net/browse/RHINENG-25556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Extend the PR auto-labeling workflow to consider both GitHub check suites and commit status APIs so labels reflect Konflux and other external CI results without flapping.

Enhancements:
- Trigger the labeling workflow on both check_suite and status events and support manual workflow_dispatch runs.
- Resolve the associated PR and head SHA for status events via the GitHub API when PR data is not provided directly.
- Aggregate results from check suites and commit statuses, excluding the workflow’s own check suite, into a single combined CI outcome used for labeling.
- Preserve existing labels when checks or statuses are still pending to avoid label flickering.


ezr-ondrej (approved):
Gotcha, nice improvement, ty! :)

PR #3975, opened 2026-04-13
perf: add lightweight request profiling middleware

rodrigonull:
## Jira
N/A — diagnostic instrumentation for performance investigation

## What
Add a lightweight request profiling middleware that logs per-request timing breakdowns for key code paths: auth (RBAC v1 / Kessel), staleness queries, host serialization, Unleash feature flag lookups, and all DB queries via SQLAlchemy engine events.

**Note:** This is temporary instrumentation. It will be reverted once we identify the source of the performance regression.

## Why
After recent deployments to stage, the IQE test suite duration nearly doubled (1h15m → 1h57m). Disabling the staleness L2 cache recovered part of the regression, but a consistent +200ms per-request overhead remains on most API endpoints. DB queries were ruled out via `EXPLAIN ANALYZE` (sub-millisecond). This profiler will identify which application-layer component (auth, feature flags, serialization, or framework overhead) is responsible.

## How
- New `api/profiling.py` module, gated behind `INVENTORY_PROFILING_ENABLED` env var (default `false`). Zero overhead when disabled.
- Monkey-patches `get_staleness_obj`, `get_rbac_filter`, `get_kessel_filter`, `serialize_host`, and `get_flag_value` with timing decorators across all importing modules.
- Hooks into SQLAlchemy `before_cursor_execute` / `after_cursor_execute` engine events to track cumulative DB time and query count.
- Logs a single DEBUG line per request: PROF GET /api/inventory/v1/hosts total=67.1ms [ staleness=3.5ms(x1) | db=13.9ms(x3) | other=49.7ms ] status=200
- `init_profiling(app)` is called from `create_app()` after `init_cache()`.
- `INVENTORY_PROFILING_ENABLED` added to all deployment blocks in `clowdapp.yml`, defaulting to `false`.

## Testing
- Unit tests pass locally (`pytest tests/test_api_hosts_get.py` — profiler disabled and enabled).
- Verified end-to-end with `docker compose -f dev.yml up` and `INVENTORY_PROFILING_ENABLED=true`: PROF log lines appear correctly for `curl` requests against `GET /hosts`.

## Summary by Sourcery

Introduce optional request profiling instrumentation to measure per-request timing of key inventory API components while keeping it disabled by default.

Enhancements:
- Add a profiling module that conditionally instruments staleness lookups, auth filters, serialization, feature flag checks, and SQLAlchemy queries to log per-request timing breakdowns.
- Wire the profiling initialization into the application startup so that timing data is collected and logged when profiling is enabled via configuration.

Deployment:
- Expose an INVENTORY_PROFILING_ENABLED environment variable across all deployment configurations, defaulting to disabled.


fstavela (commented):
Please add type annotations to `profiling.py`  (since it's a new file, you could also add `# mypy: disallow-untyped-defs` comment on top of it). Otherwise LGTM. If we want to be extra sure that it won't break things, you could temporarily set the default for `INVENTORY_PROFILING_ENABLED` to `true` and let the PR checks run, but I think it's not necessary. We can try it in Stage, and if something bad happens, it's easy to turn it off.

rodrigonull:
> Please add type annotations to `profiling.py` (since it's a new file, you could also add `# mypy: disallow-untyped-defs` comment on top of it). Otherwise LGTM. If we want to be extra sure that it won't break things, you could temporarily set the default for `INVENTORY_PROFILING_ENABLED` to `true` and let the PR checks run, but I think it's not necessary. We can try it in Stage, and if something bad happens, it's easy to turn it off.

My plan with this PR is just to merge it, trigger the tests on stage and revert it. It's temporary. Do you think we really need the type annotations?

fstavela:
> > Please add type annotations to `profiling.py` (since it's a new file, you could also add `# mypy: disallow-untyped-defs` comment on top of it). Otherwise LGTM. If we want to be extra sure that it won't break things, you could temporarily set the default for `INVENTORY_PROFILING_ENABLED` to `true` and let the PR checks run, but I think it's not necessary. We can try it in Stage, and if something bad happens, it's easy to turn it off.
> 
> My plan with this PR is just to merge it, trigger the tests on stage and revert it. It's temporary. Do you think we really need the type annotations?

OK, in that case it's not necessary.

fstavela (approved):

kruai (approved):

PR #3977, opened 2026-04-13
refactor(RHINENG-21704): remove code which reads PRS nested format

FabriciaDinizRH:
## Jira
[RHINENG-21704](https://issues.redhat.com/browse/RHINENG-21704)

## What
Removes code that reads from the nested PRS structure, since all instances in the db now have the new flattened format.

## Why
Reduce information saved to the db.

## How
Standard code clean up.

## Testing
Tests have been in place for a couple of sprints

[RHINENG-21704]: https://redhat.atlassian.net/browse/RHINENG-21704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Remove legacy support and migration job for nested per-reporter staleness data now that all records use the flattened format.

Enhancements:
- Simplify per-reporter staleness filtering to operate only on flat string values and tighten semantics around non-string entries.
- Refactor staleness timestamp helpers into a single reusable function that can compute timestamps for host-level and per-reporter use cases.
- Normalize per_reporter_staleness model values to ISO-8601 strings and validate that only flat (non-nested) values are stored.
- Update staleness utilities and imports to expose get_staleness_obj as the public helper and use it consistently across the codebase.

Build:
- Remove the flatten-per-reporter-staleness ClowdApp job definition, its CJI invocation, and associated resource parameters from deployment manifests.

Tests:
- Update staleness tests to rely on the unified timestamp helper and the new staleness utility naming, and drop tests tied to the removed flatten-per-reporter-staleness job.


FabriciaDinizRH:
@sourcery-ai review

ezr-ondrej (dismissed):
Nice clean-up, looks clean to me 👍 
I 🧡 removing 930 lines 😛

api/filtering/db_filters.py
```diff
@@ -238,28 +235,14 @@ def _stale_timestamp_per_reporter_filter(
         and_conditions = []
 
         for rep in reporter_list:
-            # For flat format: reporter is culled if computed_culled < now
             flat_format_culled = and_(
```
fstavela: I would probably rename these variables (here and in other places as well) and remove the "flat" from their name now, but it's not a blocker.

app/models/host.py
```diff
@@ -563,31 +569,23 @@ def reporter_stale(self, reporter):
         if not reporter_data:
             logger.debug("Reports from %s are stale (no check-in recorded)", reporter)
             return True
-        # Support both flat (string) and nested (dict with "last_check_in" / "stale_timestamp") formats
+
         if isinstance(reporter_data, dict):
```
fstavela: Is this necessary? If we are sure that the nested format is no longer there in Stage and Prod, I think we can remove this check.

app/models/host.py
```diff
@@ -563,31 +569,23 @@ def reporter_stale(self, reporter):
...
-            from app.common import inventory_config
-            from app.culling import Timestamps
-
-            staleness_ts = Timestamps.from_config(inventory_config())
-            staleness = get_staleness_obj(self.org_id)
+            raise ValidationException(
+                f"Invalid per_reporter_staleness for reporter {reporter!r}: nested object is no longer supported; "
+                "use a flat ISO-8601 last_check_in string."
+            )
+        if not isinstance(reporter_data, str):
```
fstavela: Same here

tests/test_models.py
```diff
@@ -585,6 +586,71 @@ def test_update_per_reporter_staleness_yupana_replacement(db_create_host, models
...
+    host.per_reporter_staleness[reporter] = {"legacy_nested": "2020-01-01T00:00:00+00:00"}
+    with pytest.raises(ValidationException, match="nested object is no longer supported"):
+        host.reporter_stale(reporter)
+
+
+def test_reporter_stale_non_string_scalar_is_stale(db_create_host):
+    host = db_create_host()
+    reporter = host.reporter
+    host.per_reporter_staleness[reporter] = 42
+    assert host.reporter_stale(reporter) is True
```
fstavela: I'm not strictly against having these tests, but I think we should make sure that this situation is impossible to happen, and then we don't need them.

For the second test (`test_reporter_stale_non_string_scalar_is_stale`), if we decide to keep it, I think it would be better if this situation errored out rather than said the host is stale.

app/models/host.py
FabriciaDinizRH: Yeah, I'm still a tiny bit paranoid about wild pieces of data, but I'll remove these checks

jpramos123 (approved):
It LGTM!

fstavela (approved):
LGTM, thank you!

PR #3979, opened 2026-04-13
fix(RHINENG-25789): Sync workspace create/update timestamps to groups

kruai:
## Jira
[RHINENG-25789](https://redhat.atlassian.net/browse/RHINENG-25789)

## What
Syncs the created/updated timestamps from Workspace mq messages to the corresponding Group.

## Why
To keep the timestamps in sync between Workspaces and Groups.

## How
Instead of always defaulting these timestamps to the current time, I made it so they're now able to be explicitly assigned.

## Testing
I updated `test_workspace_mq_create` and `test_workspace_mq_update` to test for the synced timestamps.

## Summary by Sourcery

Propagate workspace creation and update timestamps from incoming MQ workspace messages into Group records and ensure they stay in sync.

New Features:
- Allow Group creation to accept explicit created_on and modified_on timestamps from callers.
- Support passing optional created and modified timestamps in generated Kessel workspace MQ messages.

Bug Fixes:
- Ensure Group.created_on and Group.modified_on reflect workspace MQ message timestamps instead of always defaulting to the current time.
- Update Group.modified_on when a workspace MQ update message includes a modified timestamp.

Tests:
- Extend workspace MQ create and update tests to assert that Group timestamps match the workspace message timestamps.

[RHINENG-25789]: https://redhat.atlassian.net/browse/RHINENG-25789?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Sync group timestamp behavior with workspace MQ messages and RBAC v2, allowing explicit timestamp propagation and conditional modified_on updates.

New Features:
- Allow groups to be created with explicit created_on and modified_on timestamps derived from workspace MQ messages.
- Support passing optional created and modified timestamps when generating Kessel workspace MQ messages.

Bug Fixes:
- Ensure group created_on and modified_on fields mirror workspace MQ message timestamps instead of always using the current time.
- Prevent host add/remove operations from changing group modified_on when RBAC v2 (Kessel groups) is enabled, relying instead on workspace-driven updates.
- Allow workspace MQ update messages to directly set a group's modified_on timestamp via patch data.

Tests:
- Extend workspace MQ create and update tests to assert that group timestamps match workspace message timestamps.
- Add API tests for adding and removing hosts from groups to verify modified_on behavior differs based on the Kessel groups feature flag.


kruai:
@sourcery-ai review

tests/test_host_mq_service.py
```diff
@@ -2815,13 +2815,24 @@ def test_workspace_mq_event_loop(handle_message_mock, post_process_rows_mock, fl
...
     workspace_message_consumer_mock.handle_message(json.dumps(message))
     found_group = db_get_group_by_id(workspace_id)
 
     assert found_group.name == workspace_name
     assert found_group.ungrouped == (workspace_type == "ungrouped-hosts")
+    assert found_group.created_on == datetime.fromisoformat(workspace_created)
+    assert found_group.modified_on == datetime.fromisoformat(workspace_modified)
 
 
 @pytest.mark.parametrize(
```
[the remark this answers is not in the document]
kruai: The test `test_workspace_mq_update` covers the handling of the "modified" field.

lib/group_repository.py
```diff
@@ -240,9 +240,19 @@ def add_group(
     group_id: UUID | None = None,
     ungrouped: bool = False,
     session: Session | None = None,
+    created_on=None,
+    modified_on=None,
```
fstavela: Missing type annotations

app/models/group.py
fstavela, line 80: Hmm, since `onupdate=_time_now`, will this timestamp get out of sync when we update the group through HBI? Or will it first be set to `_time_now`, but then immediately rewritten to what will be in the update message from Kessel?

And will host associations affect this?

app/models/group.py
```diff
@@ -59,6 +65,9 @@ def patch(self, patch_data):
         if not patch_data:
             raise InventoryException(title="Bad Request", detail="Patch json document cannot be empty.")
 
+        if "modified" in patch_data:
+            self.modified_on = patch_data["modified"]
```
fstavela, line 68: Are we sure that "modified" will always be in `patch_data`?

fstavela (commented):
Generally LGTM, just left a couple of questions. And, I would like to try this in Kessel enabled ephemeral env before merging,

app/models/group.py
kruai: Nope, which is why I have the `if "modified" in patch_data` check :smile:

fstavela: omg I'm so blind :joy:

app/models/group.py
kruai: Nope - `onupdate` only takes effect if no value is provided for the column ([source](https://docs.sqlalchemy.org/en/21/core/defaults.html#python-executed-functions)). So if we provide a timestamp (via the MQ flow), this will be set to the value from the workspace; otherwise, it'll use the current time.

If you mean the `groups` column on the `hosts` table, that gets updated for each host in a group after the group is updated, so the updated timestamp should be present there.

lib/group_repository.py
kruai: Added!

app/models/group.py
fstavela: Thanks. I meant, what if we add or remove hosts to/from a group? Will it update the `modified_on` timestamp because of the `onupdate`? Because that would make the `modified_on` timestamp out of sync with RBAC, because I think we don't send updates to RBAC if hosts are added/removed.

kruai: Oh, good point - I forgot we explicitly update the group's `updated` timestamp when adding/removing groups. I've pushed a change to fix that.

lib/group_repository.py
```diff
@@ -161,7 +164,8 @@ def _add_hosts_to_group(group_id: str, host_id_list: list[str], org_id: str):
     ]
     db.session.add_all(host_group_assoc)
 
-    _update_group_update_time(group_id, org_id)
+    if not get_flag_value(FLAG_INVENTORY_KESSEL_GROUPS):
```
fstavela: I think we need to send the context with org id

lib/group_repository.py
```diff
@@ -444,7 +458,8 @@ def _remove_hosts_from_group(group_id, host_id_list, org_id):
             else:
                 log_host_group_delete_failed(logger, assoc.host_id, assoc.group_id, get_control_rule())
 
-    _update_group_update_time(group_id, org_id)
+    if not get_flag_value(FLAG_INVENTORY_KESSEL_GROUPS):
```
fstavela: Same here

fstavela (changes requested):

lib/group_repository.py
kruai: done

lib/group_repository.py
kruai: done

fstavela (dismissed):
LGTM

thearifismail (commented):
The code looks good to me but I could not run it locally and encountered the following error, which I did not get when I ran host-inventory code in a different branch:
```
Syncs the created/updated timestamps from Workspace mq messages to the corresponding Group.

corresponding groups in the HBI DB, right?

[1/2] STEP 4/4: RUN ./configure --prefix=/usr &&     make &&     make install
/bin/sh: line 1: ./configure: No such file or directory
/bin/sh: line 1: ./configure: No such file or directory
Error: building at STEP "RUN ./configure --prefix=/usr &&     make &&     make install": while running runtime: exit status 127

Error: building at STEP "RUN ./configure --prefix=/usr &&     make &&     make install": while running runtime: exit status 127

ERROR:podman_compose:Build command failed
(asa) ➜  asa git:(workspace-group-timestamp-sync) ppsa
```

thearifismail (commented):
podman compose error appeared on my system but not on error.  so approving it.

thearifismail (approved):

PR #4004, opened 2026-04-15
fix(RHINENG-25814): Use Kessel auth in export-service

kruai:
## Jira
[RHINENG-25814](https://redhat.atlassian.net/browse/RHINENG-25814)

## What
Updates export-service to get permissions filter from Kessel rather than RBAC when Kessel Phase 1 flag is enabled.

## Why
Export service breaks when Kessel Phase 1 flag is enabled because RBAC v1 is disallowed for orgs that are set up for RBAC v2.

## How

- Centralized the get_kessel_filter/get_rbac_filter logic into `lib.middleware.resolve_permission`
- Updated export service to use `resolve_permission`

## Testing
Added `test_handle_kessel_allowed` and `test_handle_kessel_prohibited` to `test_export_service.py`.

## PR Guidelines

## PR Guideline checks

[RHINENG-25814]: https://redhat.atlassian.net/browse/RHINENG-25814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Centralize authorization resolution on the shared permission resolver and update consumers to support Kessel-based auth for host exports and host-exists checks.

New Features:
- Enable export-service to authorize host exports using Kessel permissions when the Kessel Phase 1 flag is enabled.

Bug Fixes:
- Prevent export-service failures for organizations using RBAC v2 by correctly routing permission checks through Kessel when required.

Enhancements:
- Introduce a resolve_permission helper to encapsulate Kessel-vs-RBAC decision logic and reuse it across middleware, API handlers, and export-service.
- Simplify get_kessel_filter usage by constructing the Kessel client internally instead of requiring callers to pass it.

Tests:
- Add export-service tests covering allowed and prohibited Kessel-based authorization flows.


kruai:
@sourcery-ai review

api/host.py
```diff
@@ -593,23 +588,12 @@ def host_checkin(body, rbac_filter=None):  # noqa: ARG001, required for all API
...
-    ):
-        kessel_client = get_kessel_client(current_app)
-        allowed, kessel_data = get_kessel_filter(
-            kessel_client, current_identity, KesselResourceTypes.HOST.view, [host_id]
-        )
-        if not allowed:
-            unauthorized_ids = kessel_data.get("unauthorized_ids") if kessel_data else None
-            raise IdsNotFoundError("host", unauthorized_ids)
+
+    check_access(KesselResourceTypes.HOST.view, [host_id])
```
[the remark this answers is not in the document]
kruai: OK, I've updated it so it raises IdsNotFoundError when appropriate again.

tests/test_export_service.py
```diff
@@ -193,6 +193,30 @@ def test_handle_rbac_prohibited(mock_post, subtests, flask_app, db_create_host,
...
+@mock.patch("requests.Session.post", autospec=True)
+@mock.patch("app.queue.export_service.resolve_permission", return_value=(True, None))
+def test_handle_kessel_allowed(mock_resolve, mock_post, flask_app, db_create_host, export_service_consumer_mock):
+    with flask_app.app.app_context():
+        db_create_host()
+        export_message = es_utils.create_export_message_mock()
+        mock_post.return_value.status_code = 202
+        resp = export_service_consumer_mock.handle_message(export_message)
+        assert resp is True
+        mock_resolve.assert_called_once()
```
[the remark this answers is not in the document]
kruai: OK, I've updated the test as per your suggestion.

lib/middleware.py
```diff
@@ -575,6 +557,42 @@ def modified_func(*args, **kwargs):
...
+def resolve_permission(
+    identity: Identity,
+    permission: KesselPermission,
+    ids: list[str] | None = None,
+    rbac_request_headers: dict | None = None,
+) -> tuple[bool, dict[str, Any] | None]:
+    """
+    Resolve authorization by checking Kessel (if enabled) or falling back to RBAC v1.
+
+    This centralizes the flag-check-and-branch pattern used across the codebase.
```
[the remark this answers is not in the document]
kruai: Yes, this behavior is correct.

app/queue/export_service.py
```diff
@@ -88,8 +87,8 @@ def create_export(
 
     rbac_request_headers, request_headers = build_headers(x_rh_identity, exportUUID, inventory_config, exportFormat)
 
-    allowed, rbac_filter = get_rbac_filter(
-        RbacResourceType.HOSTS, RbacPermission.READ, identity=identity, rbac_request_headers=rbac_request_headers
+    allowed, rbac_filter = resolve_permission(
+        identity, KesselResourceTypes.HOST.view, [resourceUUID], rbac_request_headers
```
rodrigonull: `resourceUUID` is not a host_id, is it? Won't this produce a denied access or unintended behavior?

api/host.py
```diff
@@ -593,22 +589,15 @@ def host_checkin(body, rbac_filter=None):  # noqa: ARG001, required for all API
...
-    # Pass org_id context for org-specific feature flag targeting
-    if (not inventory_config().bypass_kessel) and get_flag_value(
-        FLAG_INVENTORY_KESSEL_PHASE_1, context=build_flag_context(current_identity.org_id)
-    ):
-        kessel_client = get_kessel_client(current_app)
-        allowed, kessel_data = get_kessel_filter(
-            kessel_client, current_identity, KesselResourceTypes.HOST.view, [host_id]
-        )
+
+    if not inventory_config().bypass_kessel:
```
thearifismail: This is a duplicate check because the decorator `@access(KesselResourceTypes.HOST.view)` has already been called.

lib/middleware.py
```diff
@@ -575,6 +557,42 @@ def modified_func(*args, **kwargs):
...
+        permission: A KesselPermission that carries both Kessel and RBAC v1 type/permission mappings
+        ids: Optional resource IDs to check permissions against
+        rbac_request_headers: Optional pre-built RBAC request headers; built automatically if not provided
+    """
+    if ids is None:
+        ids = []
+
+    if (
+        get_flag_value(FLAG_INVENTORY_KESSEL_PHASE_1, context=build_flag_context(identity.org_id))
+        and not inventory_config().bypass_kessel
```
thearifismail: `bypass_kessel` check should be before `get_flag_value()`.   This way if `bypass_kessel is True`, then there is no need to call `get_flage_vlaue()`

fstavela (commented):
I agree with Arif's and Rodrigo's comments, otherwise LGTM, I have nothing more to add.

fstavela:
I tested this in ephemeral and the exports are failing, probably because of the `resourceUUID` usage

app/queue/export_service.py
kruai: Good call - I missed this mistake after the refactor

lib/middleware.py
kruai: Good idea, that's more efficient!

app/queue/export_service.py
```diff
@@ -88,9 +87,7 @@ def create_export(
 
     rbac_request_headers, request_headers = build_headers(x_rh_identity, exportUUID, inventory_config, exportFormat)
 
-    allowed, rbac_filter = get_rbac_filter(
-        RbacResourceType.HOSTS, RbacPermission.READ, identity=identity, rbac_request_headers=rbac_request_headers
-    )
+    allowed, rbac_filter = resolve_permission(identity, KesselResourceTypes.HOST.view, None, rbac_request_headers)
```
fstavela:
```suggestion
    allowed, rbac_filter = resolve_permission(identity, KesselResourceTypes.HOST.view, rbac_request_headers=rbac_request_headers)
```

kruai: OK, updated it

api/host.py
```diff
@@ -593,23 +589,15 @@ def host_checkin(body, rbac_filter=None):  # noqa: ARG001, required for all API
...
-            kessel_client, current_identity, KesselResourceTypes.HOST.view, [host_id]
-        )
-        if not allowed:
-            unauthorized_ids = kessel_data.get("unauthorized_ids") if kessel_data else None
-            raise IdsNotFoundError("host", unauthorized_ids)
+
+    allowed, filter_data = resolve_permission(get_current_identity(), KesselResourceTypes.HOST.view, [host_id])
+    if not allowed:
+        unauthorized_ids = filter_data.get("unauthorized_ids") if filter_data else None
+        raise IdsNotFoundError("host", unauthorized_ids)
```
fstavela: Hmm, I think this check is still redundant. The `@access(KesselResourceTypes.HOST.view)` decorator returns `rbac_filter`, and we use it in `get_host_id_by_insights_id`, so the `host_id` should be accessible by the user.

kruai: I think you're right... It was already redundant before this change. I was focused on preserving the existing check and `IdsNotFoundError` behavior, but that should be unreachable. I'll remove it.
Edit: removed.

fstavela (approved):
LGTM

rodrigonull (approved):

PR #4010, opened 2026-04-16
test(RHINENG-25474): Update tests to make them more stable when Kessel is enabled

fstavela:
## Jira
[RHINENG-25474](https://issues.redhat.com/browse/RHINENG-25474)

## What
Update tests to make them more stable when Kessel is enabled

## Why
A few tests are flaky in Stage, sometimes failing, sometimes passing

## How
When we make a request with cert-auth, it doesn't go through Kessel (it uses `owner_id` check instead). So, when we create a new host via cert-auth and then wait for it being available, it is available immediatelly with cert auth. So, when we then make some API request with a different auth, we can sometimes get HTTP 403, because the host is not synced to Kessel yet. This PR makes the tests wait for the host to be available with basic auth as well, after it is created with cert auth,.

## Testing
I tested this against Stage

## PR Guidelines

## PR Guideline checks

[RHINENG-25474]: https://redhat.atlassian.net/browse/RHINENG-25474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Stabilize host inventory tests that rely on certificate-authenticated host creation by ensuring hosts are available via standard auth paths before assertions and cache/notification flows run.

Tests:
- Add synchronization waits after certificate-auth host creation so tests query hosts only after they are available through non-certificate auth, reducing flakiness with Kessel enabled.
- Adjust a notifications end-to-end test to fetch system profiles via the same test application client used to create the host.

### DIFF
--- a/iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/cache/test_cache_insights_client.py
+++ b/iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/cache/test_cache_insights_client.py
@@ -123,6 +123,10 @@ def test_cache_invalidation_insights_client_get_hosts_list_delete_host_by_id(
     # Prepare the second host
     host_to_delete = host_inventory_cert_auth.upload.create_host(insights_id=generate_uuid())
 
+    # Cert auth doesn't go through Kessel, so we need to do additional waiting with other auth
+    # to make sure the host got synced and is available
+    host_inventory.apis.hosts.wait_for_created(host_to_delete)
+
     # Init the cache
     response = host_inventory_cert_auth.apis.hosts.get_hosts(
         insights_id=host_to_delete.insights_id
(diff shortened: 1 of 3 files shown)


iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/cache/test_cache_insights_client.py
```diff
@@ -166,7 +174,6 @@ def test_cache_invalidation_insights_client_get_hosts_list_delete_filtered(
 
     # Delete the second host to invalidate the cache
     host_inventory.apis.hosts.delete_filtered(insights_id=host_to_delete.insights_id)
-    host_inventory.apis.hosts.wait_for_deleted(host_to_delete)
```
kruai, line 169: Is this wait no longer needed? Just making sure

kruai (approved):

rodrigonull (approved):

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/cache/test_cache_insights_client.py
fstavela: I'm not sure why it's there, but I think it never was needed. The `delete_filtered` method does wait for the host to not be retrievable anymore by default.

PR #4014, opened 2026-04-16
fix(RHINENG-25878): prevent ghost hosts from update-after-delete races

rodrigonull:
## Jira
[RHINENG-25878](https://issues.redhat.com/browse/RHINENG-25878)

## What
Add pre-publish host existence checks to all Kafka event-producing paths to prevent ghost hosts caused by concurrent delete/update races across pods.

## Why
In a multi-pod environment (50+ MQ consumers, 20+ API pods), a host can be deleted by one pod while another is about to publish an "updated" Kafka event for it. Downstream consumers (e.g. Advisor) then see the update after the delete and re-create the host as a ghost record. This was already partially fixed in the staleness path but the same race existed in API patch/check-in/facts, MQ ingress batches, group membership changes, and the host synchronizer job.

## How
- Added a shared `host_exists(host_id, org_id, session)` helper in `lib/host_repository.py` that performs an indexed point lookup using the composite `(id, org_id)` key, enabling partition pruning on the hash-partitioned hosts table.
- Placed the existence check immediately before each `write_event()` call across all five affected paths: `api/host.py`, `api/staleness.py`, `app/queue/host_mq.py`, `lib/group_repository.py`, and `lib/host_synchronize.py`.
- Made `org_id` a required parameter since all event-producing paths always have tenant context available.
- Removed the old batch-level `_get_existing_host_ids` from `api/staleness.py` and the unused `get_existing_host_ids` from `lib/host_repository.py`.

## Testing
- Added `test_write_add_update_event_message_skips_deleted_host` to verify that when `host_exists` returns False, no Kafka event, notification, success log, or cache write is produced.
- Updated `test_write_add_update_event_message` to assert `host_exists` is called with the correct `(host_id, org_id, session)` arguments.
- Updated `test_write_message_batch_flushes_once` to work with the new per-host guard placement.

## Performance overhead
Each `host_exists()` call executes a single `SELECT id FROM hbi.hosts WHERE id = ? AND org_id = ?` query using the composite primary key index with partition pruning on the hash-partitioned table.

Benchmarked on a production-like database with 6 million hosts:
- **Warm lookup (cached):** ~0.03 ms DB execution time
- **Cold lookup (disk read):** ~5 ms DB execution time
- **Application-level roundtrip:** sub-ms to low single-digit ms per call

The check runs once per host event, immediately before the Kafka `write_event()` call.

## Related
- [RHINENG-24023](https://redhat.atlassian.net/browse/RHINENG-24023) — related investigation into ghost hosts in downstream applications

## PR Guidelines

## PR Guideline checks

[RHINENG-25878]: https://redhat.atlassian.net/browse/RHINENG-25878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHINENG-24023]: https://redhat.atlassian.net/browse/RHINENG-24023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Guard Kafka host update event emission with per-host existence checks to avoid emitting updates for hosts that have been concurrently deleted.

Bug Fixes:
- Prevent ghost host records by skipping staleness, synchronization, group-change, MQ, and API patch events when the corresponding host row no longer exists in the database.

Enhancements:
- Introduce a shared host_exists helper in the host repository and apply it across event-producing code paths for consistent, efficient existence checks.

Tests:
- Extend host MQ tests to cover the deleted-host guard behavior and adjust batch-writing tests to work with per-host existence checks.

### DIFF
--- a/app/queue/host_mq.py
+++ b/app/queue/host_mq.py
@@ -1023,7 +1023,7 @@ def write_add_update_event_message(
         processing_status_message="host operation complete",
         current_operation="write_message_batch",
         inventory_id=result.row.id,
-    ):
+    ) as tracker_ctx:
         output_host = serialize_host(result.row, result.staleness_timestamps, staleness=result.staleness_object)
         insights_id = str(result.row.insights_id)
         event = build_event(result.event_type, output_host, platform_metadata=result.platform_metadata)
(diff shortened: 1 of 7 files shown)


lib/host_synchronize.py
```diff
@@ -74,16 +76,22 @@ def _synchronize_hosts_for_org(org_hosts_query, custom_staleness_dict, event_pro
...
+            events_to_produce.append((event, headers, str(host.id)))
+
+        for event, headers, host_id in events_to_produce:
+            if not host_exists(host_id, org_id=host_list[0].org_id, session=query.session):
+                logger.warning(f"Skipping sync event for host {host_id}: host no longer exists")
+                continue
             # in case of a failed update event, event_producer logs the message.
             # Workaround to solve: https://issues.redhat.com/browse/RHINENG-4856
             try:
-                event_producer.write_event(event, str(host.id), headers, wait=True)
```
jpramos123, line 80: Are we sure it is a string all the time? it may be a UUID and cause issues?

jpramos123 (commented):
One small comment!

Overall LGTM!

app/queue/host_mq.py
```diff
@@ -1037,6 +1038,10 @@ def write_add_update_event_message(
             bootc_booted,
         )
 
+    if not host_exists(result.row.id, org_id=result.row.org_id, session=db.session):
```
thearifismail: shouldn't this `if not host_exists()` be inside the `with PayloadTrackerProcessingContext()` block?  Here is an example which will require adding `tracker_ctx` the `with definition() as tracker_ctx"`

```
        if not host_exists(result.row.id, org_id=result.row.org_id, session=db.session):
            logger.warning(f"Skipping event for host {result.row.id}: host no longer exists")
            tracker_ctx._success_status_msg = "skipped – host deleted before event production"
            return
```

rodrigonull: nice catch. Updated

lib/host_synchronize.py
rodrigonull: Yes, it's always a string. str(host.id) is called when building the tuple (line 79), so host_id in the loop is already converted.

thearifismail (approved):
LGTM

jpramos123 (approved):

PR #4028, opened 2026-04-17
perf(RHINENG-25888): disable expire_on_commit during MQ batch processing

rodrigonull:
## Jira
[RHINENG-25888](https://issues.redhat.com/browse/RHINENG-25888)

## What
Disable SQLAlchemy's `expire_on_commit` during MQ batch processing to eliminate redundant per-host SELECT queries that occur when serializing hosts into event payloads after the transaction commits.

## Why
After `session_guard` commits the batch transaction, SQLAlchemy expires all ORM objects in the session. When `post_process_rows()` then accesses Host attributes to serialize them into outbox events, each access triggers a lazy-load `SELECT hbi.hosts.* WHERE id = X` round-trip to Postgres. With a batch size of 50 messages, this adds 50 unnecessary SELECT queries per batch cycle.

## How
- Added a `no_expire_on_commit()` context manager in `lib/db.py` that temporarily sets `expire_on_commit=False` on the session and restores the original value on exit.
- Wrapped `_process_batch()` in `app/queue/host_mq.py` with `no_expire_on_commit(db.session)` so that both the session guard (flush/commit) and `post_process_rows()` (event serialization) run while expire is disabled.
- This is safe because `post_process_rows()` only reads attributes from already-committed Host objects to build event payloads — it does not depend on seeing concurrent writes from other transactions.

## Testing
- All 225 MQ service tests pass (`tests/test_host_mq_service.py`).

## PR Guidelines

## PR Guideline checks

[RHINENG-25888]: https://redhat.atlassian.net/browse/RHINENG-25888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Disable SQLAlchemy expire_on_commit during MQ batch processing to avoid redundant database reloads while preserving existing message handling and offset commit behavior.

Enhancements:
- Introduce a no_expire_on_commit context manager to temporarily disable SQLAlchemy session expire_on_commit and restore it afterward.
- Wrap host MQ batch processing, including post_process_rows and Kafka offset committing, in the no_expire_on_commit context to keep ORM instances warm during event serialization.

### DIFF
--- a/app/queue/host_mq.py
+++ b/app/queue/host_mq.py
@@ -80,6 +80,7 @@ from app.staleness_serialization import AttrDict
 from lib import group_repository
 from lib import host_app_repository
 from lib import host_repository
+from lib.db import no_expire_on_commit
 from lib.db import raw_db_connection
 from lib.db import session_guard
 from lib.feature_flags import FLAG_INVENTORY_REJECT_RHSM_PAYLOADS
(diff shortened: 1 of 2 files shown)


jpramos123 (dismissed):
LGTM! Thanks

FabriciaDinizRH (approved):
LGTM!

PR #4069, opened 2026-04-22
refactor(RHINENG-25843): Update staleness serialization

FabriciaDinizRH:
## Jira
[RHINENG-25843](https://issues.redhat.com/browse/RHINENG-25843)

## What

Not much needed to get done, as previous PRs covered most of the changes, so I added some more clarity.

- Add a regression test
- Rename the local timestamps value to lifecycle_timestamps for clarity
- Expand the serialize_host docstring

## Why
Phase 1 of the culling / staleness migration keeps DB columns for compatibility, but the read path should not depend on those columns for API output. The plan called out making that contract explicit for maintainers and preventing accidental regressions.

## How

-  Serialize_host: document the behavior; use lifecycle_timestamps = get_staleness_timestamps(...) and wire field_mapping / state from that dict only.
- Test: build a host with wrong stale_timestamp / stale_warning_timestamp / deletion_timestamp and a fixed last_check_in; assert serialized fields match get_staleness_timestamps and disagree with the ORM values.

## Testing
Unit tests all passing.

## PR Guidelines

## PR Guideline checks

[RHINENG-25843]: https://redhat.atlassian.net/browse/RHINENG-25843?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Document host serialization behavior around staleness and ensure lifecycle timestamps and state derive from computed staleness values rather than persisted DB columns.

Bug Fixes:
- Prevent host lifecycle fields and state from using potentially incorrect ORM staleness columns during serialization.

Enhancements:
- Clarify the serialize_host function via an expanded docstring and clearer lifecycle_timestamps naming.

Tests:
- Add a regression test verifying serialized lifecycle timestamps and state are based on computed staleness timestamps instead of persisted column values.


app/serialization.py
```diff
@@ -186,10 +186,21 @@ def serialize_host(
...
+        for_mq: Whether to serialize for MQ consumers.
+        additional_fields: Additional fields to serialize.
+        staleness: Staleness configuration to use for staleness calculations.
+        system_profile_fields: System profile fields to serialize.
+    """
     # Ensure additional_fields is a tuple
     additional_fields = additional_fields or ()
 
-    timestamps = get_staleness_timestamps(host, staleness_timestamps, staleness)
+    lifecycle_timestamps = get_staleness_timestamps(host, staleness_timestamps, staleness)
```
fstavela: Wouldn't `staleness_timestamps` be better? I'm just worried about introducing a new term, to not create confusion.

FabriciaDinizRH: `staleness_timestamp` is already being used inside this function, so I used `st_timestamps`

fstavela (approved):
LGTM

ezr-ondrej (approved):
Really easy, nice 👍

adarshdubey-star (approved):

PR #4078, opened 2026-04-22
chore(RHINENG-25996): Require context for all feature flag checks

kruai:
## Jira
[RHINENG-25996](https://redhat.atlassian.net/browse/RHINENG-25996)

## What

Require context to be supplied for all `get_flag_value` checks.

## Why
To make it so we can't accidentally omit the `org_id` context when checking a feature flag's value.

## How

- Change `get_flag_value` so it requires `org_id`
- Build the Unleash context within `get_flag_value`

## Testing
Updated the mocks in the existing unit tests.

## PR Guidelines

## PR Guideline checks

## Summary by Sourcery

Require org-scoped context for all feature flag evaluations and update callers accordingly.

Enhancements:
- Change feature flag helpers so get_flag_value now requires an org_id and internally builds the Unleash context.
- Propagate org_id into RBAC, Kessel bulk checks, workload compatibility logic, host serialization, and RHSM payload handling where feature flags are evaluated.
- Adjust middleware and permission checks to obtain the current identity once and pass its org_id into feature flag lookups.

Tests:
- Update feature flag and Kessel tests to pass explicit context or org_id and to verify orgId is included in the Unleash context.

[RHINENG-25996]: https://redhat.atlassian.net/browse/RHINENG-25996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Require org-scoped context for all feature flag evaluations and propagate org_id through key call paths that use feature flags.

Enhancements:
- Change feature flag helpers so get_flag_value now requires an org_id and internally builds the Unleash context, while get_flag_value_and_fallback requires an explicit context.
- Pass org_id into feature flag checks across RBAC middleware, Kessel bulk permission checks, workload compatibility logic, host serialization, and RHSM payload handling.
- Cache and reuse identity/org_id within middleware and host queries when performing multiple feature flag or RBAC operations in a single request.
- Adjust the host reaper job to perform the read-only mode check using the explicit get_flag_value_and_fallback API.

Tests:
- Update feature flag utility and Kessel tests to provide explicit org_id/context and validate that the correct Unleash context is used.


kruai:
@sourcery-ai review

api/host_query_db.py
```diff
@@ -763,12 +763,14 @@ def get_sparse_system_profile(
...
+
     # Track if we need to fetch workloads for backward compatibility
     requested_sp_fields_dict: dict[str, bool] = fields.get("system_profile", {}) if fields else {}  # type: ignore[assignment]
     requested_sp_fields = list(requested_sp_fields_dict.keys()) if isinstance(requested_sp_fields_dict, dict) else []
     workloads_requested = "workloads" in requested_sp_fields
     workloads_needed_for_compat = (
-        get_flag_value(FLAG_INVENTORY_WORKLOADS_FIELDS_BACKWARD_COMPATIBILITY)
+        get_flag_value(FLAG_INVENTORY_WORKLOADS_FIELDS_BACKWARD_COMPATIBILITY, identity.org_id)
         and requested_sp_fields
         and any(field in legacy_workload_fields for field in requested_sp_fields)
```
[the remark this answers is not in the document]
kruai: OK, I've introduced a variable to reuse the flag value.

app/queue/host_mq.py
```diff
@@ -416,7 +416,7 @@ def handle_message(self, message: str | bytes, headers: list[tuple[str, bytes]]
             try:
                 host = validated_operation_msg["data"]
                 if host.get("reporter") in ["rhsm-conduit", "rhsm-system-profile-bridge"] and get_flag_value(
-                    FLAG_INVENTORY_REJECT_RHSM_PAYLOADS
+                    FLAG_INVENTORY_REJECT_RHSM_PAYLOADS, host.get("org_id", "")
```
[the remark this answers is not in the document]
kruai: Since host is required to have "org_id", I've updated this so it no longer uses `get()`, and thus avoids the scenario where it passes an empty string.

tests/test_lib_kessel.py
```diff
@@ -98,7 +98,9 @@ def test_bulk_check_with_flag_disabled_uses_bulk_api(
         kessel_client.inventory_svc.CheckBulk = Mock(return_value=mock_response)
 
         # Call the method
-        result, unauthorized_ids = kessel_client._check_bulk_resources(mock_subject_ref, test_permission, resource_ids)
+        result, unauthorized_ids = kessel_client._check_bulk_resources(
+            mock_subject_ref, test_permission, resource_ids, "test"
+        )
```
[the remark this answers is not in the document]
kruai: I think our existing tests should be sufficient.

tests/test_utils.py
```diff
@@ -149,15 +148,15 @@ def test_feature_flag_error_fallback(_enable_unleash):
...
         assert using_fallback
 
 
 @patch.dict(FLAG_FALLBACK_VALUES, {TEST_FEATURE_FLAG: False})
 def test_feature_uninitialized_fallback(_enable_unleash):
     with patch.object(UNLEASH, "client", None):
-        flag_value, using_fallback = get_flag_value_and_fallback(TEST_FEATURE_FLAG)
+        flag_value, using_fallback = get_flag_value_and_fallback(TEST_FEATURE_FLAG, {})
         assert not flag_value
         assert using_fallback
```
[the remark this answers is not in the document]
kruai: I think our existing tests should be sufficient.

jobs/host_reaper.py
```diff
@@ -117,7 +117,7 @@ def run(config, logger, session, event_producer, notification_event_producer, sh
     )
 
     with application.app.app_context():
-        if get_flag_value(FLAG_INVENTORY_API_READ_ONLY):
+        if get_flag_value_and_fallback(FLAG_INVENTORY_API_READ_ONLY, {})[0]:
```
rodrigonull, line 121:
```suggestion
        # The reaper runs across all orgs, so there is no org to scope this check to.
        if get_flag_value_and_fallback(FLAG_INVENTORY_API_READ_ONLY, {})[0]:
```

lib/feature_flags.py
rodrigonull: Should we make this function private now? It's no longer imported anywhere. Renaming it to `_build_flag_context` would signal that it's an implementation detail and discourage direct use.

lib/middleware.py
```diff
@@ -476,8 +475,12 @@ def rbac(resource_type: RbacResourceType, required_permission: RbacPermission, p
     def other_func(func):
         @wraps(func)
         def modified_func(*args, **kwargs):
+            current_identity = get_current_identity()
+
             # If the API is in read-only mode and this is a Write endpoint, abort with HTTP 503.
-            if required_permission == RbacPermission.WRITE and get_flag_value(FLAG_INVENTORY_API_READ_ONLY):
+            if required_permission == RbacPermission.WRITE and get_flag_value(
```
rodrigonull, line 484: We only use `FLAG_INVENTORY_API_READ_ONLY` as a global kill switch, so we need to be careful. Since the middleware now passes org_id into the context for this flag, adding an orgId constraint on the server side could silently scope it to specific orgs, which would be misleading if the intent is to turn it on for all users. Not an issue, as long we don't add the server side constraint.

rodrigonull (commented):
The changes LGTM. Added a few non-blocker suggestions.

lib/middleware.py
kruai: Yeah - for the feature flags that don't use org_id targeting, providing org_id context in the check shouldn't affect anything. We just need to make sure we don't set up org_id targeting for flags that need to stay global (which we already need to do)

jobs/host_reaper.py
kruai: Good idea, added

lib/feature_flags.py
kruai: Yep, better as a private function. Updated it

rodrigonull (approved):

thearifismail (approved):
LGTM

fstavela (approved):

PR #4084, opened 2026-04-23
Add inventory views management paths and schemas to OpenAPI spec

adarshdubey-star:
Jira
[RHINENG-25827](https://issues.redhat.com/browse/RHINENG-25827)

What
Add OpenAPI spec definitions for the new Inventory Views Management feature — 6 API paths, 7 schemas, and 1 path parameter under /beta/views.

Why
The Inventory Views feature allows users to save and manage custom table configurations (columns, sort, filters) for the inventory UI. This card defines the API contract so frontend and backend development can proceed in parallel.

How
Added 6 paths to swagger/api.spec.yaml:

-   GET /beta/views — list visible views (paginated)
-   POST /beta/views — create a view
-   GET /beta/views/{view_id} — get a single view
-   PUT /beta/views/{view_id} — update a view (owner only)
-   DELETE /beta/views/{view_id} — delete a view (owner only)
-   POST /beta/views/{view_id}/clone — clone any visible view

Added schemas: ViewColumnConfig, ViewSortConfig, ViewConfiguration, ViewIn, ViewPatch, ViewOut, ViewsListOut

- Added viewId UUID path parameter
- Regenerated swagger/openapi.json via redocly bundle
- All operationId values follow the pattern api.views.<function_name>
- Paths are under /beta/ prefix consistent with existing beta endpoints

Testing

- OpenAPI spec validates successfully via swagger-cli validate swagger/openapi.json
- ruff check passes with no errors
- No runtime handlers implemented yet — this is spec-only (no Python changes)

[RHINENG-25827]: https://redhat.atlassian.net/browse/RHINENG-25827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Define OpenAPI contract for the new inventory views management API under /beta/views, including endpoints for listing, creating, retrieving, updating, deleting, and cloning views.

New Features:
- Add /beta/views endpoints to list, create, retrieve, update, delete, and clone inventory views with appropriate permissions and responses.
- Introduce shared viewId UUID path parameter for inventory view resource operations.
- Define Inventory Views schemas for column, sort, and full view configuration plus create, update, single-view, and paginated-list response payloads.


swagger/api.spec.yaml
```diff
@@ -2544,6 +2718,186 @@ components:
...
+        is_system_view:
+          description: >-
+            True if this is a read-only system view (Red Hat preset).
+            System views cannot be updated or deleted, but can be cloned.
+          type: boolean
+        is_owner:
+          description: >-
+            True if the requesting user is the creator of this view.
+            Only the owner can update or delete a view.
+          type: boolean
```
kruai, line 2887: Is this field useful? We already have `created_by`

swagger/api.spec.yaml
```diff
@@ -1041,6 +1041,172 @@ paths:
         '400':
           description: "Invalid request."
 
+  /beta/views:
+    get:
+      operationId: api.views.get_views_list
```
kruai, line 1046: I think you'll need to create `api/views.py` and add a placeholder function for each of these operationIds. They can just return HTTP 501 "Not Implemented". Otherwise, the checks are going to complain that the function doesn't exist

kruai (changes requested):
Looks good so far! Got one Q and one suggestion

swagger/api.spec.yaml
adarshdubey-star: I thought about it, and I have two things here : 

1 . `is_owner` is a computed convenience field, so the frontend doesn't need to compare `created_by` against the current user's identity on every view. The frontend can just check `is_owner` to show/hide edit and delete buttons. But if we'd prefer to keep the response leaner, I can remove it the frontend can derive it from created_by.

2.  If ownership logic ever changes in Phase 2 (e.g., org admins can also edit, or Kessel grants edit access to non-creators), With is_owner, the backend updates the logic in one place, and the frontend doesn't change.

adarshdubey-star: @rodrigonull, I would love to have your opinion on this.

swagger/api.spec.yaml
adarshdubey-star: This is my fav thing to miss. I will update it : )

jpramos123 (approved):
LGTM!

swagger/api.spec.yaml
kruai: @rodrigonull got any thoughts on this? I'm approving so my review no longer blocks the merge, but I'd still like to hear opinions on it.

kruai (approved):

swagger/api.spec.yaml
rodrigonull: I agree with keeping a server-computed boolean. It avoids coupling the frontend to ownership/permission logic that will evolve (Kessel, org admin roles, etc.)

rodrigonull (approved):

PR #4092, opened 2026-04-24
feat(RHINENG-20674): Delete custom staleness when within 1h of defaults

kruai:
## Summary
When POST or PATCH `/account/staleness` receives conventional staleness values that are all within one hour (3600 seconds) of the system defaults, the API now removes an existing custom staleness row for the org instead of keeping or inserting one. This avoids storing rows that are effectively the same as defaults.

Host staleness updates are still triggered when a custom row is removed. Staleness is serialized to a plain dict in the request thread before the background host-update job runs, so ORM instances are not used from a worker thread after the request session ends.

## Jira
[RHINENG-20674](https://issues.redhat.com/browse/RHINENG-20674)

## Changes
- `lib/staleness.py`: added `staleness_equivalent_to_system_defaults` and `org_has_custom_staleness`
- `api/staleness.py`: early handling in create/update + async serialization fix
- Tests: coverage for near-default POST/PATCH and updated create tests to use non-default values when a DB row is required

## Testing
- `make style`
- `pipenv run pytest` on `tests/test_api_staleness_create.py`, `tests/test_api_staleness_update.py`, `tests/test_api_staleness_delete.py`, `tests/test_api_staleness_get.py`, `tests/test_staleness_cache.py`

[RHINENG-20674]: https://redhat.atlassian.net/browse/RHINENG-20674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Handle near-default staleness values by clearing custom org staleness and updating hosts accordingly.

Bug Fixes:
- Prevent background host staleness updates from using ORM objects after their session is closed by pre-serializing staleness data in the request thread.

Enhancements:
- Introduce a tolerance-based helper to detect when conventional staleness values are effectively equivalent to system defaults and avoid persisting redundant rows.
- Add conditional deletion of custom staleness rows with safe handling when no row exists.

Tests:
- Extend staleness API test coverage for near-default POST/PATCH behavior and boundary conditions around the one-hour tolerance window.
- Add shared test constants and helpers for conventional staleness triples and default-equivalence unit tests.


kruai:
@sourcery-ai review

kruai:
Follow-up push addressing Sourcery review:

- **`staleness_equivalent_to_system_defaults`**: Documented that `StalenessSchema` / merge supplies all three keys; added defensive handling for missing keys and non-`int` values (returns `False`).
- **Shared helper** `_response_if_staleness_equivalent_to_system_defaults` in `api/staleness.py` to centralize remove/cache/async/response for create + update.
- **`_async_update_host_staleness`**: Parameter renamed to `staleness_orm_or_defaults` and typed as `Staleness | AttrDict` (both work with `serialize_staleness_to_dict` in the request thread).
- **Tests**: `tests/test_lib_staleness_equivalence.py` for boundary offsets 0, ±3600, ±3601 and incomplete payload; POST/PATCH API tests for +3600 (no row / `system_default`) and +3601 (custom row kept).

kruai:
**Boundary fix:** “Default-equivalent” now means each conventional field differs from system defaults by **strictly less than** 3600 seconds (`abs(diff) < 3600`). A difference of **exactly** one hour creates/keeps a custom staleness row as intended.

- Implementation: `abs(v - default[k]) >= 3600` → not equivalent (was `> 3600` before).
- Tests updated: +3599 → no row / delete path; +3600 → row persisted / PATCH keeps custom row.

kruai:
@sourcery-ai review

kruai:
### Review follow-up (Sourcery + tests)

**Implemented**

1. **Single defaults object** — `_response_if_staleness_equivalent_to_system_defaults` calls `get_sys_default_staleness_api(identity)` once, passes it as `sys_defaults=` into `staleness_equivalent_to_system_defaults`, and reuses the same object for `_async_update_host_staleness` (no second fetch for equivalence + async).

2. **Atomic remove vs check-then-act** — Added `remove_staleness_if_exists()` using a bulk `DELETE` with row-count; cache clear + async run **only** when a row was actually removed. `remove_staleness()` now delegates to that and still raises `NoResultFound` when nothing was deleted (unchanged behavior for `DELETE /account/staleness`). Removed `org_has_custom_staleness` from this path.

3. **Boundary tests** — Assert all three `conventional_time_*` fields on the DB row for exactly-+3600 and +3601 create/patch cases (per review).

4. **`test_delete_existing_staleness`** — Store `org_id` before the HTTP DELETE. Bulk delete can detach the fixture’s `Staleness` instance so `saved_staleness.org_id` is unsafe to read after the request.

**Notes (not “disagree,” but scope)**

- `get_sys_default_staleness_api` does not hit the DB; it builds defaults from config. The duplication removed was extra **work** in the hot path, not a query round-trip.
- Earlier Sourcery items (partial PATCH keys, `Staleness | AttrDict` on async) were already addressed in prior commits: merge + schema always supply three keys; defensive `get`/type checks; serialized dict in the worker thread.

Pushed: `abc03131`

tests/test_api_staleness_create.py
```diff
@@ -8,18 +8,50 @@
...
+def test_create_staleness_at_defaults_does_not_persist_row(api_create_staleness, db_get_staleness_culling):
+    """POST with default triple should not add a custom staleness row (RHINENG-20674)."""
+    response_status, response_data = api_create_staleness(_INPUT_DATA)
+    assert_response_status(response_status, 201)
+    assert response_data["id"] == "system_default"
+    org_id = response_data["org_id"]
+    assert db_get_staleness_culling(org_id) is None
+
+
+def test_create_staleness_at_exactly_one_hour_creates_row(api_create_staleness, db_get_staleness_culling):
```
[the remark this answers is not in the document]
kruai: OK, this was addressed with the latest commit.

tests/test_lib_staleness_equivalence.py
```diff
@@ -0,0 +1,43 @@
...
+@pytest.mark.parametrize(
+    "offset_seconds, expect_equivalent",
+    [
+        (0, True),
+        (3599, True),
+        (-3599, True),
+        (3600, False),
+        (-3600, False),
+        (3601, False),
+        (-3601, False),
```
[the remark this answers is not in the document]
kruai: OK, this was addressed with the latest commit.

jpramos123 (dismissed):
Looks solid!

LGTM

api/staleness.py
```diff
@@ -284,6 +319,13 @@ def update_staleness(body):
 
     identity = get_current_identity()
     org_id = identity.org_id
+
+    early = _response_if_staleness_equivalent_to_system_defaults(
```
rodrigonull: Before this PR, `update_staleness` would return 404 when no custom staleness row existed. Now, if the payload is near-default, the early-return path triggers and returns 200 with system defaults. Is this change intentional?

api/staleness.py
rodrigonull: `_validate_input_data` already catches ValidationError internally and calls abort(). The outer except ValidationError would never execute because abort() raises an HTTPException, not a ValidationError. This is a pre-existing issue, not introduced by this PR, but it means the except branch is dead code. Not blocking, but worth noting. The same pattern exists in create_staleness.

api/staleness.py
kruai: Probably not a big deal, but wasn't intentional. I'll get it back to how it was.

api/staleness.py
kruai: Technically a bit out of scope, but I figured I might as well address it while we're talking about it. Pushed an update!

api/staleness.py
kruai: Updated the implementation, and corrected the test to validate the 404.

rodrigonull (approved):
LGTM

PR #4096, opened 2026-04-27
test(RHINENG-26076): Migrate test_groups_cant_update_host_groups_via_kafka to unit tests

fstavela:
## Jira
[RHINENG-26076](https://issues.redhat.com/browse/RHINENG-26076)

## What
Migrate test_groups_cant_update_host_groups_via_kafka to unit tests.

Also, fix a bug where we were passing `groups` as a positional argument for `per_reporter_staleness`. See the discussion about it here: https://github.com/RedHatInsights/insights-host-inventory/pull/4096#discussion_r3147085943

## Why
It's failing now and it should be a unit test, so instead of just fixing it I also migrated it.

## Testing
Unit tests are passing

## PR Guidelines

## PR Guideline checks

[RHINENG-26076]: https://redhat.atlassian.net/browse/RHINENG-26076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Migrate the Kafka-based group update regression test to a unit-level MQ service test while preserving the behavior that host group memberships cannot be modified via message processing.

Tests:
- Add a unit test in the MQ host service suite to verify that updating host groups via MQ raises an error and leaves groups unchanged.
- Remove the corresponding ephemeral integration test from the IQE host inventory plugin REST groups suite.


fstavela:
@sourcery-ai review

tests/test_host_mq_service.py
```diff
@@ -2416,6 +2416,39 @@ def test_groups_not_overwritten_for_existing_hosts(
...
+    original_group_id = str(original_group.id)
+    host_id = str(host.id)
+
+    other_group = db_create_group("other_group")
+    other_group_id = str(other_group.id)
+
+    update_data = minimal_host(insights_id=str(host.insights_id))
+    update_data.groups = [{"id": other_group_id, "name": "other_group", "ungrouped": False}]
+
+    with pytest.raises(ValueError):
```
rodrigonull: The ValueError here isn't a groups validation error. it's triggered because HostSchema.build_model passes groups as the 11th positional arg to Host.__init__, which maps to per_reporter_staleness, not groups. The _validate_per_reporter_staleness validator then fails on the list of group dicts. This test passes for the wrong reason and would break if that positional mismatch is ever fixed.

rodrigonull (changes requested):

tests/test_host_mq_service.py
fstavela: Yeah, for me it wasn't important that the `mq_create_or_update_host` raises an error, I just wanted to send the message to HBI and check that the host (especially the `groups` field) was not updated, so the `pytest.raises` is there only because the test would fail otherwise.

So I wouldn't say that the test passes for the wrong reason, because the following assertions still validate the important part. But you are right that the test would break if that positional mismatch is fixed, so I'll try to find a different way.

fstavela: The ValueError is raised when the test is processing the notification message generated by HBI, not when HBI processes the update message.

rodrigonull: I re-ran the test and the traceback shows me the ValueError is raised during process_message → deserialize_host, not during notification processing:

```
File "app/queue/host_mq.py", line 497, in process_message
    input_host = deserialize_host(host_data)
...
File "app/models/host.py", line 600, in _validate_per_reporter_staleness
    for reporter, raw in dict(value).items():
ValueError: dictionary update sequence element #0 has length 3; 2 is required
```

fstavela: Oh shoot, you are right! I was messing with the test and saw a different error and I've mistaken it for this one. I'll update the test.

fstavela: @rodrigonull I've decided to fix the bug instead, as it looks quite serious. I also added `*` into the init methods definitions to force the arguments to be passed as keyword arguments, to minimize the chance of a similar issue happening again.

The reason why this was unnoticed is that this would come up only if the incoming MQ message includes groups data, but none of the reporters (thankfully) do that.

kruai (approved):
LGTM, thank you for making this fix 🙏

rodrigonull (approved):
LGTM!

PR #4106, opened 2026-04-28
refactor(RHINENG-25844): Make staleness_timestamp nullable

FabriciaDinizRH:
## Jira
[RHINENG-25844](https://issues.redhat.com/browse/RHINENG-25844)

## What

- Alembic migration to drop the NOT NULL constraint on hbi.hosts.stale_timestamp.
- SQLAlchemy Host model updated so stale_timestamp is nullable=True, matching the database.

## Why

- Phase 2: allow new/updated host rows to exist without a persisted stale_timestamp once read paths no longer depend on the column being always written at check-in.
- Depends on Phase 1 (staleness reads decoupled) before production can safely stop persisting these values.

## How

- New revision 41e07d4c1092

## Testing

- -----

## PR Guidelines

## PR Guideline checks

## Summary by Sourcery

Enhancements:
- Relax the Host model stale_timestamp field to be nullable to align with updated staleness handling requirements.

[RHINENG-25844]: https://redhat.atlassian.net/browse/RHINENG-25844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/host.py
+++ b/app/models/host.py
@@ -267,7 +267,7 @@ class LimitedHost(db.Model, HostTypeDeriver):
 
 
 class Host(LimitedHost):
-    stale_timestamp = db.Column(db.DateTime(timezone=True), nullable=False)
+    stale_timestamp = db.Column(db.DateTime(timezone=True), nullable=True)
     deletion_timestamp = db.Column(db.DateTime(timezone=True))
     stale_warning_timestamp = db.Column(db.DateTime(timezone=True))
     reporter = db.Column(db.String(255), nullable=False)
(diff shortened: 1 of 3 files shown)


fstavela (changes requested):
Please just add a couple of simple unit tests that will verify that the stale timestamp is nullable. I have in mind these two tests:
- One that will try to create a new host in the DB without `stale_timestamp`, commit, and check that the host was successfully created with `stale_timestamp is NULL`.
- One that will try to modify an existing host in the DB. The existing host will have `stale_timestamp` populated, but the test will make it null, commit, and check that the update was successful.

Otherwise looks good!

fstavela (approved):
LGTM

rodrigonull (approved):

PR #4112, opened 2026-04-29
refactor(RHINENG-26150): Don't pass static information around

ezr-ondrej:
We are passing the Timestamps around, while it's a static class which doesn't take much to initiate.

## Jira
[RHINENG-26150](https://issues.redhat.com/browse/RHINENG-26150)

## What

SImplify the serialize_host method by removing the static variable input and initialize it within the function

## Why

Simplify the code
## PR Guidelines

## PR Guideline checks

[RHINENG-26150]: https://redhat.atlassian.net/browse/RHINENG-26150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Inline computation of staleness timestamps instead of passing a shared Timestamps helper, simplifying host serialization and related flows.

Enhancements:
- Refine serialize_host and related helpers to depend only on staleness configuration, instantiating Timestamps internally from inventory configuration.
- Remove the staleness_timestamps helper and associated plumbing across API, MQ, export, and synchronization layers to reduce parameter passing of static data.
- Tighten type hints and default arguments for host serialization functions for clearer interfaces.

Tests:
- Update staleness, host serialization, MQ, and export service tests to reflect the new staleness timestamp computation model and removed staleness_timestamps parameter.


app/queue/host_mq.py
```diff
@@ -1065,7 +1055,8 @@ def write_add_update_event_message(
         current_operation="write_message_batch",
         inventory_id=result.row.id,
     ) as tracker_ctx:
-        output_host = serialize_host(result.row, result.staleness_timestamps, staleness=result.staleness_object)
+        staleness = result.staleness_object or get_staleness_obj(result.row.org_id)
```
ezr-ondrej, line 1137: We need to load the staleness_object unless it's set. This might be slightly problematic, unless cached, this loads staleness for every OperationResult which doesn't have that, which is usually the case.

Previously it passed this empty staleness_object without crashing, but `serialize_host` calls `get_staleness_timestamps` which in turn accesses the staleness object without checking, so passing staleness = None should crash.

I'd be happy to hear any opinion on this :)

tests/test_host_mq_service.py
```diff
@@ -3203,7 +3202,6 @@ class FakeHostRow:
     result = OperationResult(
         row=FakeHostRow(),
         pm={"request_id": "abc"},
-        st=None,
         so=None,
         et=EventType.created,
         sl=mock_success_logger,
```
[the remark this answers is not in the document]
ezr-ondrej: I think it's a niche, I'll wait until we agree we want to take this route.

app/queue/host_mq.py
kruai: Yep, passing `staleness = None` will hit an error when it tries to read `staleness["conventional_time_to_stale"]`. However, all OperationResults for "add" and "update" Host messages should have their `staleness_object` set ([here](https://github.com/ezr-ondrej/insights-host-inventory/blob/stop_dependency_inject/app/queue/host_mq.py#L460)). So I don't think we need the `or get_staleness_obj(result.row.org_id)` part of this line

kruai (changes requested):
Looks good overall! Just one suggestion (in response to your comment)

fstavela (dismissed):
LGTM

app/queue/host_mq.py
kruai: It turns out that we need the `or get_staleness_obj(result.row.org_id)` part of this to satisfy type-hints, and to absolutely guarantee that we don't pass None to serialize_host.

kruai (dismissed):

fstavela (approved):
LGTM

kruai (approved):

PR #4115, opened 2026-04-30
refactor(RHINENG-25845): Remove staleness persistence logic

FabriciaDinizRH:
## Jira
[RHINENG-25845](https://issues.redhat.com/browse/RHINENG-25845)

## What
Stop persisting host-level staleness columns (stale_timestamp, stale_warning_timestamp, deletion_timestamp) on normal ingest paths. Adjust listing / “non-culled” semantics and test fixtures (including mq_create_deleted_hosts) so behavior stays correct when those columns are NULL.

## Why
Phase 2 of compute-on-read: eliminate write amplification from recalculating and saving staleness on every check-in while reads already use computed values.

## How

- Model/API: Delete the Host method and call sites; wire api/staleness.py to apply_computed_staleness_timestamps_to_host(); remove staleness write from host_checkin.
- Queries / SQL truth: Update HostStalenessStatesDbFilters so NULL timestamps do not disappear from default staleness ORs and culled() is only true when deletion_timestamp is non-null and in the past; narrow _excluded_hosts_filter() accordingly.
- Logging: Guard log_add_host_attempt when stale_timestamp is unset.

## Testing
Passing unit tests and PR checks.

## PR Guidelines

## PR Guideline checks

## Summary by Sourcery

Stop persisting staleness timestamps on host check-in and treat NULL staleness columns as compute-on-read, while keeping persistence only for org-level staleness updates and deleted-host scenarios.

Enhancements:
- Adjust staleness state database filters and host exclusion logic to treat NULL staleness timestamps as active hosts and gate stale states on non-NULL columns.
- Extract a reusable helper to compute and persist staleness timestamps for hosts used by org-level staleness rebasing and MQ fixtures.
- Update host model construction, update flows, and instrumentation to avoid automatic staleness persistence at check-in and handle nullable staleness fields safely.

Tests:
- Refactor host creation fixtures to optionally skip persisting staleness timestamps and update model and staleness tests to assert NULL staleness columns for fresh hosts created via check-in paths.

[RHINENG-25845]: https://redhat.atlassian.net/browse/RHINENG-25845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Stop persisting host-level staleness timestamps on normal ingest paths and treat NULL staleness columns as compute-on-read while keeping behavior of listings, counts, and org-level staleness updates consistent.

Enhancements:
- Treat hosts with NULL staleness timestamps as fresh for listing and only classify stale, warning, or culled states when corresponding timestamps are non-NULL.
- Introduce a helper to compute and persist staleness timestamps for hosts and use it in DB fixtures, org-level staleness rebasing, and MQ deleted-host setup.
- Update host model lifecycle, logging, and repository exclusion filters to avoid automatic staleness persistence on check-in and to handle nullable staleness fields safely.

Tests:
- Adjust DB and MQ fixtures plus staleness/model tests to support optional staleness persistence and to assert NULL staleness timestamps for freshly created hosts via check-in paths.


app/models/utils.py
```diff
@@ -87,5 +88,16 @@ def _create_staleness_timestamps_values(host, org_id):
...
 
+def apply_computed_staleness_timestamps_to_host(host) -> None:
+    """Persist staleness columns from config + host facts (org staleness rebaseline only)."""
+    staleness_timestamps = _create_staleness_timestamps_values(host, host.org_id)
+    host.stale_timestamp = staleness_timestamps["stale_timestamp"]
+    host.stale_warning_timestamp = staleness_timestamps["stale_warning_timestamp"]
+    host.deletion_timestamp = staleness_timestamps["culled_timestamp"]
+    orm.attributes.flag_modified(host, "stale_timestamp")
+    orm.attributes.flag_modified(host, "stale_warning_timestamp")
+    orm.attributes.flag_modified(host, "deletion_timestamp")
```
fstavela: Just a question, I thought we don't want to persist the staleness timestamps anymore, why do we need this?

app/staleness_states.py
```diff
@@ -11,13 +12,24 @@ def __init__(self):
         self.now = datetime.now(tz=UTC)
 
     def fresh(self):
-        return self.now < Host.stale_timestamp
+        # NULL stale_timestamp: staleness not persisted (compute-on-read); treat as fresh for listing.
+        return or_(Host.stale_timestamp.is_(None), self.now < Host.stale_timestamp)
```
fstavela: I don't think this is correct. If we don't want to persist staleness timestamps anymore, then `Host.stale_timestamp` will always be `None`, so with this all hosts would be treated as fresh. This looks like the reason why the IQE tests are failing - it looks like the hosts in those tests are always fresh, even though they should have a different staleness state.

app/staleness_states.py
```diff
@@ -11,13 +12,24 @@ def __init__(self):
...
 
     def stale(self):
-        return and_(self.now >= Host.stale_timestamp, self.now < Host.stale_warning_timestamp)
+        return and_(
+            Host.stale_timestamp.isnot(None),
+            Host.stale_warning_timestamp.isnot(None),
+            self.now >= Host.stale_timestamp,
+            self.now < Host.stale_warning_timestamp,
+        )
```
fstavela: Shouldn't we calculate the states by `last_seen` timestamp, instead of the staleness timestamps?

fstavela (changes requested):

app/staleness_states.py
FabriciaDinizRH: yeah, I caught this one and just pushed the fix

app/staleness_states.py
FabriciaDinizRH: yeah, distraction got the best of me here. But I fixed all of the mentions to the staleness

app/staleness_states.py
```diff
@@ -1,23 +1,74 @@
...
+            self.now >= self._computed_stale_warning_boundary(),
+            self.now < self._computed_deletion_boundary(),
+        )
+        return or_(persisted, computed)
 
     def culled(self):
-        return self.now >= Host.deletion_timestamp
+        persisted = and_(Host.deletion_timestamp.isnot(None), self.now >= Host.deletion_timestamp)
+        computed = and_(self._computed_basis(), self.now >= self._computed_deletion_boundary())
+        return or_(persisted, computed)
```
fstavela: `Host.last_check_in` can never be `None`, it always has to be set. If it isn't, it means something is wrong and I would want to know about it even if it means failures, rather than it silently going on and then causing issues in the future.

I think we should remove the persisted calculation completely and always use the computed logic. There is nothing preventing us from doing it and it will make the code simpler and the behavior consistent between different hosts.

tests/fixtures/db_fixtures.py
```diff
@@ -151,15 +151,17 @@ def _db_get_groups_for_host(host_id, org_id=None):
...
+        extra_data: dict[str, Any] | None = None,
     ) -> Host:
         identity = identity or SYSTEM_IDENTITY
         extra_data = extra_data or {}
         org_id = extra_data.pop("org_id", None) or identity["org_id"]
 
         host = host or minimal_db_host(org_id=org_id, account=identity["account_number"], **extra_data)
         db.session.add(host)
-
+        db.session.flush()
```
fstavela: commit flushes automatically, so this is not needed

tests/fixtures/mq_fixtures.py
```diff
@@ -133,35 +137,41 @@ def mq_create_edge_host(mq_create_or_update_host):
...
-    staleness_timestamps = get_staleness_timestamps()
+    staleness_states = get_staleness_timestamps().keys()
     created_hosts = {}
-    for state, timestamp in staleness_timestamps.items():
-        host = minimal_host(
-            insights_id=generate_uuid(), stale_timestamp=timestamp.isoformat(), reporter="some reporter", facts=FACTS
-        )
+    for state in staleness_states:
+        host = minimal_host(insights_id=generate_uuid(), reporter="some reporter", facts=FACTS)
         created_hosts[state] = mq_create_or_update_host(host)
```
fstavela: This just creates multiple hosts now, but it doesn't adjust the staleness in any way. I guess this has been true even before, because the `stale_timestamp` from the payload has been ignored for a long time 🤔 Is any test still using this? If yes, then it probably doesn't do what we want it to do. If no, we can just delete this fixture.

tests/fixtures/mq_fixtures.py
```diff
@@ -133,35 +137,41 @@ def mq_create_edge_host(mq_create_or_update_host):
...
+        last_check_in_culled = datetime.now(tz=UTC) - timedelta(seconds=delete_seconds + 3600)
 
-        return created_hosts
+    for state in staleness_states:
+        host = minimal_host(
+            insights_id=generate_uuid(),
+            reporter="some reporter",
+            facts=FACTS,
+        )
+        hw = mq_create_or_update_host(host)
```
fstavela: What does `hw` stand for?

fstavela (changes requested):

tests/test_custom_staleness.py
```diff
@@ -737,3 +740,46 @@ def test_serialize_per_reporter_staleness_multiple_flat_reporters(flask_app):
         yupana = result["yupana"]
         assert yupana["last_check_in"] == "2024-06-16T12:00:00+00:00"
         assert "stale_timestamp" in yupana and "culled_timestamp" in yupana
+
+
+def test_staleness_db_filter_fresh_uses_org_defaults_when_columns_null(db_create_host):
```
rodrigonull, line 503: Can you add a test for stale_warning scenario?

Something like this:

```
def test_staleness_db_filter_stale_warning_when_columns_null(db_create_host):
    host = db_create_host()
    host.last_check_in = datetime.now(tz=UTC) - timedelta(days=10)
    db.session.commit()

    expr = _staleness_filter(["stale_warning"], host.org_id)
    assert Host.query.filter(Host.id == host.id).filter(*expr).count() == 1
```

api/staleness.py
rodrigonull:
```suggestion
                        logger.debug(f"Recomputed and emitted staleness events for {processed_hosts}/{num_hosts} hosts")
```

rodrigonull (commented):
The `def _create_staleness_timestamps_values(host, org_id)` can be removed in the `utils.py` file

api/staleness.py
```diff
@@ -147,7 +147,6 @@ def _update_hosts_staleness_async(identity: Identity, app: Flask, staleness_dict
 
                     with session_guard(hosts_query.session):
                         for host in batch_hosts:
-                            host._update_staleness_timestamps()
```
kruai, line 150: Do we need `_update_hosts_staleness_async` anymore? I think I asked this on a previous PR and it wasn't safe to remove it yet, but since this PR is removing staleness persistence, it feels like the right time

kruai (commented):
I just have a question/suggestion in addition to Fero's and Rodrigo's comments. If we can remove the async update of the staleness timestamps it will further simplify things

tests/fixtures/mq_fixtures.py
FabriciaDinizRH: host wrapper. I've updated it to add clarity

api/staleness.py
FabriciaDinizRH: yeah, totally safe to remove now

fstavela (changes requested):
Hey, please look at the failing IQE tests. From what I've seen, they are failing because HBI no longer produces update events when we create/update/delete staleness settings. This makes sense, since the hosts are no longer storing the staleness timestamps, but the tests currently expect the events, you will have to update them.

For example, here is one place where waiting for the events needs to be removed: https://github.com/RedHatInsights/insights-host-inventory/blob/master/iqe-host-inventory-plugin/iqe_host_inventory/modeling/staleness_api.py#L114

But there are other places and tests that will need to be updated as well.

api/staleness.py
rodrigonull:
```suggestion
    host_details = sa.inspect(host)
    prs_changed, _, _ = host_details.attrs.per_reporter_staleness.history
    if prs_changed:
        orm.attributes.flag_modified(host, "modified_on")
```

`deletion_timestamp` is no longer written on normal paths, so `staleness_changed` will always be empty. This branch is dead code now.

rodrigonull (changes requested):

jpramos123 (approved):
with IQE tests fixed, this LGTM!

Thanks

jpramos123 (commented):
I still see some IQE tests failing, I retriggered the pipeline, if they persist, please take a deeper look into it:

```
TestResourceTypesEmptyGroups.test_resource_types_get_groups_by_name_part[case insensitive]

TestResourceTypesEmptyGroups.test_resource_types_get_groups_by_name_part[case sensitive]
```

jpramos123 (dismissed):
IQE tests are passing now!

LGTM

fstavela (dismissed):
LGTM, thanks!

api/staleness.py
FabriciaDinizRH: removed

rodrigonull (approved):

jpramos123 (approved):

PR #4128, opened 2026-05-04
feat: Add inventory_views table, model, and Marshmallow schemas

adarshdubey-star:
## Jira
[RHINENG-25824](https://redhat.atlassian.net/browse/RHINENG-25824)

## What
Add the `hbi.inventory_views` table, SQLAlchemy model, and Marshmallow schemas for the Inventory Views feature.

## Why
Users need the ability to save and manage custom inventory views (saved filters, column configurations). This lays the database and validation foundation for the views CRUD endpoints.

## How
- Created an Alembic migration that adds `hbi.inventory_views` with columns: `id` (UUID PK), `org_id`, `name`, `description`, `is_system_view` (generated column — TRUE when org_id IS NULL), `configuration` (JSONB), `org_wide`, `created_by`, `created_at`, `updated_at`. Includes indexes on `org_id`, `is_system_view`, and `created_by`.
- Added an `InventoryView` SQLAlchemy model in `app/models/views.py` following the `Group` model pattern.
- Added three Marshmallow schemas in `app/models/schemas/views.py`:
  - `InputViewSchema` — validates POST payloads (`name` and `configuration` required)
  - `PatchViewSchema` — validates PATCH payloads (all fields optional)
  - `ViewResponseSchema` — serializes responses (includes all fields, system fields are `dump_only`)
- Exported the model and schemas from `app/models/__init__.py`.

## Testing
- Ran `flask db upgrade` — migration applied cleanly.
- Verified table structure in PostgreSQL (`\d hbi.inventory_views`) — all columns, types, defaults, generated column, and indexes present.
- Verified downgrade drops the table cleanly.
- Tested all three Marshmallow schemas: required field validation, whitespace stripping, empty name rejection, partial patch payloads, and response serialization.

## PR Guidelines

## PR Guideline checks

[RHINENG-25824]: https://redhat.atlassian.net/browse/RHINENG-25824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Introduce persistence and API validation layer for saved inventory views.

New Features:
- Add hbi.inventory_views table with metadata for saved inventory views, including org scoping, configuration JSON, and audit fields.
- Add InventoryView SQLAlchemy model to represent and manage inventory views in the application.
- Add Marshmallow schemas for creating, updating, and serializing inventory view payloads and responses.

Enhancements:
- Export the InventoryView model and its Marshmallow schemas from the models and schemas packages for reuse across the application.


migrations/versions/d4f7a2b8c1e3_create_inventory_views_table.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.dialects.postgresql import UUID
+
+# revision identifiers, used by Alembic.
+revision = "d4f7a2b8c1e3"
+down_revision = "c4e8f2a91b03"
```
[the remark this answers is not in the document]
adarshdubey-star: Good catch.

app/models/views.py
```diff
@@ -0,0 +1,44 @@
...
+    description = db.Column(db.Text, nullable=True)
+    is_system_view = db.Column(db.Boolean, Computed("org_id IS NULL"), nullable=False)
+    configuration = db.Column(JSONB, nullable=False)
+    org_wide = db.Column(db.Boolean, default=False, nullable=False)
+    created_by = db.Column(db.String(255), nullable=True)
+    created_at = db.Column(db.DateTime(timezone=True), default=_time_now, nullable=False)
+    updated_at = db.Column(db.DateTime(timezone=True), default=_time_now, onupdate=_time_now, nullable=False)
+
+    def patch(self, patch_data):
+        if "name" in patch_data:
```
rodrigonull, line 41: To make it consistent with other models we have, let's add this verification:

```
if not patch_data:
            raise InventoryException(title="Bad Request", detail="Patch json document cannot be empty.")
```

app/models/schemas/views.py
```diff
@@ -0,0 +1,43 @@
...
+        return in_data
+
+
+class ViewResponseSchema(MarshmallowSchema):
+    id = fields.UUID(dump_only=True)
+    org_id = fields.Str(dump_only=True)
+    name = fields.Str()
+    description = fields.Str(allow_none=True)
+    is_system_view = fields.Bool(dump_only=True)
+    configuration = fields.Dict()
```
rodrigonull, line 63: I think we could add a more strict schema for the configuration:

```
configuration = fields.Nested(ConfigurationSchema, required=True)

class ColumnSchema(MarshmallowSchema):
    key = fields.Str(required=True, validate=marshmallow_validate.Length(min=1))
    visible = fields.Bool(required=True)

class SortSchema(MarshmallowSchema):
    key = fields.Str(required=True, validate=marshmallow_validate.Length(min=1))
    direction = fields.Str(required=True, validate=marshmallow_validate.OneOf(["asc", "desc"]))

class FiltersSchema(MarshmallowSchema):
    class Meta:
        unknown = INCLUDE  # allow arbitrary filter keys

class ConfigurationSchema(MarshmallowSchema):
    columns = fields.List(fields.Nested(ColumnSchema), required=True)
    sort = fields.Nested(SortSchema, required=False)
    filters = fields.Nested(FiltersSchema, required=False)
```
WDYT?

app/models/views.py
```diff
@@ -0,0 +1,44 @@
...
+from app.models.utils import _time_now
+
+logger = get_logger(__name__)
+
+
+class InventoryView(db.Model):
+    __tablename__ = "inventory_views"
+    __table_args__ = (
+        Index("idx_inventory_views_org_id", "org_id"),
+        Index("idx_inventory_views_is_system_view", "is_system_view"),
```
rodrigonull: I was thinking about this and a B-tree index on a boolean column with only two distinct values provides very low selectivity. PostgreSQL's query planner will almost certainly choose a sequential scan over using this index.

I think we can remove it.

app/models/views.py
```diff
@@ -0,0 +1,44 @@
...
+        Index("idx_inventory_views_created_by", "created_by"),
```
rodrigonull: Let's use a composite index here: 

```suggestion
        Index("idx_inventory_views_org_id_created_by", "org_id", "created_by"),
```

app/models/views.py
```diff
@@ -0,0 +1,44 @@
...
+    id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    org_id = db.Column(db.String(36), nullable=True)
+    name = db.Column(db.String(255), nullable=False)
+    description = db.Column(db.Text, nullable=True)
+    is_system_view = db.Column(db.Boolean, Computed("org_id IS NULL"), nullable=False)
+    configuration = db.Column(JSONB, nullable=False)
+    org_wide = db.Column(db.Boolean, default=False, nullable=False)
+    created_by = db.Column(db.String(255), nullable=True)
+    created_at = db.Column(db.DateTime(timezone=True), default=_time_now, nullable=False)
+    updated_at = db.Column(db.DateTime(timezone=True), default=_time_now, onupdate=_time_now, nullable=False)
```
rodrigonull:
```suggestion
    created_on = db.Column(db.DateTime(timezone=True), default=_time_now, nullable=False)
    modified_on = db.Column(db.DateTime(timezone=True), default=_time_now, onupdate=_time_now, nullable=False)
```

rodrigonull (changes requested):
Can you add some unit tests for the new model?

app/models/views.py
adarshdubey-star: cool

app/models/views.py
adarshdubey-star: This makes more sense

app/models/views.py
adarshdubey-star: okay I will add this

app/models/schemas/views.py
adarshdubey-star: Yeah, the previous one was a little loose; this one is more structured for validation.

tests/test_models.py
```diff
@@ -2928,3 +2928,255 @@ def test_ungrouped_group_cache_deduplicates_group_creation_rbac_v2(flask_app, mo
     mock_rbac.assert_called_once_with(mock_identity)
     mock_wait.assert_called_once()
     mock_get_rbac_ws.assert_called_once_with(str(workspace_id))
+
+
+class TestInventoryViewPatch:
+    def test_patch_updates_name(self, flask_app):  # noqa: ARG002
+        from app.models.views import InventoryView
```
rodrigonull: This could be imported at module level. To be consistent with other tests we have.

rodrigonull (dismissed):
I left just a minor suggestion. The changes LGTM!

I would suggest to have a second pair of eyes on this one before merging it.

kruai (dismissed):
Looks good! Only concern I have is that we plan to append "Copy" or "Clone" to the name of a View when the user clones it. We could either account for this in our max lengths (which we should do in this PR), or trim the end of the name if it's too long when cloning (no-op for this PR). What do you think @adarshdubey-star @jpramos123 @rodrigonull?

jpramos123 (dismissed):

jpramos123:
> Looks good! Only concern I have is that we plan to append "Copy" or "Clone" to the name of a View when the user clones it. We could either account for this in our max lengths (which we should do in this PR), or trim the end of the name if it's too long when cloning (no-op for this PR). What do you think @adarshdubey-star @jpramos123 @rodrigonull?

I would rather account to the max lenghts in this PR, and make sure it is merged with edge cases already accounted.

Good catch Asa

rodrigonull:
> > Looks good! Only concern I have is that we plan to append "Copy" or "Clone" to the name of a View when the user clones it. We could either account for this in our max lengths (which we should do in this PR), or trim the end of the name if it's too long when cloning (no-op for this PR). What do you think @adarshdubey-star @jpramos123 @rodrigonull?
> 
> I would rather account to the max lenghts in this PR, and make sure it is merged with edge cases already accounted.
> 
> Good catch Asa

Agreed

adarshdubey-star:
## Poke Review Summary

**Risk Level:** High
**Files Reviewed:** 6
**Lenses Applied:** security, migration, test

### Findings by Category

- **Migration Safety** (1 critical)

### Critical Issues

- 🛑 `migrations/versions/d4f7a2b8c1e3_create_inventory_views_table.py` (line 10): INVENTORY_SCHEMA is hardcoded as 'hbi' instead of being imported from app.models.constants. This can lead to inconsistencies if the schema name changes.

---
*Poked by Poke | 1 finding(s) total*

adarshdubey-star:
> Looks good! Only concern I have is that we plan to append "Copy" or "Clone" to the name of a View when the user clones it. We could either account for this in our max lengths (which we should do in this PR), or trim the end of the name if it's too long when cloning (no-op for this PR). What do you think @adarshdubey-star @jpramos123 @rodrigonull?

Good catch @kruai! I've added a MAX_VIEW_NAME_LENGTH = 255 constant in app/models/views.py and referenced it in both the model and Marshmallow schemas, no more magic numbers.

For the clone overflow, I'll keep the user-facing input max at 255 (no reason to penalize users for a system operation). In RHINENG-25828, the clone_view function will truncate: f"Copy of {name}"[: MAX_VIEW_NAME_LENGTH]. This handles all edge cases, including nested clones ("Copy of Copy of...") without an arbitrary reduced limit.

kruai (approved):
LGTM, thanks!

rodrigonull (approved):

jpramos123 (approved):

PR #4168, opened 2026-05-07
Remove ingress upload workaround

mshriver:
## Jira
https://redhat.atlassian.net/browse/IQE-4039

## What
return to using upload_post, keep the mimetype add

## Why
This was a workaround for an ingress bug

## How
sonnet undo the do

## Testing

`ENV_FOR_DYNACONF=stage_proxy iqe tests plugin host_inventory --critical -v -m "not ephemeral" iqe_host_inventory/tests/rest/test_upload_host_creation.py`

```
======== 9 passed, 6 warnings in 47.37s ====
```

## PR Guidelines

## PR Guideline checks

## Summary by Sourcery

Enhancements:
- Simplify ingress upload logic by switching from upload_post_without_preload_content back to the default upload_post call and removing the associated custom response handling.

## Summary by Sourcery

Enhancements:
- Simplify ingress upload logic by using the default upload_post API instead of the workaround method and removing custom response handling for the workaround.


iqe-host-inventory-plugin/iqe_host_inventory/modeling/uploads.py
```diff
@@ -115,17 +113,13 @@ def _reset_ingress_api(self) -> None:
...
         file_data = pathlib.Path(filepath).read_bytes()
         file_extension = _CONTENT_TYPE_TO_EXTENSION.get(content_type, ".bin")
 
         for attempt in range(2):
             try:
-                response = self._ingress_api.upload_post_without_preload_content(
+                self._ingress_api.upload_post(
                     file=(f"upload{file_extension}", file_data),
                     metadata={"content_type": content_type},
                 )
```
fstavela, line 125: Could we now go back to simpler
```
self._ingress_api.upload_post(file=archive.filename, content_type=_FILE_TYPE)
```
without reading the file data as bytes and setting the file extensions?

Ref: https://github.com/RedHatInsights/insights-host-inventory/pull/3832#discussion_r2993366785

https://github.com/RedHatInsights/insights-host-inventory/pull/3832/changes#diff-bdd566d28732ead069edfb91daa1bc9336323691c3bf75e6155e59b6767018b5L167

mshriver: the content_type kwarg is no longer available and has to be passed as metadata.

In testing, if we don't read the file data and use `mimetypes.add_type` then the api client will infer the wrong type, resulting in a 415 from ingress.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/uploads.py
fstavela, line 32: Do we still need this?

iqe-host-inventory-plugin/iqe_host_inventory/modeling/uploads.py
fstavela: And do we still need `mimetypes.add_type` ? I'm just asking because just passing the file path instead of the byte contents used to work in the past.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/uploads.py
mshriver: yes, if you just pass a file the v7 api client will guess the mimetype wrong because it strips one file type

fstavela (approved):
LGTM, thank you!

kruai (approved):

mshriver:
```  
FAILED iqe-local-plugin/iqe_host_inventory/tests/rest/notifications/kafka/test_notifications_kafka_stale.py::test_notifications_kafka_host_stale_toggle_via_refresh
  = 1 failed, 4647 passed, 65 skipped, 587 deselected, 13 xfailed, 143 warnings, 3 manual in 3644.48s (1:00:44) =
  ```
  
  I do not think this single failure in the insights-hbi-all/insights-host-inventory konflux pipeline is related to this change.

PR #4171, opened 2026-05-07
fix(RHINENG-15779): sqldumps decorator doesn't work

kruai:
## Summary
The `dumps_sql` decorator attached the SQLAlchemy `before_execute` listener and then returned immediately when the wrapped callable was a **generator** (e.g. `get_hosts_to_export`). The generator body runs only while the consumer iterates, so all queries ran after the listener was removed. The context manager worked because it stays entered for the whole `with` block across yields.

## Jira
[RHINENG-15779](https://issues.redhat.com/browse/RHINENG-15779)

## Changes
- Wrap generator functions with `yield from` so the listener stays active for the full iteration
- Support async and async-generator callables the same way
- Use `try`/`finally` so the listener is always removed (including on exceptions)
- Fix `SQLDump.__init__` to honor a non-default `dump_method`
- Add `sqldumps` as an alias for `dumps_sql` (ticket wording)
- Add regression tests in `tests/helpers/test_sql_dump.py`

## Testing
- `pipenv run pytest tests/helpers/test_sql_dump.py -v`
- `make style`

[RHINENG-15779]: https://redhat.atlassian.net/browse/RHINENG-15779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Ensure the SQL query dumping helper works consistently for sync, async, and generator callables and add regression coverage.

Bug Fixes:
- Keep the SQLAlchemy before_execute listener active for the full lifetime of generator, async-generator, and coroutine functions decorated with dumps_sql.
- Honor a non-default dump_method passed to SQLDump when constructing the helper.

Enhancements:
- Add sqldumps as an alias of dumps_sql for compatibility with existing references.

Tests:
- Add regression tests verifying dumps_sql logs SQL for synchronous functions and generators, including during iteration, and for async functions.

### DIFF
--- a/tests/helpers/sql_dump.py
+++ b/tests/helpers/sql_dump.py
@@ -1,3 +1,4 @@
+import inspect
 import json
 from functools import wraps
 
@@ -28,7 +29,7 @@ Usage:
     from tests.helpers.sql_dump import dumps_sql
     :
     :
-    # Or decorator for whole method
+    # Or decorator for whole method (including generators and async functions)
     @dumps_sql
     def test_find_host_using_superset_canonical_fact_match(db_create_host):
 """
@@ -36,8 +37,7 @@ Usage:
 
 class SQLDump:
     def __init__(self, dump_method=None, write_method=print):
-        if dump_method is None:
-            self.dump_method = self.dump_sql
+        self.dump_method = self.dump_sql if dump_method is None else dump_method
         self.write_method = write_method
 
     def __enter__(self):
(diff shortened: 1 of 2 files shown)


kruai:
@sourcery-ai review

kruai:
Addressed Sourcery feedback in the latest push: shared `_sqldump_enter`/`_sqldump_exit` helpers with `sys.exc_info()` for teardown, plus tests for async generators, custom `dump_method`, and listener removal after exceptions.

PR #4182, opened 2026-05-08
chore(RHINENG-26341): Migrate from pipenv to uv

kruai:
## Jira

[RHINENG-26341](https://issues.redhat.com/browse/RHINENG-26341)

## What

Migrates the **main Host Based Inventory application** from Pipenv to [uv](https://docs.astral.sh/uv/). **IQE continues to use Pipenv** via `Pipfile.iqe` (unchanged workflow for CJI pods and local IQE setup).

### Main app (uv)

- Root **`pyproject.toml`** + **`uv.lock`**; runtime and dev deps in `[project]` and `[dependency-groups]`.
- **`uv sync`** / **`uv run`** replace `pipenv install` / `pipenv run` for local dev, CI, and make targets.
- Removed root **`Pipfile`** and **`Pipfile.lock`** only.
- **`[tool.uv] required-version`** pins the uv CLI (read by CI, Dockerfiles, and hermetic scripts).

### Containers

- **`Dockerfile`**, **`Dockerfile.upstream`**, and **`dev.dockerfile`**: copy `pyproject.toml` / `uv.lock`, install uv from `required-version`, **`uv sync --frozen`** into **`$APP_ROOT/.venv`**, prepend **`.venv/bin`** to `PATH`.
- Production image uses **`--no-dev`**. Dev image uses **`ubi9/python-312`**.
- **`run_command.sh`**: runs `./*.py` entrypoints with the venv interpreter; uses **`exec`** (not `eval`) for other commands (e.g. `FLASK_APP=./manage.py flask db upgrade`).

### CI and automation

- **`checks.yaml`**: Python **3.12** (aligned with `requires-python`), **`setup-uv`** with `version-file: pyproject.toml`, **`uv sync --frozen`**, tests via **`uv run`**.
- Hermetic builds: **`uv export`** for `requirements.txt`; build deps pin uv via `pyproject.toml`.
- **Dependabot / automerge**: root **`pyproject.toml`** + **`uv.lock`**; IQE **`Pipfile.iqe`** / **`Pipfile.iqe.lock`** (not uv).

### IQE (still Pipenv)

- **`Pipfile.iqe`** and **`Pipfile.iqe.lock`** retained.
- **`setup-iqe.sh`**, **`iqe-shell.sh`**, **`docs/IQE.md`**, and **`run_cji_with_local_plugin.sh`** unchanged (pipenv locally; `pip install -e .` in the IQE pod).
- Plugin **`pyproject.toml`** stays a minimal package manifest for editable installs (not a separate uv environment).

### Docs and tooling

- README, CLAUDE.md, `.cursor` rules, `.claude` commands/hooks, and debug docs updated for the split: **uv for main app**, **pipenv for IQE**.
- Job/service shebangs use **`/usr/bin/env python`** where needed so container **`PATH`** selects the venv.

**Note:** `iqe-host-inventory-plugin/scripts/rbac_local.sh` still uses Pipenv for the separate **insights-rbac** repo.

## Why

uv gives faster installs, a standard PEP 621 lockfile, and simpler CI caching for the main service. IQE stays on Pipenv because the CJI pod and Nexus-backed IQE stack already expect that flow; migrating IQE added little value and broke pod installs.

## How

- `uvx migrate-to-uv --ignore-locked-versions` for the root app (strict Pipenv lock resolution failed across `requires-python` / transitive constraints).
- `requires-python = ">=3.12,<3.13"`; OpenTelemetry pins aligned with the lockfile (1.41.x / 0.62b1).
- `kessel-sdk[auth]` only (no invalid `console` extra).
- Follow-up fixes: drop invalid `uv sync --system`; install into `.venv` + `PATH`; `run_command.sh` venv routing for MQ/export/Cron job scripts.

## Testing

- `uv sync` then `uv run make style` (pre-commit, ruff, mypy): pass.
- `uv run pytest --cov=.`: pass.

## PR Guidelines

## PR Guideline checks

[RHINENG-26341]: https://redhat.atlassian.net/browse/RHINENG-26341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


run_command.sh
```diff
@@ -5,8 +5,14 @@ if [ "$#" -eq 0 ]; then
     exit 1
```
rodrigonull, line 5: The switch from eval "$@" to exec "$@" breaks two arg patterns in deploy/clowdapp.yml that rely on shell interpretation:

db-migration initContainer (line 1867): args: ["FLASK_APP=./manage.py", "flask", "db", "upgrade"] — exec doesn't recognize quoted VAR=value as an assignment, so it tries to execute a file literally named FLASK_APP=./manage.py.

synchronizer job (line 1413): args: ["./jobs/rebuild_events_topic.py", "&&", "./jobs/host_synchronizer.py"] — exec "$VENV_PYTHON" "$@" passes && as a literal argument to the python script; the second script never runs.

Suggested fix:

```
#!/bin/bash

if [ "$#" -eq 0 ]; then
    echo "ERROR: No command provided to run."
    exit 1
fi

APP_ROOT="${APP_ROOT:-/opt/app-root/src}"
VENV_PYTHON="${APP_ROOT}/.venv/bin/python"

# Export leading VAR=value arguments (e.g. FLASK_APP=./manage.py flask db upgrade)
while [[ $# -gt 0 && "$1" == *'='* ]]; do
    export "$1"
    shift
done

if [ "$#" -eq 0 ]; then
    echo "ERROR: Only environment assignments provided, no command to run."
    exit 1
fi

# If args contain shell operators (&&, ||, ;), delegate to sh -c for interpretation.
# This handles ClowdApp args like ["./script_a.py", "&&", "./script_b.py"].
for arg in "$@"; do
    if [[ "$arg" == "&&" || "$arg" == "||" || "$arg" == ";" || "$arg" == "|" ]]; then
        exec /bin/sh -c "$*"
    fi
done

# ClowdApp runs scripts as ./foo.py; the shebang must not use /usr/bin/python because
# that is the RPM interpreter without project deps. Run .py entrypoints with the venv.
if [ -x "$VENV_PYTHON" ] && [[ "$1" == *'.py' ]]; then
    exec "$VENV_PYTHON" "$@"
fi

exec "$@"
```

mk/base.mk
```diff
@@ -7,7 +7,7 @@ NUM_HOSTS=1
 
 
 init:
-	pipenv shell
+	uv shell
```
rodrigonull: `uv shell` does not exist, it's a [requested but unimplemented feature](https://github.com/astral-sh/uv/issues/1910). Running make init will fail with an error.

rodrigonull (changes requested):

run_command.sh
kruai: Sorry about that - I had claude address Sourcery's suggestions right before asking for reviews, and didn't realize that they introduced some breaks. I should have reviewed them more thoroughly.

rodrigonull: No worries :)

kruai: OK, I've fixed it

mk/base.mk
kruai: Whoops, thanks - replaced it.

rodrigonull (dismissed):
LGTM. Thank you!

README.md
```diff
@@ -126,13 +126,13 @@ source .env
 1. **Install dependencies**:
 
 ```bash
-pipenv install --dev
+uv sync
```
fstavela, line 129: Should we use `--frozen` or `--locked` when installing the env locally? So that we don't update the lock file unintentionally

README.md
```diff
@@ -126,13 +126,13 @@ source .env
...
 ```
 
-2. **Activate virtual environment**:
+2. **Activate virtual environment** (optional):
```
fstavela: Hmm, is it really optional? Won't `make upgrade_db` or running the unit tests fail without being in the right env?

fstavela (commented):
I just have a couple of questions, otherwise looks good, thanks for all the updates!

Edit: One more question - will dependabot and konflux bot be able to update dependencies in the uv lock file?

README.md
kruai: The lockfile should already be up-to-date each time the dependencies are updated, so it _shouldn't_ be an issue. But if the dependencies have drifted for some reason, we do want the lockfile to be updated ASAP (rather than having it raise an error)

fstavela (dismissed):
LGTM!

README.md
kruai: Well... it's technically optional, though you have to prepend other commands with `uv run` if you don't activate the venv. Probably best to keep it simple and just standardize on the venv

kruai:
> I just have a couple of questions, otherwise looks good, thanks for all the updates!
> 
> Edit: One more question - will dependabot and konflux bot be able to update dependencies in the uv lock file?

For dependabot, yes! I updated the dependabot config so it should update uv dependencies as well.
For Konflux, I confess I don't know how those builds work. Do you know where the config for those are?

fstavela:
> > I just have a couple of questions, otherwise looks good, thanks for all the updates!
> > Edit: One more question - will dependabot and konflux bot be able to update dependencies in the uv lock file?
> 
> For dependabot, yes! I updated the dependabot config so it should update uv dependencies as well. For Konflux, I confess I don't know how those builds work. Do you know where the config for those are?

No idea :(

rodrigonull (dismissed):

rodrigonull (approved):

PR #4324, opened 2026-05-25
refactor(RHINENG-25499): Drop Staleness indexes and columns

FabriciaDinizRH:
## Jira
[RHINENG-25499](https://issues.redhat.com/browse/RHINENG-25499)

## What
Phase 3 of the compute-on-read migration: drop persisted staleness columns and the covering index on hbi.hosts, and remove application code that depended on them.

## Why
Staleness is now derived at read time from last_check_in and per_reporter_staleness (Stories 1–2). The three timestamp columns and idx_hosts_org_id_deletion_ts (built for (org_id, id, deletion_timestamp)) are redundant and add write/index maintenance cost. This change completes the migration by removing them from the schema and codebase.

## How

1. Alembic upgrade() drops the partitioned index first, then the three columns on hbi.hosts (parent table; propagates to partitions).
2. Host model no longer defines or accepts persisted staleness fields.
3. Culled-host tests set host.last_check_in after create (same pattern as MQ fixtures / group host-count tests) instead of seeding stale_timestamp in extra_data.
4. IQE plugin tests and utilities updated to stop asserting or seeding removed DB columns.

## Testing

- unit tests and IQE tests were adjusted to reflect the new model.
- all tests pass.
**Note**: Downgrade restores nullable columns and the index for dev/ephemeral use only; production rollback is not supported once compute-on-read is the sole source of truth.

## PR Guidelines

## PR Guideline checks

[RHINENG-25499]: https://redhat.atlassian.net/browse/RHINENG-25499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Remove persisted staleness columns and index from the hosts data model in favor of compute-on-read based on last_check_in, and update application/test code and IQE plugin accordingly.

Enhancements:
- Drop stale_timestamp, stale_warning_timestamp, and deletion_timestamp columns from the hosts table and remove the covering index used for host staleness queries.
- Update the Host ORM model, schemas, query projections, utilities, and instrumentation to stop exposing or depending on persisted staleness columns and rely on last_check_in and per_reporter_staleness instead.
- Adjust IQE host-inventory plugin data generation, DB models, Kafka helpers, and staleness utilities to treat last_check_in as the source for lifecycle timestamps and staleness behavior.

Documentation:
- Clarify staleness-related requirements in the IQE plugin requirements to remove references to reporters setting a global stale_timestamp and persisted staleness columns.

Tests:
- Refactor unit, API, MQ, culling, staleness, and job synchronizer tests to stop seeding or asserting stale_timestamp-related fields and validate behavior based on last_check_in and per_reporter_staleness.
- Update IQE REST and cache tests to use last_check_in in payloads and expectations, and relax assertions that previously required non-null DB staleness columns.

[SWATCH-5008]: https://redhat.atlassian.net/browse/SWATCH-5008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Remove persisted staleness fields from the host data model and rely on last_check_in and per_reporter_staleness for lifecycle computation.

Enhancements:
- Drop host-level staleness columns from the ORM model, query projections, and instrumentation, and stop accepting stale_timestamp in API payloads.
- Adjust helper utilities and serializers to compute lifecycle timestamps from last_check_in and per_reporter_staleness instead of persisted staleness columns.
- Update culling, synchronization, and custom staleness flows to use last_check_in when marking or filtering culled hosts.

Documentation:
- Refresh IQE plugin requirements and helper comments to remove references to reporters setting global stale_timestamp or persisted staleness columns.

Tests:
- Refactor unit, MQ, API, culling, and job synchronizer tests to stop seeding or asserting stale_timestamp-related fields and validate behavior based on last_check_in and per_reporter_staleness.
- Update IQE plugin DB models, data generators, REST tests, and staleness utilities to remove dependencies on persisted staleness columns and to assert behavior using last_check_in-driven staleness.

Chores:
- Add an Alembic migration that drops the hosts staleness columns and the covering index, with a dev-only downgrade path that restores nullable columns and the index.

### DIFF
--- a/app/models/schemas/host.py
+++ b/app/models/schemas/host.py
@@ -460,7 +460,6 @@ class LimitedHostSchema(CanonicalFactsSchema):
 
 
 class HostSchema(LimitedHostSchema):
-    stale_timestamp = fields.AwareDateTime(required=False)
     reporter = fields.Str(required=True, validate=marshmallow_validate.Length(min=1, max=255))
 
     @staticmethod
--- a/app/models/schemas/host.py
+++ b/app/models/schemas/host.py
@@ -477,7 +477,6 @@ class HostSchema(LimitedHostSchema):
             tags=tags,
             tags_alt=tags_alt,
             system_profile_facts=data.get("system_profile", {}),
-            stale_timestamp=data.get("stale_timestamp"),
             reporter=data["reporter"],
             groups=data.get("groups", []),
             insights_id=data.get("insights_id"),
--- a/docs/entity_relationship_diagram.pu
+++ b/docs/entity_relationship_diagram.pu
@@ -23,8 +23,6 @@ table(hbi.hosts) {
   jsonb(canonical_facts): jsonb <<not null>>
   jsonb(system_profile_facts): jsonb
   column(ansible_host): varchar(255)
-  column(stale_warning_timestamp): timestamp with time zone
-  column(deletion_timestamp): timestamp with time zone
   column(reporter): varchar(255) <<not null>>
   column(display_name_reporter): varchar(255)
   jsonb(per_reporter_staleness): jsonb <<not null>>
--- a/docs/entity_relationship_diagram.pu
+++ b/docs/entity_relationship_diagram.pu
@@ -23,7 +23,6 @@ table(hbi.hosts) {
   jsonb(canonical_facts): jsonb <<not null>>
   jsonb(system_profile_facts): jsonb
   column(ansible_host): varchar(255)
-  column(stale_timestamp): timestamp with time zone
   column(stale_warning_timestamp): timestamp with time zone
   column(deletion_timestamp): timestamp with time zone
   column(reporter): varchar(255) <<not null>>
--- a/iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/cache/test_cache_insights_client.py
+++ b/iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/cache/test_cache_insights_client.py
@@ -437,7 +437,7 @@ def test_cache_invalidation_insights_client_get_hosts_list_create_staleness(
     response = host_inventory_cert_auth.apis.hosts.get_hosts(insights_id=host.insights_id)
     assert len(response) == 1
     assert response[0].id == host.id
-    assert response[0].last_check_in != initial_timestamp
+    assert response[0].stale_timestamp != initial_timestamp
 
 
 @iqe_blocker(iqe_blocker.jira("RHINENG-18758", category=iqe_blocker.PRODUCT_ISSUE))
@@ -468,7 +468,7 @@ def test_cache_invalidation_insights_client_get_hosts_list_update_staleness(
     response = host_inventory_cert_auth.apis.hosts.get_hosts(insights_id=host.insights_id)
     assert len(response) == 1
     assert response[0].id == host.id
-    initial_timestamp = response[0].last_check_in
+    initial_timestamp = response[0].stale_timestamp
 
     # Update staleness config to invalidate the cache
     host_inventory.apis.account_staleness.update_staleness(conventional_time_to_stale=100)
@@ -479,7 +479,7 @@ def test_cache_invalidation_insights_client_get_hosts_list_update_staleness(
     response = host_inventory_cert_auth.apis.hosts.get_hosts(insights_id=host.insights_id)
     assert len(response) == 1
     assert response[0].id == host.id
-    assert response[0].last_check_in != initial_timestamp
+    assert response[0].stale_timestamp != initial_timestamp
 
 
 @iqe_blocker(iqe_blocker.jira("RHINENG-18758", category=iqe_blocker.PRODUCT_ISSUE))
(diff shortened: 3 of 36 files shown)


docs/entity_relationship_diagram.pu
```diff
@@ -23,7 +23,6 @@ table(hbi.hosts) {
   jsonb(canonical_facts): jsonb <<not null>>
   jsonb(system_profile_facts): jsonb
   column(ansible_host): varchar(255)
-  column(stale_timestamp): timestamp with time zone
   column(stale_warning_timestamp): timestamp with time zone
   column(deletion_timestamp): timestamp with time zone
```
rodrigonull: This 2 should be removed too, right?

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/cache/test_cache_insights_client.py
```diff
@@ -426,7 +426,7 @@ def test_cache_invalidation_insights_client_get_hosts_list_create_staleness(
     response = host_inventory_cert_auth.apis.hosts.get_hosts(insights_id=host.insights_id)
     assert len(response) == 1
     assert response[0].id == host.id
-    initial_timestamp = response[0].stale_timestamp
```
rodrigonull, line 429: We still return the stale_timestamp in the API, should we keep the tests as it was? I am afraid this change will break the tests because last_check_in does not change when you modify the org's staleness config.

tests/test_host_mq_service.py
```diff
@@ -1684,42 +1684,6 @@ def test_delete_host_tags(mq_create_or_update_host, db_get_host_by_insights_id,
             assert Tag.create_flat_tags_from_structured(structured_tags) == record.tags_alt
```
rodrigonull, line 1668: Three places in this file that still set input_host.stale_timestamp = None on HostWrapper instances (lines 1962, 2726, 3363). Since the stale_timestamp property was removed from HostWrapper in app/utils.py, this now silently creates a regular Python attribute on the instance instead of setting a value in the underlying __data dict. It's dead code that should be cleaned up in this PR.

app/models/schemas/host.py
rodrigonull: I think this can be removed.

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/staleness/test_per_reporter_staleness.py
```diff
@@ -103,6 +103,7 @@ def verify_staleness(
     assert reporter_staleness.check_in_succeeded
 
     assert_datetimes_equal(reporter_staleness.last_check_in, host.last_check_in)
+    # update to check for the calculatd values
```
rodrigonull:
```suggestion
    # update to check for the calculated values
```

rodrigonull (changes requested):

docs/entity_relationship_diagram.pu
FabriciaDinizRH: yeah, I feel like a couple of things were put back while merge conflict resolution. I'm cleaning everything up

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/staleness/test_per_reporter_staleness.py
FabriciaDinizRH: done :)

app/models/schemas/host.py
FabriciaDinizRH: done :)

tests/test_host_mq_service.py
FabriciaDinizRH: done :)

migrations/versions/f8a9b0c1d2e3_drop_hosts_staleness_columns_and_index.py
```diff
@@ -0,0 +1,67 @@
...
+    op.add_column(
+        "hosts",
+        sa.Column("stale_warning_timestamp", sa.DateTime(timezone=True), nullable=True),
+        schema=INVENTORY_SCHEMA,
+    )
+    op.add_column(
+        "hosts",
+        sa.Column("deletion_timestamp", sa.DateTime(timezone=True), nullable=True),
+        schema=INVENTORY_SCHEMA,
+    )
```
kruai: Just to follow the same pattern you used in `upgrade()`:
```suggestion
    for column_name in STALENESS_COLUMNS:
        op.add_column(
            "hosts",
            sa.Column(column_name, sa.DateTime(timezone=True), nullable=True),
            schema=INVENTORY_SCHEMA,
        )
```

kruai (changes requested):
In addition to Rodrigo's comments, I wanted to suggest a small simplification

docs/entity_relationship_diagram.pu
FabriciaDinizRH: done :)

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/cache/test_cache_insights_client.py
FabriciaDinizRH: done :)

kruai:
@FabriciaDinizRH This PR has some commits that are not signed/verified, so it's not going to allow you to merge it. You'll need to open another one

FabriciaDinizRH:
Opened https://github.com/RedHatInsights/insights-host-inventory/pull/4459 to clean commit history

PR #4404, opened 2026-06-01
feat(RHINENG-25828): Add inventory views repository with CRUD and visibility logic

adarshdubey-star:
Jira
[RHINENG-25828](https://issues.redhat.com/browse/RHINENG-25828)

What
Add lib/views_repository.py with CRUD operations and visibility logic for inventory views.

Why
The endpoint handlers (RHINENG-25829–25833) need a data access layer to query, create, update, delete, and clone inventory views with proper visibility filtering and ownership enforcement.

How
Created lib/views_repository.py with 6 functions: get_views_list, get_view_by_id, create_view, update_view, delete_view, clone_view.
Visibility filter: system views (org_id IS NULL) + org-wide views + user's own private views.
Ownership check: only created_by user can update/delete (403).
Immutability check: system views cannot be updated/deleted (403).
Clone copies configuration (deep copy), sets name to "Copy of {name}" (truncated to 255), org_wide=False.
All responses include computed is_owner field.
Testing
29 unit tests in tests/test_views_repository.py covering visibility, pagination, CRUD, ownership/immutability guards, clone edge cases.
Full test suite: 2765 passed, 0 failed.
PR Guidelines

PR Guideline checks

 Keep PRs under 400 lines of meaningful changes, including tests, excluding auto-generated files, config, etc.

[RHINENG-25828]: https://redhat.atlassian.net/browse/RHINENG-25828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Introduce a repository layer for inventory views with CRUD, cloning, and visibility/ownership enforcement, along with unit tests covering these behaviors.

New Features:
- Add views repository module providing list, detail, create, update, delete, and clone operations for inventory views with serialization helpers.
- Support visibility rules for system, org-wide, and user-private inventory views, including ownership flags and system-view detection.
- Allow cloning of existing inventory views into user-owned, org-scoped copies with deep-copied configuration and name truncation.

Tests:
- Add unit test suite for the views repository covering visibility filtering, pagination, CRUD behaviors, ownership and immutability checks, and cloning edge cases.

### DIFF
--- a/app/serialization.py
+++ b/app/serialization.py
@@ -26,6 +26,7 @@ from app.models import LimitedHostSchema
 from app.models.constants import WORKLOADS_FIELDS
 from app.models.system_profile_transformer import DYNAMIC_FIELDS
 from app.models.system_profile_transformer import STATIC_FIELDS
+from app.models.views import InventoryView
 from app.staleness_serialization import AttrDict
 from app.staleness_serialization import get_staleness_timestamps
 from app.utils import Tag
@@ -583,6 +584,22 @@ def serialize_staleness_response(staleness):
         "immutable_time_to_delete": staleness.conventional_time_to_delete,
         "created": _serialize_datetime(staleness.created_on) if staleness.created_on is not None else None,
         "updated": _serialize_datetime(staleness.modified_on) if staleness.modified_on is not None else None,
+    }
+
+
+def serialize_view(view: InventoryView, username: str) -> dict:
+    return {
+        "id": serialize_uuid(view.id),
+        "org_id": view.org_id,
+        "name": view.name,
+        "description": view.description,
+        "is_system_view": view.is_system_view,
+        "configuration": view.configuration,
+        "org_wide": view.org_wide,
+        "is_owner": view.created_by == username,
+        "created_by": view.created_by,
+        "created_at": _serialize_datetime(view.created_on) if view.created_on else None,
+        "updated_at": _serialize_datetime(view.modified_on) if view.modified_on else None,
     }
 
 
--- a/lib/views_repository.py
+++ b/lib/views_repository.py
@@ -1,0 +1,153 @@
+from __future__ import annotations
+
+from copy import deepcopy
+
+from sqlalchemy import and_
+from sqlalchemy import or_
+
+from app.exceptions import InventoryException
+from app.logging import get_logger
+from app.models import InventoryView
+from app.models import db
+from app.models.views import MAX_VIEW_NAME_LENGTH
+from lib.db import session_guard
+
+logger = get_logger(__name__)
+
+__all__ = (
+    "get_views_list",
+    "get_view_by_id",
+    "create_view",
+    "update_view",
+    "delete_view",
+    "clone_view",
+)
+
+CLONE_NAME_PREFIX = "Copy of "
+
+
+class ViewNotFoundError(InventoryException):
+    def __init__(self, detail: str = "View not found."):
+        super().__init__(status=404, title="Not Found", detail=detail)
+
+
+class ViewPermissionError(InventoryException):
+    def __init__(self, detail: str):
+        super().__init__(status=403, title="Forbidden", detail=detail)
+
+
+def _visibility_filter(org_id: str, username: str):
+    """Return a SQLAlchemy filter for views visible to the given user.
+
+    Visible views are:
+    - System views (org_id IS NULL)
+    - Org-wide views in the same org
+    - Private views created by the user in the same org
+    """
+    return or_(
+        InventoryView.org_id.is_(None),
+        and_(
+            InventoryView.org_id == org_id,
+            or_(
+                InventoryView.created_by == username,
+                InventoryView.org_wide.is_(True),
+            ),
+        ),
+    )
+
+
+def _get_visible_view(view_id: str, org_id: str, username: str) -> InventoryView:
+    view = InventoryView.query.filter(
+        InventoryView.id == view_id,
+        _visibility_filter(org_id, username),
+    ).one_or_none()
+
+    if view is None:
+        raise ViewNotFoundError()
+
+    return view
+
+
+def get_views_list(org_id: str, username: str, page: int = 1, per_page: int = 50) -> tuple[list[InventoryView], int]:
+    query = InventoryView.query.filter(_visibility_filter(org_id, username)).order_by(InventoryView.modified_on.desc())
+
+    total = query.count()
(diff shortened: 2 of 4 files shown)


lib/views_repository.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def update_view(view_id: str, data: dict, org_id: str, username: str) -> dict:
+    view = InventoryView.query.filter(
+        InventoryView.id == view_id,
+        _visibility_filter(org_id, username),
+    ).one_or_none()
+
+    if view is None:
+        abort(404, "View not found.")
```
rodrigonull: I'd recommend refactoring this to raise domain exceptions (e.g., ViewNotFoundError, ViewPermissionError) and mapping them to HTTP codes in the endpoint handlers. We shoudn't use abort() in the repositories.

lib/views_repository.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def get_view_by_id(view_id: str, org_id: str, username: str) -> dict:
+    view = InventoryView.query.filter(
+        InventoryView.id == view_id,
+        _visibility_filter(org_id, username),
+    ).one_or_none()
+
+    if view is None:
+        abort(404, "View not found.")
```
rodrigonull: This could be extracted to a function and reused in this file:
```
def _get_visible_view_or_404(view_id: str, org_id: str, username: str) -> InventoryView:
    view = InventoryView.query.filter(
        InventoryView.id == view_id,
        _visibility_filter(org_id, username),
    ).one_or_none()
    if view is None:
        abort(404, "View not found.")
    return view
```

lib/views_repository.py
```diff
@@ -0,0 +1,165 @@
...
+    return _serialize_view(cloned, username)
+
+
+def _serialize_view(view: InventoryView, username: str) -> dict:
+    return {
+        "id": str(view.id),
+        "org_id": view.org_id,
+        "name": view.name,
+        "description": view.description,
+        "is_system_view": view.org_id is None,
```
rodrigonull:
```suggestion
        "is_system_view": view.is_system_view,
```

lib/views_repository.py
```diff
@@ -0,0 +1,165 @@
...
+def get_view_by_id(view_id: str, org_id: str, username: str) -> dict:
+    view = InventoryView.query.filter(
+        InventoryView.id == view_id,
+        _visibility_filter(org_id, username),
+    ).one_or_none()
+
+    if view is None:
+        abort(404, "View not found.")
+
+    return _serialize_view(view, username)
```
rodrigonull: Repositories shouldn't return serialized objects. Return InventoryView ORM objects. _serialize_view should be moved to app/serialization.py

lib/views_repository.py
```diff
@@ -0,0 +1,165 @@
...
+from app.logging import get_logger
+from app.models import InventoryView
+from app.models import db
+from app.models.views import MAX_VIEW_NAME_LENGTH
+from lib.db import session_guard
+
+logger = get_logger(__name__)
+
+CLONE_NAME_PREFIX = "Copy of "
+
```
rodrigonull, line 27: Add __all__ listing public functions

lib/views_repository.py
```diff
@@ -0,0 +1,165 @@
...
+from sqlalchemy import and_
+from sqlalchemy import or_
+
+from app.logging import get_logger
+from app.models import InventoryView
+from app.models import db
+from app.models.views import MAX_VIEW_NAME_LENGTH
+from lib.db import session_guard
+
+logger = get_logger(__name__)
```
rodrigonull, line 15: This is currently no used, please add logging for mutations (create, update, delete, clone)

rodrigonull (changes requested):

lib/views_repository.py
adarshdubey-star: Got it

lib/views_repository.py
adarshdubey-star: I will update this

lib/views_repository.py
adarshdubey-star: okay, makes sense

lib/views_repository.py
adarshdubey-star: Good point

app/serialization.py
```diff
@@ -562,6 +562,22 @@ def serialize_staleness_response(staleness):
     }
 
 
+def serialize_view(view, username: str) -> dict:
```
rodrigonull:
```suggestion
def serialize_view(view: InventoryView, username: str) -> dict:
```

tests/test_views_repository.py
```diff
@@ -0,0 +1,337 @@
...
+
+ORG_ID = "test-org-1"
+OTHER_ORG = "test-org-2"
+USERNAME = "testuser"
+OTHER_USER = "otheruser"
+
+VALID_CONFIG = {"columns": [{"key": "display_name", "visible": True}]}
+
+
+def _create_db_view(
```
rodrigonull: The test helpers `_create_db_view` and `_create_system_view` work fine here, but consider moving them to `tests/fixtures/db_fixtures.py` as proper pytest fixtures (like db_create_host). The upcoming endpoint tests will likely need the same setup, and having them as shared fixtures avoids duplication.

lib/views_repository.py
```diff
@@ -0,0 +1,153 @@
...
+
+    with session_guard(db.session, close=False):
+        db.session.add(view)
+
+    db.session.refresh(view)
+    logger.info("Created view %s for org %s by %s", view.id, org_id, username)
+    return view
+
+
+def update_view(view_id: str, data: dict, org_id: str, username: str) -> InventoryView:
```
jpramos123, line 102: Out of curiosity with an edge case:

What happens If I clone and update the name of a view to an existing name? Is that possible?

jpramos123 (dismissed):
Apart from the doubt with the edge case, LGTM.

Make sure all test pipeline is working before merge

lib/views_repository.py
adarshdubey-star: Yes, that's possible. There's no unique constraint on name the architecture doc explicitly states:

`Name uniqueness: Not enforced at the database level (users may have views with the same name). The UI should handle disambiguation.`

So a user can:
Clone a view → "Copy of My View"
Rename it to "My View" → now they have two views with the same name

rodrigonull (approved):
LGTM

PR #4438, opened 2026-06-05
fix(deps): Make Konflux dependency PRs update the lockfile

kruai:
## Jira
None

## What

- Make Konflux lock dependencies updating one
- Explicitly ignore hermetic builds when considering which deps to update

## Why

Konflux PRs haven't been updating the lockfile

## How
- Pin python & uv versions in constraints. Not having this may have led to silent failures while locking dependencies.
- Add `.hermetic_builds/**` to ignored paths.

## Testing
n/a

## PR Guidelines

## PR Guideline checks

## Summary by Sourcery

Build:
- Adjust Renovate configuration so Konflux dependency updates also regenerate the lockfile and exclude hermetic build artifacts from consideration.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,12 +4,23 @@
     "github>RedHatInsights/konflux-pipelines//renovate/foreman_satellite/renovate.json"
   ],
   "ignorePaths": [
-    "Pipfile.iqe"
+    "Pipfile.iqe",
+    ".hermetic_builds/**"
   ],
+  "constraints": {
+    "python": "3.12",
+    "uv": "0.11"
+  },
   "tekton": {
     "schedule": [
       "at any time"
     ]
+  },
+  "pip_requirements": {
+    "enabled": false
+  },
+  "pip-compile": {
+    "enabled": false
   },
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -4,13 +4,24 @@
     "github>RedHatInsights/konflux-pipelines//renovate/foreman_satellite/renovate.json"
   ],
   "ignorePaths": [
-    "Pipfile.iqe"
+    "Pipfile.iqe",
+    ".hermetic_builds/**"
   ],
+  "constraints": {
+    "python": "3.12",
+    "uv": "0.11"
```
thearifismail, line 12: Yesterday, I updated my dev setup to use `uv` and was asked to update my `uv` and now it is set to `uv 0.11.11`

rodrigonull (approved):

renovate.json
kruai: Yep, our project's uv version is `0.11.11`. Mintmaker is using `0.11.19`, so it's likely that it was failing to lock the dependencies due to the mismatch; using a less-specific version constraint (like `0.11`) should fix the issue

thearifismail (approved):

PR #4442, opened 2026-06-08
chore: Fix an issue with local deployment on Macbooks

fstavela:
## What
Preserve container-built .venv in hbi-web and hbi-mq dev services.

## Why
On Mac, the bind mount .:/opt/app-root/src overwrites the container's .venv with the host's macOS-native virtualenv, making Linux-incompatible packages (e.g. connexion) unavailable at runtime.

## How
Added anonymous volume mounts (/opt/app-root/src/.venv) to both services so the container-built virtualenv is preserved instead of being overwritten by the host bind mount.

## Testing
Verified podman compose -f dev.yml up starts hbi-web successfully on Mac without the connexion import error. No impact on Linux — the host .venv there is already Linux-compatible, and the anonymous volume
  simply takes precedence.

## PR Guidelines

## PR Guideline checks

## Summary by Sourcery

Deployment:
- Add anonymous volume mounts for the .venv directory in hbi-web and hbi-mq dev services so container-built virtualenvs are not overridden by host bind mounts on Mac.


kruai (approved):
Someone else with a Macbook should review as well, but I can confirm that it doesn't break on my Fedora setup!

strider (approved):
Yep, we need that on MacOS

PR #4446, opened 2026-06-08
perf(RHINENG-26748): add composite index (org_id, modified_on DESC)

rodrigonull:
## Jira
[RHINENG-26748](https://issues.redhat.com/browse/RHINENG-26748)

## What
Add a composite B-tree index `(org_id, modified_on DESC)` on the `hbi.hosts` partitioned table.

## Why
The host list endpoint queries `WHERE org_id = ? ORDER BY modified_on DESC LIMIT ? OFFSET ?`. Without a composite index, PostgreSQL may choose the standalone `modified_on_id_idx` to satisfy the ORDER BY, then scan the entire partition filtering by org_id — causing 20-30s query timeouts in production.

## How
- New Alembic migration using the `create_partitioned_table_index` helper, which in production creates indexes `CONCURRENTLY` on each partition before attaching to the parent table (zero downtime).
- Index name: `idx_hosts_org_id_modified_on`
- Downgrade drops the index from parent and all partitions.

## Testing
- All 372 `test_api_hosts_get.py` tests and 6 migration tests pass.
- Linting (ruff, mypy) and pre-commit hooks pass cleanly.

## PR Guidelines

## PR Guideline checks

[RHINENG-26748]: https://redhat.atlassian.net/browse/RHINENG-26748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Add a database migration to introduce a composite index on the hosts table to optimize org-scoped, modified-on-ordered queries.

Enhancements:
- Add an Alembic migration that creates a composite B-tree index on the partitioned hosts table for (org_id, modified_on DESC) to improve host listing query performance.
- Provide a corresponding downgrade path that removes the composite index from the parent hosts table and its partitions.


kruai (approved):
LGTM. Just to make sure: because this is using a b-tree index, it should be used for `modified_on ASC` as well, right?

rodrigonull:
> LGTM. Just to make sure: because this is using a b-tree index, it should be used for `modified_on ASC` as well, right?

That is correct, this new index should handle both ORDER BY modified_on DESC and ORDER BY modified_on ASC efficiently because B-tree indexes support backward scans.

fstavela (approved):

PR #4449, opened 2026-06-08
perf(RHINENG-26879): rewrite workloads filter to use @> containment

rodrigonull:
## Jira
[RHINENG-26879](https://issues.redhat.com/browse/RHINENG-26879)

## What
Rewrite workloads filters to use the `@>` JSONB containment operator for both boolean equality and object-type presence checks, and add a GIN index to accelerate them.

## Why
The previous filters generated `CAST(workloads #>> '{sap,sap_system}' AS BOOLEAN) IS true` for booleans and `workloads #>> '{ansible}' IS NOT NULL` for presence checks. Neither is GIN-indexable, forcing PostgreSQL into full sequential scans on `system_profiles_dynamic` and causing 28s+ query timeouts when filtering hosts by workload presence (e.g., SAP systems, Ansible-managed hosts).

## How
- **Boolean filter rewrite** (`api/filtering/db_custom_filters.py`): Boolean workload filters now generate `workloads @> '{"sap": {"sap_system": true}}'::jsonb` using SQLAlchemy's `type_coerce` with the `@>` operator. The `not_nil` existence check for booleans uses an OR of two containment checks (`@> true` OR `@> false`).
- **Object-type presence filter rewrite**: `not_nil`/`nil` on object-type workloads (ansible, mssql, sap, rhel_ai, ibm_db2, crowdstrike, intersystems, oracle_db) now generate `workloads @> '{"ansible": {}}'::jsonb` instead of `#>> IS NOT NULL`. The `nil` case handles NULL columns correctly with `workloads IS NULL OR NOT (workloads @>...)`.
- **Bug fix**: Fixed a pre-existing bug where `nil`/`not_nil` operators (e.g., `sap_sids[not_nil]`) were incorrectly overridden by the default comparator when the value was an empty string.
- **GIN index** (new Alembic migration): Adds `idx_sp_dynamic_workloads_gin` using `jsonb_path_ops` on `system_profiles_dynamic.workloads`. The `jsonb_path_ops` variant is ~60% smaller than the default operator class with less write amplification, and supports the `@>` containment operator.
- Non-boolean, non-object filters (strings, integers, arrays) continue to use the existing `#>>` path.

## Testing
- All 2773 tests pass (full suite including workload/SAP/system-profile tests).
- Verified all 8 object-type workload filters (`sap`, `ansible`, `mssql`, `rhel_ai`, `ibm_db2`, `crowdstrike`, `intersystems`, `oracle_db`) generate `@>` via OpenTelemetry traces in local Docker Compose + Tempo.
- Linting (ruff, ruff format, mypy) and pre-commit hooks pass cleanly.

## PR Guidelines

## PR Guideline checks

[RHINENG-26879]: https://redhat.atlassian.net/browse/RHINENG-26879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Optimize workloads JSONB filtering to use containment queries and support indexing for better host query performance.

Enhancements:
- Rewrite workloads boolean filters to use JSONB @> containment instead of path extraction for GIN indexability and performance.
- Add specialized handling for sap_system filters, including not_nil checks using JSONB containment.
- Introduce helpers to detect boolean workload filters and construct the appropriate containment documents.
- Add a partitioned GIN index on system_profiles_dynamic.workloads using jsonb_path_ops to accelerate containment queries.


api/filtering/db_custom_filters.py
```diff
@@ -337,29 +338,161 @@ def _validate_pg_op_and_value(pg_op: str | None, value: str | bool, field_filter
...
+
+def _build_workloads_not_nil_boolean_filter(column: Column, jsonb_path: tuple[str, ...]) -> ColumnElement:
+    """Build a 'field is not nil' filter for boolean fields.
+
+    Uses: column @> {...: true} OR column @> {...: false}
+    This tests that the field has ANY boolean value (i.e., it exists and is not null).
+    """
+    return _build_workloads_containment_filter(column, jsonb_path, True) | _build_workloads_containment_filter(
+        column, jsonb_path, False
+    )
```
kruai, line 367: Makes sense... perf-wise, is OR-ing these 2 together slower than using one "IS NOT NULL"?

kruai (approved):
One minor Q, but LGTM!

api/filtering/db_custom_filters.py
rodrigonull: No, the OR should be faster. IS NOT NULL would require (workloads #>> '{sap,sap_system}') IS NOT NULL, which uses path extraction (#>>) and can't use the GIN index.

PR #4467, opened 2026-06-11
fix: Clean host reaper's DB session memory between batches

fstavela:
## What
Fix host reaper job getting OOM killed when it runs for too long

## Why
The host reaper jobs are getting OOM killed when they process too many hosts. For example, now they are processing a bit number of hosts because a staleness config change has been made on the biggest account in Prod.

## How
Clean host reaper's DB session memory between batches via `session.expunge_all()`

## Testing
Unit tests are passing, IQE test will run in PR check

## PR Guidelines

## PR Guideline checks

## Summary by Sourcery

Clean up database session state after each host reaper batch to avoid excessive memory growth during long-running jobs.


jobs/host_reaper.py
```diff
@@ -104,6 +104,7 @@ def run(config, logger, session, event_producer, notification_event_producer, sh
                 hosts_processed = 0
```
rodrigonull, line 105: I think the actual memory growth comes from list(events) materializing every OperationResult (each holding a full Host ORM object) into a Python list. For a large account this can be hundreds of thousands of objects in memory at once.

Consider consuming the generator without keeping references, so each result is GC'd immediately after counting:

```suggestion
try:
    hosts_processed = sum(
        1
        for _ in delete_hosts(
            query,
            event_producer,
            notification_event_producer,
            config.host_delete_chunk_size,
            shutdown_handler.shut_down,
            control_rule="REAPER",
        )
    )
except InterruptedError:
    hosts_processed = 0
```

rodrigonull (commented):
Thanks for the PR. I added a suggestion

jobs/host_reaper.py
fstavela: @rodrigonull The `len(list(events))` happens only for the current batch, so even if the account has millions of hosts, this list should always contain only `config.host_delete_chunk_size` host objects at most.

However, your suggestion makes it even more efficient, so I pushed the change. Thank you for the suggestion!

rodrigonull (approved):
LGTM

kruai (approved):

PR #4478, opened 2026-06-14
RHINENG-26236: Design doc for IQE API cleanup 

thearifismail:
## Jira
[RHINENG-26236](https://issues.redhat.com/browse/RHINENG-26236)
Covered by [RHINENG-27036](https://issues.redhat.com/browse/RHINENG-27036)

## What
Design document for migrating the IQE test plugin from auto-generated OpenAPI bindings (apigen) to direct Python requests

## Why
Replacing the auto-generated OpenAPI bindings in IQE test plugin with direct Python requests for simpler maintenance and easier V2 API adoption

## How
The migration keeps the existing wrapper layer (GroupsAPIWrapper, HostsAPIWrapper, etc.) but replaces their apigen internals with direct requests.Session HTTP calls. This will be done incrementally, starting with the groups/workspaces wrapper as a proof of concept.

## Testing
Too early. The design doc itself is up for review — the testing approach should come out of that discussion, not be decided upfront.

## PR Guidelines

## PR Guideline checks

[RHINENG-26236]: https://redhat.atlassian.net/browse/RHINENG-26236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Add a design document outlining the planned migration of the IQE HBI API client from autogenerated OpenAPI bindings to direct HTTP requests using requests.Session.

Documentation:
- Document the problems with the current apigen-based IQE API client, including regeneration overhead, abstraction complexity, and tight coupling to the OpenAPI spec.
- Describe the target architecture using a wrapper layer over requests.Session instead of autogenerated bindings, with illustrative before/after examples for group APIs.
- Define the migration scope, explicitly listing in-scope wrappers, out-of-scope components, and a sequenced set of deliverable PRs for the epic.

[RHINENG-27036]: https://redhat.atlassian.net/browse/RHINENG-27036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/design-docs/RHINENG-26236-epic.md
```diff
@@ -0,0 +1,76 @@
...
+```python
+# groups_api.py wrapper
+def get_groups_response(self, *, name=None, per_page=None, page=None,
+                        order_by=None, order_how=None):
+    params = {k: v for k, v in {
+        "name": name, "per_page": per_page, "page": page,
+        "order_by": order_by, "order_how": order_how,
+    }.items() if v is not None}
+
+    response = self.session.get("/api/inventory/v1/groups", params=params)
```
fstavela: I think there should be some helper for building the URLs. Either a function that will build the URL, or maybe even better would be having some method like `self.api_request` or `self.get`, which would build the URL for us and we would pass just `"/groups"`, instead of the whole path.

docs/design-docs/RHINENG-26236-epic.md
```diff
@@ -0,0 +1,76 @@
...
+    response.raise_for_status()
+    return response.json()
+```
+
+### Scope
+
+**In scope (this epic):**
+- New base HTTP client/session wrapper using `requests.Session` with auth headers, retries, and base URL config
```
fstavela: I think we shouldn't create a new `requests.Session` object, but we can reuse IQE's [app.http_client](https://gitlab.cee.redhat.com/insights-qe/iqe-core/-/blob/main/iqe/base/application/__init__.py?ref_type=heads#L400). It returns a [RobustSession](https://gitlab.cee.redhat.com/insights-qe/iqe-core/-/blob/main/iqe/base/http.py?ref_type=heads#L25) object, which is a child of `requests.Session`, so it can be used in the same way. The advantage is that it will take care of all the authentication and config for us.

docs/design-docs/RHINENG-26236-epic.md
```diff
@@ -0,0 +1,76 @@
...
+- Migrate wrapper classes directly impacted by V2 work (groups/workspaces, hosts, host-views) as proof of concept
+- Return plain dicts from the wrappers instead of Pydantic model objects (tests assert on dict keys)
```
fstavela: For PoC, do you think it would be better to migrate just the new endpoints (host-views, and anything that will be in the `/api/inventory/v2/...` path), but keep all of the old endpoints (`/api/inventory/v1/...`) as-is (until full migration)? One of the reasons for this is that when we migrate already existing endpoints, we will have to take care of other IQE plugins as well, because they are currently using our wrappers which are returning auto-generated objects, so if we migrate to returning dicts, the tests of other teams would break. I think the work on other IQE plugins increases the scope of the PoC, so that's why I'm asking.

PR #4481, opened 2026-06-15
docs(RHINENG-27036): ADR migrate off of IQE generated bindings

thearifismail:
## Jira
[RHINENG-27036](https://redhat.atlassian.net/browse/RHINENG-27036)
[RHINENG-27113](https://redhat.atlassian.net/browse/RHINENG-27113)

## What
Document IQE API Client Migration: Apigen to app.http_client.
It also adds a `BaseAPIWrapper` and a `host_checkin_response`, and `test_host_checkin` test_suite

## Why
IQE relies on automatically generated API bindings to connect to the API layer.  This JIRA card documents how to migrate the IQE API client from Apigen to app_client which uses Python request.session for direct connection.
## How
Provide a new BaseAPIWrapper, which can accept the API version to generate a proper URL. The default is `"v2"` for the API version.

## PR Guidelines

## PR Guideline checks

[RHINENG-27036]: https://redhat.atlassian.net/browse/RHINENG-27036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Introduce a base HTTP API wrapper to enable direct IQE client calls to the inventory API and add coverage for the host check-in endpoint.

New Features:
- Add a BaseAPIWrapper class for versioned, direct HTTP access to inventory API endpoints without using apigen.
- Expose a host_checkin_response method on HostsAPIWrapper for calling POST /hosts/checkin via the new HTTP wrapper.

Enhancements:
- Configure clowder_smoke host-inventory main service connection settings for use with the new HTTP client wrapper.

Documentation:
- Add documentation describing the migration strategy from apigen-generated IQE API clients to the new app.http_client-based BaseAPIWrapper.

Tests:
- Add REST tests verifying POST /hosts/checkin behavior, including staleness timestamp updates, custom checkin_frequency handling, and validation of required canonical facts.

[RHINENG-27113]: https://redhat.atlassian.net/browse/RHINENG-27113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/iqe-api-client-migration-doc.md
```diff
@@ -0,0 +1,99 @@
...
+    )
+```
+
+**After (using `app.http_client`):**
+```python
+# base wrapper — URL helper builds full path from versioned base
+class BaseAPIWrapper:
+    def __init__(self, app, api_version="v2"):
+        self._app = app
+        self._base_path = f"/api/inventory/{api_version}"
```
fstavela: This will need to include protocol, hostname, and port (optional) as well. All these things can be found in the IQE plugin config and accessed through app.host_inventory.config

docs/iqe-api-client-migration-doc.md
```diff
@@ -0,0 +1,99 @@
...
+
+    @property
+    def client(self):
+        return self._app.http_client
+
+    def get(self, path, **kwargs):
+        return self.client.get(f"{self._base_path}{path}", **kwargs)
+
+# groups_api.py wrapper
+def get_groups_response(self, *, name=None, per_page=None, page=None,
```
fstavela: groups -> workspaces

fstavela (commented):
Hey, I left two comments, otherwise LGTM

docs/iqe-api-client-migration-doc.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+### Scope
+
+**In scope (this epic):**
+- Base wrapper using IQE's `app.http_client` (`RobustSession`) with a URL helper for versioned path construction
+- Migrate **only new V2 endpoints** (host-views, and any new endpoints under `/api/inventory/v2/...`) as proof of concept
+- Return plain dicts from the new V2 wrappers (tests assert on dict keys)
+
+**Out of scope:**
```
ezr-ondrej: I believe this should be Phase 2, rather then out of scope. We'll need to do it as well, just in second phase

docs/iqe-api-client-migration-doc.md
```diff
@@ -0,0 +1,99 @@
...
+- Migrating existing V1 endpoints — other IQE plugins depend on the current auto-generated model objects, so migrating them would require coordinating changes across plugins
```
ezr-ondrej: This is combining the thigs:
1. migrating V1 endpints in our codebase, which is what we want to target and should be relatively straightforward
2. migrating other IQE plugins depending on our bindings, which is indeed a phase in its own as it will require bit more work

ezr-ondrej (commented):
I like the writup of the problem and I'm fine with scoping this narrowly for PoC, but I thin we should describe all the phases. Until we've removed the bindings from our code. It doesn't need to be scope for your first deliverable, but I think this document should describe the full migration (can be just high level steps split into phases with clear outcome of each phase, I don't think it needs much more detail then you currently have).

docs/iqe-api-client-migration-doc.md
```diff
@@ -0,0 +1,108 @@
...
+
+**Phase 2: Migrate V1 endpoints in this plugin**
+
+Migrate the existing V1 wrapper classes (`hosts_api.py`, `groups_api.py`, tags, system_profile, staleness, etc.) to use `BaseAPIWrapper` with `api_version="v1"`. The base class is ready for this today. This is relatively straightforward because it is contained within this plugin's codebase.
+
+Outcome: all HBI IQE wrappers use `app.http_client` directly; the generated apigen bindings are no longer called from any wrapper.
+
+**Phase 3: Remove the generated bindings and coordinate with other plugins**
+
+Once all wrappers in this plugin are migrated, other IQE plugins that depend on our auto-generated model objects (`iqe_host_inventory_api/`, `iqe_host_inventory_api_v7/`) need to be updated. This requires cross-team coordination before the generated packages can be removed.
```
fstavela: Some plugins don't use just the auto-generated objects directly, but they also use our API wrappers from `hosts_api.py`, `groups_api.py`, `kafka_interactions.py` etc., so I think that Phase 2 and Phase 3 should be in a reverse order. We first have to make sure that other plugins don't use our internal wrappers (they can migrate to iqe-bindings, for example), and only then we can change the return type on the wrappers from auto-generated objects to dicts. Or, we need to work with other teams to update their code to expect the dicts as return types and keep using our wrappers, but then we need to coordinate merging all of the PRs, which seems to be too complicated.

fstavela (dismissed):

docs/iqe-api-client-migration-doc.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+### Phase 1: Proof of Concept — Migrate a small set of V1 endpoints
+
+Establish the pattern using existing, live V1 endpoints. This is immediately testable and unblocks the migration without waiting for V2 to be deployed.
+
+**Deliverables:**
+
+1. **PR 1** — Introduce `BaseAPIWrapper` using `app.http_client` with the URL helper. Migrate a single V1 endpoint (e.g., `GET /groups`) as proof of concept, with one test passing end-to-end. The apigen packages remain installed; only this one wrapper changes internally.
+2. **PR 2** — Migrate the test suites directly impacted by the V2 work to the new pattern.
```
fstavela: This will need to go in in 1 PR. If we change the wrapper for an endpoint, it would break the tests, so we need to fix them in the same PR.

docs/iqe-api-client-migration-doc.md
```diff
@@ -0,0 +1,119 @@
...
+    }.items() if v is not None}
+
+    response = self.get("/groups", params=params)
+    response.raise_for_status()
+    return response.json()
+```
+
+### Phase 1: Proof of Concept — Migrate a small set of V1 endpoints
+
+Establish the pattern using existing, live V1 endpoints. This is immediately testable and unblocks the migration without waiting for V2 to be deployed.
```
fstavela: If we are going to migrate a V1 endpoint in the first phase, we need to make sure that no other plugin uses our wrappers for this endpoint, otherwise the changed return type would break their tests.

docs/iqe-api-client-migration-doc.md
```diff
@@ -0,0 +1,144 @@
...
+
+- `POST /hosts/checkin` exists in the V1 spec (operationId: `api.host.host_checkin`). It accepts canonical facts (`insights_id`, `fqdn`, `subscription_manager_id`, etc.) plus an optional `checkin_frequency` (minutes), and returns a `HostOut`.
+- `HostsAPIWrapper` has **no** `host_checkin` method today — there are no existing callers to break.
+- It is a host write operation in the same domain as Create Hosts, demonstrating `BaseAPIWrapper` on a V1 write path without any cross-team risk.
+- Adding a new `BaseAPIWrapper`-backed method introduces zero risk to downstream services or other IQE plugins.
+
+**Out of scope for this POC:**
+
+- Migrating any existing V1 endpoints (GET, PATCH, DELETE /hosts, etc.) — other IQE plugins depend on the apigen model objects returned by those wrappers
+- V2 endpoint wrappers (deferred; original design doc covers these — see `RHINENG-26236-epic.v1.md`)
```
fstavela: I don't see this file

docs/iqe-api-client-migration-doc.md
```diff
@@ -0,0 +1,144 @@
...
+- Removing `iqe_host_inventory_api/` and `iqe_host_inventory_api_v7/` packages
+
+
+
+### Future: Broader V1 Migration
+
+`BaseAPIWrapper` is version-aware — it accepts `api_version` as a parameter, so both V1 and V2 wrappers can use it. The base class is ready for broader use today.
+
+**V2 endpoints:** Can be added as new `BaseAPIWrapper`-backed methods immediately (see original design doc `RHINENG-26236-epic.v1.md`).
```
fstavela: Same here

docs/iqe-api-client-migration-doc.md
```diff
@@ -0,0 +1,144 @@
...
+- **Apply this new method only in Create Hosts / checkin tests**
+- Return plain dicts from the new method (tests assert on dict keys)
+
+**Why "Create Hosts" tests cannot use `BaseAPIWrapper`:**
+
+The initial POC plan was to target Create Hosts tests because adding a new host-creation wrapper would not affect other services. However, HBI does **not** expose a `POST /hosts` REST endpoint in V1 — the V1 spec only defines `GET` and `DELETE` on `/hosts`. Hosts are created exclusively via:
+- **Kafka ingestion pipeline** — `kafka_interaction.py` sends host messages to the ingress topic
+- **Ingress upload** — `uploads.py` uploads insights archives via `IngressApi`
+
+Neither path goes through a REST endpoint that `BaseAPIWrapper` could wrap. There is nothing to migrate.
```
fstavela: I don't think this section, or any other mention of "Create hosts" needs to be in this document. It's true that we don't support creating hosts via API and that there's nothing to migrate related to this, but then I don't think it needs to even be mentioned in this migration document.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/base_api_wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    def __init__(self, app: Application, api_version: str = "v1") -> None:
+        self._app = app
+        # Derive the base URL from the existing apigen client configuration.
+        # configuration.host is the full versioned base path, e.g.:
+        #   "https://console.redhat.com/api/inventory/v1"
+        # Strip the trailing version component and replace with the desired version.
+        apigen_host: str = app.host_inventory.rest_client.client.configuration.host
+        base = apigen_host.rsplit("/", 1)[0]  # -> "https://host/api/inventory"
```
fstavela: We shouldn't do this. We want to get rid of the apigen client, so we can't use the generated client's config in order to configure a different client that will replace the apigen client.

The IQE config already has everything we need (scheme, hostname, port) per-environment: https://gitlab.cee.redhat.com/insights-qe/iqe-core/-/blob/main/iqe/conf/settings.default.yaml?ref_type=heads#L540-546

It can be accessed via `app.config.main.hostname`, etc. We can build the base_url from that. The only problem will be in ephemeral (`clowder_smoke`), where we make the API requests directly against the services, not a central gateway, so this needs to be configured per-service. The apigen client in ephemeral has the config here: https://github.com/RedHatInsights/insights-host-inventory/blob/master/iqe-host-inventory-plugin/iqe_host_inventory/conf/host_inventory.default.yaml#L723-L726

This config is for the apigen client, so I think we should copy the same config under `main`, to keep it consistent with the stage/prod config. It can be then accessed through `app.host_inventory.config.main.hostname` etc.. So in the `__init__`, we will have to check the current environment via `app.config.current_env`, and use `app.host_inventory.config.main` if it's `clowder_smoke`, or `app.config.main` otherwise.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/hosts_api.py
```diff
@@ -140,6 +141,53 @@ def raw_api(self) -> HostsApi:
...
+    def _base_wrapper(self) -> BaseAPIWrapper:
+        return BaseAPIWrapper(self.application)
+
+    def host_checkin_response(
+        self,
+        *,
+        insights_id: str | None = None,
+        fqdn: str | None = None,
+        subscription_manager_id: str | None = None,
+        checkin_frequency: int | None = None,
```
fstavela, line 180: Are these all of the possible parameters? This method should have all of the available parameters of the API endpoint.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/hosts_api.py
```diff
@@ -140,6 +141,53 @@ def raw_api(self) -> HostsApi:
     def api_client(self) -> ApiClient:
         return self.application.host_inventory.rest_client.client
 
+    @cached_property
+    def _base_wrapper(self) -> BaseAPIWrapper:
+        return BaseAPIWrapper(self.application)
```
fstavela, line 166: I would move this to the top of the class

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/test_host_checkin.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    host = host_inventory.kafka.create_host()
+
+    response = host_inventory.apis.hosts.host_checkin_response(
+        insights_id=host.insights_id,
+        checkin_frequency=60,
+    )
+
+    assert response["id"] == host.id
+    assert response["insights_id"] == host.insights_id
```
fstavela: I don't think that the test names or descriptions should mention BaseAPIWrapper, that is an implementation detail, the important thing is what does the test test. It tests the `POST /hosts/checkin` endpoint, not the `BaseAPIWrapper`, that is just a tool that is used for testing.

Also, I would add a check that the `last_seen` and staleness timestamps have been updated after we've used the endpoint, because that's the main point why the endpoint exists and it should be tested.

fstavela (commented):
I think the direction of this looks good, but I left a few technical comments. The most important thing from this review is about how we derive the `base_url` in the `BaseAPIWrapper`.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/hosts_api.py
fstavela: Ignore this comment, I just noticed that it already is at the top.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/base_api_wrapper.py
```diff
@@ -0,0 +1,66 @@
...
+        self._app = app
+        # Build the base URL from IQE config rather than the apigen client.
+        # In clowder_smoke the gateway is the direct service, so we use the
+        # per-plugin main config; everywhere else the central gateway config
+        # lives in app.config.main (scheme/hostname/port per environment).
+        if app.config.current_env == "clowder_smoke":
+            cfg = app.host_inventory.config.main
+        else:
+            cfg = app.config.main
+        self._base_url = f"{cfg.scheme}://{cfg.hostname}:{cfg.port}/api/inventory/{api_version}"
```
fstavela: This won't work in Stage and Prod, because we don't use port there. The port should be optional and used only in `clowder_smoke`.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/hosts_api.py
```diff
@@ -140,6 +141,81 @@ def raw_api(self) -> HostsApi:
...
+            "ip_addresses",
+            "fqdn",
+            "mac_addresses",
+            "provider_id",
+            "provider_type",
+        )
+        if not any(k in body for k in canonical_facts):
+            raise ValueError(
+                "At least one canonical fact must be provided: " + ", ".join(canonical_facts)
+            )
```
fstavela: This verification shouldn't be here. The restriction is on the app side. The tests should be able to execute the API request even with invalid input data, and then we can check the response from the app.

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/test_host_checkin.py
```diff
@@ -0,0 +1,83 @@
...
+    """POST /hosts/checkin updates the host's last_check_in and staleness timestamps.
+
+    metadata:
+        requirements: inv-host-checkin
+        assignee: aarif
+        importance: medium
+        title: POST /hosts/checkin updates last_check_in and staleness timestamps
+    """
+    host = host_inventory.kafka.create_host()
+    sleep(1)  # Ensure timestamps differ after checkin
```
fstavela: The sleep shouldn't be needed. The timestamps include microseconds, so the timestamps will differ even without the sleep.

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/test_host_checkin.py
```diff
@@ -0,0 +1,83 @@
...
+    """POST /hosts/checkin with an explicit checkin_frequency updates staleness timestamps.
+
+    metadata:
+        requirements: inv-host-checkin
+        assignee: aarif
+        importance: medium
+        title: POST /hosts/checkin with custom checkin_frequency updates staleness timestamps
+    """
+    host = host_inventory.kafka.create_host()
+    sleep(1)  # Ensure timestamps differ after checkin
```
fstavela: Same here

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/test_host_checkin.py
```diff
@@ -0,0 +1,83 @@
...
+def test_host_checkin_requires_canonical_fact(host_inventory: ApplicationHostInventory) -> None:
+    """POST /hosts/checkin requires at least one canonical fact.
+
+    metadata:
+        requirements: inv-host-checkin
+        assignee: aarif
+        importance: low
+        title: POST /hosts/checkin rejects requests with no canonical facts
+    """
+    with pytest.raises(ValueError, match="At least one canonical fact must be provided"):
```
fstavela: This will probably need to catch a different error once we remove the canonical facts restriction from the IQE's `host_checkin_response` method.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/base_api_wrapper.py
```diff
@@ -0,0 +1,66 @@
...
+        return self.client.post(f"{self._base_url}{path}", **kwargs)
+
+    def patch(self, path: str, **kwargs: Any) -> Any:
+        return self.client.patch(f"{self._base_url}{path}", **kwargs)
+
+    def put(self, path: str, **kwargs: Any) -> Any:
+        return self.client.put(f"{self._base_url}{path}", **kwargs)
+
+    def delete(self, path: str, **kwargs: Any) -> Any:
+        return self.client.delete(f"{self._base_url}{path}", **kwargs)
```
fstavela: I would add a log about the API request that we are making, something similar to this: https://gitlab.cee.redhat.com/insights-qe/iqe-core/-/blob/main/iqe/base/rest_client.py?ref_type=heads#L181-184

fstavela (changes requested):

iqe-host-inventory-plugin/iqe_host_inventory/modeling/hosts_api.py
```diff
@@ -159,6 +160,64 @@ def raw_api(self) -> HostsApi:
...
+        subscription_manager_id: str | None = None,
+        satellite_id: str | None = None,
+        bios_uuid: str | None = None,
+        ip_addresses: list[str] | None = None,
+        fqdn: str | None = None,
+        mac_addresses: list[str] | None = None,
+        provider_id: str | None = None,
+        provider_type: str | None = None,
+        checkin_frequency: int | None = None,
+    ) -> Any:
```
fstavela: I think we should use a better type annotation here, the `RobustSession` (`requests.Session`) surely returns some kind of Response object.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/base_api_wrapper.py
```diff
@@ -0,0 +1,68 @@
...
```
fstavela: Same in these, we can use a more strict type annotation.

fstavela (commented):
There is one outstanding comment about logging + I'm adding a couple about type annotations. Other than that, looks good to me now!

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/test_host_checkin.py
```diff
@@ -0,0 +1,82 @@
...
+    response = host_inventory.apis.hosts.host_checkin_response(
+        insights_id=host.insights_id,
+        checkin_frequency=60,
+    )
+
+    assert response.status_code == 201
+    assert_checkin_response(response.json(), host)
+
+
+@pytest.mark.ephemeral
```
fstavela: This test doesn't use Kafka, so it doesn't need the `ephemeral` marker and it can run against Stage and Prod as well.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/hosts_api.py
thearifismail: Now, it returns `requests.Response`

iqe-host-inventory-plugin/iqe_host_inventory/modeling/base_api_wrapper.py
thearifismail: Now `request.Response` is returned.
```    def get(self, path: str, **kwargs: Any) -> requests.Response:
        return self._request("get", path, **kwargs)

    def post(self, path: str, **kwargs: Any) -> requests.Response:
        return self._request("post", path, **kwargs)

    def patch(self, path: str, **kwargs: Any) -> requests.Response:
        return self._request("patch", path, **kwargs)

    def put(self, path: str, **kwargs: Any) -> requests.Response:
        return self._request("put", path, **kwargs)

    def delete(self, path: str, **kwargs: Any) -> requests.Response:
        return self._request("delete", path, **kwargs)
```

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/test_host_checkin.py
thearifismail: Done! Removed @pytest.mark.ephemeral and successfully ran this test against the Stage and Prod clusters

fstavela (commented):
The code LGTM, but please revert the changes to `pyproject.toml` and `uv.lock`, I don't think we need `python-docx` as a dependency. Or do we?

fstavela (dismissed):
LGTM, thank you!

docs/iqe-api-client-migration-doc.md
```diff
@@ -0,0 +1,172 @@
+## Epic: IQE API Client Migration — Apigen to Direct Requests
```
ezr-ondrej: This currently describes only what this PR does.
Is the plan to extend this to an ADR once we implement the rest?

Currently it has a lot of implementation details on top of what I'd expect in Architectural decision record. It is probably not very interesting as implementation is already easily readable in the code itself.

ezr-ondrej (commented):
I'm good with the code, but the document just repeats the code, which IMO doesn't add too much value.

The Problem definition is great, the Architecture is great, the rest of the document seems redundant and could be just code or a PR description instead.

ezr-ondrej (approved):

PR #4485, opened 2026-06-16
fix(RHINENG-27044): add total advisories sort fields

rodrigonull:
## Jira
[RHINENG-27044](https://issues.redhat.com/browse/RHINENG-27044)

## What
Add new computed sort fields `patch:advisories_total_installable` and `patch:advisories_total_applicable` to the `/host-views` endpoint that sort hosts by the sum of all advisory types.

## Why
When sorting by "Installable advisories" ASC, hosts with 0 RHSA but non-zero RHBA/other advisories appeared among "No installable advisories" hosts. This happened because the endpoint sorted by a single advisory type (e.g. `rhsa_installable`) while the UI displays a combined total. The Patch API sorts by a group of advisory counts together, and our sorting should match.

## How
- Added `__computed_sortable_fields__` support to the app data sorting infrastructure, allowing models to define virtual sort fields backed by SQL expressions rather than plain columns.
- Defined two computed expressions on `HostAppDataPatch` that return `COALESCE(rhsa, 0) + COALESCE(rhba, 0) + COALESCE(rhea, 0) + COALESCE(other, 0)`, wrapped in a `CASE` to preserve NULL (NULLS LAST) when no patch data row exists.
- Updated `resolve_app_sort()` to detect computed fields and invoke the classmethod to produce the expression.
- Added the new fields to the OpenAPI spec enum and documentation.

## Testing
- Unit tests: verify `resolve_app_sort` returns the correct model and a valid SQL expression containing all four advisory columns.
- Integration tests: verify correct ASC/DESC ordering by total sum, including the exact bug scenario (host with `rhsa=0, rhba=1, other=1` sorts after `total=0` hosts), and NULLS LAST behavior for hosts without patch data.
- All 355 tests pass, `make style` clean.

## PR Guidelines

## PR Guideline checks

[RHINENG-27044]: https://redhat.atlassian.net/browse/RHINENG-27044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Add support for computed sort fields in host app data and expose new total patch advisory sorting options in the host-views API.

New Features:
- Expose patch:advisories_total_installable and patch:advisories_total_applicable as sortable fields on the /host-views endpoint to sort by total advisories across all types.

Enhancements:
- Extend the app data sorting infrastructure to support model-defined computed sortable fields backed by SQL expressions.
- Implement computed total installable and applicable advisory expressions on HostAppDataPatch and wire them into resolve_app_sort and the dynamic field map.

Documentation:
- Document the new total advisories patch sort fields in the OpenAPI specification.

Tests:
- Add unit tests for resolving the new computed patch advisory sort fields to the correct model and SQL expression.
- Add integration tests verifying correct ordering and NULLS LAST behavior when sorting hosts by total installable advisories.


tests/test_api_host_views.py
```diff
@@ -1365,6 +1365,215 @@ def test_sort_by_patch_advisories_rhsa_installable(self, api_get, db_create_host
...
+
+        url = build_host_view_url(query="?order_by=patch:advisories_total_applicable&order_how=DESC")
+        response_status, response_data = api_get(url)
+
+        assert_response_status(response_status, 200)
+        results = response_data["results"]
+        assert results[0]["display_name"] == "host1.example.com"
+        assert results[1]["display_name"] == "host2.example.com"
+
+    def test_sort_by_patch_advisories_total_applicable_nulls_last(
```
fstavela, line 1459: We could parametrize this test to test both asc and desc

fstavela (commented):
Thank you for the PR, I left just one suggestion

tests/test_api_host_views.py
fstavela: @rodrigonull Thanks for parametrizing the tests, but I meant the `nulls_last` tests. Those should be easier to parametrize, because I think the host with no data should be always last, regardless of `order_how`, right?

app/models/host_app_data.py
```diff
@@ -197,6 +207,42 @@ class HostAppDataPatch(HostAppDataMixin, db.Model):
...
+        )
+
+    @classmethod
+    def _advisories_total_applicable_expr(cls) -> ColumnElement:
+        """Sum of all applicable advisory types for sorting.
+
+        Returns NULL when no patch data row exists (LEFT JOIN produced no match),
+        so that nullslast() correctly places those hosts at the end.
+        """
+        return case(
```
adarshdubey-star, line 235: Just a suggestion this feels easier on eyes something like: 
```
@classmethod
def _advisories_total_applicable_expr(cls) -> ColumnElement:
    total = (
        func.coalesce(cls.advisories_rhsa_applicable, 0)
        + func.coalesce(cls.advisories_rhba_applicable, 0)
        + func.coalesce(cls.advisories_rhea_applicable, 0)
        + func.coalesce(cls.advisories_other_applicable, 0)
    )

    return case(
        (cls.host_id.isnot(None), total),
        else_=None,
    )
```

tests/test_api_host_views.py
rodrigonull: My bad, NULLs are indeed always last regardless of sort direction. I updated the tests to add the parametrization.

fstavela (approved):

adarshdubey-star (approved):

app/models/host_app_data.py
adarshdubey-star: Feel free to ignore and merge : )

PR #4535, opened 2026-06-30
chore(test): Use iqe-bindings for export-service

mshriver:
## Jira
[IQE-4098](https://redhat.atlassian.net/browse/IQE-4098)

## What
Move to iqe-bindings for export-service API consumption.

Re-opened previous branch under new PR, couldn't open one that was closed for staleness. We were waiting on the ingress/export fixes to go to prod. 

## Why
Far better than relying on the plugin as a dependency.

## How
Using service object definition, same as other iqe-bindings consumed clients

## Testing
testing stage/prod from local env on this commit

## PR Guidelines

## PR Guideline checks

[IQE-4098]: https://redhat.atlassian.net/browse/IQE-4098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Switch export-service integration to use iqe-bindings REST client and wire it into the host-inventory application configuration.

New Features:
- Introduce a v7_export_v1 REST client for export-service in the host-inventory application configuration.

Enhancements:
- Refine export download handling to work with raw zip bytes via temporary files and tighten completion wait defaults.
- Remove legacy iqe_export_service dependency and associated mypy ignore configuration.
- Update pre-commit configuration to use the ruff-check hook name.


iqe-host-inventory-plugin/iqe_host_inventory/modeling/exports_api.py
```diff
@@ -209,16 +215,13 @@ def create_export(
...
-
         export_id = self.raw_api.create_export(export_request=request).id
 
         if register_for_cleanup:
             self._host_inventory.cleanup.add_exports({export_id}, scope=cleanup_scope)
 
         if wait_for_completion:
             status = self.wait_for_completion(export_id, **kwargs)
-
-            # Terminal states are "failed" and "complete".
```
fstavela, line 221: I think the comment is useful and we should keep it.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/exports_api.py
```diff
@@ -130,25 +130,31 @@ class ExportsAPIWrapper(BaseEntity):
...
+            return tmp.name
 
     def wait_for_completion(
         self,
         export_id: str,
         *,
-        retries: int = 20,
-        delay: float = 0.5,
+        retries: int = 30,
+        delay: float = 1,
```
fstavela: Is this increase really needed? I would at least keep the `delay` to half a second.

iqe-host-inventory-plugin/pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
   "faker",
   "google-api-python-client",
   "google-auth",
-  "iqe-bindings",
+  "iqe-bindings>=26.06.30.0",
   "iqe-core>=26.06.22.0",
   "iqe-export-service-plugin",
```
fstavela: Can we remove this now?

iqe-host-inventory-plugin/iqe_host_inventory/modeling/exports_api.py
mshriver: Also added to the wait_for_completion docblock so it's more clear that it's not only waiting for success

fstavela (dismissed):
LGTM, thank you! The pipelines are failing because of `kessel-inventory` deployment issue, we'll need to wait for that to get fixed.

It would also be good if we re-locked the uv.lock file to remove the `iqe-export-service-plugin` dependency, but it can also be done during the next lock update.

fstavela (approved):

PR #4576, opened 2026-07-08
chore(RHINENG-26247): Add ansible playbook that creates users for RBAC testing

fstavela:
## Jira
[RHINENG-26247](https://issues.redhat.com/browse/RHINENG-26247)

## What
Add Ansible playbooks that create users for RBAC testing

## Why
We want to start using pre-configured users for our RBAC tests to avoid having to change permissions during the tests. Changing the permissions is time-consuming when Kessel is enabled (due to caching), so this will save time on Kessel-enabled accounts.

## How
Create 4 playbooks:
- create_users.yml — auto-discovers the org admin's username, email, and account ID from the SSO token, then creates any missing test users via the Red Hat Account API. Each user's email is derived from the admin's email using + aliasing so all confirmation emails land in one inbox.
- setup_rbac_v1.yml — strips the "Default access" group of all roles (so test users don't inherit unwanted permissions), then for each user: verifies the user exists, removes them from non-designated groups, creates a dedicated test-<suffix> group with the correct role assigned.
- setup_rbac_v2.yml — same as v1 but uses the RBAC v2 API (role bindings on workspaces instead of roles on groups). Kept as a separate playbook because making any v2 API call permanently disables v1 for that org.
- teardown_rbac.yml — deletes all test groups (clearing role bindings first in v2) and custom roles.

All playbooks share common task files (tasks_auth.yml, tasks_user_cleanup.yml, tasks_group_setup.yml) and are idempotent. User definitions live in vars/users.yml — 15 users covering preconfigured roles, custom single-permission roles, and a no-permissions user.

## Testing
I ran the playbooks on both Kessel-enabled and Kessel-disabled accounts in Stage and Prod, and they worked fine.

## PR Guidelines

## PR Guideline checks

[RHINENG-26247]: https://redhat.atlassian.net/browse/RHINENG-26247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Add Ansible-based tooling to create and manage dedicated RBAC test users and their permissions for HBI across RBAC v1 and v2.

New Features:
- Introduce Ansible playbook to create Red Hat user accounts for HBI RBAC testing via the Red Hat Account API.
- Add RBAC v1 and RBAC v2 setup playbooks that configure per-user roles, groups, and workspace bindings for deterministic permissions.
- Provide a teardown playbook to remove test groups and custom roles for both RBAC v1 and v2 configurations.

Enhancements:
- Share common Ansible tasks for authentication, user cleanup, group setup, and per-user configuration between RBAC v1 and v2 workflows.
- Define a reusable users roster and environment-specific configuration for prod and stage, including proxy and API endpoints.

Documentation:
- Document how to run the RBAC setup, teardown, and user-creation playbooks, including prerequisites, v1 vs v2 guidance, and required variables.

### DIFF
--- a/scripts/rbac-setup/tasks_auth.yml
+++ b/scripts/rbac-setup/tasks_auth.yml
@@ -19,7 +19,7 @@
   retries: "{{ api_retries }}"
   delay: "{{ api_delay }}"
   until: sso_response is not failed
-  failed_when: false
+  ignore_errors: true
 
 - name: "Fail if token exchange failed"
   ansible.builtin.fail:
--- a/scripts/rbac-setup/tasks_auth.yml
+++ b/scripts/rbac-setup/tasks_auth.yml
@@ -1,0 +1,51 @@
+---
+# Shared authentication and identity discovery tasks
+# Required variables: offline_token, sso_token_url, account_api_base,
+#                     proxy_env, api_retries, api_delay
+# Sets: auth_header, org_admin_username, org_admin_email
+
+- name: "Exchange offline token for access token"
+  ansible.builtin.uri:
+    url: "{{ sso_token_url }}"
+    method: POST
+    body_format: form-urlencoded
+    body:
+      grant_type: "refresh_token"
+      client_id: "rhsm-api"
+      refresh_token: "{{ offline_token }}"
+    status_code: 200
+  environment: "{{ proxy_env }}"
+  register: sso_response
+  retries: "{{ api_retries }}"
+  delay: "{{ api_delay }}"
+  until: sso_response is not failed
+  failed_when: false
+
+- name: "Fail if token exchange failed"
+  ansible.builtin.fail:
+    msg: >-
+      Failed to authenticate with SSO. Is your offline token valid/expired?
+      HTTP {{ sso_response.status }}: {{ sso_response.json | default(sso_response.msg) }}
+  when: sso_response is failed or sso_response.status != 200
+
+- name: "Set auth header"
+  ansible.builtin.set_fact:
+    auth_header:
+      Authorization: "Bearer {{ sso_response.json.access_token }}"
+
+- name: "Discover org admin identity"
+  ansible.builtin.uri:
+    url: "{{ account_api_base }}/user"
+    method: GET
+    headers: "{{ auth_header }}"
+    status_code: 200
+  environment: "{{ proxy_env }}"
+  register: admin_identity
+  retries: "{{ api_retries }}"
+  delay: "{{ api_delay }}"
+  until: admin_identity is not failed
+
+- name: "Set org admin username and email"
+  ansible.builtin.set_fact:
+    org_admin_username: "{{ admin_identity.json.body.username }}"
+    org_admin_email: "{{ admin_identity.json.body.email }}"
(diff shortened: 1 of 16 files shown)


fstavela:
@sourcery-ai review

scripts/rbac-setup/tasks_v2_per_user.yml
```diff
@@ -0,0 +1,141 @@
...
+
+    - name: "Set target workspace ID"
+      ansible.builtin.set_fact:
+        target_workspace_id: "{{ workspace_ids[user.workspace_type] }}"
+
+    - name: "Check existing role bindings for group"
+      ansible.builtin.uri:
+        url: "{{ rbac_v2_base }}/role-bindings/?subject_type=group&subject_id={{ group_uuid }}&limit=100"
+        method: GET
+        headers: "{{ auth_header }}"
```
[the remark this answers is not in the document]
fstavela: This won't happen in our test orgs

scripts/rbac-setup/tasks_auth.yml
```diff
@@ -0,0 +1,51 @@
...
+      grant_type: "refresh_token"
+      client_id: "rhsm-api"
+      refresh_token: "{{ offline_token }}"
+    status_code: 200
+  environment: "{{ proxy_env }}"
+  register: sso_response
+  retries: "{{ api_retries }}"
+  delay: "{{ api_delay }}"
+  until: sso_response is not failed
+  ignore_errors: true
```
adarshdubey-star, line 22: Should we ignore errors here ?

adarshdubey-star, line 22: This task sends offline_token in the body and registers sso_response, which contains the access_token in its JSON. Without no_log: true, Ansible will print the entire request body (including refresh_token) and the entire response (including access_token) when running in verbose mode, or if the task fails (since ignore_errors: true still prints the output).

```suggestion
- name: "Exchange offline token for access token"
  ansible.builtin.uri:
    url: "{{ sso_token_url }}"
    method: POST
    body_format: form-urlencoded
    body:
      grant_type: "refresh_token"
      client_id: "rhsm-api"
      refresh_token: "{{ offline_token }}"
    status_code: 200
  environment: "{{ proxy_env }}"
  register: sso_response
  retries: "{{ api_retries }}"
  delay: "{{ api_delay }}"
  until: sso_response is not failed
  ignore_errors: true
  no_log: true            # This kinda prevents printing the token : )

- name: "Set auth header"
  ansible.builtin.set_fact:
    auth_header:
      Authorization: "Bearer {{ sso_response.json.access_token }}"
  no_log: true            # Same as above : )
```

adarshdubey-star (dismissed):
Thanks for adding this, left some suggestions

scripts/rbac-setup/tasks_auth.yml
fstavela: This is on purpose. The next task checks if the `sso_response` is failed and provides a human-readable error with the API response.

fstavela: Nice catch! I've added `no_log`.

adarshdubey-star (approved):
Ack

PR #4596, opened 2026-07-14
disable host filtering by staleness

pfreyburg:
## Jira
[RHINENG-28769](https://redhat.atlassian.net/browse/RHINENG-28769)

## What
The host-inventory is filtering hosts on Satellite by staleness but these hosts aren't deleted there because there is no such job. These filtered and not deleted hosts can cause failing INSERT on re-inserting existing primary key.

## Why
On the Satellite there is no inventory reaper job running. So when the host gets into culled state it is not deleted from the database but the SELECT is ignoring these host. When there is run `insights-client --register` on the host the inventory-host service fails because of the attempt to reinsert primary key.

## How
Removed the filtering of culled hosts from `multiple_canonical_facts_host_query`.

## Testing
I provisioned Satellite 6.19 instance and registered RHEL host. I manually changed all staleness timestamps of the host in database two months back. Then I tried reregister the host using `insights-client --register` and the inventory-host service failed.

After that I replaced the iop-core-host-inventory.service's image with the new image built from this PR. After this reregistering was successful. Without failing iop-core-host-inventory service.

## PR Guidelines

## PR Guideline checks

[RHINENG-28769]: https://redhat.atlassian.net/browse/RHINENG-28769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/host_repository.py
+++ b/lib/host_repository.py
@@ -202,7 +202,9 @@ def multiple_canonical_facts_host_query(
     )
     if restrict_to_owner_id:
         query = update_query_for_owner_id(identity, query)
-    return find_non_culled_hosts(query, identity.org_id)
+    # NOTE: Culled hosts are intentionally not filtered out here to prevent 
+    # service failures caused by attempts to re-insert existing primary keys.
+    return query
 
 
 def multiple_canonical_facts_host_query_in_memory(


raswonders (approved):
lgtm

lib/host_repository.py
```diff
@@ -202,7 +202,7 @@ def multiple_canonical_facts_host_query(
     )
     if restrict_to_owner_id:
         query = update_query_for_owner_id(identity, query)
-    return find_non_culled_hosts(query, identity.org_id)
+    return query # return query without filtering culled hosts
```
fstavela: We should probably mention why we are not filtering out culled hosts, so that we know why we have this deviation from the upstream HBI code.

fstavela (commented):
There are other places in the app where we filter out the culled hosts. Won't that be a problem?

pfreyburg:
> There are other places in the app where we filter out the culled hosts. Won't that be a problem?

It probably won't. The problem is the whole service now fails without this fix.

lib/host_repository.py
pfreyburg: @fstavela I have added longer comment. Is it okay this way?

fstavela: yes, looks good now, thank you!

fstavela (approved):

fstavela (approved):

PR #4606, opened 2026-07-15
fix(RHINENG-28788): Handle Deadlock exceptions gracefully

kruai:
## Jira
[RHINENG-28788](https://issues.redhat.com/browse/RHINENG-28788)

## What
Adds handling for DB deadlock exceptions.

## Why
The mq-workspaces pod restarts due to this exception a few times per day.

## How
Adds code to detect when an OperationalError is a DeadlockDetected error. If so, it retries; if not, it allows the pod to restart.

## Testing
Adds a scenario to test_event_loop_handles_invalid_request_error_gracefully
Adds test_deadlock_operational_error_is_reraised and test_non_deadlock_operational_error_still_exits

## PR Guidelines

## PR Guideline checks

[RHINENG-28788]: https://redhat.atlassian.net/browse/RHINENG-28788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Handle database deadlock OperationalErrors without terminating the MQ consumer process by treating them as retriable in the event loop while preserving existing behavior for other OperationalErrors.

Bug Fixes:
- Allow PostgreSQL deadlock OperationalErrors to bubble up so the event loop can retry instead of exiting the consumer process.
- Ensure non-deadlock OperationalErrors still cause the MQ consumer to exit with the appropriate status code.

Tests:
- Add tests to verify deadlock OperationalErrors are re-raised and do not trigger process exit in both span-enabled and span-disabled paths.
- Extend event loop error-handling tests to cover PostgreSQL deadlock scenarios and confirm they are retried.


rodrigonull (commented):
Thanks for the PR! 

I have just one observation: each retry calls _process_batch() → consume(), which fetches new messages rather than replaying the failed batch. The deadlocked message's offset stays uncommitted (no data loss), but it won't be reprocessed until the next pod restart or rebalance. Are we okay with that behavior?

kruai:
> Thanks for the PR!
> 
> I have just one observation: each retry calls _process_batch() → consume(), which fetches new messages rather than replaying the failed batch. The deadlocked message's offset stays uncommitted (no data loss), but it won't be reprocessed until the next pod restart or rebalance. Are we okay with that behavior?

Nah, I'd rather it retries the failed messages immediately afterwards. It looks like that was an issue for DeadlockDetected, but also for IntegrityError and StaleDataError. I've pushed an update to our batch processing that should fix this.

rodrigonull (approved):
LGTM!

adarshdubey-star (approved):

PR #4607, opened 2026-07-15
refactor: Change nginx image to a var

SteveHNH:
In the govcloud environment, we need to be able modify the nginx image that we use. By making this change, it leaves the commercial deployment untouched, but allows us to make changes in app-interface to override the image more easily.

## What

Set the nginx image to be a variable that defaults to the latest nginx-124 image

## Why

If it's a variable, we can override it in the govcloud environment when necessary and more easily.

## How

Add the vars to the template, and set the default parameters to the current image as it stands today.

## PR Guidelines

## PR Guideline checks

## Summary by Sourcery

Enhancements:
- Replace the hardcoded nginx image reference in the deployment template with parameterized NGINX_IMAGE and NGINX_IMAGE_TAG values.


thearifismail (commented):
LGTM

kruai (approved):

PR #4608, opened 2026-07-15
feat(RHINENG-28583): Migrate account staleness from api bindings to api wrapper

thearifismail:
## Jira
[RHINENG-28583](https://issues.redhat.com/browse/RHINENG-28583)

## What
Migrate Account Staleness API Wrapper to use BaseAPIWrapper

## Why
`BaseAPIWrapper` talks to HBI through IQE’s http_client as plain HTTP, so you avoid regenerating/reviewing apigen bindings and debugging through a generated client stack. Spec churn (especially operationId renames for V2) no longer breaks call sites when the endpoint itself hasn’t changed.

## How
`AccountStalenessAPIWrapper` holds a `BaseAPIWrapper`, which builds /api/inventory/v1 from IQE config and sends requests through app.http_client (RobustSession). Wrapper methods call that with a path and optional JSON body (e.g. post("/account/staleness", json=body)), get back a raw requests.Response, and parse with .json() instead of going through apigen models/clients.

## Testing
Validate the acceptance criteria.

[RHINENG-28583]: https://redhat.atlassian.net/browse/RHINENG-28583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Migrate account staleness modeling and tests from the OpenAPI-generated client to a BaseAPIWrapper that uses plain HTTP responses, while centralizing org_id leak checks on JSON responses.

New Features:
- Expose raw POST and PATCH helpers on the account staleness wrapper to send arbitrary JSON bodies for negative validation scenarios.

Bug Fixes:
- Ensure org_id leak detection also applies to plain HTTP JSON responses by validating response bodies before use.

Enhancements:
- Refactor account staleness wrapper methods to operate on requests.Response objects and JSON payloads instead of OpenAPI-generated models.
- Simplify staleness utilities to work directly with JSON dictionaries and remove OpenAPI-specific types.
- Update staleness, RBAC, and Kafka notification tests to assert on HTTP status codes and JSON content instead of apigen exceptions and models.

Tests:
- Add unit tests for the new assert_response_org_id_matches helper to verify correct behavior for matching, mismatched, and unsuccessful responses.
- Adjust existing staleness and RBAC test suites to cover the new HTTP-based wrapper behavior and negative scenarios.


iqe-host-inventory-plugin/iqe_host_inventory/modeling/staleness_api.py
```diff
@@ -33,47 +30,45 @@ def _host_inventory(self) -> ApplicationHostInventory:
...
+    def _build_staleness_body(
+        self,
+        conventional_time_to_stale: int | None,
+        conventional_time_to_stale_warning: int | None,
+        conventional_time_to_delete: int | None,
+        immutable_time_to_stale: int | None,
+        immutable_time_to_stale_warning: int | None,
+        immutable_time_to_delete: int | None,
+    ) -> dict[str, int]:
+        return {
```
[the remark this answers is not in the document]
thearifismail: Rejecting it because:
get_staleness_fields() only covers the 3 conventional fields, while this helper builds bodies for all 6 (including immutable). Zipping them would drop immutable values. Request body construction and response field extraction intentionally differ; hardcoding the keys here is the clear/safe approach.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/staleness_api.py
```diff
@@ -117,44 +101,29 @@ def update_staleness(
...
+    def delete_staleness(self, **api_kwargs: Any) -> requests.Response:
         with self._host_inventory.apis.measure_time("DELETE /account/staleness"):
-            return self.raw_api.api_staleness_delete_staleness(**api_kwargs)
+            return self._base_wrapper.delete("/account/staleness", **api_kwargs)
 
     @contextmanager
     def cleanup_before_and_after(self) -> Generator[None]:
-        with suppress(ApiException):
+        with suppress(Exception):
             self.delete_staleness()
```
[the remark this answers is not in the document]
thearifismail: Leaving suppress(Exception) in best-effort test cleanup is fine — it only affects teardown noise, not production behavior, so it shouldn’t block the PR.

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/notifications/kafka/test_notifications_kafka_registered.py
```diff
@@ -290,15 +290,12 @@ def test_notifications_kafka_registered_create_staleness_dont_produce(
...
-        )
+    staleness = host_inventory.apis.account_staleness.create_staleness(
+        conventional_time_to_stale=1000,
+        headers={"x-rh-insights-request-id": generate_uuid()},
+    )
 
     with pytest.raises(KafkaMessageNotFoundError):
         host_inventory.kafka.wait_for_filtered_registered_notification_message(
-            RegisteredNotificationWrapper.inventory_id, staleness.id.actual_instance, timeout=3
+            RegisteredNotificationWrapper.inventory_id, staleness.json()["id"], timeout=3
```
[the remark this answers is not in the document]
thearifismail: added 
```
    assert staleness.status_code in (200, 201)
    assert "id" in staleness.json()
```

iqe-host-inventory-plugin/iqe_host_inventory/modeling/staleness_api.py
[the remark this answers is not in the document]
thearifismail: Not worth blocking due to:
BaseAPIWrapper.delete() does not raise on HTTP status codes — it returns a Response. So suppress(Exception) mostly catches network/transport failures (requests.RequestException) and real bugs. The old ApiException path is gone.

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/staleness/test_staleness_post.py
```diff
@@ -134,8 +131,9 @@ def test_staleness_create_existing_record(host_inventory: ApplicationHostInvento
     logger.info("Creating account record")
     host_inventory.apis.account_staleness.create_staleness(**test_data)
 
-    with raises_apierror(400, match_message="already exists"):
-        host_inventory.apis.account_staleness.create_staleness(**test_data)
+    resp = host_inventory.apis.account_staleness.create_staleness(**test_data)
+    assert resp.status_code == 400
+    assert "already exists" in resp.text
```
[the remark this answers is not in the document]
thearifismail: Skip it. Nice-to-have, not a real gap introduced by this PR.

The pre-migration tests only asserted the 400 via raises_apierror — they never checked “state unchanged” either. Switching to status_code / resp.text didn’t drop that coverage; it was never there.

Also, Sourcery’s snippet is wrong: get_staleness() already returns dict[str, int], so .json() on it would blow up.

For test_staleness_create_value_too_big, there’s often no custom record beforehand — “unchanged” just means still defaults. That’s weak signal for what the test is about (validation rejection).

If you ever add it, do it only where it matters (e.g. duplicate create): capture pre = get_staleness() before the failing POST, then assert get_staleness() == pre. Don’t apply their broken sample as-is.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/staleness_api.py
```diff
@@ -33,47 +30,45 @@ def _host_inventory(self) -> ApplicationHostInventory:
...
+            return self._base_wrapper.get("/account/staleness", **api_kwargs)
 
     def get_staleness(self, **api_kwargs: Any) -> dict[str, int]:
-        """Retrieve the current staleness settings and extract staleness fields from the response.
+        return extract_staleness_fields(self.get_staleness_response(**api_kwargs).json())
 
-        :return dict[str, int]: Extracted staleness fields from the retrieved current staleness
-        """
-        return extract_staleness_fields(self.get_staleness_response(**api_kwargs))
+    def _build_staleness_body(
```
ezr-ondrej, line 37: This private function is in the middle of public methods. Can we organise the file a bit and have private methods first (or last)?

iqe-host-inventory-plugin/iqe_host_inventory/modeling/staleness_api.py
```diff
@@ -117,44 +101,29 @@ def update_staleness(
...
-        """
+    def delete_staleness(self, **api_kwargs: Any) -> requests.Response:
         with self._host_inventory.apis.measure_time("DELETE /account/staleness"):
-            return self.raw_api.api_staleness_delete_staleness(**api_kwargs)
+            return self._base_wrapper.delete("/account/staleness", **api_kwargs)
 
     @contextmanager
     def cleanup_before_and_after(self) -> Generator[None]:
-        with suppress(ApiException):
+        with suppress(Exception):
```
ezr-ondrej: This would now suppress anything that goes wrong. Why we needed this change?

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/rbac/staleness/test_rbac_staleness_cert_auth.py
```diff
@@ -112,15 +107,13 @@ def test_rbac_staleness_cert_auth_bypass_checks_update_staleness(
...
-        with raises_apierror(
-            403, "You don't have the permission to access the requested resource"
-        ):
-            host_inventory_non_org_admin_cert_auth.apis.account_staleness.update_staleness(
-                **settings
-            )
+        resp = host_inventory_non_org_admin_cert_auth.apis.account_staleness.update_staleness(
+            **settings
+        )
+        assert resp.status_code == 403
```
ezr-ondrej, line 116: Minor NIT: We are not verifying the message now (probably not big deal given these are E2E tests)

ezr-ondrej (commented):
I like the direction here 🧡 I've left few comments.
One caveat is those commented out lines, where I don't understand how re-enabling strict validation will help to get to the right solution. Can you please provide more details there?

thearifismail:
> I like the direction here 🧡 I've left few comments. One caveat is those commented out lines, where I don't understand how re-enabling strict validation will help to get to the right solution. Can you please provide more details there?

Fair catch. Those TODOs incorrectly linked body unknown-field handling to RHINENG-14003 (which is about Connexion strict_validation for undefined query params / deep-object filters — unrelated to this migration and to unknown JSON body keys).

I’ve removed the misleading commented blocks from this PR and kept only the assertion that matches current soft behavior ("bad_field" not persisted). Any hard 400 for unknown body fields can land in a follow-up if/when we decide that product change.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/staleness_api.py
thearifismail: Since it was done while cleaning up after running the test, I let it go. Just narrowed the exception to catch `request.RequestException`

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/rbac/staleness/test_rbac_staleness_cert_auth.py
thearifismail: Yes, since asserting for 403 is enough as it means no permission, asserting for message itself does not add much value.

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/staleness/test_staleness_patch.py
```diff
@@ -194,14 +193,10 @@ def test_staleness_update_invalid_field(host_inventory: ApplicationHostInventory
...
-    with api_disabled_validation(host_inventory.apis.account_staleness.raw_api) as api:
-        # TODO: Uncomment this when https://issues.redhat.com/browse/RHINENG-14003 is done
-        # with raises_apierror(400, match_message="Unknown field"):
-        #    api.api_staleness_update_staleness(test_data)
-        # TODO: And delete these 3 lines (temporary replacement for above)
-        api.api_staleness_update_staleness(test_data)
-        settings = host_inventory.apis.account_staleness.get_staleness()
-        assert "bad_field" not in settings
+    # Current API ignores unknown body fields; assert they are not persisted.
+    host_inventory.apis.account_staleness._base_wrapper.patch("/account/staleness", json=test_data)
```
rodrigonull: Accessing `_base_wrapper` directly couples tests to a private implementation detail. Consider exposing a public `raw_post`/`raw_patch` method (or just making `_base_wrapper` public as `base_wrapper`) on `AccountStalenessAPIWrapper` for these negative-validation cases.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/staleness_api.py
```diff
@@ -33,47 +30,45 @@ def _host_inventory(self) -> ApplicationHostInventory:
...
+            }.items()
+            if v is not None
+        }
+
+    def get_default_staleness_response(self, **api_kwargs: Any) -> requests.Response:
         with self._host_inventory.apis.measure_time("GET /account/staleness/defaults"):
-            return self.raw_api.api_staleness_get_default_staleness(**api_kwargs)
+            return self._base_wrapper.get("/account/staleness/defaults", **api_kwargs)
 
     def get_default_staleness(self, **api_kwargs: Any) -> dict[str, int]:
```
rodrigonull, line 64: `get_default_staleness` and `get_staleness` call `.json()` without checking the status code first. If the API returns a non-2xx response, this will raise a `KeyError` from `extract_staleness_fields` (or a `JSONDecodeError` if the body isn't JSON). Adding `resp.raise_for_status()` before `.json()` would surface the actual HTTP error immediately.

rodrigonull (commented):
Thanks for the PR @thearifismail !

The old wrapper had `@check_org_id` on `get_default_staleness_response`, `get_staleness_response`, `create_staleness`, `update_staleness`, and `delete_staleness`. This decorator automatically validated that the `org_id` in every API response matches the calling tenant and the migration removes it without a replacement. The RBAC tests still do spot-check `org_id` assertions, but the automatic, always-on validation is gone.

Do you think we should add a replacement for that?

iqe-host-inventory-plugin/iqe_host_inventory/modeling/staleness_api.py
thearifismail: moved it up with the other private functions.

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/staleness/test_staleness_patch.py
thearifismail: Fixed.

iqe-host-inventory-plugin/iqe_host_inventory/modeling/staleness_api.py
thearifismail: Fixed

thearifismail (commented):
@ezr-ondrej @rodrigonull you comments addressed

rodrigonull:
The implementation LGTM but I still have this concern about the @check_org_id decorator that has been removed: https://github.com/RedHatInsights/insights-host-inventory/pull/4608#pullrequestreview-4737465593

ezr-ondrej (dismissed):
TY, LGTM 👍

ezr-ondrej:
> The implementation LGTM but I still have this concern about the @check_org_id decorator that has been removed: [#4608 (review)](https://github.com/RedHatInsights/insights-host-inventory/pull/4608#pullrequestreview-4737465593)

The concern is not addressed (comment what we plan to do about it), so it cannot be merged yet

thearifismail:
> The concern is not addressed (comment what we plan to do about it), so it cannot be merged ye

thearifismail:
> The implementation LGTM but I still have this concern about the @check_org_id decorator that has been removed: [#4608 (review)](https://github.com/RedHatInsights/insights-host-inventory/pull/4608#pullrequestreview-4737465593)

`@check_org_id` decorator use was not valid anymore because now we get back `request.Response`.  To restore the org_id check, I added `assert_response_org_id_matches()` and updated its users to make explicit calls.

rodrigonull (approved):
LGTM

ezr-ondrej (approved):

PR #4613, opened 2026-07-16
feat(RHINENG-28773): enforce per-service RBAC on /hosts-view app_data

jkrocil:
## Jira
[RHINENG-28773](https://issues.redhat.com/browse/RHINENG-28773)

## What
Add per-service RBAC checks to the `/beta/hosts-view` endpoint so that each service's `app_data` is only included when the user has the minimum read permission for that service.

## Why
The endpoint aggregates data from 6 downstream services into a single response but only checked `inventory:hosts:read`. Any user with inventory access could see all service data regardless of their permissions in those services.

## How
Each `HostAppData` model declares its required permission (`__v1_read_permission__`, `__kessel_relation__`). A new `get_allowed_app_services()` in `lib/middleware.py` checks all services in a single RBAC v1 call or per-service `ListAllowedWorkspaces` for Kessel v2. The endpoint handler enforces three points: unauthorized service keys omitted from `app_data`, sorting by an unauthorized field returns 403, and filtering by an unauthorized field returns 403. The response includes a `denied_services` list so the frontend can distinguish denied services from services with no data.

## Testing
- 21 unit tests covering wildcard matching, RBAC bypass, identity types, RBAC v1, and Kessel v2 paths
- 10 integration tests covering partial/no/full access, sort 403, filter 403, authorized sort/filter, non-app sort, sparse fields, and denied_services
- All 182 tests passing

## PR Guidelines

## PR Guideline checks

[RHINENG-28773]: https://redhat.atlassian.net/browse/RHINENG-28773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jkrocil:
@sourcery-ai review

api/host_views.py
```diff
@@ -74,13 +94,19 @@ def _filter_app_data_fields(app_data: dict, requested_fields: list[str] | None)
     return {k: v for k, v in app_data.items() if k in requested_fields}
 
 
-def _fetch_app_data_for_hosts(host_ids: list, org_id: str, fields: dict | None = None) -> dict:
-    """Fetch app data for hosts. Returns all apps by default if fields is omitted."""
+def _fetch_app_data_for_hosts(
```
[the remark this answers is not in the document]
jkrocil: @sourcery-ai

- Suggestion 1: Not applicable - we now return `denied_services` in the response, which requires `allowed_apps` at the response-building level, not hidden away in a data-fetching helper.

- Suggestion 2: The combined validation function is 10 lines. Splitting it would add more code than it removes for no real benefit.

tests/test_host_views_rbac.py
```diff
@@ -0,0 +1,294 @@
...
+            result = get_allowed_app_services()
+            assert result is None
+
+    def test_system_identity_returns_none(self):
+        """System identity bypasses RBAC — return None."""
+        from lib.middleware import get_allowed_app_services
+
+        with (
+            patch("lib.middleware.inventory_config") as mock_config,
+            patch("lib.middleware.get_current_identity") as mock_identity,
```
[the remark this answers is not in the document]
jkrocil: Addressed in 6e59b62d.

api/host_views.py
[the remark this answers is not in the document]
jkrocil: Never mind I ended up splitting the validation, I guess it's more readable that way... don't have a strong opinion 🤷

app/models/host_app_data.py
```diff
@@ -25,6 +25,13 @@ class HostAppDataMixin:
     Subclasses should define:
         __app_name__: str - The application name used in API responses (e.g., "advisor")
         __sortable_fields__: tuple[str, ...] - Fields that can be used for sorting (optional)
+        __v1_app__: str - RBAC v1 application name for the permissions query (optional)
+        __v1_read_permission__: str - RBAC v1 permission required to view this data (optional)
+        __kessel_relation__: str - Kessel v2 contingent permission relation (optional)
```
kruai: I assume we need it set for all subclasses, so I don't think they should be optional

tests/test_host_views_rbac.py
```diff
@@ -0,0 +1,290 @@
...
+        ):
+            mock_config.return_value.bypass_rbac = False
+            identity = Mock(spec_set=["identity_type", "org_id", "service_account"])
+            identity.identity_type = IdentityType.SERVICE_ACCOUNT
+            identity.org_id = "test_org"
+            mock_identity.return_value = identity
+            mock_rbac.return_value = create_mock_rbac_response(
+                "tests/helpers/rbac-mock-data/inv-hosts-read-advisor-only.json"
+            )
+            result = get_allowed_app_services()
```
kruai: There's a lot of repetition in this file; it could do with a refactor and probably some parametrization.

lib/middleware.py
```diff
@@ -693,6 +693,106 @@ def check_access(permission: KesselPermission, ids: list[str] | None = None) ->
     return rbac_filter
 
 
+def _has_rbac_permission(rbac_permissions: list[str], required: str) -> bool:
```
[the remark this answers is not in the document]
jkrocil: @sourcery-ai

- Suggestion 1: The two permission matchers serve different domains (cross-service strings vs inventory enums). Unifying them would couple pre-existing code to our new feature for minimal benefit. I don't see the point.

- Suggestion 2: The v1 and v2 paths are intentionally separate - they do different things (HTTP vs gRPC). A shared descriptor would add abstraction without removing real duplication. Again - I don't see the point.

- Suggestion 3: Addressed.

app/models/host_app_data.py
```diff
@@ -116,6 +128,9 @@ class HostAppDataAdvisor(HostAppDataMixin, db.Model):
     __app_name__ = "advisor"
     __sortable_fields__ = ("recommendations", "incidents", "critical", "important", "moderate", "low")
     __filterable_fields__ = ("recommendations", "incidents", "critical", "important", "moderate", "low")
+    __v1_app__ = "advisor"
+    __v1_read_permission__ = "advisor:*:read"
```
kruai: For the `__v1_read_permission__` values that have wildcards, I'm worried that if a user has the equivalent of that app's read-all permission (i.e. they have `advisor:resource1:read` + `advisor:resource2:read` etc for all resources), it won't qualify as having adequate permissions. Do you think that would be an issue?

kruai (changes requested):
Great start! Got a few suggestions:

- `denied_services` is now included in the API response, but it's not defined in the API spec. Please add it to `swagger/api.spec.yaml`.
- I'm a bit concerned about how the extra auth requests are going to affect response latency; we should keep an eye on the before/after.
- We should really include some IQE tests for this kind of change, since it's dealing with integration.

app/models/host_app_data.py
jkrocil: Agreed - done!

tests/test_host_views_rbac.py
jkrocil: Done.

app/models/host_app_data.py
jkrocil: Good catch and specifically for Advisor I just made an adjustment to a specific read permission so that was a gap, for Vulnerability, Compliance and Remediations we already were using a specific permission but with Patch and Malware there is no specific permission to tie it to so we have to use a wildcard.

Full list:
- Advisor - `advisor:recommendation-results:read`
- Vulnerability - `vulnerability:vulnerability_results:read`
- Compliance - `compliance:system:read`
- Remediations - `remediations:remediation:read`
- Patch - `patch:*:read`
- Malware - `malware-detection:*:read`

So the risk with this approach obviously comes from external services changing used read permissions or adding new ones for the ones that don't have them but I would say the probability of such changes in RBAC v1 is low and the impact is low as well since the worst case scenario is that users would (until fixed) just lose access to some column data rather than gain access to data they shouldn't since we deny by default.

jkrocil:
> Great start! Got a few suggestions:
> 
> * `denied_services` is now included in the API response, but it's not defined in the API spec. Please add it to `swagger/api.spec.yaml`.
> * I'm a bit concerned about how the extra auth requests are going to affect response latency; we should keep an eye on the before/after.
> * We should really include some IQE tests for this kind of change, since it's dealing with integration.

- re denied_services: Done - spec updated.
- re response latency: we use a batched call for all 6 services for v1 so I wouldn't expect a problem there but we do make 6 sequential calls to ListAllowedWorkspaces for Kessel / RBAC v2 and I don't know (yet) how big a problem that is. I should be able to get some measurements when working on the IQE tests.
- re IQE: Makes sense - looking into this now.

jkrocil:
**New commits since last review - feature flag, tests, fixes, telemetry:**

Feature flag & unit test updates (ced5c65) - Added `hbi.inventory-views-rbac` feature flag (shared between backend and frontend, following platform.rbac.workspaces pattern). Added flag-off and bypass unit tests.

IQE tests (cfcee47) - 10 IQE tests (15 with parametrization) and Unleash flag support. Uses system "Malware detection viewer" role since RBAC v1 rejects custom role creation for malware-detection, otherwise we use custom roles for the remaining 5 services.

IQE client (e801ac4) - Regenerated via `iqe apigen` with `denied_services` field.

Prometheus metric (3a3e67e) - Added `inventory_views_kessel_rbac` label to `outbound_http_response_time` histogram for tracking the 6 ListAllowedWorkspaces calls. If we are stuck with the 6 separate calls let's at least have a way to monitor how this solution performs.

--

I'm testing this in ephemeral now, would like to make this work together with the frontend PR (which should be pretty much ready as well) before merging.

Update:
Ephemeral testing with RBACv1 with the frontend PR went well but I wasn't able to test Kessel so I'll be doing that in Stage. All of this work is behind `hbi.inventory-views-rbac` which is currently disabled everywhere.

lib/middleware.py
```diff
@@ -693,6 +694,112 @@ def check_access(permission: KesselPermission, ids: list[str] | None = None) ->
...
+
+    allowed: set[str] = set()
+
+    with outbound_http_response_time.labels("inventory_views_kessel_rbac").time():
+        for app_name, model in app_models.items():
+            kessel_relation = getattr(model, "__kessel_relation__", "")
+            if not kessel_relation:
+                continue
+            try:
+                workspaces = kessel_client.ListAllowedWorkspaces(identity, kessel_relation)
```
rodrigonull: We should consider doing these call in parallel, but that can be addressed in a follow up PR.

lib/middleware.py
```diff
@@ -693,6 +694,112 @@ def check_access(permission: KesselPermission, ids: list[str] | None = None) ->
...
+                if workspaces:
+                    allowed.add(app_name)
+            except Exception as e:
+                details = e.details() if hasattr(e, "details") else str(e)
+                logger.warning(f"Kessel check failed for {app_name}: {details}")
```
rodrigonull: Consider using:

```suggestion
                logger.warning("Kessel ListAllowedWorkspaces failed", extra={"app_name": app_name, "details": details})
```

This makes it greppable in Kibana / Cloudwatch

lib/middleware.py
jkrocil: Good call and I'm addressing it in this PR, commit below - that said, I do want to spend more time on Kessel testing once in Stage. 👍

lib/middleware.py
jkrocil: Done! Thank you!

lib/middleware.py
```diff
@@ -693,6 +694,126 @@ def check_access(permission: KesselPermission, ids: list[str] | None = None) ->
...
+            return None
+
+    allowed: set[str] = set()
+    with outbound_http_response_time.labels("inventory_views_kessel_rbac").time():
+        with ThreadPoolExecutor(max_workers=len(services_to_check)) as executor:
+            futures = {executor.submit(_check_service, name, rel): name for name, rel in services_to_check}
+            for future in futures:
+                result = future.result(timeout=10)
+                if result:
+                    allowed.add(result)
```
rodrigonull, line 754: `future.result(timeout=10)` raises `TimeoutError` if a thread exceeds the timeout, which would propagate unhandled and 500 the request. Consider wrapping it:

```suggestion
            for future in futures:
            try:
                result = future.result(timeout=10)
            except Exception:
                result = None
            if result:
                allowed.add(result)
```

rodrigonull (commented):
Just a minor suggestion. Otherwise LGTM.

Thank you for the PR :)

kruai (dismissed):
Looks good now, thanks!

lib/middleware.py
jkrocil: Done. :)

kruai (approved):

PR #4628, opened 2026-07-20
refactor(RHINENG-25830): Remove visible flag from column configuration

adarshdubey-star:
Jira
[RHINENG-25830](https://issues.redhat.com/browse/RHINENG-25830)

What
Remove visible flag from view column configuration. Presence in the columns array now determines visibility.

Why
Team agreed that visible: true/false is redundant — if a column is in the array it's shown, if it's not it's hidden. Simplifies the API contract and aligns with frontend's usage pattern.

How
Removed visible field from ColumnSchema in Marshmallow (sending it now returns 400 "Unknown field")
Removed visible property from ViewColumnConfig in OpenAPI spec
Updated all test data and DB fixtures to use {"key": "field_name"} without visible
Existing views with visible in their JSONB are unaffected (backend never reads the field)

Testing
69 unit/API tests passing
Verified via curl: creating without visible returns clean {"key": "..."} objects
Verified sending visible is rejected with 400
Verified existing views with old visible data still return correctly (backward compatible read)

PR Guidelines

PR Guideline checks

 Keep PRs under 400 lines of meaningful changes, including tests, excluding auto-generated files, config, etc.

## Summary by Sourcery

Simplify view column configuration by relying solely on column presence for visibility and removing the deprecated visible flag from the API and schema.

Enhancements:
- Remove the visible flag from ColumnSchema and view column configuration, making column visibility implicit based on inclusion in the columns array.
- Update OpenAPI specification and test/DB fixtures to reflect the new column configuration shape without the visible property.

Tests:
- Adjust API, schema, and repository tests to use the new column configuration format and validate behavior without the visible field.

[RHINENG-25830]: https://redhat.atlassian.net/browse/RHINENG-25830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rodrigonull (dismissed):
LGTM

rodrigonull (approved):

PR #4652, opened 2026-07-22
feat(RHINENG-29084): enable parent-child context propagation for MQ

rodrigonull:
## Jira
[RHINENG-29084](https://issues.redhat.com/browse/RHINENG-29084)

## What
Enable parent-child context propagation for per-message Kafka spans and propagate the trace context to outbound Kafka events, so HBI's entire processing — from message consumption to event production — becomes part of the upstream producer's trace.

## Why
To gain full visibility on how long an Insights client archive takes to be processed end-to-end (ingress → puptoo → HBI → advisor/vulnerability/patch). Without parent-child propagation, each service creates an isolated trace. With it, all services share a single trace, allowing us to measure total host registration time and identify processing bottlenecks directly in Grafana/Tempo.

## How
- Added `_extract_producer_context()` method to `HBIMessageConsumerBase` that parses W3C `traceparent`/`tracestate` headers from Kafka messages.
- Modified `_message_span()` and `_retroactive_span()` to use the extracted producer context as the parent span context when available, making the message processing span a child of the upstream producer's trace.
- Added `otel_context` field to `OperationResult` to capture the per-message OTel context during processing and re-attach it during event production via `_use_otel_context` context manager. This ensures outbound Kafka events (`platform.inventory.events`, `platform.notifications.ingress`) carry the upstream trace context.
- Replaced the Confluent Kafka consumer auto-instrumentation with a manual batch-level span, eliminating the cluttered auto-generated span links on the batch `process` span.
- Added per-message `msg N` child spans under the batch span, each with a single link to its upstream trace for easy navigation from the batch view to the end-to-end trace.
- Optimized by extracting producer context once per message and threading it through the processing chain to avoid double header parsing.

## Testing
- All existing unit tests pass (213 tests in `test_host_mq_service.py` + 17 in `test_mq_message_spans.py`).
- Pre-commit hooks pass (ruff, ruff format, mypy).
- Verified locally with docker-compose: produced a batch of 10 messages with unique W3C trace contexts and confirmed 10/10 outbound events carry the correct upstream `trace_id`.
- Verified in Tempo that the end-to-end trace shows: `puptoo.process_archive → mq.process → platform.inventory.events send + platform.notifications.ingress send` as a single trace.
- Batch trace shows clean per-message structure (`msg 0` through `msg 9`) with individual "View linked span" navigation.
## PR Guidelines

## PR Guideline checks

[RHINENG-29084]: https://redhat.atlassian.net/browse/RHINENG-29084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Enable Kafka per-message tracing to optionally join upstream producer traces for end-to-end visibility across services.

New Features:
- Add support for extracting W3C trace context from Kafka message headers and using it as the parent for per-message spans.
- Introduce OTEL_MQ_PROPAGATE_CONTEXT configuration flag to control whether message spans join the upstream producer trace.

Enhancements:
- Update per-message and retroactive span creation to conditionally use extracted producer context while preserving existing behavior when disabled.
- Extend OpenTelemetry initialization logging to report the MQ context propagation configuration state.


deploy/clowdapp.yml
```diff
@@ -3378,3 +3392,6 @@ parameters:
 - name: OTEL_MQ_SLOW_MESSAGE_MS
   description: When message spans are disabled, emit spans for messages exceeding this threshold (ms). 0 = disabled.
   value: '0'
+- name: OTEL_MQ_PROPAGATE_CONTEXT
+  description: When true, per-message spans use the upstream producer trace context as parent (end-to-end distributed tracing). When false, spans stay local to the batch.
+  value: 'true'
```
kruai: When will we want this to be False? Is it worth having the env var?

kruai (dismissed):

deploy/clowdapp.yml
rodrigonull: I was thinking of having it just for a mechanism to switch between the old way and the new way in case this causes unexpected behavior in Tempo, like traces become too wide, query performance degrade, etc, but the goal is to have this full trace accommodating all services, so I will remove it and make it as default = true

app/queue/host_mq.py
```diff
@@ -362,7 +438,7 @@ def _process_message_with_span(self, msg) -> None:
                 self.failure_metric.inc()
                 logger.exception("Unable to process message", extra={"incoming_message": msg.value()})
 
-    def _process_message_without_span(self, msg) -> None:
+    def _process_message_without_span(self, msg, producer_ctx=None) -> None:
```
kruai: `producer_ctx` isn't used in this function - shouldn't it be?

app/queue/host_mq.py
```diff
@@ -85,7 +89,6 @@
 from app.telemetry import OTEL_MQ_MESSAGE_SPANS_ENABLED
 from app.telemetry import OTEL_MQ_SLOW_MESSAGE_MS
 from app.telemetry import get_tracer
-from app.telemetry import instrument_kafka_consumer
```
kruai, line 88: `instrument_kafka_consumer` is effectively dead code now, and is only tested. It can be removed

kruai (changes requested):
In addition to my inline suggestions, I want to suggest that we set `producer_ctx` on the MQ consumer itself (i.e. `self.producer_ctx = self._extract_producer_context(msg)`); that way we don't have to pass it as an arg through multiple levels of function calls. Not required, I just think it would be a bit cleaner

app/queue/host_mq.py
rodrigonull: It is still used for the export service.

app/queue/host_mq.py
rodrigonull: It shouldn't. I will remove it.

app/queue/host_mq.py
kruai: You're right, I missed that 👍

kruai (approved):

PR #4659, opened 2026-07-23
perf(RHINENG-28940): move tag aggregation from Python to SQL

rodrigonull:
## Jira
[RHINENG-28940](https://issues.redhat.com/browse/RHINENG-28940)

## What
Move tag aggregation from Python to SQL for the `GET /api/inventory/v1/tags` endpoint, reducing response time by 70% and eliminating the memory-heavy `.all()` fetch pattern.

## Why
The old implementation fetched ALL hosts into Python memory and aggregated tags in a loop — O(hosts × tags_per_host). For large orgs, this caused high memory usage and ~3.7s response times. 

## How
- Replace Python-side iteration with a PostgreSQL lateral join using `jsonb_array_elements` on `tags_alt`, performing aggregation, search filtering, sorting, and pagination entirely in SQL.
- Use `count(*)` instead of `count(DISTINCT id)` — verified via production `EXPLAIN ANALYZE` that this enables HashAggregate with parallel workers (215ms vs 3705ms for 152k hosts, avoids 154MB disk spill).
- Add deduplication safeguard in both `tags_alt` write paths (`_populate_tags_alt_from_tags` and `_replace_tags_alt_in_namespace`) to structurally guarantee `count(*) == count(DISTINCT id)`. Confirmed zero duplicates exist across all production data via GABI.
- Fix `Tag.create_tags_from_nested` to correctly handle string tag values (was iterating characters).

## Testing
- All 46 existing tag tests pass with no behavioral changes.
- Added unit test for multi-value tag aggregation correctness.
- Added unit tests for `tags_alt` deduplication on create and update paths.
- Local Docker benchmark (10k hosts, 950 unique tags): OLD 320ms → NEW 95ms (70% faster).
- Production `EXPLAIN ANALYZE` via GABI (152k hosts): 3705ms → 215ms (17× faster).

## PR Guidelines

## PR Guideline checks

[RHINENG-28940]: https://redhat.atlassian.net/browse/RHINENG-28940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Move tag aggregation for the inventory tags endpoint from in-memory Python processing to SQL-based aggregation over tags_alt, improving performance and scalability while preserving behavior.

Bug Fixes:
- Correct handling of string tag values when creating tags from nested structures so they are treated as single values rather than iterated characters.

Enhancements:
- Aggregate tags directly in PostgreSQL using tags_alt JSONB, including search, sorting, and pagination, eliminating full host fetches and reducing response time.
- Guarantee uniqueness of (namespace, key, value) entries in tags_alt on both create and update paths to support efficient SQL aggregation.

Tests:
- Add tests covering multi-value tag aggregation behavior and verifying tags_alt deduplication on host creation and update.

### DIFF
--- a/api/host_query_db.py
+++ b/api/host_query_db.py
@@ -496,6 +496,44 @@ def _convert_null_string(input: str | list):
     return input
 
 
+def _build_tags_aggregation_query(filtered_hosts, search: str | None, order_by: str, order_how: str | None):
+    """Build the SQL aggregation query for tag listing using jsonb_array_elements."""
+    tag_lateral = (
+        func.jsonb_array_elements(func.coalesce(filtered_hosts.c.tags_alt, literal("[]").cast(JSONB)))
+        .table_valued(column("value", JSONB))
+        .lateral("t")
+    )
+
+    namespace_expr = tag_lateral.c.value["namespace"].astext
+    key_expr = tag_lateral.c.value["key"].astext
+    value_expr = tag_lateral.c.value["value"].astext
+    # count(*) is safe because tags_alt write paths deduplicate (ns, key, value) entries per host
+    host_count_expr = func.count()
+    tag_string_expr = func.concat(
+        func.coalesce(namespace_expr, "None"), "/", func.coalesce(key_expr, "None"), "=", func.coalesce(value_expr, "")
+    )
+
+    agg_stmt = (
+        select(
+            namespace_expr.label("namespace"),
+            key_expr.label("key"),
+            value_expr.label("value"),
+            host_count_expr.label("host_count"),
+        )
+        .select_from(filtered_hosts)
+        .join(tag_lateral, literal(True))
+        .group_by(namespace_expr, key_expr, value_expr)
+    )
+
+    if search:
+        agg_stmt = agg_stmt.where(tag_string_expr.op("~*")(search))
+
+    order_col = tag_string_expr if order_by == "tag" else host_count_expr
+    agg_stmt = agg_stmt.order_by(order_col.desc()) if order_how == "DESC" else agg_stmt.order_by(order_col.asc())
+
+    return agg_stmt
+
+
 def get_tag_list(
     display_name: str | None,
     fqdn: str | None,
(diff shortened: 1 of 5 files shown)


kruai (approved):
LGTM. `_build_tags_aggregation_query` is a bit hard for me to follow, but I don't think we can really improve that much.
The reason we originally moved the tag aggregation to python was DB load/performance. Can we do a comparison `EXPLAIN ANALYZE` on this before and after it gets to Stage?

rodrigonull:
> LGTM. `_build_tags_aggregation_query` is a bit hard for me to follow, but I don't think we can really improve that much. The reason we originally moved the tag aggregation to python was DB load/performance. Can we do a comparison `EXPLAIN ANALYZE` on this before and after it gets to Stage?

## EXPLAIN ANALYZE comparison - Prod (largest org: 152k hosts, 182 unique tags)

### BEFORE: Old approach (fetch all rows → Python aggregation)

```
Seq Scan on hosts_pN  (actual time=0.007..1463.364 rows=167,913)
  Filter: org_id = '<redacted>'
  Rows Removed by Filter: 174,799
Planning Time: 71.863 ms
Execution Time: 1,471.803 ms
```

**DB time: 1,472ms** — then Python iterates 167k rows × ~7 tags each, aggregates in a dict, sorts, and paginates. Total end-to-end estimate: **~2.5-3.5s**.

### AFTER: New approach (SQL aggregation with lateral join + `count(*)`)

```
Limit (actual time=255.357..256.077 rows=50)
  → Sort: quicksort Memory: 31kB
    → Finalize GroupAggregate (actual time=254.398..255.363 rows=182)
      → Gather Merge (Workers Launched: 2)
        → Partial HashAggregate (Batches: 1, Memory: 64kB)
          → Nested Loop + Memoize (Cache Hits: 167k, Misses: 636)
Planning Time: 13.162 ms
Execution Time: 277.439 ms
```

**DB time: 277ms** — returns the final 50-row paginated result directly. No Python post-processing needed.

### Summary

| | DB Execution | Python post-processing | Data transferred | Memory |
|---|---|---|---|---|
| **Before** | 1,472ms | ~1-2s (iterate 167k rows) | ~50MB (all rows) | ~200MB in app RAM |
| **After** | 277ms | None | ~5KB (50 rows) | 64KB HashAggregate |

### DB Load Analysis

Here's why the new approach should be **neutral or better** for RDS:

1. **Shorter connection hold time**: 277ms vs 1,472ms — each request releases the DB connection 5× faster, improving concurrency under load.
2. **Less I/O**: Transfers 50 rows (~5KB) vs 167k rows (~50MB) per request — reduces network I/O and buffer pool pressure.

**Net effect**: Less total RDS time per request, less I/O, less buffer pool pressure. The trade-off is slightly more CPU-per-millisecond during execution, but for a much shorter duration. Under concurrent load, the shorter hold time should reduce overall DB pressure.

PR #4668, opened 2026-07-24
feat(RHINENG-28671): Add V2 API spec skeleton

thearifismail:
## Jira
[RHINENG-28671](https://issues.redhat.com/browse/RHINENG-28671)

## What
Add new spec skeleton for V2 API with shared schemas, RFC 7807 errors, and unified pagination

## Why
This skeleton is need to start V2 API implementation.

## How
Add V2 API spec skeleton with Foundation-only schemas.

#### Error & Pagination: 
  - ErrorResponse (RFC 7807)
  - PaginationOut, PaginationLinks

#### Primitive/ID types:
  - AccountNumber, OrgId, HyphenatedUuid
  - NonNullableString, NonNullableStringArray
  - TimeStamp, HostId, GroupId, GroupName
#### Canonical Facts:
  - InsightsId, SubscriptionManagerId, SatelliteId, BiosUuid
  - IpAddresses, Fqdn, MacAddresses
  - ProviderId, ProviderType
  - CanonicalFactsOut, CanonicalFactsIn

#### Supporting:
  - FactSet, PerReporterStaleness, GroupOut
  - StructuredTag
  - SystemProfile, SystemProfileNestedObject

#### Shared Responses:
  - NoContent, BadRequestError, ForbiddenError, NotFoundError, ConflictError, ServiceUnavailableError

The Phase 1, 2, and 3 schemas are mentioned by name in the api_v2.spec.yaml

## Testing
Check the Acceptance Criteria in [RHINENG-28671](https://issues.redhat.com/browse/RHINENG-28671)

## PR Guidelines

## PR Guideline checks

[RHINENG-28671]: https://redhat.atlassian.net/browse/RHINENG-28671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHINENG-28671]: https://redhat.atlassian.net/browse/RHINENG-28671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Introduce an initial OpenAPI 3.0 V2 inventory specification with shared components but no concrete endpoints yet.

New Features:
- Define V2-wide shared query parameters for pagination, sorting, filtering, and host identification.
- Add common V2 schemas for RFC 7807-compliant error responses and a unified pagination envelope.
- Introduce canonical host fact schemas, host app data structures, and host/system profile/tag query result models for V2.
- Provide reusable V2 responses for standard HTTP error conditions aligned with the new error schema.

Enhancements:
- Align V2 API contract with design documentation, including stricter field validation and length constraints for host updates.

## Summary by Sourcery

Introduce an initial OpenAPI 3.0 specification file for the Inventory V2 REST API with shared components but no paths defined yet.

New Features:
- Define common V2 query parameters for pagination, sorting, host filtering, and identification.
- Add shared V2 schemas for RFC 7807-compliant error responses and a unified pagination envelope.
- Introduce foundational canonical fact and supporting data schemas for hosts, groups/workspaces, system profiles, and tags.
- Provide reusable V2 error and no-content response definitions referencing the new error schema.

Enhancements:
- Align V2 API contract with design docs by tightening ID formats, timestamp types, and field nullability for foundational schemas.


swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+    (RFC 7807 errors, stricter schemas, unified sorting and pagination) across
+    a phased endpoint migration.
+  version: 2.0.0
+  title: Insights Host Inventory V2 REST Interface
+servers:
+  - url: /api/inventory/v2
+    description: V2 API server
+externalDocs:
+  description: How to use the Red Hat Insights API
+  url: https://docs.redhat.com/en/documentation/red_hat_insights/1-latest/html/using_the_red_hat_insights_api/index
```
kruai: This URL gives me a 404

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+    sortParam:
+      name: sort
+      in: query
+      required: false
+      schema:
+        type: string
+      description: >-
+        Comma-separated list of sort fields. Prefix a field with `-` for
+        descending order. Examples: `?sort=-updated`, `?sort=display_name,-updated`,
+        `?sort=-vulnerability.critical_cves,display_name`.
```
kruai, line 78: We could also specify a `pattern` here to filter out things like `sort=+display_name`, `sort=-5`, etc

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+      type: string
+      nullable: true
+      example: i-05d2313e6b9a42b16
+    ProviderType:
+      description: >-
+        Type of external source e.g. Alibaba, AWS EC2, Azure, GCP, IBM, etc.
+        This field is a canonical fact and requires provider_id.
+      type: string
+      nullable: true
+      example: aws
```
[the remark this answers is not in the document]
thearifismail: No, this is not issue.

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+      description: Filter by FQDN (case-insensitive substring match).
+    hostnameOrId:
+      name: hostname_or_id
+      in: query
+      required: false
+      schema:
+        type: string
+      description: >-
+        Filter by display_name, fqdn, or id (case-insensitive substring match
+        on the first two, exact match on id).
```
kruai: This param has always felt a little awkward/hacky. We needed it in v1 bc it's effectively the general "search" filter, passed by the UI. From personal experience, I've basically only expect that search bar to filter on the display name... in which case, we could just use displayName. It would be worth checking with UI to see if this is still needed in v2.

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+      type: string
+      format: uuid
+      example: 22cd8e39-13bb-4d02-8316-84b850dc5136
+    GroupName:
+      description: >-
+        A group/workspace human-readable name.
+        Must contain only letters, numbers, spaces, hyphens, and underscores.
+      type: string
+      nullable: false
+      example: sre-group
```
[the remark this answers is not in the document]
thearifismail: Groups and workspaces will be dealt with as the very last step as the effort has to be coordinated with RBAC team, if it is to be worked on or nto

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+
+  # ---------------------------------------------------------------------------
+  # Shared Parameters
+  # ---------------------------------------------------------------------------
+  parameters:
+    # -- Pagination -----------------------------------------------------------
+    pageParam:
+      name: page
+      in: query
+      required: false
```
kruai: Pretty sure `required: false` is redundant, and that we can drop it from all query params.

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+      schema:
+        type: string
+        format: date-time
+      description: Only show hosts whose last check-in is before this timestamp.
+    tagsParam:
+      name: tags
+      in: query
+      required: false
+      style: form
+      explode: true
```
kruai, line 167: Why `style: form` and `explode: true`? I see that in a few other places too

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+      schema:
+        type: array
+        items:
+          type: string
```
kruai, line 171: We should also specify `pattern` here to ensure that they're in the right format

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+          enum:
+            - fresh
+            - stale
+            - stale_warning
+      description: >-
+        Filter by host staleness state. V2 removes "unknown" from the allowed
+        values (RHINENG-5893).
```
kruai: Not useful to have in the API doc.
```suggestion
        Filter by host staleness state.
```

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+  # Schemas
+  # ---------------------------------------------------------------------------
+  schemas:
+    # -- RFC 7807 Error Response ----------------------------------------------
+    ErrorResponse:
+      title: RFC 7807 Problem Details
+      description: >-
+        Standard error response following RFC 7807 Problem Details for HTTP APIs.
+      type: object
+      required:
```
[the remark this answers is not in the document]
thearifismail: the nullable gap sourcery-ai flagged doesn't exist in practice — it's structurally blocked by the allOf composition in
  CanonicalFactsIn, even though the standalone schemas remain nullable: true for output use (CanonicalFactsOut).

swagger/api_v2.spec.yaml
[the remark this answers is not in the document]
thearifismail: To be addressed as the last step if needed.

thearifismail (commented):
Ready for review

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+          enum:
+            - puptoo
+            - yupana
+            - satellite
+            - discovery
+            - rhsm-conduit
+            - cloud-connector
+      description: >-
+        Filter by reporter that registered the host. V2 removes "insights"
+        from the allowed values (RHINENG-4846).
```
kruai: It's not just the reporter that registered the host; it's the reporter that has ever reported the host before.
Also, the V2 changelog doesn't belong in the API endpoint description.
```suggestion
        Filter by reporter that has ever reported the host before.
```

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+        type: array
+        items:
+          type: string
+          enum:
+            - puptoo
+            - yupana
+            - satellite
+            - discovery
+            - rhsm-conduit
+            - cloud-connector
```
kruai, line 213: What about the inverses like `!puptoo`, `!yupana`, etc?

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+    systemTypeParam:
+      name: system_type
+      in: query
+      required: false
+      schema:
+        type: string
+        enum:
+          - nil
+          - not_nil
+          - edge
```
kruai, line 231: Just checking - are we still supporting `edge`?

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+      in: query
+      required: false
+      style: form
+      explode: true
+      schema:
+        type: array
+        items:
+          type: string
+      description: >-
+        Filter by workspace name. Replaces V1's deprecated group_name parameter.
```
kruai:
```suggestion
        Filter by workspace name.
```

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+      required: false
+      style: form
+      explode: true
+      schema:
+        type: array
+        items:
+          type: string
+          format: uuid
+      description: >-
+        Filter by workspace ID. Replaces V1's deprecated group_id parameter.
```
kruai:
```suggestion
        Filter by workspace ID.
```

swagger/api_v2.spec.yaml
kruai, line 290: What happens if any of these fields (status, title, detail) is NOT returned in the error response?

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+          type: string
+          example: /api/inventory/v2/hosts/query
+
+    # -- Pagination -----------------------------------------------------------
+    PaginationOut:
+      title: Paginated Response Envelope
+      description: >-
+        V2 unified pagination envelope. All V2 collection endpoints use this
+        structure, including resource-types (which used a different envelope in V1).
+        The links field provides navigation URLs with full query string preservation.
```
kruai:
```suggestion
        Unified pagination envelope. All collection endpoints use this
        structure, including resource-types.
        The links field provides navigation URLs with full query string preservation.
```

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+          description: The number of items on the current page.
+          type: integer
+        page:
+          description: The current page number.
+          type: integer
+        per_page:
+          description: The number of items per page.
+          type: integer
+        links:
+          $ref: '#/components/schemas/PaginationLinks'
```
kruai, line 349: Got some questions about combining these pagination formats.

- Doesn't this break the expected format for resource-types endpoints? The current response has `links`, `meta`, and `data` as top-level fields in the response
- Is `links` actually useful for the UI?

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+      description: The Org ID of the tenant that owns the host.
+      type: string
+      example: '000102'
+    HyphenatedUuid:
+      type: string
+      nullable: true
+      pattern: '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
+      minLength: 36
+      maxLength: 36
+      example: 22cd8e39-13bb-4d02-8316-84b850dc5136
```
kruai: Is this type helpful when we've already got `format: uuid`?

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+          description: Whether the group is the "ungrouped hosts" group.
+          type: boolean
+        created:
+          description: A timestamp when the entry was created.
+          type: string
+          format: date-time
+        updated:
+          description: A timestamp when the entry was last updated.
+          type: string
+          format: date-time
```
kruai: Don't we want to get rid of the distinction between Groups and Workspaces?

swagger/api_v2.spec.yaml
kruai: I thought we were getting rid of these "Group" namings?

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+      type: string
+      nullable: true
+      example: my.host.example.com
+    MacAddresses:
+      description: >-
+        Host's network interfaces MAC addresses. This field is a canonical fact.
+      type: array
+      nullable: true
+      items:
+        type: string
```
kruai, line 457: This could use a `format` or at least a `pattern`

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+      type: string
+      nullable: true
+      example: 22cd8e39-13bb-4d02-8316-84b850dc5136
+    IpAddresses:
+      description: >-
+        Host's network IP addresses. This field is a canonical fact.
+      type: array
+      nullable: true
+      items:
+        type: string
```
kruai, line 441: This could use a `format` or at least a `pattern`

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+        A Red Hat Satellite ID of a RHEL host. This field is a canonical fact.
+      type: string
+      nullable: true
+      example: 22cd8e39-13bb-4d02-8316-84b850dc5136
+    BiosUuid:
+      description: >-
+        A UUID of the host machine BIOS. This field is a canonical fact.
+      type: string
+      nullable: true
+      example: 22cd8e39-13bb-4d02-8316-84b850dc5136
```
kruai, line 434: These should all have `format: uuid`, right?

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+                - $ref: '#/components/schemas/SubscriptionManagerId'
+                - $ref: '#/components/schemas/HyphenatedUuid'
+            satellite_id:
+              allOf:
+                - $ref: '#/components/schemas/SatelliteId'
+                - $ref: '#/components/schemas/HyphenatedUuid'
+            bios_uuid:
+              allOf:
+                - $ref: '#/components/schemas/BiosUuid'
+                - $ref: '#/components/schemas/HyphenatedUuid'
```
kruai: Probably simpler to drop the `allOf` and `HyphenatedUuid` for each of these, and just specify what we need in each ID schema

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+            ip_addresses:
+              $ref: '#/components/schemas/NonNullableStringArray'
+            fqdn:
+              $ref: '#/components/schemas/NonNullableString'
+            mac_addresses:
+              $ref: '#/components/schemas/NonNullableStringArray'
+            provider_id:
+              $ref: '#/components/schemas/NonNullableString'
+            provider_type:
+              $ref: '#/components/schemas/NonNullableString'
```
kruai: Can we standardize these to use the more specific schemas like IPAddresses, MacAddresses, etc?

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+        last_check_in:
+          $ref: '#/components/schemas/TimeStamp'
+        stale_timestamp:
+          $ref: '#/components/schemas/TimeStamp'
+        stale_warning_timestamp:
+          $ref: '#/components/schemas/TimeStamp'
+        culled_timestamp:
+          $ref: '#/components/schemas/TimeStamp'
+        check_in_succeeded:
+          type: boolean
```
kruai: These are being dropped, right? Also, isn't its format an array of reporters, with each item being a dict with the reporter name & last_check_in?

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+    #    be added with Phase 3 endpoints:
+    #    SystemProfileByHostOut, HostSystemProfileOut
+
+  # ---------------------------------------------------------------------------
+  # Shared Responses
+  # ---------------------------------------------------------------------------
+  responses:
+    NoContent:
+      description: >-
+        The operation completed successfully. No content returned.
```
kruai: Needs a `status` at least

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+# V2 API spec — built from DESIGN_DOC_HBI_V2_API.docx.pdf.
+# OpenAPI 3.0.0 because Connexion 3.x does not support 3.1+.
+# Paths are empty; Phase 1/2/3 tasks add endpoint definitions.
+openapi: 3.0.0
+info:
+  description: >-
+    REST interface for the Insights Platform Host Inventory application — V2 API.
+    V2 is the full successor to V1, delivering contract-quality improvements
+    (RFC 7807 errors, stricter schemas, unified sorting and pagination) across
+    a phased endpoint migration.
```
kruai: This is all internal commentary, none of which is useful to users.
```suggestion
```

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+    description: V2 API server
+externalDocs:
+  description: How to use the Red Hat Insights API
+  url: https://docs.redhat.com/en/documentation/red_hat_insights/1-latest/html/using_the_red_hat_insights_api/index
+
+paths: {}
+
+components:
+  # ---------------------------------------------------------------------------
+  # Security Schemes (same as V1)
```
kruai:
```suggestion
  # Security Schemes
```

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,699 @@
...
+      in: query
+      required: false
+      schema:
+        type: integer
+        minimum: 1
+        maximum: 100
+        default: 50
+      description: A number of items to return per page.
+
+    # -- Sorting (replaces V1 order_by + order_how) ---------------------------
```
kruai:
```suggestion
    # -- Sorting ---------------------------
```

kruai (changes requested):

swagger/api_v2.spec.yaml
thearifismail: Done!

swagger/api_v2.spec.yaml
thearifismail: This is intentional for now and planning to handle the groups-to-workspaces rename as a separate pass at the end.

swagger/api_v2.spec.yaml
thearifismail: Oh, it got dropped when updating documents and revisions.  Fixed to this:
```
    PerReporterStaleness:
      type: array
      items:
        type: object
        required:
          - reporter
          - last_check_in
        properties:
          reporter:
            type: string
            description: Name of the reporting source.
          last_check_in:
            $ref: '#/components/schemas/TimeStamp'
```

swagger/api_v2.spec.yaml
thearifismail: Fixed

swagger/api_v2.spec.yaml
thearifismail: Fixed.
```
    CanonicalFactsIn:
      allOf:
        - properties:
            insights_id:
              $ref: '#/components/schemas/InsightsId'
            subscription_manager_id:
              $ref: '#/components/schemas/SubscriptionManagerId'
            satellite_id:
              $ref: '#/components/schemas/SatelliteId'
            bios_uuid:
              $ref: '#/components/schemas/BiosUuid'
            ip_addresses:
              $ref: '#/components/schemas/IpAddresses'
            fqdn:
              $ref: '#/components/schemas/Fqdn'
            mac_addresses:
              $ref: '#/components/schemas/MacAddresses'
            provider_id:
              $ref: '#/components/schemas/ProviderId'
            provider_type:
              $ref: '#/components/schemas/ProviderType'
```

swagger/api_v2.spec.yaml
thearifismail: MAC formats vary and HBI doesn't  validate them today.  Adding a strict pattern could reject MAC addresses that are currently accepted.

swagger/api_v2.spec.yaml
thearifismail: IP addresses can be IPv4, IPv6, link-local, with or without CIDR notation, and HBI doesn't validate them today. Adding a pattern that covers all valid formats would be complex and fragile, and a strict one risks rejecting addresses that currently work.  These are reporter-supplied canonical facts, not user input — adding format constraints in V2 would be a silent breaking change for check-in clients. Not worth the risk for fields that are just stored and returned as-is.

@kruai what do you think?

swagger/api_v2.spec.yaml
thearifismail: Fixed already `hyphenatedUuid` was removed.

swagger/api_v2.spec.yaml
thearifismail: This will be dealt with when Groups/workspaces will be updated in the last stage.

swagger/api_v2.spec.yaml
thearifismail: `HyphenatedUuid` gone!

swagger/api_v2.spec.yaml
thearifismail: About the first point:
Yes, this is an intentional V2 breaking change. Resource-types will adopt the same unified pagination envelope (total, count, page,  per_page, links, results) as all other V2 endpoints. The V1 {meta, links, data} format stays unchanged in V1.

Yes, `links` should be useful for the UI.  Links are part of json api standard.  I took it in think it would help the UI

swagger/api_v2.spec.yaml
thearifismail: accepted

swagger/api_v2.spec.yaml
thearifismail: They should be. If not, then the HBI should produce an RFC 7807 compliant message based on the information available and make up the information not available in the response.

swagger/api_v2.spec.yaml
thearifismail: Done!

swagger/api_v2.spec.yaml
thearifismail: This doesn't look right.  It should match the systemTypeParam in V1.  Fixed

swagger/api_v2.spec.yaml
thearifismail: Fixed.

swagger/api_v2.spec.yaml
thearifismail: Added `pattern: '^([^=/]+/)?[^=/]+(=[^=/]+)?$'`

swagger/api_v2.spec.yaml
thearifismail: `explode: true` happens by default in OpenAPI 3.0.0, i.e. implicitly.  Making them explicit for clarity and communicating intent.

swagger/api_v2.spec.yaml
thearifismail: @johnsonm325, @raswonders, @computercamplove can you guys please provide your input?  Thanks.

swagger/api_v2.spec.yaml
thearifismail: using this `pattern: '^-?[a-z][a-z0-9_.]*(,-?[a-z][a-z0-9_.]*)*$'`

swagger/api_v2.spec.yaml
thearifismail: Yes, you're right.  There were 20 `required: false` occurrences; removed them all.

swagger/api_v2.spec.yaml
thearifismail: Fixed

swagger/api_v2.spec.yaml
thearifismail: The link had changed.  Fixed.

swagger/api_v2.spec.yaml
thearifismail: Set `status: pending`

thearifismail (commented):
@kruai thank you very much for the thorough review.  I addressed all comments/questions except one for the UI folks. Tagged Michael, Rasti, and Zhamilya

rodrigonull (commented):
**provider_type enum mismatch between spec and code**

The `providerType` query parameter (line 131) lists `[alibaba, aws, azure, gcp, ibm]`, but the authoritative enum in the codebase (`app/models/constants.py`) also includes `discovery`:

```python
class ProviderType(StrEnum):
    ALIBABA = "alibaba"
    AWS = "aws"
    AZURE = "azure"
    DISCOVERY = "discovery"
    GCP = "gcp"
    IBM = "ibm"
```

This means a host ingested with `provider_type: "discovery"` would be unfilterable by the V2 query param.

Additionally, the `ProviderType` schema (line 496) used in `CanonicalFactsOut`/`CanonicalFactsIn` is an unconstrained `type: string`. Since the code already enforces the enum at ingestion time (in `app/models/schemas/host.py`), the spec should reflect that constraint too — both in the query parameter and in the schema itself. This keeps the contract honest and prevents drift between code and spec.

Suggestion: add `discovery` to the query param enum, and also add the same enum to the `ProviderType` schema so that the spec is the single source of truth for valid values.

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,709 @@
+---
+# V2 API spec — built from DESIGN_DOC_HBI_V2_API.docx.pdf.
```
rodrigonull:
```suggestion
# V2 API spec
```

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,709 @@
...
+          $ref: '#/components/schemas/Fqdn'
+        mac_addresses:
+          $ref: '#/components/schemas/MacAddresses'
+        provider_id:
+          $ref: '#/components/schemas/ProviderId'
+        provider_type:
+          $ref: '#/components/schemas/ProviderType'
+    CanonicalFactsIn:
+      allOf:
+        - properties:
```
rodrigonull:
```suggestion
        - type: object
          properties:
```

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,709 @@
...
+  parameters:
+    # -- Pagination -----------------------------------------------------------
+    pageParam:
+      name: page
+      in: query
+
+      schema:
+        type: integer
+        minimum: 1
+        maximum: 21474837
```
rodrigonull, line 53: Can we add a comment explaining the derivation? Otherwise it looks like a typo of 2147483647

```suggestion
        # (page - 1) * per_page must be less than 2,147,483,647 (signed int32 max).
        # per_page max is 100, so page max is 21,474,837.
        maximum: 21474837
```

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,709 @@
...
+          $ref: '#/components/schemas/IpAddresses'
+        fqdn:
+          $ref: '#/components/schemas/Fqdn'
+        mac_addresses:
+          $ref: '#/components/schemas/MacAddresses'
+        provider_id:
+          $ref: '#/components/schemas/ProviderId'
+        provider_type:
+          $ref: '#/components/schemas/ProviderType'
+    CanonicalFactsIn:
```
rodrigonull: The `anyOf` requires at least one canonical fact key to be present, but all referenced schemas (InsightsId, SubscriptionManagerId, etc.) are `nullable: true`. This means {"insights_id": null} satisfies the constraint while providing no usable identifier. V1 solves this by overriding input properties with non-nullable types (e.g. `HyphenatedUuid`, `NonNullableString`). I think we should do the same here, either add type: object with inline nullable: false overrides, or compose each property with a non-nullable variant. WDYT?

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,709 @@
...
+      description: Only show hosts whose last check-in is after this timestamp.
+    lastCheckInEnd:
+      name: last_check_in_end
+      in: query
+
+      schema:
+        type: string
+        format: date-time
+      description: Only show hosts whose last check-in is before this timestamp.
+    tagsParam:
```
rodrigonull, line 163: Design doc: "maxItems added to unbounded list parameters" as a V2 baseline fix. Currently none of the array params (tags, staleness, registered_with, workspace_name, workspace_id, host_id_list) have maxItems. Please add reasonable bound. e.g. V1's host_id_list had maxItems: 100; tags/workspace lists likely need similar caps to prevent oversized queries.

swagger/api_v2.spec.yaml
```diff
@@ -0,0 +1,709 @@
...
+      schema:
+        type: array
+        items:
+          type: string
+          enum:
+            - fresh
+            - stale
+            - stale_warning
+      description: >-
+        Filter by host staleness state.
```
rodrigonull, line 196: Missing default: [fresh, stale, stale_warning]. V1 declares this default in the spec, and the code relies on it. Without the default, clients omitting staleness will receive culled hosts in V2, which is a silent behavior change.

```suggestion
    stalenessParam:
        name: staleness
        in: query
        style: form
        explode: true
        schema:
          type: array
          items:
            type: string
            enum:
              - fresh
              - stale
              - stale_warning
          default:
            - fresh
            - stale
            - stale_warning
        description: >-
          Filter by host staleness state.
```

rodrigonull (changes requested):

swagger/api_v2.spec.yaml
thearifismail: Agreed — added `nullable: false` overrides on each `CanonicalFactsIn` property via `allOf` composition. The shared schemas stay `nullable: true` for output (absent facts are legitimately null) but input now rejects null values, so `{"insights_id": null}` no longer passes the `anyOf` constraint.

swagger/api_v2.spec.yaml
thearifismail: done

swagger/api_v2.spec.yaml
thearifismail: Done!

swagger/api_v2.spec.yaml
thearifismail: Agreed that V2 should enforce `maxItems` on unbounded list parameters — the design doc calls for it. However, none of these parameters are referenced by any path yet (`paths: {}`), so I'd prefer to add `maxItems` in each Phase PR alongside the endpoint that exercises it — that way the bound can be reviewed against actual usage. For `stalenessParam` specifically, `maxItems` is redundant since the enum already constrains it to 3 valid values.

swagger/api_v2.spec.yaml
thearifismail: Done!

thearifismail (commented):
@rodrigonull all your comments addressed

swagger/api_v2.spec.yaml
kruai: OK, what intent are you communicating?

swagger/api_v2.spec.yaml
kruai: Does that mean we do still need to support `edge`?

swagger/api_v2.spec.yaml
kruai: I'm just noting that those 3 fields are all marked as required. Currently when the API tries to respond with a format that doesn't align with the schema, it gets replaced with a generic HTTP 500 with no detail/description, which is tough to debug. Just want to make sure this was intentional

swagger/api_v2.spec.yaml
kruai: - It's an intentional breaking change? I'm confused, because the only reason we have the resource-types endpoints is for RBAC v1. Wouldn't that mean we _don't_ want to break the API contract here?
- I'd like UI to confirm that adding `links` is beneficial, and worth changing how we do our pagination. @johnsonm325 @computercamplove can you please weigh in here?

swagger/api_v2.spec.yaml
kruai: If we don't have a plan for groups/workspaces yet, then I don't think we should include them in the API spec yet. Let's figure out what we need first

swagger/api_v2.spec.yaml
kruai: Sure, it's not a huge deal. I just thought it would be a good time to add guidance and validation when possible.

swagger/api_v2.spec.yaml
kruai: I posted a similar comment elsewhere, so I'll continue in that thread, and we can resolve this one

swagger/api_v2.spec.yaml
kruai: The `pattern`, `minlength`, and `maxlength` are redundant on each of these, since we're using `format: uuid`

swagger/api_v2.spec.yaml
kruai: do our mac formats vary? I've only seen hosts use the colon format (e.g. `00:60:46:bd:1a:14`)

kruai (changes requested):

swagger/api_v2.spec.yaml
thearifismail: The way I understand is:
- sort=+display_name is rejected because ? only permits an optional minus sign —, + doesn't match.  
- sort=-5 is rejected because the first character after the optional '-' must be a lowercase letter ([a-z]), not a digit

swagger/api_v2.spec.yaml
thearifismail: The intent is to make the serialization format immediately visible to anyone reading the spec `— explode: true` means `?tags=a&tags=b (repeated keys)`, not `?tags=a,b (comma-separated)`. Since it's the default, a reader who doesn't have the OpenAPI 3.0 defaults memorized would have to look it up to know which format applies. Making it explicit removes that guesswork.

rodrigonull:
@thearifismail The only remaining thing from me is this comment: https://github.com/RedHatInsights/insights-host-inventory/pull/4668#pullrequestreview-4821626023. Please take a look

swagger/api_v2.spec.yaml
thearifismail: I don't know.  I am just carrying over from V1.  IIRC something was said about `edge`, was it about not supporting it?

swagger/api_v2.spec.yaml
thearifismail: Yes, intentional. The V2 error handler ([RHINENG-29105](https://redhat.atlassian.net/browse/RHINENG-29105)) will be the single point that produces all error responses, so it can guarantee all three fields are always present — status is always known from the HTTP code, title can be derived from the status (e.g. "Not Found"), and detail can fall back to a generic message if the exception doesn't provide one. The handler will never emit a response missing any of them, so the Connexion schema validation won't replace it with a bare 500.

swagger/api_v2.spec.yaml
thearifismail: The pagination links are being added because @ezr-ondrej wanted (paraphrasing here) to follow `JSON:API` [standard](https://jsonapi.org/format/).  IMO, pagination links are useful for navigation.  It is upto the UI to make use of it or not.  Do you see any harm in including these links.

As far `ResourceTypes` link is concerned I had the same concern that it was needed only for RBAC V1.  The reason I included was that it was being used by one or more external customers.

swagger/api_v2.spec.yaml
thearifismail: You're right about `minLength` and `maxLength` and they've been removed.  Did you say `pattern` by mistake?  Somewhere else I set the UUID values to be hyphenated.

swagger/api_v2.spec.yaml
thearifismail: MAC addresses can come in colon-separated (00:60:46:bd:1a:14), hyphen-separated (00-60-46-BD-1A-14), dot-separated (0060.46bd.1a14), or no separator at all — and case can vary. Since HBI accepts hosts from multiple reporters (insights-client, Satellite, RHSM, etc.), I'd rather not assume they all use the colon format without checking. Adding a pattern that rejects a valid host submission isn't worth the risk for a field we don't query or filter on.

thearifismail:
> @thearifismail The only remaining thing from me is this comment: [#4668 (review)](https://github.com/RedHatInsights/insights-host-inventory/pull/4668#pullrequestreview-4821626023). Please take a look

oh, I must have gotten distracted because I had set it for `registered_with` but didn't update the `providerType`. Now **done**.

thearifismail (commented):
@kruai PTAL

swagger/api_v2.spec.yaml
rodrigonull, line 467: Consider adding the same enum here as in the providerType query param (alibaba, aws, azure, discovery, gcp, ibm).

rodrigonull (changes requested):

swagger/api_v2.spec.yaml
thearifismail: Done

rodrigonull (dismissed):

swagger/api_v2.spec.yaml
kruai: Seems good; maybe slightly more complex than it needs to be, since I don't think _any_ of the characters would be numbers. But this pattern should be fine

swagger/api_v2.spec.yaml
johnsonm325: This feels like more of a question for @jeromemarc. I agree with Asa that I use the search bar exclusively for display name searches. But what I don't know is if customers use it for id searches or not, or if Jerome wants the option to remain.

I personally have no problem not having it in v2.

swagger/api_v2.spec.yaml
kruai: Even though it's not important to distinguish `edge` hosts anymore, we may still have hosts with that system type. So, we may as well keep it until we hear otherwise.

swagger/api_v2.spec.yaml
kruai: OK, the new class should help us avoid that situation.

swagger/api_v2.spec.yaml
kruai: `format: uuid` requires the input to be in hyphenated UUID format, so `pattern` is redundant.

swagger/api_v2.spec.yaml
kruai: Right, I just meant that I've only seen the colon format in our host data. It's not worth blocking on, though, so I'm just going to resolve this one.

swagger/api_v2.spec.yaml
kruai: OK, thanks. In that case, can we please remove `hostnameOrId` for now? If Jerome really wants this filter, we can add it later.

swagger/api_v2.spec.yaml
kruai: json-api standard has support for `links` as an optional field; we comply with the standard whether we include it or not. I just wanted to make sure that we're making the intentional decision to add this to our responses because it will actually be useful, and not because we want to use the same pagination schema as resource-types (which we can't do).
I won't block on this because it's valid either way, but I do still want to hear back from UI devs about whether these links would in fact be useful to have in all of our paginated responses.

kruai (changes requested):

swagger/api_v2.spec.yaml
thearifismail: > This feels like more of a question for @jeromemarc. I agree with Asa that I use the search bar exclusively for display name searches. But what I don't know is if customers use it for id searches or not, or if Jerome wants the option to remain.
> 
> I personally have no problem not having it in v2.

`hostnameOrId` removed as a filter from V2.  Thanks @johnsonm325

kruai (approved):

rodrigonull (approved):

PR #4685, opened 2026-07-28
tests(RHINENG-27061): Use pre-configured RBAC accounts

fstavela:
## Jira
[RHINENG-27061](https://issues.redhat.com/browse/RHINENG-27061)

## What
Use pre-configured RBAC accounts in tests instead of changing the RBAC permissions before every test class

## Why
Modifying the permissions during the tests is a time expensive operation when RBAC v2 is enabled, because we have to wait 20s every time we do it to clear the cache. In the previous [PR](https://github.com/RedHatInsights/insights-host-inventory/pull/4576) I've created ansible playbooks which set up various test users, and in this PR I'm modifying the tests to start using them.

## How
- Add new staleness RBAC users to the ansible playbooks
- Refactor all RBAC tests to use a single package-scoped fixture for setting up hosts and workspaces
- Add the pre-configured users into the IQE config file
- Add new fixtures that create a testing `host_inventory` application that uses the pre-configured users
- Modify RBAC tests to use these fixtures instead of modifying the RBAC permissions in the test setup
- When running in ephemeral, set up the accounts at the beginning of the test session and wait 20s only once (`conftest.py`, `rbac_api.py`)

## Testing
I ran the modified tests against Stage, Prod, and Ephemeral environments with RBAC v1 and RBAC v2 accounts and all passed.

## PR Guidelines

You can find the documentation of the guidelines [here](https://redhat.atlassian.net/wiki/spaces/RHIN/pages/384311342/HBI+PR+Guideline)

## PR Guideline checks

- [ ] Keep PRs under 400 lines of meaningful changes, including tests, excluding auto-generated files, config, etc.

[RHINENG-27061]: https://redhat.atlassian.net/browse/RHINENG-27061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Refactor RBAC tests to use pre-configured RBAC accounts and shared fixtures while extending RBAC user configuration and Ansible setup to cover granular hosts, groups, and staleness scenarios across environments.

New Features:
- Introduce a comprehensive set of pre-configured RBAC test users for hosts, groups, staleness, and combined permissions in IQE configuration and Ansible RBAC setup.
- Add application fixtures that expose host-inventory clients authenticated as specific RBAC roles, enabling role-based test parametrization without modifying permissions at runtime.

Enhancements:
[...]

PR #4697, opened 2026-07-29
tests: Use both RBAC v2 feature flags

fstavela:
## What
Use both `platform.rbac.workspaces` and `hbi.rbac-v2` feature flags at the places where they should be used.

## Why
Before this PR, the IQE would check `hbi.rbac-v2` to know if it should use RBAC v2 write endpoints for setting up RBAC permissions. If that happens, we can't go back to using v1, even if we decide to lower the gradual rollout of the flag. Then, we wouldn't have a testing account for v1.

## How
- Use `platform.rbac.workspaces` in `rbac_workspaces_flag` method
- Add new methods for `hbi.rbac-v2`
- Start referencing the correct flags in other places in the IQE plugin

## Testing
This was tested against Prod, Stage, and ephemeral as part of this PR: https://github.com/RedHatInsights/insights-host-inventory/pull/4685

## PR Guidelines

## PR Guideline checks

## Summary by Sourcery

Align RBAC-related feature flag usage so workspace behavior and RBAC v2 authorization checks are controlled by their respective flags.

Enhancements:
- Use the platform.rbac.workspaces flag for workspace-related behavior instead of hbi.rbac-v2.
- Expose dedicated helpers for the hbi.rbac-v2 flag and its enabled state for RBAC v2 authorization checks.
- Update RBAC sync utilities to rely on the hbi.rbac-v2 flag when deciding whether to wait for Kessel synchronization.

Tests:
- Toggle the new hbi.rbac-v2 flag in test setup and remove an obsolete RBAC workspaces-enabled session fixture from group listing tests.

### DIFF
--- a/iqe-host-inventory-plugin/iqe_host_inventory/modeling/unleash.py
+++ b/iqe-host-inventory-plugin/iqe_host_inventory/modeling/unleash.py
@@ -67,7 +67,7 @@ class HBIUnleash(BaseEntity):
 
     @cached_property
     def rbac_workspaces_flag(self) -> str:
-        feature_flag = "hbi.rbac-v2"
+        feature_flag = "platform.rbac.workspaces"
 
         if isinstance(self._unleash, UnleashBackend) and not self._unleash.has_flag(feature_flag):
             flag_request = UnleashFlagRequest(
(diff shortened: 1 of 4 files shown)


kruai (approved):
LGTM!

PR #4709, opened 2026-07-30
fix: Fix System Profile filter examples on our API docs

kruai:
## Jira
None

## What
Fix the `filter_param` description to show correct filtering examples, and to align the URL params with the associated JSON examples.

## Why
Incorrect documentation misleads users. [This user](https://redhat-internal.slack.com/archives/C022YV4E0NA/p1785395571304989) ran into such an issue today.

## How
Update `filter_param.description`.

## Testing
N/A

## PR Guidelines

## PR Guideline checks

## Summary by Sourcery

Documentation:
- Update system profile filter examples in swagger docs to reflect the correct workloads.sap.sap_system URL structure and operating_system.rhel.version JSON/URL mappings.


swagger/api.spec.yaml
```diff
@@ -1556,13 +1556,13 @@ components:
         <br /><br />
         &nbsp;&nbsp;&nbsp;&nbsp;"?filter[system_profile][host_type][eq]=edge"
         <br /><br />
-        To get hosts with an specific operating system, use this explicit filter:
+        To get hosts with a specific operating system, use this explicit filter:
         <br /><br />
-        &nbsp;&nbsp;&nbsp;&nbsp;{"system_profile": {"operating_system": {"name": {"eq": "rhel"}}}}
```
rodrigonull, line 1561: This example is also valid, should we keep it as a second example?

fstavela (dismissed):

swagger/api.spec.yaml
kruai: Is it valid? It wasn't working for me - it fails because "name" isn't any of `RHEL`, `CentOS`, or `CentOS Linux`

rodrigonull: The `{"system_profile": {"operating_system": {"name": {"eq": "rhel"}}}}` filter should be valid.

The url example was wrong: `?filter[system_profile][name][eq]=rhel`

it should be: ` ?filter[system_profile][operating_system][name][eq]=rhel `

kruai: You're right, it's valid. I've updated it to include both versions

rodrigonull (approved):

PR #4730, opened 2026-08-04
feat(RHINENG-28674): Add V2 test infrastructure

thearifismail:
## Jira
[RHINENG-28674](https://issues.redhat.com/browse/RHINENG-28674)

## What
Add V2 API test infrastructure

## Why
To enable adding tests for testing the V2 API code.

## How
Created:

- test_api_v2_smoke.py (new)
- v2_api_fixture.py (new)
- v2_api_utils.py (new)
- conftest.py (added one line for v2_api_fixtures)
- app_fixutres.py (addd two lines for HBI_V2_API_ENABLED handling)

**Why V2 is enabled on the shared test app:** 
The test suite uses a single session-scoped Flask app because `db` (`SQLAlchemy()`) is a process-wide singleton — calling `create_app()` twice rebinds it via `db.init_app()` and corrupts database sessions for V1 tests. Enabling V2 on the shared app is safe since V2 routes (`/api/inventory/v2/...`) are namespaced and invisible to V1 tests. Tests that need to verify behavior with V2 *disabled* (e.g., 404 when the flag is off) should use a self-contained unit test with a mocked `create_app()`, following the existing `test_v2_api_registration` pattern.

## Testing
See the Accpetance Criteria in[RHINENG-28674](https://issues.redhat.com/browse/RHINENG-28674)
 

## PR Guidelines

## PR Guideline checks

[RHINENG-28674]: https://redhat.atlassian.net/browse/RHINENG-28674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Introduce initial test infrastructure for the V2 API and ensure the test app runs with V2 enabled.

Enhancements:
- Configure the test Flask app to run with the V2 API feature flag enabled via environment patching.
- Add shared V2 API URL constants and a helper for asserting RFC 7807-compliant error responses.

Tests:
- Add V2 API-specific fixtures for Flask app, client, and GET helper for exercising V2 endpoints.
- Add a smoke test that verifies the V2-enabled application starts and handles V2 base route requests without crashing.


thearifismail:
> Hey - I've left some high level feedback:
> 
>     * The `v2_flask_client` fixture is annotated as returning `starlette.testclient.TestClient` but actually returns `flask_app.test_client()`; consider updating the type hint to match the concrete Flask client type or use a protocol/interface if both clients should be supported.
> 
>     * In `v2_api_get`, `USER_IDENTITY` is used as a default dict argument; to avoid accidental mutations across tests, consider defaulting to `None` and assigning `USER_IDENTITY` inside the function when `identity` is `None`.

Both follow existing patterns — TestClient is correct for Connexion 3.x (see flask_client
  fixture), and identity=USER_IDENTITY matches all V1 fixtures (api_post, api_patch, etc.).

tests/fixtures/v2_api_fixtures.py
```diff
@@ -0,0 +1,39 @@
...
+
+from tests.helpers.api_utils import do_request
+from tests.helpers.db_utils import clean_tables
+from tests.helpers.test_utils import USER_IDENTITY
+
+
+@pytest.fixture(scope="function")
+def v2_flask_app(flask_app: FlaskApp) -> Generator[FlaskApp]:
+    yield flask_app
+    clean_tables()
```
rodrigonull: flask_app fixture already calls clean_tables(). You can remove this line

thearifismail: hmmm, this was removed. did I not push the change.  Let me do that. thank you!!!

thearifismail (commented):
@rodrigonull Please take a look now.

tests/fixtures/v2_api_fixtures.py
thearifismail: done!

rodrigonull (dismissed):

tests/test_api_v2_smoke.py
```diff
@@ -0,0 +1,8 @@
+from tests.helpers.v2_api_utils import V2_BASE_URL
+
+
+def test_v2_routing_does_not_crash(v2_api_get):
+    response_status, _ = v2_api_get(V2_BASE_URL)
+    # With paths: {} in the V2 spec, no V2 routes exist yet.
+    # A non-500 response proves the V2-enabled app starts and serves requests.
+    assert response_status != 500
```
kruai: Just FYI, we can remove this test after we have other v2 API tests, since it'll be implied by any API tests passing (not requesting a change right now ofc)

tests/fixtures/v2_api_fixtures.py
```diff
@@ -0,0 +1,37 @@
...
+        url: str,
+        identity: dict[str, Any] = USER_IDENTITY,
+        query_parameters: dict[str, Any] | None = None,
+        extra_headers: dict[str, Any] | None = None,
+    ) -> tuple[int, dict]:
+        return do_request(
+            v2_flask_client.get, url, identity, query_parameters=query_parameters, extra_headers=extra_headers
+        )
+
+    return _v2_api_get
```
kruai: Is any of this necessary? Why not just use the existing flask_app, flask_client, and api_get functions?
The only things that are changing are the endpoint paths, which you've covered with the constants in `v2_api_utils.py`.

kruai (changes requested):

thearifismail (commented):
@kruai @rodrigonull updated the code with Asa's comments.  PTAL. Thanks.

tests/test_api_v2_smoke.py
thearifismail: yes

tests/fixtures/v2_api_fixtures.py
thearifismail: v2 fixture removed

kruai (approved):

rodrigonull (approved):

PR #4751, opened 2026-08-07
feat(RHINENG-25830): Add character validation for view names

adarshdubey-star:
## Jira
[RHINENG-25830](https://issues.redhat.com/browse/RHINENG-25830)

## What
Add character validation for inventory view names on create and update endpoints.

## Why
View names need consistent character restrictions to prevent broken UI rendering (control characters), confusing names (punctuation-only), and to align with the existing Group name validation rules across the product.

## How
- Added `^[a-zA-Z0-9 _-]+$` regex validation to `InputViewSchema` and `PatchViewSchema` using `marshmallow_validate.And()`, matching the existing `GROUP_NAME_PATTERN`.
- Updated OpenAPI spec (`api.spec.yaml` and `openapi.json`) with the `pattern` constraint on `ViewIn` and `ViewPatch` name fields.
- Added 5 new API tests: special characters, control characters, punctuation-only, and valid hyphen/underscore names for both POST and PATCH.

## Testing
- Schema validation verified directly via Python (valid/invalid name matrix).
- Curl tests against local running instance: all 12 scenarios pass (valid names return 201/200, invalid names return 400 with clear error message).
- Ruff lint and format checks pass.

## PR Guidelines

## PR Guideline checks

[RHINENG-25830]: https://redhat.atlassian.net/browse/RHINENG-25830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Add stricter validation rules for inventory view names across create and update APIs to prevent invalid or confusing names and align with existing product naming constraints.

Enhancements:
- Introduce a shared view name validation schema enforcing allowed characters and length for both create and patch view payloads.
- Document view name character restrictions in the OpenAPI specification via updated descriptions and regex patterns for relevant fields.

Tests:
- Add API tests covering rejection of special characters, control characters, and punctuation-only view names, plus acceptance of names with hyphens and underscores for both POST and PATCH endpoints.


computercamplove (dismissed):

fstavela (changes requested):
`test_creates_view_with_system_profile_os_filter` fails, because it has dots in the name. This test needs to be updated.

Other than that LGTM. Sourcery is right that `VIEW_NAME_PATTERN` and `GROUP_NAME_PATTERN` are identical, so it might be good to reuse the same constant in both places, but I'm also OK with keeping them separate.

ezr-ondrej (dismissed):
We could possibly improve it so we don't allow '   ' (just spaces), but that's a minor thing :)
`^[a-zA-Z0-9]+[a-zA-Z0-9 _-]*$` might be slightly better (still doesn't solve for not ending with space, but that would be uglier regexp.

I'm fine with keeping as is tho. Thank you @adarshdubey-star 🙏

adarshdubey-star:
> `test_creates_view_with_system_profile_os_filter` fails, because it has spaces in the name. This test needs to be updated.
> 
> Other than that LGTM. Sourcery is right that `VIEW_NAME_PATTERN` and `GROUP_NAME_PATTERN` are identical, so it might be good to reuse the same constant in both places, but I'm also OK with keeping them separate.

I also felt the same here when I thought of it , but it's just a single line so kept it as it is.

adarshdubey-star:
> We could possibly improve it so we don't allow ' ' (just spaces), but that's a minor thing :) `^[a-zA-Z0-9]+[a-zA-Z0-9 _-]*$` might be slightly better (still doesn't solve for not ending with space, but that would be uglier regexp.
> 
> I'm fine with keeping as is tho. Thank you @adarshdubey-star 🙏

Yup I agree this feels more than enough to me.

fstavela:
> We could possibly improve it so we don't allow ' ' (just spaces), but that's a minor thing :) `^[a-zA-Z0-9]+[a-zA-Z0-9 _-]*$` might be slightly better (still doesn't solve for not ending with space, but that would be uglier regexp.
> 
> I'm fine with keeping as is tho. Thank you @adarshdubey-star 🙏

There is a `@pre_load` method that strips the whitespaces from the beginning and end of the name, so if someone tries to use just whitespaces, it would end up being an empty string, which would not be accepted.

fstavela (approved):
LGTM, thank you!

PR #4780, opened 2026-08-11
fix(RHINENG-29751): flush event messages synchronously

rodrigonull:
## Jira
[RHINENG-29751](https://issues.redhat.com/browse/RHINENG-29751)

## What
Switch create/update event production to synchronous flush (`wait=True`) and remove the now-redundant batch flush and deferred span handling from `write_message_batch`.

## Why
ADR-0002 (event-based system attributes replication) requires the producer to use synchronous sends to the broker for all events on `platform.inventory.events`. Delete events already complied (`wait=True`), but create/update events were produced with `wait=False` and only flushed once at the end of the batch. This created a window where a delete event could reach the broker before a buffered update for the same host, breaking the ordering guarantee that downstream consumers (Compliance, Vulnerability, Advisor) depend on after the Cyndi migration.

## How
- Changed `wait=False` to `wait=True` in `write_add_update_event_message()` so each create/update event is individually flushed to the broker before proceeding.
- Removed the trailing `event_producer.flush()` call in `write_message_batch()` since each event is now flushed per-message.
- Removed the `_end_deferred_span()` helper function, which was only needed for the batched post-processing path. Span lifecycle is now handled by `_end_all_deferred_spans()` in `_run_batch()`.

## Testing
- All 290 MQ-related unit tests pass (`test_host_mq_service.py`, `test_mq_message_spans.py`, `test_notifications.py`, `test_outbox_end_to_end_cases.py`).
- Updated `test_write_message_batch_flushes_once` → `test_write_message_batch_produces_all_messages` to assert flush is no longer called at the batch level.
- Updated `test_end_deferred_span_enriches_with_threadctx` → `test_end_all_deferred_spans_enriches_with_threadctx` to test enrichment through the remaining code path.
- End-to-end trace verified locally via Tempo/Grafana: uploaded an archive through ingress → puptoo → hbi-mq → events topic → insights-engine/vulnerability. Confirmed `traceparent` propagation and span parenting are intact.
- `make style` passes (ruff, mypy, pre-commit hooks).

## PR Guidelines

## PR Guideline checks

[RHINENG-29751]: https://redhat.atlassian.net/browse/RHINENG-29751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Flush create and update host events synchronously and streamline MQ span handling, updating tests to align with the new event production behavior.

Bug Fixes:
- Ensure create and update host events are flushed synchronously to the broker to maintain ordering guarantees with delete events.

Enhancements:
- Simplify MQ span lifecycle handling by removing per-message deferred span ending in batch processing.

Tests:
- Adjust MQ tests to reflect per-message flushing and validate span enrichment via the remaining _end_all_deferred_spans path.

### DIFF
--- a/app/queue/host_mq.py
+++ b/app/queue/host_mq.py
@@ -1351,7 +1351,7 @@ def write_add_update_event_message(
             tracker_ctx._success_status_msg = "skipped – host deleted before event production"
             return
 
-        event_producer.write_event(event, str(result.row.id), headers, wait=False)
+        event_producer.write_event(event, str(result.row.id), headers, wait=True)
 
         if result.event_type.name == HOST_EVENT_TYPE_CREATED:
             # Notifications are expected to omit null canonical facts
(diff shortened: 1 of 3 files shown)


kruai (approved):
I'm fine with this change, but note that this will impact performance. We need to keep an eye on our MQ metrics after we release this to Prod, especially during the nightly upload window.

PR #4743, opened 2026-08-06
test(RHINENG-25835): Add IQE happy-path tests for Inventory Views CRUD

adarshdubey-star:
## Jira
[RHINENG-25835](https://issues.redhat.com/browse/RHINENG-25835)

## What
Add IQE happy-path integration tests for the Inventory Views Management CRUD API (`/beta/views`), along with the `ViewsAPIWrapper` modeling layer and cleanup support needed to run them.

## Why
The Views CRUD endpoints (list, get, create, update, delete, clone) are now implemented and merged. We need IQE integration tests to validate these endpoints in deployed environments (Stage/Prod/ephemeral) and establish a regression safety net.

## How
- Created `ViewsAPIWrapper` in `modeling/views_api.py` wrapping the auto-generated `ViewsApi` client with high-level methods for create, get, update, delete, and clone.
- Registered the wrapper in `HBIApis` and added `view_ids_to_clean` / `clean_views()` to `HBICleanUp` for automatic test cleanup.
- Added 15 happy-path IQE tests in `tests/rest/views/test_views_crud.py` covering:
  - **List** — paginated response, created view appears in list
  - **Get by ID** — fetch by UUID, 404 for nonexistent
  - **Create** — with full configuration, validation rejects invalid column key
  - **Update (PATCH)** — name update preserves other fields, configuration update
  - **Delete** — delete + confirm 404, 404 for nonexistent
  - **Clone** — "Copy of" prefix, config preserved, 404 for nonexistent, independent modification
  - **Cross-org isolation** — view not visible to another org

## Testing
- All 15 IQE test scenarios written following existing patterns (groups, host_views).
- Ruff lint and format checks pass.
- Syntax verified via `py_compile`.

## PR Guidelines

You can find the documentation of the guidelines [here](https://redhat.atlassian.net/wiki/spaces/RHIN/pages/384311342/HBI+PR+Guideline)

## PR Guideline checks

- [ ] Keep PRs under 400 lines of meaningful changes, including tests, excluding auto-generated files, config, etc.

## Summary by Sourcery

Add integration support for inventory views CRUD and align the views update endpoint with PATCH semantics.

Enhancements:
- Introduce a ViewsAPIWrapper modeling layer with high-level helpers for listing, creating, updating, deleting, and cloning views, plus bulk delete and cleanup integration.

Tests:
- Add happy-path IQE tests for views list, get-by-id, create, patch update, delete, clone, and cross-org isolation on the /beta/views API.

PR #4798, opened 2026-08-13
refactor(iqe): migrate ViewsAPIWrapper to BaseAPIWrapper

adarshdubey-star:
Follow up for : https://github.com/RedHatInsights/insights-host-inventory/pull/4743#pullrequestreview-4904930286

## Summary by Sourcery

Refactor the views API modeling to use the shared BaseAPIWrapper and JSON-based responses while preserving test coverage and cleanup behavior.

Enhancements:
- Replace the OpenAPI-generated ViewsApi client in ViewsAPIWrapper with a BaseAPIWrapper that issues raw HTTP requests to /beta/views endpoints.
- Change ViewsAPIWrapper methods to work with plain JSON dictionaries and requests.Response objects instead of typed ViewOut models, including new helpers for raw POST/PATCH/DELETE calls.
- Update cleanup utilities to accept generic iterables of view IDs for more flexible bulk operations.

Tests:
- Adapt views CRUD tests to the new JSON-based API wrapper, asserting on dict fields and HTTP status codes using requests exceptions instead of custom raises_apierror helpers.


iqe-host-inventory-plugin/iqe_host_inventory/modeling/views_api.py
```diff
@@ -163,58 +160,86 @@ def update_view(
...
+        """
+        return self._base_wrapper.patch(path, **kwargs)
+
+    def delete_raw(self, path: str, **kwargs: Any) -> requests.Response:
+        """Send a raw DELETE request for negative validation testing.
+
+        :param str path: API path appended to the base URL (e.g. ``/beta/views/{id}``)
+        :return requests.Response: Raw response (no raise_for_status)
+        """
+        return self._base_wrapper.delete(path, **kwargs)
```
fstavela: These methods don't really do anything more than the BaseAPIWrapper does. We could either make the `_base_wrapper` public and let the tests use it directly, or we could let these "raw" methods do something for us, for example at least set the path, so that the called doesn't have to. Or, we could even delete these methods completely and use `with pytest.raises...` in the tests with the "non-raw" methods.

adarshdubey-star: I removed these methods and replaced it with pytes.raises instead.

fstavela (approved):
LGTM, thank you!

PR #4841, opened 2026-08-24
feat(RHINENG-30193): Add missing column keys to views validation allowlist

adarshdubey-star:
## Jira
[RHINENG-30193](https://issues.redhat.com/browse/RHINENG-30193)

## What
Add missing core inventory column keys to the views validation allowlist so the frontend can save views that include `tags`, `status`, `infrastructure`, `vendor`, `workload`, and `created`.

## Why
The frontend was failing to save view configurations with errors like `Invalid column key 'status'` and `Invalid column key 'tags'`. `VALID_CORE_VIEW_FIELDS` only allowed 5 core keys, but the SystemsView column catalogue uses 11. These keys are display metadata stored in the view config JSONB — the backend does not use them to fetch data.

## How
Expanded `VALID_CORE_VIEW_FIELDS` in `api/views_validation.py` with the 6 missing keys from the frontend column catalogue. No sort allowlist changes — the frontend maps `status` sorting to `updated` internally and does not send `status` as an `orderBy` to the backend.

## Testing
- Added `test_creates_view_with_all_core_columns` covering all 11 core column keys in a single POST
- `uv run pytest tests/test_api_views.py` — 85 passed
- `uv run pytest tests/test_views_repository.py` — 34 passed
- Existing invalid-column-key rejection tests still pass (`nonexistent_field`, `advisor:nonexistent`, etc.)

## PR Guidelines

## PR Guideline checks

[RHINENG-30193]: https://redhat.atlassian.net/browse/RHINENG-30193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Expand the views validation allowlist to accept the complete set of core inventory column keys.

Bug Fixes:
- Allow view configurations to be saved when they include the missing core inventory columns: tags, status, infrastructure, vendor, workload, and created.

Tests:
- Add coverage for creating a view containing all 11 supported core columns.


kruai (dismissed):

fstavela (dismissed):
The code changes LGTM, however, I have a question about the general direction and architecture.
You mentioned that we are adding these frontend columns and that the frontend will map them to the backend fields when requesting the hosts. Is this the right approach? Shouldn't the backend store information about backend columns? This creates a tighter coupling between the backend and frontend microservices. This doesn't have to mean that there is an issue, but I just want to make sure this was considered.
An even higher-level question is if frontend should even need to request the hosts with all the filters and field parameters every time it wants to get hosts based on a specific view. If the backend stored backend columns in the view configuration, then we could have a `GET /hosts-view/<view-id>` endpoint which would return all hosts data based on a specific view and frontend wouldn't have to map all the filters. Also, then the views would also be usable just via backend, for example by some scripts.
But I'm just thinking out loud to make sure that these things were considered at some point.

adarshdubey-star:
> The code changes LGTM, however, I have a question about the general direction and architecture. You mentioned that we are adding these frontend columns and that the frontend will map them to the backend fields when requesting the hosts. Is this the right approach? Shouldn't the backend store information about backend columns? This creates a tighter coupling between the backend and frontend microservices. This doesn't have to mean that there is an issue, but I just want to make sure this was considered. An even higher-level question is if frontend should even need to request the hosts with all the filters and field parameters every time it wants to get hosts based on a specific view. If the backend stored backend columns in the view configuration, then we could have a `GET /hosts-view/<view-id>` endpoint which would return all hosts data based on a specific view and frontend wouldn't have to map all the filters. Also, then the views would also be usable just via backend, for example by some scripts. But I'm just thinking out loud to make sure that these things were considered at some point.

@fstavela The view configuration is a UI layout descriptor it tells the frontend which columns to render and in what order, not how to query the backend. Keys like tags and status are display concerns (status isn't even a stored field, it's computed from modified_on). The backend validates them to prevent garbage from being saved, but the actual mapping from display key to API params lives in the frontend, which is where that knowledge belongs.

If we stored backend-native column keys instead, the view config would leak backend implementation details and the frontend would still need a mapping layer to know which columns to show in the UI. So the coupling would shift, not reduce.

On the GET /hosts-view/<view-id> idea, I actually think that's worth exploring as a future enhancement. It would make views usable from scripts and automation without having to reconstruct the query. But it would also mean the backend needs to understand every column/filter key and how to translate it into a query, which is a bigger scope change. For now the frontend already handles this translation, so the current approach works

api/views_validation.py
```diff
@@ -10,11 +10,17 @@
 
 VALID_CORE_VIEW_FIELDS = frozenset(
     {
-        "updated",
+        "created",
         "display_name",
         "group_name",
```
computercamplove, line 15: should it be `workspace_name`?

adarshdubey-star: workspace_name is the filter key, not the column key

kruai (approved):

PR #4859, opened 2026-08-26
fix: Catch export-service "resource already processed" responses

kruai:
## Jira
None

## What
Catch export-service's "already been processed" responses, even when the response status isn't HTTP 400

## Why
export-service's "already been processed" responses don't have HTTP 400 response status, so they're not being correctly handled; these are falling through as exceptions, and leading to alerts triggering more frequently. I'm not sure if the response code changed recently, or if it was never HTTP 400 to begin with; either way, this change makes it catch these responses more broadly.

## How
Remove the `response.status_code` condition from the check

## Testing
Update `test_already_processed_does_not_raise` with parametrized status codes

## PR Guidelines

## PR Guideline checks

## Summary by Sourcery

Handle export-service already-processed responses independently of their HTTP status code.

Bug Fixes:
- Treat export-service responses indicating that a resource has already been processed as successful regardless of the HTTP status code, preventing duplicate deliveries from raising exceptions and triggering unnecessary alerts.

Tests:
- Parameterize already-processed response coverage across bad request, conflict, and internal server error status codes.


rodrigonull (approved):

tests/test_export_service.py
```diff
@@ -323,9 +323,13 @@ class TestHandleExportResponse:
     def test_accepted_response(self):
         _handle_export_response(_make_response(HTTPStatus.ACCEPTED, "payload delivered"), uuid4(), "json")
 
-    def test_already_processed_does_not_raise(self):
+    @pytest.mark.parametrize(
+        "status_code",
+        [HTTPStatus.BAD_REQUEST, HTTPStatus.CONFLICT, HTTPStatus.INTERNAL_SERVER_ERROR],
```
fstavela, line 328: Just checking, what is the response code that we see in the logs now? Are these all that can happen? Should we add OK and CREATED as well?

tests/test_export_service.py
```diff
@@ -323,9 +323,13 @@ class TestHandleExportResponse:
...
+    )
+    def test_already_processed_does_not_raise(self, status_code):
         resp = _make_response(
-            HTTPStatus.BAD_REQUEST,
+            status_code,
             '{"detail": "this resource has already been processed", "status": 400}',
```
fstavela, line 333:
```suggestion
            f'{"detail": "this resource has already been processed", "status": {status_code}}',
```

fstavela (commented):
Just a couple of comments on the tests

tests/test_export_service.py
kruai: We don't currently see what the response code was; we just know that it wasn't ACCEPTED or BAD_REQUEST. We can test for OK and CREATED as well; doesn't matter as long as it has "already been processed" in the response.

PR #4874, opened 2026-08-31
feat(RHINENG-30304): add user_view_preferences table for default view

adarshdubey-star:
## Jira
[RHINENG-30304](https://issues.redhat.com/browse/RHINENG-30304)

## What
Add `hbi.user_view_preferences` table, SQLAlchemy model, and tests for per-user default view pinning.

## Why
Foundation for the default view per user feature. Users currently must manually switch to their preferred view every time they open Inventory. This table stores which view each user has pinned as their default.

## How
- Alembic migration creates `user_view_preferences` with composite PK `(org_id, user_id)`, FK to `inventory_views.id` with `ON DELETE CASCADE`, and an index on `default_view_id`.
- `UserViewPreference` model added to `app/models/views.py` and registered in `app/models/__init__.py`.
- Test fixture and 5 tests covering: basic CRUD, PK uniqueness, FK constraint, cascade delete.

## Testing
- `make style` — all checks pass
- `pytest tests/test_views_repository.py` — 39/39 pass (5 new)
- `pytest tests/test_api_views.py` — 84/84 pass (no regressions)

## PR Guidelines

## PR Guideline checks

[RHINENG-30304]: https://redhat.atlassian.net/browse/RHINENG-30304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Add persistent per-user default view preferences to support pinned Inventory views.

New Features:
- Add per-user default inventory view preferences scoped by organization and user.

Enhancements:
- Expose the user view preference model for application use and enforce view relationship integrity with cascading cleanup.

Tests:
- Add coverage for preference creation, uniqueness, foreign-key validation, multi-user defaults, and cascade deletion.


tests/test_views_repository.py
```diff
@@ -303,3 +305,56 @@ def test_system_view_serialization(self, db_create_system_view):
...
+
+    def test_cascade_deletes_preference_when_view_deleted(self, db_create_view, db_create_user_view_preference):
+        view = db_create_view(org_id=ORG_ID, created_by=USER_ID)
+        db_create_user_view_preference(ORG_ID, USER_ID, view.id)
+
+        db.session.delete(view)
+        db.session.commit()
+
+        pref = UserViewPreference.query.filter_by(org_id=ORG_ID, user_id=USER_ID).one_or_none()
+        assert pref is None
```
computercamplove, line 360: @adarshdubey-star  when the default view is removed, should we fall back to "All systems" as the default?
When a user hasn't set a custom view as their default yet, i think "All systems" should act as the default.

I think there should always be exactly one default, because the UI always needs a view to load - the Systems page depends on a view to load the systems table, so we should fall back to the default one.

computercamplove: or maybe from tests i missed somehow the fallback part

adarshdubey-star: Yes, the fallback to "All systems" is planned. This PR is just the foundation: the table, model, and constraint tests. The fallback logic comes in the upcoming PR's

computercamplove: thank you for clarification!

fstavela (approved):
LGTM, thank you!

PR #4892, opened 2026-09-02
feat(RHINENG-28021): apply saved view filters to export 

rodrigonull:
## Jira
[RHINENG-28021](https://issues.redhat.com/browse/RHINENG-28021)

## What
Apply the active Inventory View's saved filters to the export query, so exports reflect the same host subset the user sees in the UI.

## Why
Phase 3 of Inventory Views requires exports to be "view-aware." This is the first step: when the frontend sends a `view_id` in the export request filters, the backend loads the saved view and applies its filters to the host query. Dirty (unsaved) view state is intentionally deferred.

## How
- Extract `view_id` from the Kafka export message's `filters` dict in `create_export()`.
- Added `_load_view_config()` to load an `InventoryView` and split its configuration into host-level filters (staleness, tags, date ranges, etc.) and app-data/system_profile filters.
- Added `_get_user_id()` to extract the user ID from User or ServiceAccount identities, needed for view ownership lookup.
- Passed both filter dicts through `_non_empty_hosts_iter()` → `get_hosts_to_export()` → `query_filters()`.
- The view's staleness setting overrides the default "all staleness states" behavior.
- Error handling: missing `user_id` → 403, view not found → 404, permission denied → 403, all reported back to the export service.

## Testing
- Added `TestLoadViewConfig` (3 tests): verifies correct splitting of host vs. app-data filters and view-not-found error.
- Added `TestCreateExportWithView` (4 tests): end-to-end tests for export with valid view, nonexistent view, missing user_id, and backward compatibility without view_id.
- All 48 tests pass (existing + new). Ruff, mypy, and pre-commit hooks pass.

## PR Guidelines

## PR Guideline checks

[RHINENG-28021]: https://redhat.atlassian.net/browse/RHINENG-28021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Make inventory exports honor the filters from the selected saved view.

New Features:
- Apply saved Inventory View filters to host exports when a view ID is included in the export request.
- Report clear authorization and not-found errors when view-aware exports cannot load the requested view.

Enhancements:
- Centralize identity user ID access and separate host-level filters from query filters for export processing.
- Preserve export behavior for requests without a view ID.

Tests:
- Add coverage for saved-view filter handling, workspace normalization, missing views, missing user IDs, and backward-compatible exports.


adarshdubey-star (approved):

app/auth/identity.py
```diff
@@ -114,6 +114,14 @@ def _asdict(self):
...
             return ident
 
+    @property
+    def user_id(self) -> str | None:
+        if self.identity_type == IdentityType.USER:
+            return getattr(self, "user", {}).get("user_id")
+        if self.identity_type == IdentityType.SERVICE_ACCOUNT:
+            return getattr(self, "service_account", {}).get("user_id")
+        return None
+
```
kruai, line 124: Ayyyy I appreciate this logic consolidation :+1:

kruai (approved):
Overall LGTM. Just one question, though...
This implementation requires the view to have been saved so we can retrieve it. This restriction is fine, but what about if a user makes changes to their current view without saving, and then hits Export? I expect that it would export based on the view without the pending changes. If that's the case, I want to make sure that's acceptable requirements-wise.

rodrigonull:
> Overall LGTM. Just one question, though... This implementation requires the view to have been saved so we can retrieve it. This restriction is fine, but what about if a user makes changes to their current view without saving, and then hits Export? I expect that it would export based on the view without the pending changes. If that's the case, I want to make sure that's acceptable requirements-wise.

This was an intentional scoping decision we agreed on with @ezr-ondrej. Start with saved-state-only exports, get that working and validated, and then tackle dirty view state as a follow-up. The Kafka export message contract already supports this future extension where the filters dict can carry inline filter/column overrides alongside the view_id when we're ready to implement it.

PR #4903, opened 2026-09-04
feat(30305):Repository layer for default view preferences

adarshdubey-star:
## Jira
[RHINENG-30305](https://issues.redhat.com/browse/RHINENG-xxxx)

## What
Add three functions in lib/views_repository.py:

get_default_view_id(org_id, user_id)  returns the user's pinned default view UUID. If no preference exists, or the pinned view is no longer visible, falls back to the "All systems" system view UUID. System default is cached in a module-level variable after first lookup.

set_default_view(org_id, user_id, view_id) validates the view exists and is visible to the user, then upserts into user_view_preferences. Returns the view object. Raises 404 if not visible.

delete_default_view(org_id, user_id) deletes the preference row. Idempotent (no error if row doesn't exist).

## Why
For the API endpoints work.

## Testing
Added test scenarios.

## PR Guidelines

## PR Guideline checks

[RHINENG-30305]: https://redhat.atlassian.net/browse/RHINENG-30305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Implement repository support for managing user default view preferences.

New Features:
- Add repository operations for retrieving, pinning, and removing users’ default inventory views.
- Fall back to the seeded “All systems” view when no valid visible preference exists.

Enhancements:
- Cache the system default view lookup and enforce visibility when setting a default view.

Tests:
- Add coverage for default-view fallback, caching, visibility validation, upsert behavior, and idempotent deletion.


lib/views_repository.py
```diff
@@ -154,3 +187,61 @@ def clone_view(view_id: str, org_id: str, user_id: str) -> InventoryView:
...
+
+    visible = InventoryView.query.filter(
+        InventoryView.id == pref.default_view_id,
+        _visibility_filter(org_id, user_id),
+    ).one_or_none()
+
+    if visible is None:
+        return _get_system_default_view_id()
+
+    return str(pref.default_view_id)
```
rodrigonull: Here we don't need a redundant 2-query round trip and full ORM hydration. We can simplify it to:

```suggestion
def get_default_view_id(org_id: str, user_id: str) -> str:
    """Return the user's pinned default view ID, or fall back to 'All systems'.

    If a preference exists but the pinned view is no longer visible to the user,
    falls back to the system default.
    """
    view_id = (
        db.session.query(UserViewPreference.default_view_id)
        .join(
            InventoryView,
            InventoryView.id == UserViewPreference.default_view_id,
        )
        .filter(
            UserViewPreference.org_id == org_id,
            UserViewPreference.user_id == user_id,
            _visibility_filter(org_id, user_id),
        )
        .scalar()
    )
    if view_id is not None:
        return str(view_id)

    return _get_system_default_view_id()
```

lib/views_repository.py
```diff
@@ -21,9 +24,39 @@
...
+            InventoryView.name == ALL_SYSTEMS_VIEW_NAME,
+        ).one_or_none()
+        if view is None:
+            raise InventoryException(
+                status=500,
+                title="Internal Server Error",
+                detail=f"System default view '{ALL_SYSTEMS_VIEW_NAME}' not found.",
+            )
+        _SYSTEM_DEFAULT_VIEW_ID = str(view.id)
+    return _SYSTEM_DEFAULT_VIEW_ID
```
rodrigonull: Let's use python built-in cache instead of a global var:

```suggestion
@cache
def _get_system_default_view_id() -> str:
    """Return the UUID of the seeded 'All systems' system view (cached after first lookup)."""
    view_id = (
        db.session.query(InventoryView.id)
        .filter(
            InventoryView.org_id.is_(None),
            InventoryView.name == ALL_SYSTEMS_VIEW_NAME,
        )
        .scalar()
    )
    if view_id is None:
        raise InventoryException(
            status=500,
            title="Internal Server Error",
            detail=f"System default view '{ALL_SYSTEMS_VIEW_NAME}' not found.",
        )
    return str(view_id)
```

lib/views_repository.py
```diff
@@ -154,3 +187,61 @@ def clone_view(view_id: str, org_id: str, user_id: str) -> InventoryView:
...
+def delete_default_view(org_id: str, user_id: str) -> None:
+    """Unpin the user's default view. Idempotent if no preference exists."""
+    pref = UserViewPreference.query.filter_by(org_id=org_id, user_id=user_id).one_or_none()
+    if pref is None:
+        return
+
+    with session_guard(db.session):
+        db.session.delete(pref)
+
+    logger.info("Deleted default view preference for org %s user %s", org_id, user_id)
```
rodrigonull: Let's avoid unnecessary SELECT before DELETE here:
```suggestion
def delete_default_view(org_id: str, user_id: str) -> None:
    """Unpin the user's default view. Idempotent if no preference exists."""
    with session_guard(db.session):
        deleted = UserViewPreference.query.filter_by(org_id=org_id, user_id=user_id).delete()

    if deleted:
        logger.info("Deleted default view preference for org %s user %s", org_id, user_id)
```

lib/views_repository.py
```diff
@@ -154,3 +187,61 @@ def clone_view(view_id: str, org_id: str, user_id: str) -> InventoryView:
...
+        org_id=org_id,
+        user_id=user_id,
+        default_view_id=view.id,
+        updated_on=_time_now(),
+    )
+    stmt = stmt.on_conflict_do_update(
+        index_elements=["org_id", "user_id"],
+        set_={
+            "default_view_id": stmt.excluded.default_view_id,
+            "updated_on": _time_now(),
```
rodrigonull: Calling `_time_now()` twice generates slightly different timestamps for insert vs update definitions within the same statement. In PostgreSQL ON CONFLICT DO UPDATE, referencing `stmt.excluded.updated_on` is both more idiomatic and guarantees timestamp consistency.

```suggestion
            "updated_on": stmt.excluded.updated_on,
```

lib/views_repository.py
```diff
@@ -21,9 +24,39 @@
...
     "clone_view",
+    "get_default_view_id",
+    "set_default_view",
+    "delete_default_view",
 )
 
 CLONE_NAME_PREFIX = "Copy of "
+ALL_SYSTEMS_VIEW_NAME = "All systems"
+
+_SYSTEM_DEFAULT_VIEW_ID: str | None = None
```
rodrigonull: This can be removed if we use `@functools.cache`

rodrigonull (changes requested):

tests/test_views_repository.py
```diff
@@ -421,3 +426,133 @@ def test_cascade_deletes_preference_when_view_deleted(self, db_create_view, db_c
...
         pref = UserViewPreference.query.filter_by(org_id=ORG_ID, user_id=USER_ID).one_or_none()
         assert pref is None
+
+
+class TestGetDefaultViewId:
+    @pytest.fixture(autouse=True)
+    def _clear_cache(self):
+        _clear_system_default_view_cache()
+        yield
+        _clear_system_default_view_cache()
```
rodrigonull: I would move this fixture to be a module level fixture, so it would be used for all tests in this file and we would avoid duplicates.

tests/test_views_repository.py
```diff
@@ -421,3 +426,133 @@ def test_cascade_deletes_preference_when_view_deleted(self, db_create_view, db_c
 
         pref = UserViewPreference.query.filter_by(org_id=ORG_ID, user_id=USER_ID).one_or_none()
         assert pref is None
+
+
+class TestGetDefaultViewId:
+    @pytest.fixture(autouse=True)
+    def _clear_cache(self):
+        _clear_system_default_view_cache()
```
rodrigonull:
```suggestion
        _get_system_default_view_id.cache_clear()
```

tests/test_views_repository.py
rodrigonull:
```suggestion
       _get_system_default_view_id.cache_clear()
```

lib/views_repository.py
```diff
@@ -21,9 +25,38 @@
...
+    "delete_default_view",
 )
 
 CLONE_NAME_PREFIX = "Copy of "
+ALL_SYSTEMS_VIEW_NAME = "All systems"
+
+
+def _clear_system_default_view_cache() -> None:
+    """Reset the cached All systems view ID. Intended for tests."""
+    _get_system_default_view_id.cache_clear()
```
rodrigonull: This is just a wrapper function used in the tests. It should be removed.

rodrigonull (changes requested):

kruai (commented):
How will `get_default_view_id` be used? Are we going to use that to get the ID, and then immediately fetch the InventoryView by ID? Just making sure we actually need the ID here rather than the view itself, so we don't end up double-querying when we don't need to.

adarshdubey-star:
> How will `get_default_view_id` be used? Are we going to use that to get the ID, and then immediately fetch the InventoryView by ID? Just making sure we actually need the ID here rather than the view itself, so we don't end up double-querying when we don't need to.

@kruai The plan is to return only the default view ID (embedded in GET /views), not the full view object. The frontend already has the view list from that response and just needs which ID is pinned, so we don’t fetch the InventoryView again no double-query.

This PR is the repository helpers only. Wiring the ID into GET /views and the pin/unpin endpoints will come in a follow-up.

lib/views_repository.py
```diff
@@ -21,9 +25,33 @@
...
+            InventoryView.name == ALL_SYSTEMS_VIEW_NAME,
+        )
+        .scalar()
+    )
+    if view_id is None:
+        raise InventoryException(
+            status=500,
+            title="Internal Server Error",
+            detail=f"System default view '{ALL_SYSTEMS_VIEW_NAME}' not found.",
+        )
```
kruai, line 56: If we're going to return HTTP 500 here, can we add a failure metric that increments when this happens? This situation occurring implies something is broken that needs to be fixed ASAP.

kruai:
> > How will `get_default_view_id` be used? Are we going to use that to get the ID, and then immediately fetch the InventoryView by ID? Just making sure we actually need the ID here rather than the view itself, so we don't end up double-querying when we don't need to.
> 
> @kruai The plan is to return only the default view ID (embedded in GET /views), not the full view object. The frontend already has the view list from that response and just needs which ID is pinned, so we don’t fetch the InventoryView again no double-query.
> 
> This PR is the repository helpers only. Wiring the ID into GET /views and the pin/unpin endpoints will come in a follow-up.

OK, thanks. Just wanted to make sure the front-end wasn't querying the ID here, only to then fetch the view using that ID (since we could just do that in one trip rather than two).

kruai (changes requested):
Looks good, I just want to make sure we have a failure metric for the HTTP 500 condition for alerting purposes.

lib/views_repository.py
adarshdubey-star: Sure , I'll add this

kruai (approved):
LGTM, thank you!

rodrigonull (approved):

PR #4906, opened 2026-09-04
test(RHINENG-30477): Fix failing tests in Stage/Prod

fstavela:
## Jira
[RHINENG-30477](https://issues.redhat.com/browse/RHINENG-30477)

## What
Fix tests that are failing in Stage and Prod pipelines

## Why
The pipelines are unreliable now

## How
Various little fixes

## Testing
All of the updated tests passed in Stage and Prod when I ran them locally

## PR Guidelines

## PR Guideline checks

[RHINENG-30477]: https://redhat.atlassian.net/browse/RHINENG-30477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Stabilize integration test coverage across Stage and Prod by aligning assertions and synchronization behavior with environment-specific service responses.

Bug Fixes:
- Stabilize Stage and Prod integration tests by accommodating environment-specific workspace responses, allowing asynchronous Kessel synchronization, and skipping unsupported legacy certificate paths.

Enhancements:
- Reduce test fixture sizes and improve test setup consistency, typing, and environment-aware assertions.
- Mark RBAC and Kessel-dependent tests appropriately for ephemeral environments and update host creation and synchronization handling.

Tests:
- Update group, resource type, Kessel replication, RBAC, and legacy path tests to reflect current Stage and Prod behavior and reduce pipeline flakiness.

### DIFF
--- a/iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/kessel/test_kessel_host_replication.py
+++ b/iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/kessel/test_kessel_host_replication.py
@@ -320,6 +320,10 @@ def test_kessel_repl_delete_hosts_by_id(
     for host in hosts:
         hbi_kessel_relations_grpc.verify_created_or_updated(host, ungrouped_group)
 
+    if host_inventory.application.config.current_env.lower() != "clowder_smoke":
+        logger.info("Waiting 10 seconds to work around hosts availability flapping issue")
+        sleep(10)
+
     host_inventory.apis.hosts.delete_by_id(hosts[:host_count])
 
     for host in hosts[:host_count]:
@@ -358,6 +362,10 @@ def test_kessel_repl_delete_hosts_by_filter(
 
     for host in hosts:
         hbi_kessel_relations_grpc.verify_created_or_updated(host, ungrouped_group)
+
+    if host_inventory.application.config.current_env.lower() != "clowder_smoke":
+        logger.info("Waiting 10 seconds to work around hosts availability flapping issue")
+        sleep(10)
 
     host_inventory.apis.hosts.delete_filtered(display_name=hosts_data[0]["display_name"])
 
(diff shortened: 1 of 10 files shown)


iqe-host-inventory-plugin/iqe_host_inventory/utils/rbac_utils.py
```diff
@@ -67,6 +67,11 @@ def wait_for_kessel_sync(host_inventory: ApplicationHostInventory) -> None:
     Wait for RBAC -> Kessel sync if the hbi.rbac-v2 feature flag is enabled.
     """
     wait_seconds = 3 if host_inventory.application.config.current_env == "clowder_smoke" else 21
-    if host_inventory.unleash.is_hbi_rbac_v2_enabled:
+    if (
+        host_inventory.unleash.is_hbi_rbac_v2_enabled
+        # We need to wait in all environments where BYPASS_KESSEL=false, because all workspace
+        # write operations go through Kessel even when the hbi.rbac-v2 FF is disabled
+        or host_inventory.application.config.current_env != "clowder_smoke"
+    ):
```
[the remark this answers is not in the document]
fstavela: We can't efficiently gate it on the `BYPASS_KESSEL` env var in Stage and Prod, because we don't have access to it. This env var is always enabled in Stage and Prod, so there is no risk of unnecessary waiting in these environments.

iqe-host-inventory-plugin/iqe_host_inventory/tests/rest/kessel/test_kessel_host_replication.py
```diff
@@ -359,6 +363,10 @@ def test_kessel_repl_delete_hosts_by_filter(
     for host in hosts:
         hbi_kessel_relations_grpc.verify_created_or_updated(host, ungrouped_group)
 
+    if host_inventory.application.config.current_env.lower() != "clowder_smoke":
```
rodrigonull, line 364: This test has the ephemeral mark, which means it only runs on ephemeral envs, right? If that is true, this if condition is useless.

fstavela: You are right, nice catch! There isn't a good reason for this test to run only in ephemeral, though, so instead of removing the if condition, I updated the test to be able to run in Stage and Prod and removed the ephemeral marker.

rodrigonull (approved):

PR #4922, opened 2026-09-09
feat(RHINENG-28294): align export with inventory views columns

rodrigonull:
## Jira
[RHINENG-28294](https://issues.redhat.com/browse/RHINENG-28294)

## What
Export now respects the active View's column configuration, so the exported data includes only the columns defined in the View (in the same order) instead of dumping every field.

## Why
Users expect the export to match what they see on screen. Previously, exports always emitted the full default field set regardless of which columns the View was configured with. This story wires the saved column configuration into the export pipeline for both CSV and JSON formats.

## How
- Renamed `_load_view_filters` → `_load_view_config` to also return the View's `columns` list alongside the existing host/query filters.
- Added `resolve_export_columns()` in `export_service.py` — maps View column keys (including `app:field` app-data keys) to the ordered flat export field list, filtering out columns for applications the requester lacks permission to view.
- Extended `get_allowed_app_services()` in `lib/middleware.py` to accept optional `identity` and request headers, enabling per-app RBAC enforcement in the export consumer. View filters referencing unauthorized apps reject the export with a 403 error.
- Extended `get_hosts_to_export()` in `host_query_db.py` with `export_fields` and `app_data_fields` parameters:
  - System profile tables (static and dynamic) are conditionally `LEFT JOIN`ed only when referenced by export fields or filters (`_export_needs_profile_joins`).
  - App-data columns are queried in batches via `_fetch_app_data_batch()` instead of joining every app table onto the main scan.
- Formatted nested JSON fields for flat export:
  - `workloads` and `per_reporter_staleness` extract root keys as comma-separated strings.
  - `compliance:policies` extracts policy names from the JSON array.
  - `operating_system` maps to `os_release`.
- Updated `_StreamingExportBody` and `serialize_host_row_for_export` to accept a custom field list and uniformly access row attributes, falling back to the default set for legacy exports.
- Added `CORE_VIEW_FIELDS_TO_EXPORT_FIELDS` and `ALWAYS_INCLUDED_EXPORT_FIELDS` constants in `serialization.py` to centralize the column-key → export-field mapping.

## Testing
Added comprehensive unit tests covering:
- `TestLoadViewConfig` — view config loading with filters + columns
- `TestResolveExportColumns` — core/app-data/unknown/duplicate column resolution, ordering, and RBAC app filtering
- `TestStreamingExportBodyCustomFields` — CSV header and JSON output with custom fields
- `TestGetHostsToExportWithColumns` — app-data batching, key ordering, conditional profile joins, and root key / policy name extraction
- `TestCreateExportWithView` — end-to-end export with view columns in JSON and CSV formats, including RBAC permission enforcement and rejection
- `TestFormatCompliancePolicies` — extracting policy names from compliance policy lists

## PR Guidelines

## PR Guideline checks

[RHINENG-28294]: https://redhat.atlassian.net/browse/RHINENG-28294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Make inventory exports match the active View columns while preserving ordering, app-data access controls, and legacy export behavior.

New Features:
- Align CSV and JSON exports with the selected inventory View's ordered column configuration, including supported app-data columns.
- Apply app-service permissions to View export columns and reject exports that filter on unauthorized applications.

Bug Fixes:
- Prevent exports from including fields or app data that are not represented by the active View or permitted for the requester.

Enhancements:
- Optimize export queries by conditionally loading system profiles, joining only app-data tables needed for filters, and batch-fetching app-data columns.
- Add consistent flat formatting for nested system-profile and compliance data and preserve legacy exports when no View columns are configured.

Tests:
- Add unit and integration coverage for View configuration loading, column resolution, RBAC enforcement, custom CSV/JSON output, optimized joins, batched app-data loading, and nested field formatting.


rodrigonull:
@sourcery-ai review

rodrigonull:
@sourcery-ai review

app/serialization.py
```diff
@@ -283,39 +300,69 @@ def serialize_host(
...
-        "satellite_managed": row.satellite_managed,
-        "cloud_provider": row.cloud_provider,
-        "is_marketplace": row.is_marketplace,
-        "ip_addresses": row.ip_addresses,
+        "os_release": getattr(row, "os_release", None),
+        "updated": _serialize_datetime(modified_on) if modified_on else None,
+        "created": _serialize_datetime(created_on) if created_on else None,
+        "last_check_in": _serialize_datetime(last_check_in) if last_check_in else None,
+        "data_collector": list(reporters) if reporters else None,
+        "per_reporter_staleness": _extract_root_keys(getattr(row, "per_reporter_staleness", None) or reporters),
```
kruai: Doesn't this just return a list of reporters, rather than per-reporter staleness values? I think there's some additional transformation that's missing here

tests/test_export_service.py
kruai, line 1: Can the new tests be refactored or simplified? I want to make sure we need all >800 added lines before I go through them more thoroughly :)

kruai (changes requested):
Thanks for the PR! Mostly looks good - I just see one issue and have one question.

app/serialization.py
rodrigonull: @kruai the view column is `per_reporter_staleness`, but in the UI it is displayed as `Data Collector` and it only renders reporter names from Object.keys(...), not the nested staleness JSON. Exporting the keys matches what’s on screen (and stays CSV-friendly), so I am not sure we should change that. WDYT?

kruai: The column `per_reporter_staleness` corresponds to `Data Collector` in the UI, even though we're also returning a `data_collector` field? That's really confusing... I think we should return `per_reporter_staleness` in full if we're calling it by name, or rename it to `reporters` in the response if we're just returning the list of reporters.
Also is there a lookup happening on the front-end? Because "puptoo" isn't a Data Collector, but "insights-client" is.

rodrigonull: You are right, the view column key stays `per_reporter_staleness` (that’s what the UI saves), but now I changed the export service to export the `data_collector` instead of `per_report_staleness`.

You’re also right about the labels. I've changed it now to apply the same frontend mapping (puptoo → insights-client, rhsm-* → subscription-manager, etc.), so the CSV/JSON match the Data collector column.

api/host_query_db.py
```diff
@@ -952,44 +1017,82 @@ def get_hosts_to_export(
...
+            for row in base_query:
+                host_row = _row_to_export_host(row, labels)
+                serialized = serialize_host_row_for_export(
+                    host_row, staleness=staleness, fields=resolved_export_fields
+                )
+                batch_hosts.append((host_row.id, serialized))
+
+                if len(batch_hosts) >= effective_batch_size:
+                    host_ids = [h[0] for h in batch_hosts]
+                    app_data = _fetch_app_data_batch(host_ids, identity.org_id, app_data_fields)
```
kruai, line 1076: I'm concerned about combining `yield_per` (line 1059) with the query being called in `_fetch_app_data_batch`. We might need to use a second DB connection. Can you please add a test with batch_size=1 and 3 (or more) hosts with app_data to validate this?

kruai (changes requested):
Found one more thing I'm concerned about before I approve. I was briefly concerned that "yupana" was not in the data-collectors lookup list, but since it doesn't map cleanly to one specific DC, we can stick with the fallback behavior (just using "yupana").

api/host_query_db.py
rodrigonull: Good catch. I added a batch_size=1 case with 3 hosts, and _fetch_app_data_batch now uses a separate Session so those lookups don’t share the streaming connection.

kruai (approved):
LGTM 👍

PR #4956, opened 2026-09-15
fix(iqe): gracefully handle RBAC setup failures in ephemeral environments

adarshdubey-star:
## Summary by Sourcery

Keep ephemeral test sessions running after RBAC setup failures while clearly skipping tests that depend on RBAC.

Bug Fixes:
- Allow ephemeral-environment test sessions to continue when RBAC role setup fails by skipping only RBAC-dependent tests.

Enhancements:
- Record RBAC setup failures for reuse across test fixtures and avoid Kessel synchronization when setup is incomplete.

Tests:
- Add automatic skipping for RBAC tests when session-level RBAC setup fails.


fstavela (changes requested):
I wouldn't skip the tests, then we can miss bugs. Especially in the RBAC integration, which can't be very well tested in unit tests.

PR #4980, opened 2026-09-18
Avoid unconditional writes and full-row hydration in bulk_tag_hosts

pablomh:
## Summary

`POST /tags` (`bulk_tag_hosts`) issues an unconditional `UPDATE` for every host in a batch, even when the merged tag data is byte-identical to what's already stored, and hydrates full `Host` ORM rows (including large unused `facts`/`system_profile_facts`/`canonical_facts` JSONB columns) to do it.

## Root cause

- `update_host_tags()` calls `combine_tags()` to merge incoming tags into the host's existing tag dict (additive-only), then unconditionally calls `host._update_tags(current_tags)`.
- `_update_tags()` → `_replace_tags_in_namespace()` always calls `orm.attributes.flag_modified(self, "tags")`, forcing SQLAlchemy to emit an `UPDATE` for that host on commit, regardless of whether the tag data actually changed.
- Separately, `process_host_batch()`'s query loads full `Host` entities, hydrating every column even though only `id`/`tags`/`tags_alt` are used on the common (no-op) path.

On a large inventory, a caller that re-tags the same hosts with the same values repeatedly (e.g. an RBAC-scoping tag applied on every forwarded API request) pays full write + full-row-read cost every time, for no actual data change.

## Changes

- `update_host_tags()`: snapshot the tags dict before merging, compare after, and skip `_update_tags()`/`flag_modified()` entirely when nothing changed.
- `process_host_batch()`: use `load_only(Host.id, Host.tags, Host.tags_alt)` instead of fetching full rows.

## Verification

Verified live against a 6,241-host Satellite/IoP environment, issuing the same tag value repeatedly:
- `UPDATE` statements per no-op `/tags` call: 6,305 → 0
- Wall-clock per call: ~5.6s → ~0.47s

## Test plan

- [x] Added `test_bulk_tag_hosts_noop_skips_write` asserting `modified_on` doesn't bump on a repeat call with identical tags
- [x] Existing `tests/test_app_tags.py` suite covers the unchanged behavior (success, not-found, empty tags, too-many-tags)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Optimize bulk host tagging by avoiding redundant updates and unnecessary host data loading.

Bug Fixes:
- Prevent bulk host tagging from issuing database writes when the requested tags already match the stored tags.

Enhancements:
- Reduce bulk tagging query overhead by loading only the host fields required for tag processing.

Tests:
- Add coverage for no-op tagging and hosts with null tag data.


app/tags_blueprint.py
```diff
@@ -54,8 +57,11 @@ def combine_tags(input_list, existing_dict=None):
 
 def update_host_tags(session, host, tags):
     try:
+        original_tags = deepcopy(host.tags)
         current_tags = host.tags
         combine_tags(tags, current_tags)
```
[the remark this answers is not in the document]
pablomh: Confirmed and fixed, thanks - applied your suggested `current_tags = combine_tags(tags, current_tags)`.

Verified live against a real host row with `tags` temporarily set to `None` (rolled back, not persisted): before the fix, `update_host_tags()` returned `True` without ever calling `_update_tags()`; after the fix, it correctly returns `False`, since `_update_tags()`'s own "tags cannot be null" guard runs again.

app/tags_blueprint.py
vkrizan, line 84: @rodrigonull  I'd suggest to check whether there would be a possibility to do a true bulk update of hosts, avoiding issuing a one UPDATE per host. Is this achievable?

tests/test_app_tags.py
```diff
@@ -61,6 +63,43 @@ def test_bulk_tag_hosts_empty_host_list(api_post):
...
+    db.session.commit()
+
+    payload = {
+        "tags": [{"namespace": "test", "key": "env", "value": "prod"}],
+        "host_id_list": [host_id],
+    }
+
+    status_code, data = api_post("/tags", host_data=payload)
+    assert status_code == 200
+    assert data["processed_hosts"] == 0
```
fstavela, line 100: Why do we need this test? This sounds like something that will never happen. And if it does happen, shouldn't the API return an error instead?

fstavela (dismissed):
LGTM, I left just one question.

app/tags_blueprint.py
```diff
@@ -65,7 +70,12 @@ def update_host_tags(session, host, tags):
 
 
 def process_host_batch(session, identity, batch_ids, tags):
-    hosts = session.query(Host).filter(and_(Host.org_id == identity.org_id, Host.id.in_(batch_ids))).all()
+    hosts = (
+        session.query(Host)
+        .options(load_only(Host.id, Host.tags, Host.tags_alt))
```
vkrizan, line 75: Does it need to pull in the `tags_alt`?

app/tags_blueprint.py
```diff
@@ -54,8 +57,10 @@ def combine_tags(input_list, existing_dict=None):
 
 def update_host_tags(session, host, tags):
     try:
-        current_tags = host.tags
-        combine_tags(tags, current_tags)
+        original_tags = deepcopy(host.tags)
```
vkrizan, line 60: A room for improvement here would be to improve the `combine_tags` to not change the structure but create a new one. Then the deepcopy would be unnecessary.

app/tags_blueprint.py
```diff
@@ -54,8 +57,10 @@ def combine_tags(input_list, existing_dict=None):
...
+        current_tags = combine_tags(tags, host.tags)
+        if current_tags == original_tags:
```
vkrizan, line 62: Just to double check here, this compares complex structures. If the structures are just primitives (list and dict) then the comparison would be straightforward, with one caveat of dependence on the order. Should the order of tags trigger an update?

tests/test_app_tags.py
pablomh: This locks in a regression: without it, a null-tags host would silently report as processed (the exact bug Sourcery caught). On the error-handling point: today's behavior (silent catch, logged, not counted) is pre-existing, this PR doesn't change it - worth a separate discussion on its own.

app/tags_blueprint.py
pablomh: No - `tags_alt` is only read on the changed path (`_replace_tags_alt_in_namespace`), never during the no-op check. Dropped it; SQLAlchemy lazy-loads it transparently for hosts that actually change.

app/tags_blueprint.py
pablomh: Moved the deepcopy inside `combine_tags` so it never mutates `existing_dict`. Removes the exact mutate/don't-mutate asymmetry that caused the null-tags bug in the first place.

app/tags_blueprint.py
pablomh: Checked - no, `combine_tags` only appends a value if it's not already in the list, so a genuine no-op never reorders existing entries. Order could only diverge if something outside this function reordered `tags` directly, which nothing does.

app/tags_blueprint.py
vkrizan: This is not what I've meant. If the combine_tags would not mutate the structure, but rather recreate it, then the deepcopy would not be needed.

pablomh: Rebuilt the result explicitly (`{namespace: {key: [values]}}` with fresh dict/list containers at every level) instead of deepcopy. `deepcopy` is gone from the file entirely now.

vkrizan: @pablomh This is still misunderstood. The recent change essentially made the code do what basically deepcopy does. The suggestion was to change the algorithm on how it walks the structures and builds a new one during the walk.

Anyhow, I'll let the you (or the Inventory) team decide whether you want to improve it. The current state or the deepcopy would do the job. My intent was to share better coding practices.

pablomh: Thanks for the patience @vkrizan. I now get it.

vkrizan: Thanks as well. After consideration (of handling with complex structure), I'm happy with the last changes.

pablomh:
On `result = existing_dict or {}`: that's intentional, not an oversight - it's what lets a true no-op return `existing_dict` itself, unmodified, with zero copies. `result` only gets replaced with new dict/list objects for the specific namespace/key a given item actually touches; if nothing in `input_list` changes anything, `result` is never reassigned and the original object comes back as-is.

I also built and benchmarked the flat-set-of-triplets idea, since it's a reasonable thing to try:

```python
existing_triplets = {(ns, k, v) for ns, kv in existing_dict.items() for k, vs in kv.items() for v in vs}
combined = set(existing_triplets)
combined.add((namespace, key, value))
# ...then rebuild {namespace: {key: [values]}} from combined
```

Measured against the current approach, same realistic shape (a handful of small namespaces + one namespace/key holding N accumulated values):

| N | flat-set | current (copy-on-write) |
|--:|--:|--:|
| 20 | 5.05μs | 0.37μs |
| 500 | 33.32μs | 0.36μs |
| 2,000 | 165.76μs | 0.37μs |
| 10,000 | 1,141.30μs | 0.38μs |

~3,000x slower at scale. The membership check was never the bottleneck (a single key's value list is small either way) - the cost is that building a flat set requires walking the *entire* existing structure once, and the `{namespace: {key: [values]}}` shape everything downstream expects has to be rebuilt from that flat set afterward, another full walk. Two full passes over data the request never touched.

**Conclusion:** `combine_tags`'s loop only needs to touch the namespace/key path a given item changes. Starting from the existing object and forking off new containers just along that path is what makes a true no-op cost nothing, without a separate rebuild or flatten/re-nest step. Any version that processes the whole structure up front - rebuild or flat-set alike - pays for the whole structure on every call, no matter how fast its internal lookups are.

